PR #18, opened 2014-09-22
Add condition to assign tracked host name in settings panel

rememberlenny:
We have a common problem where Parsely tracks our non-production environment URLs. We have had multiple instances where staging or non-production ready URLs pop up in our Parsely powered related links widgets. To prevent this, we are explicitly declaring the host name which Parsely should track.

We created an optional field in the Settings panel to declare the preferred hostname. If this is declared, then Parsely will not generate the <meta> tag to track the host name. 

screen shot 2014-09-22 at 11 37 40 am

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -55,6 +55,12 @@ class Parsely {
     public function __construct() {
         // Run upgrade options if they exist for the version currently defined
         $options = $this->get_options();
+
+        $should_parsely_track = (
+            !isset($options['explicit_domain_accepted']) || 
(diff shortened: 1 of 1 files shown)


joshuarrrr:
@hbbtstar This PR is really old. I think we could find a way to merge it, but I tend to think that disallowing subdomains is now better handled by the CSM on any given account through django (so that we don't even collect from those dev, staging, or qa sites.

rememberlenny:
I don’t even remember it! I would recommend closing it.

—
Leonard Bogdonoff
RememberLenny.com | 949-322-8496

joshuarrrr:
ok, closing due to age

PR #20, opened 2015-01-06
Add a `wp_parsely_post_tags` filter for tags

adamsilverstein:
Allows developers to adjust the tag list, for example adding all terms from custom taxonomies.

Example usage:

```php
function add_all_tags_to_parsely( $wpTags, $postId ){
	$taxonomies = get_object_taxonomies( 'post' );
	// Remove the default taxonomies which are already accounted for
	$taxonomies = array_diff( $taxonomies, array( 'category', 'post_tag' ) );
	$terms      = wp_get_post_terms( $postId, $taxonomies );

	return array_merge( $terms, $wpTags );
}
add_filter( 'wp_parsely_post_tags', 'add_all_tags_to_parsely', 10, 2 );
```

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -523,6 +523,7 @@ class Parsely {
     */
     private function get_tags_as_string($postId, $parselyOptions) {
         $wpTags = wp_get_post_tags($postId);
+        $wpTags = apply_filters( 'wp_parsely_post_tags', $wpTags, $postId );
         $tags = array();
         foreach ( $wpTags as $wpTag ) {
             if ( $parselyOptions['lowercase_tags'] === true ) {


msukmanowsky:
This is awesome @adamsilverstein! Thanks for the contribution, super elegant approach.

In another plugin for Parse.ly, we are considering adding support for custom taxonomies via a settings screen where users could select the taxonomies that are supposed to be included for tags. Way more involved than the one line change we have here, but allows us to better support teams who may not understand how to use WordPress filters. Thoughts?

adamsilverstein:
Great, thanks for the kind words, glad to contribute.

I can definitely see a use case for adding something on the options screen if its a common request (although most users will likely be able to ignore the option). 

In my case, custom taxonomies are added programmatically so I'm comfortable using a filter to add in my own and this approach is widely used by WordPress plugins to make them more developer-extendable. Its almost a convention to provide extensibility via hooks vs. an options screen if those options are only going to get used by a small number of serious developer types - it keeps the options screen simple while provide the extensibility developers need!

adamsilverstein:
ps, last line of code sample above has a type, should be 
`return array_merge( $terms, $wpTags );`

PR #21, opened 2015-01-13
Add a parsely_post_authors filter Add a parsely_post_authors filter

adamsilverstein:
Allows themes to extend the reported authors, for example adding authors from a custom taxonomy.

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -686,7 +686,6 @@ class Parsely {
             $authors = array(get_user_by('id', $post->post_author));
         }
         $authors = array_map(array($this, 'get_author_name'), $authors);
-        $authors = array_map(array($this, 'get_author_name'), $authors);
         $authors = apply_filters( 'wp_parsely_post_authors', $authors, $post );
         $authors = array_map(array($this, 'get_clean_parsely_page_value'), $authors);
         return $authors;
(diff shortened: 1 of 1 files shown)


msukmanowsky:
@adamsilverstein ditto here, can we move the filter to before the `get_clean_parsely_page_value` call for the values in the array?

adamsilverstein:
Sure, makes sense. I will resubmit.

adamsilverstein:
Ok, updated this as well. Here is an example of how we used it, to pull in our custon person taxonomy... feel free to use or link this in your docs:

https://gist.github.com/adamsilverstein/1c1291440f935faab085

adamsilverstein:
@msukmanowsky ^^ updated :)

wp-parsely.php
```diff
@@ -686,6 +686,8 @@ private function get_author_names($post) {
             $authors = array(get_user_by('id', $post->post_author));
         }
         $authors = array_map(array($this, 'get_author_name'), $authors);
+        $authors = array_map(array($this, 'get_author_name'), $authors);
```
msukmanowsky: Whoops, looks like we now have a duplicated `$authors = array_map(array($this, 'get_author_name'), $authors);` line @adamsilverstein.

adamsilverstein:
@msukmanowsky Oops! Fixed.

msukmanowsky:
Perfect, thanks!

PR #22, opened 2015-01-19
Add a 'wp_parsely_post_category' filter

adamsilverstein:
Allows developers to customize the category reported for a post.

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -599,8 +599,9 @@ class Parsely {
             $category = 'Uncategorized';
         }
 
+        $category = apply_filters( 'wp_parsely_post_category', $category, $postObj, $parselyOptions );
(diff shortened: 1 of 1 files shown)


msukmanowsky:
@adamsilverstein could we move the filter to before the `get_clean_parsely_page_value` call?

adamsilverstein:
sure, that would work, but you wouldn’t be able to override the return of that function, maybe thats fine I can't recall what that function does.

I will work up an updated version with the filter before the call.

adamsilverstein:
@msukmanowsky I updated this pull request to shift the filter to before the get_clean_parsely_page_value call.

msukmanowsky:
Awesome, thanks @adamsilverstein!

adamsilverstein:
You are welcome, thanks for open sourcing your plugin and also for your spiffy analytics service!

PR #23, opened 2015-02-05
Make parselyPage a class member and expose a new filter

eyalgursoy:
Make parselyPage a class member.
Extract from insert_parsely_page the part which handles the parselyPage init.
Expose a new filter called after_set_parsely_page - handing over the parselyPage member to enable full control for the user to change it.
Adjust the template to use parselyPage class member.


jtotoole:
Thanks for this, @eyalgursoy! This was a really useful suggestion and code change, which we actually pulled into [a larger release](https://github.com/Parsely/wp-parsely/pull/50). We're managing this project more actively at present, so please feel free to submit additional PRs/issues going forward.

PR #27, opened 2016-04-11
[WIP] googleamp metadata + section taxonomies

hbbtstar:
PR to add section taxonomies + googleamp metadata to Wordpress plugin

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -44,11 +44,11 @@ class Parsely {
     const MENU_PAGE_TITLE     = 'Parse.ly > Settings'; // Text shown in <title></title> when the settings screen is viewed
     const OPTIONS_KEY         = 'parsely';             // Defines the key used to store options in the WP database
     const CAPABILITY          = 'manage_options';      // The capability required for the user to administer settings
-    const CATEGORY_DELIMITER  = '~-|@|!{-~';
 
     private $optionDefaults     = array('apikey' => '',
                                         'content_id_prefix' => '',
                                         'use_top_level_cats' => false,
+                                        'custom_taxonomy_section' => 'category',
                                         'cats_as_tags' => false,
                                         'track_authenticated_users' => true,
                                         'lowercase_tags' => true);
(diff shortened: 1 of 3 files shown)


parsely-parsely-page.php
```diff
@@ -4,6 +4,7 @@
    <script type="application/ld+json">
    <?php echo json_encode($parselyPage); ?>
    </script>
+   <meta name='parsely-post-id' content='<?php echo $postId ?>' />
```
hbbtstar: This errors out on pages with no $postId, likely because pages / section etc. don't have postIds but can have headlines, thus fulfilling the if condition above. Might be worth checking if $postid is set and is not empty before echoing this

wp-parsely.php
```diff
@@ -152,19 +152,33 @@ public function initialize_settings() {
...
+             'populate the section value instead. <br>';
+        add_settings_field('custom_taxonomy_section',
+                           'Use Custom Taxonomy for Section  <div class="help-icons"></div>',
+                           array($this, 'print_select_tag'),
+                           Parsely::MENU_SLUG, 'optional_settings',
+                           array('option_key' => 'custom_taxonomy_section',
+                                 'help_text' => $h,
+                                 // filter Wordpress taxonomies under the hood that should not appear in dropdown
+                                 'select_options' => array_diff(get_taxonomies(), array('post_tag', 'nav_menu', 'link_category', 'post_format')),
+                                 'requires_recrawl' => false));
```
hbbtstar: might want to flag requires_recrawl as true here since if they change it in an existing install it will require a recrawl

wp-parsely.php
```diff
@@ -312,7 +328,7 @@ public function insert_parsely_page() {
         // TODO: Maping of an install's post types to Parse.ly post types (namely page/post)
         $parselyPage = array(
             "@context" => "http://schema.org",
-            "@type" => "WebPage"
+            "@type" => "NewsArticle"
```
hbbtstar: This should stay WebPage as the default- it gets overridden in 357 if the post is in fact a post, and if this is WebPage that means that non-post pages will get NewsArticle as their @type too.

wp-parsely.php
```diff
@@ -339,13 +355,30 @@ public function insert_parsely_page() {
...
+                '@type' => 'ImageObject',
+                'url' => $image_url
+            );
             $parselyPage['dateCreated']    = gmdate('Y-m-d\TH:i:s\Z', get_post_time('U', true));
+            $parselyPage['datePublished']  = gmdate('Y-m-d\TH:i:s\Z', get_post_time('U', true));
+            $parselyPage['dateModified']   = gmdate('Y-m-d\TH:i:s\Z', get_the_modified_date('U', true));
             $parselyPage['articleSection'] = $category;
+             $parselyPage['author']        = array(
+                '@type' => 'Person',
+                'name' => $authors
```
hbbtstar: how does Google want this expressed? should there be a new person object for each author? if so you'll have to expand the array here to have a new type = Person for each author

parsely-parsely-page.php
jtotoole: Away it goes!

wp-parsely.php
jtotoole: Good thinking

wp-parsely.php
jtotoole: truth

wp-parsely.php
jtotoole: Did some Google sleuthing but couldn't come up with anything more definitive than [this StackExchange post](http://webmasters.stackexchange.com/questions/90187/schema-org-book-with-multiple-authors-using-json-ld), which recommends passing a list of author objects to a single `author` key, so I went that route. 

I guess we could test this by adding the updated code in this PR to a sample WP blog that uses our plugin and see if Google validates it okay before we merge in this repo.

hbbtstar:
Looks good to me! :+1:

PR #32, opened 2016-06-06
Add a `wp_parsely_post_title` filter to change the headline

moraleida:
This is useful in case you have different versions of the Post Title, say a long form and a short one, or in any case you want a modified title to be sent. Passes the result of `get_the_title()` as the first parameter and the post ID as the second.

Example:
```php
function filter_article_title( $title, $obj ) {
    $shortform = get_post_meta( $obj->ID, 'shortform_title', true );
    return $shortform;
}
add_filter( 'wp_parsely_post_title', 'filter_article_title' `,` 10, 2 );
```

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -341,6 +341,7 @@ class Parsely {
             $authors    = $this->get_author_names($post);
             $category   = $this->get_category_name($post, $parselyOptions);
             $postId     = $parselyOptions['content_id_prefix'] . (string)get_the_ID();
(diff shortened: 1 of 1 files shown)


amontalenti:
@moraleida Thank you for this contribution. We'll be reviewing this shortly and update here with more info. cc @hbbtstar

jtotoole:
Thanks for this, @moraleida! This was a really useful sugguestion, which we actually pulled into [a larger release](https://github.com/Parsely/wp-parsely/pull/50). We're managing this project more actively at present, so please feel free to submit additional PRs/issues going forward.

PR #42, opened 2016-07-20
Use cached function (slight performance gain)

jasonbahl:
During a VIP GO code review of our project, it was suggested that we update all instances of uncached functions, and this was on the list of updates.

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -602,6 +602,7 @@ class Parsely {
         foreach ( $wpTags as $wpTag ) {
             array_push($tags, $wpTag->name);
         }
+        return $tags;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


jasonbahl:
checking why the test was failing. Looks like it might be an issue with the test itself. Committing a few things to re-test.

hbbtstar:
@jasonbahl thanks for this! it's actually quite timely, as we got the same feedback from WP VIP and were looking to integrate this in the next release. Would you do me a favor and open this PR against the release/v1.10 branch so we can merge it there? I'll merge the whole kit and kaboodle into master once the rest of the milestone issues are addressed and then deploy it to Wordpress plugin directory /  VIP. Thanks!

jtotoole:
Thanks again for this, @jasonbahl—we incorporated as part of [a larger recent release](https://github.com/Parsely/wp-parsely/pull/50). Please feel free to submit additional issues/pull requests going forward, they're always appreciated.

PR #47, opened 2016-09-14
Added AMP tracking and unit tests

cayci:
Added AMP tracking.  Added tracking note to wp-parsely settings.  Added unit test.

### DIFF
--- a/parsely-amp-tracking.php
+++ b/parsely-amp-tracking.php
@@ -1,0 +1,25 @@
+add_action( 'pre_amp_render_post', 'parsely_amp_add_actions' );
+function parsely_add_amp_actions() {
+    add_filter( 'amp_post_template_analytics', 'parsely_amp_add_analytics' );
+}
+
+function parsely_amp_add_analytics( $analytics ) {
+    global $parsely;
+    $parsely_options = $parsely->get_options();
+
+    if ( empty( $parsely_options['apikey'] ) ) {
+        return $analytics;
+    }
+
+    $analytics['wp-parsely'] = array(
+        'type' => 'parsely',
+        'attributes' => array(),
+        'config_data' => array(
+            'vars' => array(
+                'apikey' => $parsely_options['apikey'],
+            )
+        ),
+    );
+
+    return $analytics;
+}
--- a/tests/test-wp-parsely.php
+++ b/tests/test-wp-parsely.php
@@ -250,7 +250,19 @@ class SampleTest extends WP_UnitTestCase {
         $ppage = self::$parsely->insert_parsely_page();
         $this->assertTrue(strpos($ppage['url'], 'https', 0) == 0);
     }
-
+    
+    
+      
+    function test_amp_analytics() {  
+      	$options = get_option('parsely');
+        $post_array = $this->create_test_post_array();
+        $post = $this->factory->post->create($post_array);
+        $this->go_to('/?p=' . $post);
+        $ppage = self::$parsely->insert_parsely_amp_tracking();
+        $this->assertTrue(strpos($ppage['url'], '/amp') == 0);
+        $this->assertFalse(strpos($ppage['url'], '/amp'));
+        $this->assertTrue(strpos($ppage['url'], 'amp=1') == 0);
+        $this->assertFalse(strpos($ppage['url'], 'amp=1'));       
+    }
 }
-
 
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -79,10 +79,18 @@ class Parsely {
         $basename = plugin_basename(__FILE__);
         add_filter('plugin_action_links_' . $basename,
                    array($this, 'add_plugin_meta_links'));
-
-        // inserting parsely code
-        add_action('wp_head', array($this, 'insert_parsely_page'));
-        add_action('wp_footer', array($this, 'insert_parsely_javascript'));
+		
+		// Add AMP tracking if this is an AMP page 
+		$currentURL = $this->get_current_url();
+	  	if ((strpos($currentURL, '/amp') !== false) || (strpos($currentURL,'amp=1') !== false)) {
+			// This is an AMP page; add AMP tracking
+		  	insert_parsely_amp_tracking();
+		} else {
+			// This is not an AMP page; do not add AMP tracking
+        	// inserting parsely code
+        	add_action('wp_head', array($this, 'insert_parsely_page'));
+        	add_action('wp_footer', array($this, 'insert_parsely_javascript'));
+        }
     }
 
     public function add_admin_header() {
@@ -248,6 +256,12 @@ class Parsely {
         add_settings_field('dynamic_tracking_note', 'Note: ',
                             array($this, 'print_dynamic_tracking_note'),
                             Parsely::MENU_SLUG, 'optional_settings');
+            
+                            
+        // AMP tracking note
+        add_settings_field('amp_tracking_note', 'Note: ',
+                            array($this, 'print_amp_tracking_note'),
+                            Parsely::MENU_SLUG, 'optional_settings');
 
     }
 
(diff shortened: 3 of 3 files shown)


jtotoole:
Hey @cayci! I played around with this code in my WordPress dev environment with the AMP plugin installed and realized that we'll want to take a different approach here. I've left some more detailed comments in line!

parsely-amp-tracking.php
```diff
@@ -0,0 +1,25 @@
+add_action( 'pre_amp_render_post', 'parsely_amp_add_actions' );
```
jtotoole: We'll actually want to get rid of this as a separate file and instead insert this code into `wp-parsely.php`, as part of the `Parsely` class. 

To start, let's make this line `add_action('pre_amp_render_post', array($this, 'parsely_add_amp_actions'));` and add it beneath [this line](https://github.com/Parsely/wp-parsely/blob/release/v1.10/wp-parsely.php#L85) in the `Parsely` class's `__construct` function. We'll need to pass that second callback as an array with `$this` since we'll be including the functions below within the `Parsely` class

parsely-amp-tracking.php
```diff
@@ -0,0 +1,25 @@
...
+function parsely_add_amp_actions() {
```
jtotoole: This will become `public function parsely_add_amp_actions()` once we've got it within the `Parsely` class

parsely-amp-tracking.php
```diff
@@ -0,0 +1,25 @@
...
+    add_filter( 'amp_post_template_analytics', 'parsely_amp_add_analytics' );
+}
+
+function parsely_amp_add_analytics( $analytics ) {
```
jtotoole: This will become `public function parsely_add_amp_analytics($analytics)` once we've got it within the `Parsely` class

parsely-amp-tracking.php
```diff
@@ -0,0 +1,25 @@
+add_action( 'pre_amp_render_post', 'parsely_amp_add_actions' );
+function parsely_add_amp_actions() {
+    add_filter( 'amp_post_template_analytics', 'parsely_amp_add_analytics' );
```
jtotoole: This will become `add_filter('amp_post_template_analytics', array($this, 'parsely_add_amp_analytics'));`, again since it's in the `Parsely` class.

Note again the consitency for function naming--let's name this callback `parsely_add_amp_analytics`

parsely-amp-tracking.php
```diff
@@ -0,0 +1,25 @@
...
+}
+
+function parsely_amp_add_analytics( $analytics ) {
+    global $parsely;
```
jtotoole: This line can be deleted since we won't need `parsely` as a global within the `Parsely` class

parsely-amp-tracking.php
```diff
@@ -0,0 +1,25 @@
...
+    if ( empty( $parsely_options['apikey'] ) ) {
+        return $analytics;
+    }
+
+    $analytics['wp-parsely'] = array(
+        'type' => 'parsely',
+        'attributes' => array(),
+        'config_data' => array(
+            'vars' => array(
+                'apikey' => $parsely_options['apikey'],
```
jtotoole: 'apikey' => $parselyOptions['apikey']

parsely-amp-tracking.php
```diff
@@ -0,0 +1,25 @@
+add_action( 'pre_amp_render_post', 'parsely_amp_add_actions' );
+function parsely_add_amp_actions() {
+    add_filter( 'amp_post_template_analytics', 'parsely_amp_add_analytics' );
+}
+
+function parsely_amp_add_analytics( $analytics ) {
+    global $parsely;
+    $parsely_options = $parsely->get_options();
+
+    if ( empty( $parsely_options['apikey'] ) ) {
```
jtotoole: if ( empty( $parselyOptions['apikey'] ) ) {

parsely-amp-tracking.php
```diff
@@ -0,0 +1,25 @@
+add_action( 'pre_amp_render_post', 'parsely_amp_add_actions' );
+function parsely_add_amp_actions() {
+    add_filter( 'amp_post_template_analytics', 'parsely_amp_add_analytics' );
+}
+
+function parsely_amp_add_analytics( $analytics ) {
+    global $parsely;
+    $parsely_options = $parsely->get_options();
```
jtotoole: The plugin calls this object `$parselyOptions`; we'll set it here with `$parselyOptions = $this->get_options();`

wp-parsely.php
```diff
@@ -79,10 +79,18 @@ public function __construct() {
...
+		$currentURL = $this->get_current_url();
+	  	if ((strpos($currentURL, '/amp') !== false) || (strpos($currentURL,'amp=1') !== false)) {
+			// This is an AMP page; add AMP tracking
+		  	insert_parsely_amp_tracking();
+		} else {
+			// This is not an AMP page; do not add AMP tracking
+        	// inserting parsely code
+        	add_action('wp_head', array($this, 'insert_parsely_page'));
+        	add_action('wp_footer', array($this, 'insert_parsely_javascript'));
+        }
```
jtotoole: We'll actually want to get rid of the new code added on lines 82-90, removing the `if` block and reverting this section to the way it appeared previously. We'll then simply add the line `add_action('pre_amp_render_post', array($this, 'parsely_add_amp_actions'));` below the two existing `add_action` calls.

wp-parsely.php
```diff
@@ -476,6 +495,25 @@ public function insert_parsely_page() {
...
+        global $post;
+        $display = TRUE;
+        if ( is_single() && $post->post_status != 'publish' ) {
+            $display = FALSE;
+        }
+        if ( $display ) {
+            include('parsely-amp-tracking.php');
+        }
+    }
+
```
jtotoole: This whole function can be deleted. Let's instead make this the place in `wp-parsely.php` where we define our new `public function parsely_add_amp_actions()` and `public function parsely_add_amp_analytics($analytics)`

tests/test-wp-parsely.php
```diff
@@ -250,7 +250,19 @@ function test_http_canonicals()
...
-
+    
+    
+      
+    function test_amp_analytics() {  
+      	$options = get_option('parsely');
+        $post_array = $this->create_test_post_array();
+        $post = $this->factory->post->create($post_array);
+        $this->go_to('/?p=' . $post);
+        $ppage = self::$parsely->insert_parsely_amp_tracking();
```
jtotoole: Instead of generating the `ppage` object, let's change this line to `$output = self::$parsely->parsely_add_amp_actions();`, so that we can test just for the existence of the parsely amp analytics code

tests/test-wp-parsely.php
```diff
@@ -250,7 +250,19 @@ function test_http_canonicals()
...
+        $this->assertTrue(strpos($ppage['url'], '/amp') == 0);
+        $this->assertFalse(strpos($ppage['url'], '/amp'));
+        $this->assertTrue(strpos($ppage['url'], 'amp=1') == 0);
+        $this->assertFalse(strpos($ppage['url'], 'amp=1'));
```
jtotoole: We'll then remove these four assertions and replace them with the following two: 

```
$this->assertTrue(strpos($output, 'amp-analytics id="wp-parsely"') > 0);
$this->assertTrue(strpos($output, '{"vars":{"apikey":"blog.parsely.com"}}') > 0);
```

`blog.parsely.com` is the value given for `apikey` in the test stub [here](https://github.com/Parsely/wp-parsely/blob/master/tests/test-wp-parsely.php#L64)

jtotoole:
Hey @cayci! Thanks again for your work here. After pulling down the code and playing with it in my local WordPress environment with the AMP plugin installed, I realized we'll want to take a bit of a different approach, which I outline in the comments above. 

There's a lot of annoying switching around of lines and variable names, so if my feedback is confusing to follow, I can also just take the work you've done here and incorporate it in a new pull request. Let me know what would work best--you'll be due credit and thanks in any case!

jtotoole:
Closed via https://github.com/Parsely/wp-parsely/pull/50

PR #48, opened 2016-09-15
Feature/facebook instant analytics tracking

hbbtstar:
@jtotoole, check this out and let me know reviews / comments!

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -83,6 +83,7 @@ class Parsely {
         // inserting parsely code
         add_action('wp_head', array($this, 'insert_parsely_page'));
         add_action('wp_footer', array($this, 'insert_parsely_javascript'));
+        add_action('instant_articles_compat_registry_analytics', array($this, 'insert_parsely_tracking_fbia'));
     }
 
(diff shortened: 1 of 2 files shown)


wp-parsely.php
```diff
@@ -83,6 +83,7 @@ public function __construct() {
         // inserting parsely code
         add_action('wp_head', array($this, 'insert_parsely_page'));
         add_action('wp_footer', array($this, 'insert_parsely_javascript'));
+        add_action('instant_articles_compat_registry_analytics', array($this, 'insert_parsely_tracking_fbia'));
```
jtotoole: Do we need a conditional here to check whether a post is indeed being published in FBIA format? I'm wondering whether the present implementation will add this FBIA code + tracker on every page after having already inserted our standard code with `insert_parsely_page()` and `insert_parsely_javascript`

jtotoole: Scratch that! I didn't initially understand the mechanism there for hooking into `instant_articles_compat_registry_analytics`

jtotoole:
Looks great to me! Good looking out on that test.

PR #49, opened 2016-09-16
Enhancement/parsely page filter

hbbtstar:
This adds a filter for parselyPage so that publishers can modify it before it gets sent (allowing for changing canonical, headline, etc. also custom metadata)

cc @jtotoole

### DIFF
--- a/parsely-parsely-page.php
+++ b/parsely-parsely-page.php
@@ -4,6 +4,9 @@
    <script type="application/ld+json">
    <?php echo json_encode($parselyPage); ?>
    </script>
+    <?php if (isset($parselyPage["custom_metadata"])) : ?>
+        <meta name='parsely-metadata' content='<?php echo json_encode($parselyPage['custom_metadata'])?>'>
+        <?php endif; ?>
 <?php else: ?>
     <!-- parsleyPage is not defined / has no attributes.  What kind of page are you loading? -->
 <?php endif; ?>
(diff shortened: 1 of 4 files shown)


parsely-parsely-page.php
```diff
@@ -4,6 +4,9 @@
    <script type="application/ld+json">
    <?php echo json_encode($parselyPage); ?>
    </script>
+    <?php if (isset($parselyPage["custom_metadata"])) : ?>
+        <meta name='parsely-metadata' content='<?php echo json_encode($parselyPage['custom_metadata'])?>'>
+        <?php endif; ?>
```
jtotoole: Where are we exposing the option for them to set the custom metadata and to input the custom values?

hbbtstar: they'll do it via an add_action in their theme / plugin, like:

```
add_action('after_set_parsely_page', 'custom_metadata', 10. 3);

function custom_metadata($ppage, $post, $parselyOptions) {
$ppage['custom_metadata'] = get_custom_metadata_for_post($post);
return $ppage;
}
```

jtotoole: I see, and then `get_custom_metadata_for_post` would return values that they've set elsewhere by modifying their theme and adding inputs for those? Do you think it would be worth adding some documentation on this to explain how to configure this?

hbbtstar: yeah, exactly. Documentation is a good idea: it's a bog-standard filter, but we should at least point out it's an option, and what arguments it gets called with so people don't have to dig around to find it

PR #69, opened 2017-09-14
Bugfix: smarter detection for multisite

hbbtstar:
There was an issue people were running into where is_user_logged_in() would give a false positive for users who were logged into other wordpress.com domains (for example, is_user_logged_in() would return true for someone visiting a.wordpress.com if they were logged in to b.wordpress.com).

Not sure if this is a bug on WP's end or ours (this doesn't seem like expected behavior) but we can go ahead and make that check a little smarter: this checks to see if the user is actually a member of the current blog they're on instead of just a generic is_user_logged_in() call.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,16 +10,21 @@ branches:
     - master
 
 php:
-  - 5.3
   - 5.6
+  - 7.0
+  - 7.1
 
 env:
   - WP_VERSION=latest WP_MULTISITE=0
-  - WP_VERSION=4.0 WP_MULTISITE=0
+  - WP_VERSION=4.5 WP_MULTISITE=0
+  - WP_VERSION=latest WP_MULTISITE=1
+  - WP_VERSION=4.5 WP_MULTISITE=1
 
 matrix:
   include:
-    - php: 5.3
+    - php: 5.6
+    - php: 7.0
+    - php: 7.1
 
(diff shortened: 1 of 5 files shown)


msukmanowsky:
Thanks @hbbtstar! Anyway we can test this one out? I know you had troubles replicating.

hbbtstar:
@msukmanowsky I was able to reproduce this in the wild (logged into a test wordpress.com instance I created, then visited techcrunch, no pixel- logged out, pixel appeared again). I'm unfortunately not 100% if the fix above actually fixes anything: I added a test to make sure the pixel doesn't fire if the user is logged in to the blog using the new method, and that passes, so we at least don't have a regression. 

There are some function factories that indicate multiple blogs in the WP unit testing libraries; I'll see if I can reproduce this bug using those and, if so, add a test using them

hbbtstar:
@msukmanowsky @jtotoole I fussed with this a bit this morning and I think this is the best I can do: it uses wordpress' multisite testing libraries to make sure that our is_parsely_user_logged_in is expressly specific to one blog and no others and updated the travis files to test against it!

hbbtstar:
I can't get the PHP 7 tests to pass with some travis nonsense (phpunit complains with Wordpress versions under 4.8) but otherwise this is good to go

PR #74, opened 2017-10-31
Enhancement: track custom post types

hbbtstar:
Many publishers use custom post types to differentiate between different types of content. This PR enables publishers to select which kinds of posts and pages they want to track as posts vs. non-posts, giving them greater flexibility in their data collection!

Note that there are no new tests for this functionality: I am painfully aware there should be, but I can't figure out how to get register_post_type to work in tests. Given the timing of 1.11, I figured it was better to get the functionality out (I tested it manually) rather than be held up by a couple of minor tests for a few more days. If anybody wants to implement those as a separate PR that would be excellent!

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -52,7 +52,8 @@ class Parsely {
                                         'cats_as_tags' => false,
                                         'track_authenticated_users' => true,
                                         'lowercase_tags' => true,
-                                        'force_https_canonicals' => false);
+                                        'force_https_canonicals' => false,
+                                        'track_types' => ['post']);
     private $implementationOpts = array('standard' => 'Standard',
                                         'dom_free' => 'DOM-Free');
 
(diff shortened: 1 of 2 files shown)


wp-parsely.php
```diff
@@ -268,13 +270,49 @@ public function initialize_settings() {
                 'help_text' => $h,
                 'requires_recrawl' => true));
 
+        // Allow use of custom taxonomy to populate articleSection in parselyPage; defaults to category
+        $h = 'By default, Parsely only tracks the default post type as a post page. ' .
+                'If you want to track custom post types, select them here!<br>';
+        add_settings_field('track_post_types',
```
jtotoole: Wondering if we can be more explicit/less jargon-y here. Maybe say "tracks the 'default' WordPress post type as an article page, meaning that metadata values like title, author, etc. are displayed in the dashboard record. If you want to track custom WordPress post types as article pages as well..."

Thoughts? I feel like we can't assume too much Parse.ly knowledge for plugin users.

PR #75, opened 2017-11-27
Ensure modified date is never earlier than the creation date

joshuarrrr:
I decided to just add the logic in-place. Let me know if I should take a different approach.


hbbtstar:
makes sense to me :shipit:

PR #83, opened 2018-02-23
Only set ImageObject if we have a URL for image

davisshaver:
See https://github.com/Parsely/wp-parsely/issues/84

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -567,11 +567,13 @@ class Parsely {
 			);
 			$parsely_page['headline']         = $this->get_clean_parsely_page_value( get_the_title() );
 			$parsely_page['url']              = $this->get_current_url( 'post' );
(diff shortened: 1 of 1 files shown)


jtotoole:
Thanks @davisshaver! It's great to get a PR from the community. Please keep in touch with any additional feedback, and feel free to send more suggestions/PRs our way. We'll push this change up to WordPress in the next couple weeks with the next version of the plugin. 

Are y'all WordPress VIP customers, or do you use the standard plugin? We can push updates to the standard plugin immediately, but changing the VIP version generally takes longer, as it requires code review from the WordPress team that isn't always swift in coming. In any case, I'd suggest configuring your WordPress plugin settings to auto-update so you can get that update as soon as it's available.

davisshaver:
@jtotoole I'm using the standard plugin, so no worries there! Thanks for reviewing & merging.

PR #89, opened 2018-03-24
Add $author data when on author page

davisshaver:
Prevents error `Notice: Trying to get property of non-object`


jtotoole:
nice one, thx!

PR #91, opened 2018-04-04
Linter fixes

richeymichael:
A lot of this was adding comments. Where I really changed how things work is in wp_parsely.php

- Changed help texts by adding $help_link where needed and `sprintf`ing html & the url to get around the need to escape all output
- Changed the output of radio buttons, text fields, and selects to just spit out stuff when we had it instead of building up a variable to output to get around escaping all output
- Sanitize and check for existence of $_SERVER stuff, please look at that


class-parsely-recommended-widget.php
```diff
@@ -9,24 +34,32 @@ public function __construct() {
...
+	 * @subpackage Parse.ly
+	 * @param array $args Widget Arguments.
+	 * @param array $instance Values saved to the db.
+	 */
 	public function widget( $args, $instance ) {
 		$title = apply_filters( 'widget_title', $instance['title'] );
 
 		$instance['display_options'] = ! empty( $instance['display_options'] ) ? $instance['display_options'] : array();
-		echo $args['before_widget'] . $args['before_title'] . $title . $args['after_title'];
+		echo esc_html( $args['before_widget'] . $args['before_title'] . $title . $args['after_title'] );
```
jtotoole: Were you able to smoke-test the `/profile` and `/related` api calls in your local environment? They way to do so would be to enter "elevatedtoday.com" as your API key in the WordPress CMS settings page as well as the elevatedtoday.com API secret, enable the widget, and then check to see if it's rendering correctly with elevatedtoday posts and whether they API calls in the dev tools network tab look right

richeymichael: Good eye! This broke the widget, but I've got a solution now (that would have helped me much earlier if I had known about it).  For future reference wp_kses() is helpful with the escaping function messages from the linter.

class-parsely-recommended-widget.php
```diff
@@ -9,24 +34,34 @@ public function __construct() {
...
+	 * @param array $instance Values saved to the db.
+	 */
 	public function widget( $args, $instance ) {
 		$title = apply_filters( 'widget_title', $instance['title'] );
 
 		$instance['display_options'] = ! empty( $instance['display_options'] ) ? $instance['display_options'] : array();
-		echo $args['before_widget'] . $args['before_title'] . $title . $args['after_title'];
+		$allowed_tags                = wp_kses_allowed_html( 'post' );
+		$title_html                  = $args['before_widget'] . $args['before_title'] . $title . $args['after_title'];
+		echo wp_kses( $title_html, $allowed_tags );
```
joshuarrrr: For some reason this seems to be stripping the thumbnail images from the widget. But when I log `$allowed_tags`, it includes image elements and src attributes (which is the only attribute we use. So I haven't figured out yet why they're getting stripped.

wp-parsely.php
```diff
@@ -1,49 +1,66 @@
...
+ * Authors: Mike Sukmanowsky ( mike@parsely.com), Xand Lourenco ( xand@parsely.com ), James O'Toole (james.otoole@parsely.com )
+ *
+ * @package    WordPress
+ * @subpackage Parse.ly
+ */
 
-/* TODO List:
- * Wordpress Network support - going to hold off on any specific support here as content id prefix should work ok for now
+/*
+TODO List:
```
joshuarrrr: not necessarily for this release, but we should clean this up

class-parsely-recommended-widget.php
richeymichael: Weird, I was able to see the thumbnail images in the widget. I'll look at that.

joshuarrrr: hmm - maybe it's something in my environment. Maybe we can look at it together.

joshuarrrr: confirmed it was css-related handling of the no thumbnail setting

PR #102, opened 2018-05-21
makes amp integration optional, fixes #101

### LINKED ISSUE #101: Add setting/make amp integration optional
Some customers prefer to load trackers via tag managers (such as Google Tag Manager and Tealium), both on primary sites and on amp. While the settings currently provide an option for not loading the tracker (for sites that only want it for metadata), there's no way to opt out of the amp integration.

joshuarrrr:


joshuarrrr:
@hbbtstar added you as a reviewer, as I was reusing a lot of the code you added for the disable_javascript option. I decided to not even add the filter in case it's set, but I could be persuaded to move that elsewhere.

@richeymichael I'm sure this test could be more robust, so I'm open to testing what it really does vs this side effect.

PR #104, opened 2018-05-24
Release 1.13

richeymichael:
Merging 1.13 into master.


wp-parsely.php
```diff
@@ -1,14 +1,16 @@
 <?php
-/*
+/**
```
joshuarrrr: Is this intentional or for linting reasons? I wonder if it's what's causing the weird github highlights

PR #107, opened 2018-09-11
[DNM] Handle home pages that are posts pages

katjwinter:
Need to treat like archive pages instead of like a static page, because it loops through posts so we can't just grab the post information like we normally would, or else we'll just wind up with metadata for the first listed post and not the actual index page itself.

Addresses problems such as: https://github.com/Parsely/web/issues/8201

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -721,13 +721,29 @@ class Parsely {
 
 		$current_url = $this->get_current_url();
 
-		/**
-		 * Check if the post is a trackable post status.
-		 *
-		 * @param bool $is_public Flag if the post status is public.
-		 * @param WP_Post $post Post object.
-		 */
-		if (
+		if ( is_home() ) {
+			$parsely_page['@type'] = 'WebPage';
+			// If front page is set to display a static page, get the URL of the posts page.
+			if ( 'page' === get_option( 'show_on_front' ) ) {
+				$parsely_page['url'] = get_permalink( get_option( 'page_for_posts' ) );
+			} else {
+				$parsely_page['url'] = get_home_url();
+			}
+			$parsely_page['headline'] = $this->get_clean_parsely_page_value( get_bloginfo( 'name', 'raw' ) );
+		} elseif ( is_front_page() ) {
+			$parsely_page['headline'] = $this->get_clean_parsely_page_value( get_bloginfo( 'name', 'raw' ) );
+			$parsely_page['url']      = get_home_url();
+			$parsely_page['@type']    = 'WebPage';
+		} elseif ( is_archive() ) {
+			$parsely_page['@type'] = 'WebPage';
+			$parsely_page['url']   = $this->get_current_url();
+			if ( is_author() ) {
+				$author                   = get_user_by( 'slug', get_query_var( 'author_name' ) );
+				$parsely_page['headline'] = $this->get_clean_parsely_page_value( 'Author - ' . $author->display_name );
+			} else {
+				$parsely_page['headline'] = get_the_archive_title();
+			}
+		elseif (
 			in_array( get_post_type(), $parsely_options['track_post_types'], true )
 			&& apply_filters( 'wp_parsely_is_public_status', 'publish' === $post->post_status, $post )
 		) {
@@ -809,25 +825,9 @@ class Parsely {
 					'url'   => $parsely_options['logo'],
 				),
 			);
-
 		} elseif ( in_array( get_post_type(), $parsely_options['track_page_types'], true ) && 'publish' === $post->post_status ) {
 			$parsely_page['headline'] = $this->get_clean_parsely_page_value( get_the_title() );
 			$parsely_page['url']      = $this->get_current_url( 'post' );
-		}
-		if ( is_front_page() ) {
-			$parsely_page['headline'] = $this->get_clean_parsely_page_value( get_bloginfo( 'name', 'raw' ) );
-			$parsely_page['url']      = get_home_url();
-			$parsely_page['@type']    = 'WebPage';
-		}
-		if ( is_archive() ) {
-			$parsely_page['@type'] = 'WebPage';
-			$parsely_page['url']   = $this->get_current_url();
-			if ( is_author() ) {
-				$author                   = get_user_by( 'slug', get_query_var( 'author_name' ) );
-				$parsely_page['headline'] = $this->get_clean_parsely_page_value( 'Author - ' . $author->display_name );
-			} else {
-				$parsely_page['headline'] = get_the_archive_title();
-			}
 		}
 
 		$parsely_page = apply_filters( 'after_set_parsely_page', $parsely_page, $post, $parsely_options );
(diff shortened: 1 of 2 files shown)


wp-parsely.php
```diff
@@ -721,7 +721,29 @@ public function insert_parsely_page() {
 
 		$current_url = $this->get_current_url();
 
-		if ( in_array( get_post_type(), $parsely_options['track_post_types'], true ) && 'publish' === $post->post_status ) {
+		if ( is_home() ) {
```
joshuarrrr: good find!

wp-parsely.php
```diff
@@ -721,7 +721,29 @@ public function insert_parsely_page() {
...
+			$parsely_page['@type'] = 'WebPage';
+			// If front page is set to display a static page, get the URL of the posts page.
```
joshuarrrr: I'm having a hard time visualizing these two situations and the urls they will return. I'm sure I'm missing something obvious about why we can't just use `get_home_url` for both. Docstring examples would be nice, but test cases are probably preferable.

wp-parsely.php
```diff
@@ -721,7 +721,29 @@ public function insert_parsely_page() {
...
+				$parsely_page['url'] = get_permalink( get_option( 'page_for_posts' ) );
+			} else {
+				$parsely_page['url'] = get_home_url();
+			}
+			$parsely_page['headline'] = $this->get_clean_parsely_page_value( get_bloginfo( 'name', 'raw' ) );
+		} elseif ( is_front_page() ) {
+			$parsely_page['headline'] = $this->get_clean_parsely_page_value( get_bloginfo( 'name', 'raw' ) );
+			$parsely_page['url']      = get_home_url();
+			$parsely_page['@type']    = 'WebPage';
+		} elseif ( is_archive() ) {
```
joshuarrrr: I know we discussed re-ordering these. But now that I'm looking at it, I'm wondering if one of the reasons was to allow customers to override the defaults of whether we treat something as a page or a post. I guess I'm wondering if there are cases where `is_home`, `is_front_page`, or `is_archive` returns true AND it's also in the `track_post_types` array. :thinking:

joshuarrrr:
I know it's a lot more work, but I think we need to make test cases for this. I haven't written any phpunit tests before, so we can work through it together (or maybe enlist james next week).

wp-parsely.php
katjwinter: I'm not sure how to explain the situations, but what I can say is this:
for a page where it's using home.php as a posts page (in the Readings settings), and with us having code of:
```
// If front page is set to display a static page, get the URL of the posts page.
            if ( 'page' === get_option( 'show_on_front' ) ) {
                $parsely_page['url'] = get_permalink( get_option( 'page_for_posts' ) );
            } else {
                $parsely_page['url'] = get_home_url();
            }
```
in place, we get:
```
    {"@context":"http:\/\/schema.org","@type":"WebPage","url":"http:\/\/vccw.test\/front-page","headline":"Welcome to the VCCW"}
</script>
```
(note the url, which in this case is the *correct* url for the page we're on)

but if our code is using `get_home_url()` like:
```
$parsely_page['@type'] = 'WebPage';
            $parsely_page['url'] = get_home_url();
```
then the metadata we get is:
```
    {"@context":"http:\/\/schema.org","@type":"WebPage","url":"http:\/\/vccw.test","headline":"Welcome to the VCCW"}
</script>
```
(note the url, which in this case is *not* the correct url - it's just the base url of the home page, but not the page we're on that has all the posts).

I _think_ this has to do with using home.php as a template for a page that is listing posts, as opposed to operating as a static page, but basically it's something I googled because I wasn't getting the correct url, and this was the solution, and it seems to work?

Obviously this is exactly the kind of thing we need to put unit tests in place for, but this is an example anyway of why we can't just use `get_home_url` for both situations.

Does this make sense given your WP experience? I'm kind of fumbling in the dark, so if there's a better approach, I am totally open to it!

joshuarrrr: I _think_ I understand now. Tell me if this is sounds right: in the `is_home()` check, wordpress isn't really checking if you're on the homepage (website root); it's checking whether you're on a page generated by the `home.php` template. In contrast, `get_home_url()` always just returns the root url of the site.

Assuming that's right, I understand why ``get_home_url()` won't work. But now that makes me think functions like `get_home_url()` and `get_permalink( get_option( 'page_for_posts' ) )` are both equally dangerous, in that they rest on assumptions of relationships that may not be true (like that the homepage template will always _be_ the homepage). I think what we want is a more generic function that simply gets the url we're currently on.

A little bit of research led to `home_url( $wp->request )`, which, ugh, seems like a horrible hack.

katjwinter: @joshuarrrr I think you've got it right (and put it into words way better than I could, so thank you!). I agree, too, that those function calls seem suspect due to the assumptions they're making. Because indeed, in the case of millionmilesecrets.com, they're using home.php but not for the actual homepage, so a more generic function would be great... even if it's a horrible hack. I'd rather have a hack that works than something more "elegant" that doesn't!

joshuarrrr: yeah, mostly, I'm just annoyed that there seems to be no standard function that just gets the current link - like, c'mon wordpress!

joshuarrrr: If we wanted to be ultra thorough about it, I guess we could look at the source code for the official AMP plugin, which also needs to generate ld+json.

GaryJones:
To determine the root page of a site, I can offer the following, which is pulled/adapted from the Genesis theme (which has used it without complaint for nearly 6 years):

```php
/**
 * Check if the root page of the site is being viewed.
 *
 * `is_front_page()` returns false for the root page of a website when
 * - the WordPress "Front page displays" setting is set to "A static page"
 * - "Front page" is left undefined
 * - "Posts page" is assigned to an existing page
 *
 * This function checks for is_front_page() or the root page of the website
 * in this edge case.
 *
 * @return bool `true` if this is the root page of the site, `false` otherwise.
 */
function is_root_page() {

	return is_front_page() || ( is_home() && get_option( 'page_for_posts' ) && ! get_option( 'page_on_front' ) && ! get_queried_object() );

}
```

GaryJones:
https://www.parse.ly/help/integration/jsonld contains the list of required properties, and the list of accepted post (articles, reports, blog posts, etc.) or a non-post (**homepages**, index pages, section pages, etc.) types.

GaryJones:
I'm still writing the tests, but I think I've found the issue.

The `get_post_type()` on [this line](https://github.com/Parsely/wp-parsely/blob/269fd6a30520c29426e2e72295f6926e80730bf1/wp-parsely.php#L890) should be `get_post_type( $post )`, so that it's looked at the post object that has been passed into the method, and not whatever happens to have been set as the global `$post` (which for a Page for Posts, is indeed likely to be either the first Post, or the last Post, (and not the Page) depending on when the code runs.

With the existing code, `headline` and `url` properties are not set in the metadata. With the fix, they are. This issue would seem to apply to all regular Pages.

GaryJones:
Closing in favour of #152.

PR #109, opened 2018-09-19
Fix for infinite loop

emrikol:
`get_term_by()` can return `false` if the term does not exist, causing an infinite loop.


joshuarrrr:
@emrikol Thanks for the contribution! We'll want to write a test case for this before merging, but we'll try to get this fix in the next release.

joshuarrrr:
@emrikol Thanks again for spotting this! I was looking at the other bits of code that use this function, and it seemed like we needed to better handle the case where we return false. Here's what I came up with so far: https://github.com/Parsely/wp-parsely/pull/121

joshuarrrr:
merged to release candidate via #121

PR #110, opened 2018-09-26
check that api widget hasn't already been loaded

jtotoole:
fix for https://github.com/Parsely/web/issues/8240

### DIFF
--- a/class-parsely-recommended-widget.php
+++ b/class-parsely-recommended-widget.php
@@ -75,9 +75,7 @@ class Parsely_Recommended_Widget extends WP_Widget {
 					}
 				}
 
-
 				function widgetLoad() {
-
 					var parsely_results = [];
 
 					uuid = false;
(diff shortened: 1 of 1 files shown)


class-parsely-recommended-widget.php
```diff
@@ -103,6 +103,11 @@ function widgetLoad() {
 					if (parentDiv.length === 0) {
 						parentDiv = jQuery.find('.Parsely_Recommended_Widget');
 					}
+					
+					// make sure widget is not already on page
+					if (jQuery.find("div.parsely-recommendation-widget").length != 0) {
```
joshuarrrr: I think you actually want `parentDiv.find()` instead of `jQuery.find()`. Because we only care about not duplicating the widget within a parent div where it already exists. But it's totally ok (and expected by some customers) to have multiple instances of the widget in different locations on the page. `jQuery.find()` will do a global page search and just return the first element it finds.

PR #118, opened 2018-12-05
Dec 2018 vip updates

joshuarrrr:
Fixes https://github.com/Parsely/web/issues/8568

All tests still pass, but I'll still test-run it a bit with andy.


jtotoole:
looks pretty straightforward 👍

PR #119, opened 2018-12-07
Handle situations where expected data doesnt exist

harrisdavid:
We're using spoofed data from a range of sources, some where just post data is imported into WordPress, we don't create WordPress users for imported post authors, dealing with it by hooking into filters on the display layer. The updates here prevent the PHP notice errors "Trying to get property of non-object". Ideally we'll add in some filters which could short circuit these calls. I'll submit additonal request with those changes when we have time to make these updates.

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -1219,6 +1219,10 @@ class Parsely {
 	 * @param string $author The author of the post.
 	 */
 	private function get_author_name( $author ) {
+		//gracefully handle situation where no author is available
+		if( empty( $author ) || ! is_object( $author ) ){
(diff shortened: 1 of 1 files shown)


joshuarrrr:
@harrisdavid I made some minor linting changes, which you can see here: https://github.com/Parsely/wp-parsely/pull/125/commits/11a2756208786619dba8e1ea74bb92c6d9027576

I think those escape functions I added should be ok, because they're similarly in use in the following block - but let me know if you think they should be escaped differently.

Otherwise this is scheduled to be deployed with version 1.14 of the plugin.

PR #120, opened 2018-12-10
Add option to use original image or thumbnail in widget

joshuarrrr:
fixes https://github.com/Parsely/web/issues/8663#issuecomment-445863529

### DIFF
--- a/class-parsely-recommended-widget.php
+++ b/class-parsely-recommended-widget.php
@@ -190,12 +190,14 @@ class Parsely_Recommended_Widget extends WP_Widget {
 		$sort                = ! empty( $instance['sort'] ) ? $instance['sort'] : 'score';
 		$boost               = ! empty( $instance['boost'] ) ? $instance['boost'] : 'views';
 		$personalize_results = ! empty( $instance['personalize_results'] ) ? $instance['personalize_results'] : false;
+		$img_src             = ! empty( $instance['img_src'] ) ? $instance['img_src'] : 'parsely_thumb';
 
 		$instance['return_limit']        = $return_limit;
 		$instance['published_within']    = $published_within;
 		$instance['sort']                = $sort;
 		$instance['boost']               = $boost;
 		$instance['personalize_results'] = $personalize_results;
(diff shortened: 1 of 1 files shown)


joshuarrrr:
Sorry for the relatively large diff - it made sense to make "no image" an option in the `img_src` dropdown, which then meant it could be removed from the display options list, which meant _that_ no longer needed to be a list. Then I also noticed that the layout for widget items wasn't quite right/what I intended.

joshuarrrr:
also note that this settings format change means that this will be a (semi) breaking change, in that users who had display_options configured would need to update the new settings after upgrading. For this reason, it should definitely be a new minor version.

jtotoole:
looks good to me, nice work! only thing to consider is whether it makes sense to add a note regarding the `display_options` issue as part of the settings menu. maybe that's too much UI clutter and it should just be in the release notes, but definitely worth being careful when it comes to any potential reader-facing changes

joshuarrrr:
@jtotoole yeah, agreed that it's prudent to careful with breaking changes. :thinking: 

Maybe it makes sense to leave some of the checking code (for the previous settings) in place/use previous settings to fill the new values. That clutters up the codebase, but I think it's probably safest.

On the other hand any CSS/HTML changes to the widget itself are also potentially breaking...

jtotoole:
Would it make sense to instead set `no_image` via a checkbox, separate from the dropdown, that would default to `False` but would override the value from the `display_options` dropdown when set to `True`? That makes for a slightly less straightforward user experience, but would eliminate the need to make a new selection in that dropdown upon changing versions, and thereby the breaking change worry.

joshuarrrr:
@jtotoole I decided to properly migrate the old fields: see https://github.com/Parsely/wp-parsely/pull/120/commits/5de6743b636f54c91a3a8a826545c56477793820

The good thing with this approach is that it only really does anything the first time after update. Can you take a quick second look?

jtotoole:
ah that's really smart—yeah, seems like the right approach 👍

PR #122, opened 2018-12-14
Add a check to see if the post status is actually public

srtfisher:
- Check if the post status is public, not hard coding to `publish`.

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -727,9 +727,10 @@ class Parsely {
 		 * @param bool $is_public Flag if the post status is public.
 		 * @param WP_Post $post Post object.
 		 */
-		$is_public_post_status = apply_filters( 'wp_parsely_is_public_status', get_post_status_object( $post->post_status )->public, $post );
-
-		if ( in_array( get_post_type(), $parsely_options['track_post_types'], true ) && $is_public_post_status ) {
+		if (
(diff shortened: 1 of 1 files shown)


joshuarrrr:
@srtfisher Thanks for the PR! Can you provide a little more context on why you needed this change? I'm just trying to better understand the use case.

srtfisher:
@joshuarrrr the use case we're solving for here is for public post statuses that aren't `publish`. For example, we define a custom post status that works with our editorial workflow and Edit Flow. The post status is `'public' => true` and viewable on the front-end. However the Parse.ly schema isn't being included on the page because the `post_status` is not equal to `publish`.

joshuarrrr:
Perfect, thanks for the explanation - I think we'll need to maintain the existing hard-coded check for backwards compatibility, but agreed that a filter to override the post status makes sense.

srtfisher:
@joshuarrrr okay -- I pushed an update that will check the post status for `publish` specifically but will allow a filter to override that logic. Looks good?

joshuarrrr:
thanks! I'll merge it into the release candidate and take a look.

joshuarrrr:
@srtfisher OK, merged - this will go out as part of 1.14. If there are any other changes that would make it easier for Edit Flow users, feel free to open up issues - we're targeting a substantial re-write and update of this plugin for Q1 2019.

PR #129, opened 2019-02-11
Update to use config-free tracking

joshuarrrr:
Note that we're using this opportunity to automatically migrate most of our customers, so no optional support of the old method.

### DIFF
--- a/parsely-javascript.php
+++ b/parsely-javascript.php
@@ -54,23 +54,6 @@ if ( ! isset( $parsely_options['apikey'] ) || empty( $parsely_options['apikey'] 
 	</script>
 <?php endif; ?>
 
-<div id="parsely-root" style="display: none">
-	<div id="parsely-cfg" data-parsely-site="<?php echo esc_attr( $parsely_options['apikey'] ); ?>"></div>
-</div>
-<script data-cfasync="false">
-	(function (s, p, d) {
-		var h = d.location.protocol, i = p + "-" + s,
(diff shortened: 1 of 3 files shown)


tests/test-wp-parsely.php
```diff
@@ -193,24 +193,7 @@ public function setUp() {
 		);
 		update_option( 'parsely', $option_defaults );
 		self::$parsely_html = <<<PARSELYJS
-<div id="parsely-root" style="display: none">
-	<div id="parsely-cfg" data-parsely-site="blog.parsely.com"></div>
-</div>
-<script data-cfasync="false">
```
hbbtstar: we need to keep the data-cfasync thing, or cloudfront will kablammo our script

parsely-javascript.php
```diff
@@ -54,23 +54,6 @@ function uuidProfileCall() {
 	</script>
 <?php endif; ?>
 
-<div id="parsely-root" style="display: none">
-	<div id="parsely-cfg" data-parsely-site="<?php echo esc_attr( $parsely_options['apikey'] ); ?>"></div>
-</div>
-<script data-cfasync="false">
```
hbbtstar: same comment as in the test re: data-cfasync

tests/test-wp-parsely.php
joshuarrrr: ah, totally missed that, thanks for the catch.

joshuarrrr:
@hbbtstar can you take a final look to make sure I did what you meant?

PR #138, opened 2020-12-22
Un-nest nested functions

GaryJones:
Nested functions are generally a bad idea, especially when used within a constructor that will be called multiple times during tests.

This change un-nests the functions into their own methods, and updates the `add_*()` calls to be the correct callable.

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -112,20 +112,7 @@ class Parsely {
 			array( $this, 'add_plugin_meta_links' )
 		);
 
-		/**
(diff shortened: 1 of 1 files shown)


hbbtstar:
hi @GaryJones!

Thanks for the PRs, and my apologies that we've been a bit late to the party here! I'll be taking a look at these this week and into next, I just didn't want you to think I'd forgotten about them.

GaryJones:
@hbbtstar - you're welcome, and thanks for the acknowledgement :-)

PR #145, opened 2020-12-22
Remove unnecessary local variable

GaryJones:
Assigning a value to a variable, and then returning the variable uses more memory than just returning the value.


GaryJones:
The changes in the test file here are made redundant with #148. If #148 is merged first, I can rebase this PR to remove the redundant changes.

PR #152, opened 2021-03-07
Fix Home page metadata

### LINKED ISSUE #114: Don't alias '.../page' or '.../index' urls to homepage
See problem example here: https://www.thesun.co.uk/page/1392/

Metadata: 
```
{"@context":"http:\/\/schema.org","@type":"WebPage","headline":"The Sun","url":"https:\/\/www.thesun.co.uk"
````

`url` should be https://www.thesun.co.uk/page/1392/

Another offender: https://www.thesun.co.uk/index/s/page/8/

### PR DESCRIPTION
**Pull request by GaryJones**
## 🟢 Tests: Add structured data tests

Add the start of some structured data unit tests (only for posts and pages).

Check for required properties for posts and non-posts, and checks how various front page setting combinations are handled to ensure page_as_posts gives the correct metadata.

As part of this, I was able to identify #151 as a bug regarding the expected metadata URLs when pretty permalinks are not enabled.

## 🔧 Home page: Check is_front_page() earlier

Because the front page is a special circumstance for Pages, then the `is_front_page()` check needs to come before the check for regular page types and post types.

With the check now earlier, the headline and URL will correctly represent the home page of a site (site name, rather than Page title).

Resolves #107, resolves #114 

Also Handle misconfigured front page settings

When the settings are to show a page on the front, but not Page has been set, then `is_front_page()` returns false.

In terms of generating the Parse.ly metadata though, whatever is showing at the root of the site, should still show the site name and home URL.

## Before
Note the `home blog` `<body>` element classes, which indicates that a home page displays the latest posts. The metadata contains details about the a post on the page, instead of the page itself.

Screenshot 2021-03-07 at 04 32 25

## After

Same body classes, but now the metadata is for the page itself (which then displays data about the site)

Screenshot 2021-03-07 at 04 32 39

If a Static Page is set, then for a Homepage value of a page (see the updated body classes indicating this, but the metadata still showing about the page/site):

Screenshot 2021-03-07 at 04 37 36

### CODE REVIEW

**Comment by GaryJones** [comment]
Need to check that it also works for paged queries for the homepage as well - see https://github.com/Parsely/wp-parsely/issues/114#issuecomment-808766674.

**Comment by jblz** [comment]
@hbbtstar @mikeyarce can you please give this one a run through when you get a sec?

**Comment by mikeyarce** [comment]
> Need to check that it also works for paged queries for the homepage as well - see [#114 (comment)](https://github.com/Parsely/wp-parsely/issues/114#issuecomment-808766674).

This PR doesn't currently fix the problem with paged queries.  It currently outputs the `home_url()` for those as well instead of the paged URL.  

Another problem that I found was that if you have a static Home page and a separate Page for your posts, the posts page also uses the first blog post for the metadata.

Here are my proposed changes: 

**Change `is_front_page()` to `is_front_page() && ! is_paged()`**
This extra check would prevent the `home_url()` output if the current page is `paged`

**Add this conditional just below** 

```
} elseif ( is_front_page() && is_paged() ) {
			$parsely_page['headline'] = $this->get_clean_parsely_page_value( get_bloginfo( 'name', 'raw' ) );
			$parsely_page['url']      = home_url( add_query_arg( null, null ) );
		} elseif ( is_home() ) {
			$parsely_page['headline'] = get_the_title( get_option('page_for_posts', true) );
			$parsely_page['url']      = home_url( add_query_arg( null, null ) );
``` 

What this would do is check if it's the front page and is paged (thesun.co.uk/paged/1392), addressing #114.  This would also address the problem where it's a static page for posts.

## Summarizing

1. Metadata for home page when it's "Latest Posts" -> properly set to home_url ✅   
2. Metadata for home page when it's static page -> properly set to home_url ✅ 
3. Metadata for posts page when not home page -> properly set URL and title for `page_for_posts` ✅ 
4. Metadata for paged home of static posts -> properly uses paged URL ✅ 
5. Metadata for paged posts page with static home -> properly uses paged URL✅

File: wp-parsely.php [hunk 1]
```diff
@@ -817,7 +817,10 @@ public function construct_parsely_metadata( array $parsely_options, $post ) {
 			'@type'    => 'WebPage',
 		);
 		$current_url  = $this->get_current_url();
-		if ( in_array( get_post_type( $post ), $parsely_options['track_post_types'], true ) && 'publish' === $post->post_status ) {
+		if ( is_front_page() || ( 'page' === get_option( 'show_on_front' ) && ! get_option( 'page_on_front' ) ) ) {
```
mikeyarce, line 820:
```suggestion
		if ( is_front_page() ! is_paged() || ( 'page' === get_option( 'show_on_front' ) && ! get_option( 'page_on_front' ) ) ) {
```

wp-parsely.php
```diff
@@ -817,7 +817,10 @@ public function construct_parsely_metadata( array $parsely_options, $post ) {
...
+			$parsely_page['headline'] = $this->get_clean_parsely_page_value( get_bloginfo( 'name', 'raw' ) );
+			$parsely_page['url']      = home_url();
+		} elseif ( in_array( get_post_type( $post ), $parsely_options['track_post_types'], true ) && 'publish' === $post->post_status ) {
```
mikeyarce, line 823:
```suggestion
		} elseif ( is_front_page() && is_paged() ) {
			$parsely_page['headline'] = $this->get_clean_parsely_page_value( get_bloginfo( 'name', 'raw' ) );
			$parsely_page['url']      = home_url( add_query_arg( null, null ) );
		} elseif ( is_home() ) {
			$parsely_page['headline'] = get_the_title( get_option('page_for_posts', true) );
			$parsely_page['url']      = home_url( add_query_arg( null, null ) );
		} elseif ( in_array( get_post_type( $post ), $parsely_options['track_post_types'], true ) && 'publish' === $post->post_status ) {
```

wp-parsely.php
GaryJones: Missing an `&&`

mikeyarce: doh!

mikeyarce:
@GaryJones looking through the unit tests, I'm not totally understanding `test_home_page_for_posts`.
It looks like it creates a page, goes to the root of the site, and checks that the structured data for that previously created page.  However, since we're on the homepage (which is just the posts archive) it doesn't look at that pages data and just uses the `home_url` and site title for the metadata.  

Should we also set the newly created Page to the homepage for the unit test?   I feel like maybe that's what the intention here was.

```
update_option('show_on_front', 'page');
update_option( 'page_on_front', $page_id );
```

GaryJones:
@mikeyarce 

> Edit: I think this might be meant to check a homepage with the posts archive, in which case should we add a few posts instead of pages and make sure the structured data of the homepage is still correct? Or does the factory automatically do that?

The `setUp()` method sets the show_on_front = posts, so on review, it does look like the single page that's inserted via the factory is redundant there. 

I thought that as per any WP install, the Hello World post would be present, but if you want to add another post to be sure, and then leave the rest of the checks as they are (i.e. Test Blog is still the headline), then 👍🏻

mikeyarce:
> I thought that as per any WP install, the Hello World post would be present, but if you want to add another post to be sure, and then leave the rest of the checks as they are (i.e. Test Blog is still the headline), then 👍🏻

Thanks @GaryJones !  I did check and you are right, there is a default post already so it's all good 😄

jblz (approved):
This tests well for me.

I noticed phpcs caught a slew of minor whitespace issues... maybe worth running `phpcbf` on these files before landing.

mikeyarce:
Noting that this fixes #114

PR #153, opened 2021-03-07
Tests: Refresh scaffolded plugin test files

GaryJones:
WP-CLI has improved the code that is generated from `wp scaffold plugin-tests --dir=wp-content/plugins/wp-parsely` so this is a refresh of some files. Without these, local running of integration tests was giving an error.


mikeyarce:
Tested this out locally and I can still run tests without problems.  I also tested the `skip-database-creation` parameter and it worked as expected!

mikeyarce (approved):

PR #162, opened 2021-03-20
Improve widget settings UI

### LINKED ISSUE #159: Refresh widget settings
Here's a screenshot of the existing widget settings, and includes the native Recent Posts widget above it:

Screenshot 2021-03-15 at 10 53 11

Some observations and suggestions on how the Parsely widget settings could be improved:

1. Add the `widefat` class to the Title input to make it wider.
2. Add some units to the "Published Within" label field - it's not clear if that means minutes, hours, days or something else.
3. Change the wording of "Number of entries to return ( Max 20 )" to "Number of posts to show (max 20):" so it more closely matches the wording from the native widget.
4. Remove the line-breaks between the short number fields and the labels.
5. Use nicer looking names for dropdown (use capitals, avoid underscores, etc.) and radio button visible values.
6. Remove `width: 70%` style from Image Source field, since it cuts off part of the text.
7. Put checkboxes before their labels, not after, remove the line-breaks, and remove trailing `:`.

After the changes, it would then look like:

Screenshot 2021-03-15 at 11 10 29

GaryJones:
Fixes a range of small accessibility and usability issues. See individual commits for more info.

Fixes #159.

With this PR, it now looks like:
Screenshot 2021-03-20 at 12 12 45

### DIFF
--- a/class-parsely-recommended-widget.php
+++ b/class-parsely-recommended-widget.php
@@ -242,29 +242,7 @@ class Parsely_Recommended_Widget extends WP_Widget {
 		$instance['img_src']             = $img_src;
 		$instance['display_author']      = $display_author;
 
-		$boost_params = array(
-			'views',
-			'mobile_views',
-			'tablet_views',
-			'desktop_views',
-			'visitors',
-			'visitors_new',
-			'visitors_returning',
-			'engaged_minutes',
-			'avg_engaged',
-			'avg_engaged_new',
-			'avg_engaged_returning',
-			'social_interactions',
-			'fb_interactions',
-			'tw_interactions',
-			'li_interactions',
-			'pi_interactions',
-			'social_referrals',
-			'fb_referrals',
-			'tw_referrals',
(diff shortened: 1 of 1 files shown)


GaryJones:
I'd disagree with it needing a Design Review for the small changes here - its simply bringing the fields up to consistency with core widgets, and core already has plenty of designers looking at things like this :-)

class-parsely-recommended-widget.php
```diff
@@ -352,6 +329,32 @@ public function update( $new_instance, $old_instance ) {
 		$instance['img_src']             = trim( $new_instance['img_src'] );
 		return $instance;
 	}
+
+	private function get_boost_params() {
```
jblz, line 333: These labels are so much more usable!

jblz (approved):
I left a note about a whitespace readability concern on a long line that got longer, but not a blocker.

Great improvements!

GaryJones:
Added the line break.

PR #163, opened 2021-03-20
Widget: Move registration outside of widget file

### LINKED ISSUE #157: Widget: class file should not also register the widget class
It's a [good practice](https://www.php-fig.org/psr/psr-1/#1-overview) to that files should either declare symbols (classes, functions, constants, etc.) or cause side-effects (e.g. generate output, change .ini settings, etc.) but should not do both.

As such, the widget class should be the only thing in that file - the [registering callback and `add_action()`](https://github.com/Parsely/wp-parsely/blob/269fd6a30520c29426e2e72295f6926e80730bf1/class-parsely-recommended-widget.php#L364-L368) call should be moved to a different file.

GaryJones:
Fixes #157.


jblz (approved):
tests well for me. 🚀

PR #165, opened 2021-03-28
Add LICENSE file

GaryJones:
This is GPL v2, though it is actually licensed under `GPL-2.0-or-later` (via the `readme.txt` file).

WordPress itself is also `GPL-2.0-or-later` but includes the GPL v2 license text, so this is following standard behaviour.


jblz (approved):
lgtm.

PR #166, opened 2021-03-28
Add .gitattributes file

GaryJones:
This reduces the files that are included in exports for Composer, or when using a GitHub action to push to WordPress.org SVN.

There is no need to include tests and other development files and directories in production releases.

(There are a few files included here that do not yet exist in the repo, but they will do shortly.)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -17,9 +17,11 @@
 
 # Files
 /.editorconfig       export-ignore
+/.eslintrc           export-ignore
 /.gitattributes      export-ignore
 /.gitignore          export-ignore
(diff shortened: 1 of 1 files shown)


jblz (approved):
lgtm.  Should we consider including the top-level png files in this list to keep plugin size down? They are only sourced in the readme.

GaryJones:
> Should we consider including the top-level png files in this list to keep plugin size down? They are only sourced in the readme.

No, because those screenshot files will be moved to the `.wordpress-org` directory which is already included in this `.gitattributes` file. See #186.

PR #171, opened 2021-03-28
PHP: Specify minimum version

GaryJones:
Can be used for highlighting incompatible syntax from newer versions of PHP, and let's users know what versions of PHP the plugin works on.


jblz (approved):
Tests fine for me on PHP 5.6, 7.3, & 7.4 👍

Issue #172, opened 2021-03-28
Migrate Travis CI to to GitHub Actions workflow


GaryJones:
Since Travis-CI.org is shutting down in the near future, now would be a good time to move to GitHub Actions for the build steps.

As well as running tests, and checking code standards, this can also be used to [automate sending new releases to WordPress.org SVN](https://github.com/10up/action-wordpress-plugin-deploy) whenever a new tag is pushed.

GaryJones:
Moved the task of adding a new SVN-push workflow to #175.

PR #178, opened 2021-04-02
CI: Allow Composer validation to fail

GaryJones:
It was previously allowed to fail because there were some missing required fields. See annotation at https://github.com/Parsely/wp-parsely/actions/runs/698922725

These fields have since been added.

See #172.
See #165.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,6 @@ jobs:
       # Validate the composer.json file.
       # @link https://getcomposer.org/doc/03-cli.md#validate
       - name: Validate Composer installation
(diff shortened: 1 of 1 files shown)


jblz (approved):
lgtm, except, should the title be something like "Stop Allowing Composer Validation to Fail"?

GaryJones:
> should the title be something like "Stop Allowing Composer Validation to Fail"?

No - if we stop allowing it to fail, then it would never fail (which is what there is right now, out of necessaity).

By removing the `continue-on-error`, we allow it to fail, which is a good thing if there is indeed something wrong.

Issue #158, opened 2021-03-15
Widget: Move JavaScript to JS file


GaryJones:
All of [this JS](https://github.com/Parsely/wp-parsely/blob/269fd6a30520c29426e2e72295f6926e80730bf1/class-parsely-recommended-widget.php#L66-L172) should be moved to a cacheable `.js` file, enqueued when the widget is displayed, and the values from PHP provided to it via `wp_localize_script()`.

The JS object that the values become properties of, should include a hash to identify which widget instance it is for - this should allow multiple instances of the widget to be used, providing multiple sets of values from PHP to JS, but only needing one enqueued JS file.

Issue #134, opened 2020-07-28
Remove jQuery dependency


joshuarrrr:
We're not using it heavily, so it should be fairly straightforward to replace jQuery calls with vanilla JS.

jase-d-ace:
does it need to be removed from anywhere besides [here](https://github.com/Parsely/wp-parsely/blob/9fb6245ae7e0ba6bde41df892faa55958e48ab1b/parsely-javascript.php#L28)?

joshuarrrr:
Yeah, `jQuery.find()` is also used: https://github.com/Parsely/wp-parsely/search?q=jQuery&unscoped_q=jQuery

jblz:
#236 got us a lot of the way there, but the recommended widget still relies on jQuery:

https://github.com/Parsely/wp-parsely/blob/3cca62a4c228af6bd8893e744e710fb6b4269e94/src/class-parsely-recommended-widget.php#L95-L116

pauarge:
jQuery has been removed

PR #183, opened 2021-04-03
Simple JS WP build env + prettier & eslint

jblz:
This is a step toward a more robust JS environment for the front end and admin (ES6+ syntax, importing npm packages, etc.).

It leverages the [`wp-scripts` package](https://www.npmjs.com/package/@wordpress/scripts) so we don't need a custom webpack config (yet).

Additionally, this includes prettier and eslint configurations to enable IDE auto-formatters.

Right now, the src/index.js file is only a placeholder for the entrypoint. We'll want to replace it with an actual implementation prior to sourcing the built file in the plugin :)

## To Test

`npm i` to install dev dependencies.

### Development

* `npm run start`
* Make changes to `src/index.js`, add and link other source files (using `import _ from '__'` format, etc.)
* Source changes should automatically generate a new development `build/index.js`

### Release

* `npm run build`
* It should generate a new "production" `build/index.js`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,29 @@
+{
+  "name": "wp-parsely",
+  "version": "2.4.0",
+  "description": "The official WordPress plugin for Parse.ly - makes it a snap to add the required tracking code to enable Parse.ly on your WordPress site.",
+  "main": "build/index.js",
+  "directories": {
+    "test": "tests"
+  },
+  "scripts": {
+    "build": "wp-scripts build",
+    "start": "wp-scripts start",
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/Parsely/wp-parsely.git"
+  },
+  "author": "",
+  "license": "GPL-2.0-only",
+  "bugs": {
+    "url": "https://github.com/Parsely/wp-parsely/issues"
+  },
+  "homepage": "https://github.com/Parsely/wp-parsely#readme",
+  "devDependencies": {
+    "@wordpress/eslint-plugin": "^9.0.1",
+    "@wordpress/scripts": "14.0.1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -0,0 +1,29 @@
...
+    "build": "wp-scripts build",
+    "start": "wp-scripts start",
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/Parsely/wp-parsely.git"
+  },
+  "author": "",
+  "license": "GPL-2.0-only",
```
GaryJones, line 19: This doesn't match what the plugins was already licensed under, which is GPL-2.0-or-later, which is what WP core uses. Keeping it to only 2.0 would technically stop someone using it with a plugin that is only GPL-3.0. 
```suggestion
  "license": "GPL-2.0-or-later",
```

GaryJones (changes requested):
Incorrect license SPDX tag spotted. 

As a general question, the existing plugin obviously has a little bit of JS, but is the plan to introduce a lot more?

jblz:
Thanks for taking a look @GaryJones.  Fixed up the license.

> is the plan to introduce a lot more (js)?

I don't think it'll be _a lot_ off the bat, but as I started tackling issues like #158 & #134, I really missed the ability to write modern JS.

In particular, when we extract the widget script from PHP (#158) it's an opportune time to refactor as we'll be pulling a bunch values from PHP `echo`s into variables.

Further, modularizing the front end code enables us to cover it with tests more readily and things like destructuring assignments (particularly in function arguments) are just so nice to be able to use 😄.

I think having a build step in place will enable feature development in the future as well, so I think it's a pretty good time to make the jump.  Happy to hear counter arguments, of course!

GaryJones (approved):
👍🏻

PR #185, opened 2021-04-04
Add CODEOWNERS file

### LINKED ISSUE #184: Add .github/CODEOWNERS file
This will help to streamline who should be pinged for reviews. I've created a team, @Parsely/wp-parsely, so the CODEOWNERS file would need to contain something like

```
* @Parsely/wp-parsely 
```

Once the code is a little more restructured, then per-directory assignments may be possible, but the plugin isn't yet big enough to warrant it.

GaryJones:
See https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners

Closes #184.


jblz (approved):

jblz:
Thanks!

Issue #155, opened 2021-03-09
Category pages, 404 pages incorrectly displaying post metadata


arhine:
The plugin will often return metadata from the first post on the page rather than the metas associated with the page itself. This causes incorrect data to be tracked in the dashboard for both the page as well as the post providing the incorrect metadata.

This behavior is also occurring on some 404 pages, which ideally would not be tracked at all.

This is similar to the behavior for home pages being addressed in https://github.com/Parsely/wp-parsely/pull/152.

Current examples:

* Category page: https://www.eatthis.com/weight-loss/
* 404 page: https://techcrunch.com/my_account

GaryJones:
> 404 page: techcrunch.com/my_account

Techcrunch are running version 2.0 of the plugin; the fix for metadata being added for 404 pages, was [fixed](https://github.com/Parsely/wp-parsely/commit/924cc90be0a239b10705e64c25d8c559ef4e8739) in the 2.2 release.

jblz:
Fixed in #192

PR #192, opened 2021-04-06
Fix category metadata

hbbtstar:
This is branched off of #152 and that should be merged first before this one is merged. WIP because it has no tests yet!

Similar situation to #152, but for index pages instead of the home page: this swaps the order of conditionals to make sure that we check to see if a page is any of the index types first (author, tag, category, etc.) before outputting metadata.

### DIFF
--- a/tests/structured-data/archive-post-test.php
+++ b/tests/structured-data/archive-post-test.php
@@ -99,9 +99,40 @@ final class Archive_Post_Test extends TestCase {
 		self::assertEquals( get_permalink( $page_id ) . 'page/2', $structured_data['url'] );
 	}
 
+	public function test_term_archive() {
+		// Set permalinks, as Parsely currently strips ?page_id=... from the URL property.
+		// See https://github.com/Parsely/wp-parsely/issues/151
+		$this->set_permalink_structure( '/%postname%/' );
+
+		// Setup Parsley object.
+		$parsely         = new \Parsely();
+		$parsely_options = get_option( \Parsely::OPTIONS_KEY );
+
+		// Insert a single category term, and a Post with that category.
+		$category = self::factory()->category->create( [ 'name' => 'Test Category' ] );
+		self::factory()->post->create( [ 'post_category' => [ $category ] ] );
+
+		// Make a request to that page to set the global $wp_query object.
+		$cat_link = get_category_link( $category );
+		$this->go_to( $cat_link );
+
+		// Reset permalinks to default.
+		$this->set_permalink_structure( '' );
+
+		// Create the structured data for that category.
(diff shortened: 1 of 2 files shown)


jblz:
Should this have a "Fixes #155" in the description?

..or is the 404 part a separate issue that needs fixing?

mikeyarce:
Looks good to me!  Thanks for beautifying my code 😄

jblz:
I think I correctly solved all the merge conflicts, but let's give this one a solid review including the use cases in #152 to make sure no reversions were introduced.

GaryJones:
Marking this as Closes #155 as the [metadata fix for `is_404()`](https://github.com/Parsely/wp-parsely/commit/924cc90be0a239b10705e64c25d8c559ef4e8739) was added in `2.2`.

tests/structured-data/archive-post-test.php
```diff
@@ -99,17 +99,79 @@ public function test_blog_page_for_posts_paged() {
...
+		// Set permalinks, as Parsely currently strips ?page_id=... from the URL property.
+		// See https://github.com/Parsely/wp-parsely/issues/151
+		$this->set_permalink_structure( '/%postname%/' );
+
+		// Setup Parsley object.
+		$parsely         = new \Parsely();
+		$parsely_options = get_option( \Parsely::OPTIONS_KEY );
+
+		// Insert a single user, and a Post assigned to them.
+		$user = self::factory()->user->create( [ 'user_login' => 'parsely' ] );
```
jblz, line 112: Is there a reason this is using a static method call instead of `$this->factory`?  Seems like other usages are all the latter.  There's a few other places in this change I see `self::factory()` as well.

GaryJones: `self::factory()->...` is what WP core uses. See https://core.trac.wordpress.org/browser/trunk/tests/phpunit/tests/query/taxQuery.php

Just like `$this->assert...` and `self::assert...` (see "A comment question" paragraph at the top of https://phpunit.readthedocs.io/en/9.5/assertions.html) there doesn't seem to be any difference - but to me, it helps signify that it's a utility method being called, rather than a protected / private method in the test class itself.

GaryJones: Besides - the actual `factory()` method on the WP unit test case, is static: https://core.trac.wordpress.org/browser/trunk/tests/phpunit/includes/abstract-testcase.php#L41

jblz: Cool.  Definitely makes sense to me to standardize on the static syntax.

jblz (approved):
Tests good for me!

Issue #156, opened 2021-03-15
Widget: Re-think approach


GaryJones:
Right now, the widget class is [included](https://github.com/Parsely/wp-parsely/blob/269fd6a30520c29426e2e72295f6926e80730bf1/wp-parsely.php#L1719) and [registered](https://github.com/Parsely/wp-parsely/blob/269fd6a30520c29426e2e72295f6926e80730bf1/class-parsely-recommended-widget.php#L364-L368) regardless of whether any API key and API secret has been registered.

There is a [check](https://github.com/Parsely/wp-parsely/blob/269fd6a30520c29426e2e72295f6926e80730bf1/class-parsely-recommended-widget.php#L52) for these before outputting the widget front-end content, and it then [shows a message](https://github.com/Parsely/wp-parsely/blob/269fd6a30520c29426e2e72295f6926e80730bf1/class-parsely-recommended-widget.php#L177) to all visitors, which isn't a great look (doesn't event start with a capital). Visitors can't change add keys or secrets, so there is no point showing the message to them.

I'd like to suggest one of the following two approaches are used:

1. Check for the existence of the saved API key and API secret before even registering the widget - so if one or both is missing, there is no front-end output (which would be blank anyway), and no back-end widget form.
2. Allow the widget to always be registered, but update the widget class methods so that the message is shown in the back-end form if an API key or secret is missing, and if so, show nothing on the front-end.

Which approach is used should take into at least two factors - what the Support burden will be for "I can't find the widget" (approach 1), and what happens to widget settings if a widget becomes unregistered (add key and secret -> add widget -> save widget settings -> remove key or secret).

PR #193, opened 2021-04-06
Widget: Hide the recommendation widget until the element is populated

jblz:
On my test site, I noticed that the Recommendations widget still shows even though no results are present to fill it.  The `getJSON` call (predictably) fails as this site is not registered in the system resulting in an empty div with my configured Title:

Screen Shot 2021-04-06 at 4 45 28 PM

This change defaults the widget's outermost div to `display: none;` and removes the pertinent class once the script has populated the div with content.

I'm sure this is an edge case that most customers will never hit, but, in the event of a failed API call, it'd probably be better to not show an empty widget.


GaryJones (dismissed):
As per #156, I think we can do an even better approach, which is to show nothing on the front-end (like, no output at all), if the API details aren't populated in the options, and also change the widget form to a simple "Please head to the settings page and enter your API credentials"-type of message; that way, there's no expectation that the widget should be working but nothing appearing on the front-end.

I'm not keen on approach that visually hides the div with extra CSS (on every page that needs it) just for this edge case; I'd rather have less code sent to the browser to render when it's never going to work, not more.

Edit: Just realised that this was for a failed API call, not a lack of API credentials. Need to re-think my above response for this context.

GaryJones (commented):
I've rescinded my previous Request Changes review, as I now understand better what the goal of this PR is. `#156` is a different issue.

Some thoughts:

 - the added class is `wpparsely_hidden`, but all of the other classes in the CSS file start with `parsely-recommended-widget` - so consider `.parsely-recommended-widget-hidden {...}` or something like `.parsely-recommended-widget.is-hidden {...}` (a la [SMACSS](http://smacss.com/book/type-state), and then remove `is-hidden`) to keep more of a consistency.
 - If the JSON arrival and rendering takes a second or so, would this "hide and then show" affect the layout shift? Next month Google makes CLS and other [Core Web Vitals](https://web.dev/vitals/) a ranking factor. Maybe it doesn't make a difference, unless with the new behaviour the whole container is hidden that was previously retaining space to avoid such a dramatic shift.

jblz:
> If the JSON arrival and rendering takes a second or so, would this "hide and then show" affect the layout shift? Next month Google makes CLS and other Core Web Vitals a ranking factor. Maybe it doesn't make a difference, unless with the new behaviour the whole container is hidden that was previously retaining space to avoid such a dramatic shift.

This is a good point...

Perhaps we should have some fallback items that fill this space that we enable users to override or something.

> As per #156, I think we can do an even better approach, which is to show nothing on the front-end

Yeah, I definitely think we can do better than this during the effort to re-conceptualize the widget.  I mostly just noticed and wanted to make sure the issue was captured somewhere and I'd already spent the 5 minutes to figure out what was going on so flushed to a PR.

Let's go ahead and leave this on hold for now pending the other work, I think.

pauarge:
I've migrated @jblz's changes to the new project structure (without jQuery and in a separate JS file).

pauarge (approved):

GaryJones (approved):

PR #194, opened 2021-04-06
Changelog for 2.3.1

GaryJones:
Pre-emptive change log for https://github.com/Parsely/wp-parsely/pull/152 and https://github.com/Parsely/wp-parsely/pull/192 which are planned to go out as a patch release.

As the `CHANGELOG.md` didn't exist in the 2.3.0 tag, then we can't point this PR to the `release/2.3.1` branch without adding the whole file, which doesn't seem quite right. So left pointing to `develop`, and it can be merged after the `release/2.3.1` branch is merged into `master`.

It means that the 2.3.1 release won't contain a `CHANGELOG.md`, but then neither did any version before, and the details can be added to the GitHub Release anyway. 

This is only an oddity because we're doing a hotfix release immediately after the `CHANGELOG.md` file was added; in future releases (2.4.0 and beyond), then a separate PR like this can be based on the `release/x.y.z` branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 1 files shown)


hbbtstar (approved):
LGTM!

PR #195, opened 2021-04-08
Adjust plugins headers

GaryJones:
- Use a format which supports both WordPress plugin headers requirements, and phpDoc file-level DocBlock format.
- Add a text-domain header to allow translations on translate.wordpress.org.
- Change the plugin URL to be specific to the WordPress integration help docs.
- Add a header to allow updating directly from GitHub via the https://github.com/afragen/git-updater plugin.
- Clarify the license as being GPL v2 _or later_, which matches WordPress core.
- Add "Requires PHP" header, which has no real effect as it's currently the same low level as WordPress core, but it clarifies that a higher version of PHP is not needed.
- Set the author as Parse.ly and author URL as Parse.ly website.
- Removed To-Do items, as there are better places to put these (GitHub issues).

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -1,43 +1,26 @@
 <?php
 /**
- * Plugin Name: Parse.ly
- * Plugin URI: http://www.parsely.com/
- * Description: This plugin makes it a snap to add Parse.ly tracking code to your WordPress blog.
- * Author: Mike Sukmanowsky ( mike@parsely.com )
- * Version: 2.3
- * Requires at least: 4.0.0
- * Author: Parse.ly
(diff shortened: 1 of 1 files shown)


mikeyarce:
Thanks @GaryJones !  This looks great 👍🏽 
I think everything is good as long as we've got the `Parse.ly` WordPress.org user in place.

mikeyarce (approved):

GaryJones:
> I think everything is good as long as we've got the `Parse.ly` WordPress.org user in place.

That would be something we update in the README; the Author name in the plugin headers are just a plain string visible on the wp-admin plugins page.

PR #202, opened 2021-04-11
Don't give metadata output for is_search()

### LINKED ISSUE #201: Apply correct metadata settings for is_search()
There is currently no logic for constructed metadata, for search results pages (`is_search()`). The current output is:

```html
<!-- BEGIN wp-parsely Plugin Version 2.3 -->
	<meta name="wp-parsely_version" id="wp-parsely_version" content="2.3"/>
	<!-- parsleyPage is not defined / has no attributes.  What kind of page are you loading? -->
	<!-- END wp-parsely Plugin Version 2.3 -->
```

If nothing else, the comment should be tided up. But what is the ideal tracking state here @hbbtstar? Should it not be tracked? If it should, what headline should it be? What schema type? https://schema.org/SearchResultsPage ?

GaryJones:
There is logic on the Parse.ly end to exclude search results pages when crawling, so it doesn't make sense to have the request for those pages going through the logic of potentially constructing meta data only to output a redundant meta tag and some HTML comments.
Only for posts and pages.

Closes #201.

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -776,8 +776,19 @@ class Parsely {
 	public function insert_parsely_page() {
 		$parsely_options = $this->get_options();
 
-		// If we don't have an API key or if we aren't supposed to show to logged in users, there's no need to proceed.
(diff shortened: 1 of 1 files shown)


jblz (approved):
lgtm

PR #203, opened 2021-04-11
Widget: allow no boost to Recommendation widget

### LINKED ISSUE #161: Widget: Support sorting by score only, no boost
The [API Docs](https://www.parse.ly/help/api/recommendations#get-related) say that the default value for `boost` is undefined (ie. score only). 

However, the dropdown field in the widget has no option to choose this; so there's always a boost value.

We should add a new "No boost (Score only)" option to the top of the Boost dropdown, and adjust the generation of the URL so that the `boost` parameter is only included if needed (it is an optional parameter).

GaryJones:
To get to the point of adding a dropdown field value, I had to make a few other changes along the way. Each item corresponds to a commit.

## Extract recommendation API URL
The generation of the URL used for the API call (technically it is passed to JavaScript where it gets amended before actual use) was rolled up in the `widget()` method, so I did the bare minimum to extract this into it's own method. Added a couple of basic tests to cover the existing behaviour.

## Refactor construction of API URL

Collected keys and values in an array, and let WordPress construct the URL. No change in tests - that's why the basic tests were written first!

## Make `boost` arg conditional in API URL

As per [the docs](parse.ly/help/api/recommendations#get-related), the `boost` parameter is only usable when the `sort=score`. When it is `sort=pub_date`, then it is redundant, so there is no need to include it.

Its presence is now conditional on `sort=score`.

## Allow no boost to scores

The [API Docs](https://www.parse.ly/help/api/recommendations#get-related) also say that the default value for `boost` is undefined (ie. score only).

However, the dropdown field in the widget had no option to choose this; so there was always a boost value.

A "No boost" option has now been added, and the `boost` parameter is not included if it is set to this.

There is no change to the default value of this field - it is still `views` (Page views).

Existing sites would also not be affected, as their existing widget settings would continue to be used, regardless of what new options there now were.

Closes #161.


jblz (approved):
This looks good to me.  Thanks for the nice refactor and added test coverage :)

I fixed the merge conflict in 30c3ee4... please check to make sure I fully captured your intent.

PR #204, opened 2021-04-11
Add $post_id to wp_parsely_permalink filter hook

### LINKED ISSUE #168: Make $post_id available to wp_parsely_permalink filter
The `wp_parsely_permalink` filter currently only provides the existing permalink to be filtered, and a `$post`, which appears to be either "post" or "nonpost" strings.

Much more useful would be to also provide the `$post_id`, which is the ID of the specific post that the `get_current_url()` is working with.

Remember to update the filter DocBlock when adding it.

GaryJones:
The `wp_parsely_permalink` filter only provided the existing permalink to be filtered, and a $post variable, which is really just a "post" or "nonpost" string.

We now provide the $post_id, which is the ID of the specific post that the `get_current_url()` is working with. If the `$post_id` was not provided to the parent function, then the value will be 0, and the permalink will be for the `global $post`.

Closes #168.

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -1606,11 +1606,14 @@ class Parsely {
 			 * Filters the list of author names for a post.
 			 *
 			 * @since 1.14.0
+			 * @since 2.5.0  Added $post_id.
 			 *
 			 * @param string $permalink The permalink URL or false if post does not exist.
 			 * @param string $post      Post object type group ("post" or "nonpost").
+			 * @param int    $post_id   ID of the post you want to get the URL for. May be 0, so $permalink will be
+			 *                          for the global $post.
 			 */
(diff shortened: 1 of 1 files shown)


jblz (approved):
Nice enhancement.

The naming of the `$post` argument is certainly unfortunate. Since it's only positional (i.e. not a named param) & a private methoc renaming it should not break anything.  What do you think about changing it to something more semantically-clear in a follow up change?

It seems like it's treated like a boolean, so maybe we rename it `$is_post`, or if we need the flexibility of a string, maybe something like `$parsely_content_type`.

/shrug

GaryJones:
I like `$parsely_content_type`.

As of PHP 8, then someone (outside VIP) _could_ already be named arguments, but as you said, this is a private method, so they can't even access via a child class. Changing that method parameter name is outside the scope of this "add arg to hook" PR though.

PR #205, opened 2021-04-11
Widget: Replace form and hide front-end output when API settings are missing

### LINKED ISSUE #156: Widget: Re-think approach

GaryJones:
The widget requires the "Parse.ly Site ID" (API key) and "Parse.ly API Secret" to be populated.

This PR contains two commits - one that replaces the widget form in the back-end, and one that removes the content in the front-end - when the settings are not populated.

_The second commit unwraps a large block of code from a conditional, so it is advisable to ignore whitespace changes when reviewing (ensure `?w=1` is added to the files view URL)._

## Hide form when settings are missing

If either field were not populated, the widget form would still display the regular widget settings, even though there was zero chance of the widget working correctly.

This has now switched to checking for the settings, and giving a message to the user with a link to the settings page.

Here's what it now looks like when the API key or secret setting is missing:

Screenshot of new widget with replaced form

This draws the user's attention to the fact they need to populate the two settings.

## Hide front-end when settings are missing.

Prior to this PR, the widget would still display the `before_widget`, `before_title`, `title`, `after_title`, and `after_widget` content, even though the widget body would never show any content.

This has now switched to checking for the settings, and displaying nothing at all if the settings are missing.

The prior art here is the WP core RSS Widget, which [returns early](https://github.com/WordPress/WordPress/blob/270f2011f8ec7265c3f4ddce39c77ef5b496ed1c/wp-includes/widgets/class-wp-widget-rss.php#L55-L57) if the URL setting is not populated.

With no output, it technically does change the front-end of the website - previously the partial widget output acted as a placeholder:

Screenshot 2021-04-11 at 23 30 39

Now, it doesn't, so subsequent widgets get shifted.

Screenshot 2021-04-11 at 23 30 51

Given the rarity of this, since it's only likely to be when setting the plugin up, and the widget form draws attention to the fact that required settings are missing, I think this lack of placeholder is an acceptable situation; a completely missing output hints at a different situation (not set up correctly), than having no data showing in the widget body (is my key correct? is Parse.ly API broken?) (which will be handled by #193).

This second commit is optional though, and could be adapted to leave a HTML comment in the output, or even a similar visible message to what the form shows, though this would show up as soon as someone added the widget to the widget area, before they then had a chance to see the form message and populate the settings.

Closes #156.

### DIFF
--- a/class-parsely-recommended-widget.php
+++ b/class-parsely-recommended-widget.php
@@ -222,6 +222,19 @@ class Parsely_Recommended_Widget extends WP_Widget {
 	public function form( $instance ) {
 		$this->migrate_old_fields( $instance );
 
+		if ( ! $this->api_key_and_secret_are_populated() ) {
+			$settings_page_url = add_query_arg( 'page', 'parsely', get_admin_url() . 'options-general.php' );
+
+			$message = sprintf(
+				__( 'The <i>Parse.ly Site ID</i> and <i>Parse.ly API Secret</i> fields need to be populated on the <a href="%s">Parse.ly settings page</a> for this widget to work.', 'wp-parsely' ),
+				esc_url( $settings_page_url )
+			);
+
+			echo '<p>', wp_kses_post( $message ), '</p>';
+
+			return;
+		}
+
 		// editable fields: title.
 		$title               = ! empty( $instance['title'] ) ? $instance['title'] : '';
 		$return_limit        = ! empty( $instance['return_limit'] ) ? $instance['return_limit'] : 5;
@@ -358,5 +371,26 @@ class Parsely_Recommended_Widget extends WP_Widget {
 			'li_referrals'          => __( 'Page views where the referrer was linkedin.com', 'wp-parsely' ),
 			'pi_referrals'          => __( 'Page views where the referrer was pinterest.com', 'wp-parsely' ),
 		);
+	}
+
+	private function api_key_and_secret_are_populated() {
+		$options = get_option( 'parsely' );
+
+		// No options are saved, so API key is not available.
+		if ( ! is_array( $options ) ) {
+			return false;
+		}
+
+		// Parse.ly Site ID settings field is not populated.
+		if ( ! array_key_exists( 'apikey', $options ) || $options['apikey'] === '' ) {
+			return false;
+		}
+
+		// Parse.ly API Secret settings field is not populated.
+		if ( ! array_key_exists( 'api_secret', $options ) || $options['api_secret'] === '' ) {
+			return false;
+		}
+
+		return true;
 	}
 }
--- a/class-parsely-recommended-widget.php
+++ b/class-parsely-recommended-widget.php
@@ -44,6 +44,10 @@ class Parsely_Recommended_Widget extends WP_Widget {
 	 * @param array $instance Values saved to the db.
 	 */
 	public function widget( $args, $instance ) {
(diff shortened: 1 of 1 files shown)


class-parsely-recommended-widget.php
```diff
@@ -359,4 +365,25 @@ private function get_boost_params() {
...
+
+		// Parse.ly Site ID settings field is not populated.
+		if ( ! array_key_exists( 'apikey', $options ) || $options['apikey'] === '' ) {
+			return false;
+		}
+
+		// Parse.ly API Secret settings field is not populated.
+		if ( ! array_key_exists( 'api_secret', $options ) || $options['api_secret'] === '' ) {
+			return false;
+		}
```
jblz, line 385: Unless we need the specificity of checking for an empty string, we can simplify this by consulting the `empty` language construct & save a few bytes:

```suggestion
		if ( empty( $options['apikey'] ) ) {
			return false;
		}

		// Parse.ly API Secret settings field is not populated.
		if ( empty( $options['api_secret'] ) ) {
			return false;
		}
```

Even with error reporting set to `E_ALL` it won't generate a warning or notice if the variable or key is not set.

GaryJones: I would favour clarity of intent with the code, than saving a few bytes.

https://beberlei.de/2021/02/19/when_to_use_empty_in_php_i_say_never.html

jblz: Agree there are lots of times when empty is problematic, but IMO this is one of those cases where it's suited (checking that the variable & key are set and the value isn't empty. We're not overly concerned w/ data types as it's decidedly a string (stored in options, [validated as string](https://github.com/Parsely/wp-parsely/blob/1a398fac2ebc4935199299977522c5a817b62505/wp-parsely.php#L596)) ).

That said, I don't feel strongly about the suggestion, so feel free to leave as is if you feel it adds clarity or safety.

jblz (approved):
This is working well!  Looks like it should be a clean rebase. 👍

PR #207, opened 2021-04-12
Tests: Call the TestCase factory method statically

### LINKED ISSUE #208: Tests: Standardize on static access for factory() and assertions
See https://github.com/Parsely/wp-parsely/pull/192#discussion_r611664371

> Is there a reason this is using a static method call instead of $this->factory? Seems like other usages are all the latter. There's a few other places in this change I see self::factory() as well.

> `self::factory()->...` is what WP core uses. See https://core.trac.wordpress.org/browser/trunk/tests/phpunit/tests/query/taxQuery.php
> 
> Just like `$this->assert...` and `self::assert...` (see "A common question..." paragraph at the top of https://phpunit.readthedocs.io/en/9.5/assertions.html) there doesn't seem to be any difference - but to me, it helps signify that it's a utility method being called, rather than a protected / private method in the test class itself.
>
> Besides - the actual `factory()` method on the WP unit test case, is static: https://core.trac.wordpress.org/browser/trunk/tests/phpunit/includes/abstract-testcase.php#L41

jblz:
As [discussed](https://github.com/Parsely/wp-parsely/pull/192#discussion_r611664371) in #192, the factory method of the `TestCase` utility is declared static, so it makes sense to call it as such:

https://core.trac.wordpress.org/browser/trunk/tests/phpunit/includes/abstract-testcase.php?rev=50697#L41

## To Test

All PHPUnit tests in affected files should continue to pass

Note: This is currently based on ~~the branch in #192~~ `release/2.4.0` to catch the tests added in that branch, so it will have to be rebased against develop once it lands.

### DIFF
--- a/tests/structured-data/archive-post-test.php
+++ b/tests/structured-data/archive-post-test.php
@@ -31,8 +31,8 @@ final class Archive_Post_Test extends TestCase {
 		$parsely_options = get_option( \Parsely::OPTIONS_KEY );
 
 		// Insert 2 posts.
-		$page_id = $this->factory()->post->create();
-		$this->factory()->post->create();
+		$page_id = self::factory()->post->create();
+		self::factory()->post->create();
 		$page = get_post( $page_id );
 
 		// Set permalinks, as Parsely currently strips ?page_id=... from the URL property.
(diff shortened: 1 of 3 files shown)


GaryJones (approved):
Changes are good.

jblz:
Changed to the 2.5.0 milestone so we don't need to update anything else for 2.4.0 -- this just touches tests and doesn't change them functionally.

PR #214, opened 2021-04-13
2.4.1.

mikeyarce:
### Fixed
- Fix the version number set in the main plugin file.


jblz:
Let's try #215 instead.

PR #216, opened 2021-04-14
Docs: Remove HTML from screenshot caption

GaryJones:
The WPORG readme parser / display page seems to break with it.


jblz (approved):
I think you're right to suspect this is causing the issue. It's worth a shot, anyway!

PR #217, opened 2021-04-14
Docs: Use standard markdown for code block

GaryJones:
Standard markdown code blocks use four-space indentation to delimit code blocks, and this is what's needed for correct viewing on WordPress.org.


mikeyarce (approved):
Thanks, Gary!

Do we need to also specify language?  like ` ```php `?

mikeyarce:
Oh wait, nevermind - it's just four spaces with no backticks!

GaryJones:
The ```php format is very much non-regular Markdown :-)

Issue #225, opened 2021-04-15
Support translations


GaryJones:
- [x] Load the plugin text domain (`wp-parsely`), since the plugin supports earlier than WP 4.6 (#268).
 - [x] Add internationalization functions to strings in PHP (#226).
 - [x] Fix `WordPress.WP.I18n` PHPCS violations, which includes the use of numbered placeholders, adding translator comments, and text strings not being a single string literal (avoid concatenations). Run `./vendor/bin/phpcs --standard=WordPress --sniffs=WordPress.WP.I18n . --ignore=vendor,node_modules -sp` to see what is outstanding (#226).
 - [x] Add internationalization functions to strings in JS (#236).
 - [x] Configure [ESLint rules](https://pascalbirchler.com/wordpress-javascript-i18n-eslint/) to check internationalized strings in JS (#183).
 - [x] ~Arrange translations from colleagues / Team Global to be added to translate.wordpress.org, and consider requesting them to be translation editors for this plugin (@GaryJones).~ Moving to post-2.5.0 release.

jblz:
> Configure ESLint rules to check internationalized strings in JS.

While I was working on #236, eslint was not happy until I added a translator comment (per [@wordpress/i18n-translator-comments](https://github.com/WordPress/gutenberg/blob/trunk/packages/eslint-plugin/docs/rules/i18n-translator-comments.md)). 

Our setup extends `plugin:@wordpress/eslint-plugin/recommended-with-formatting` since #183, so we get the "recommended" settings by virtue:

https://github.com/Parsely/wp-parsely/blob/3cca62a4c228af6bd8893e744e710fb6b4269e94/.eslintrc#L3

jblz:
Looks like all that's left here is to arrange translations!  Should we pull that out to the next milestone, @GaryJones?

GaryJones:
> Looks like all that's left here is to arrange translations! Should we pull that out to the next milestone, @GaryJones?

Yes, happy to do that - we'd want to get this 2.5.0 release published so that translate.wordpress.org can enable translations anyway. Also, I've been chatting to the A8C Team Global, who have a way to get paid translations done.

GaryJones:
> While I was working on #236, eslint was not happy until I added a translator comment (per [@wordpress/i18n-translator-comments](https://github.com/WordPress/gutenberg/blob/trunk/packages/eslint-plugin/docs/rules/i18n-translator-comments.md)).

Ah, I didn't realise it was already all configured!

[...]

PR #226, opened 2021-04-15
Translations: Internationalize strings in PHP

GaryJones:
This PR is made up of four commits which make a big dent into supporting translations (#225).

To see all PHP strings are internationalized, apply the PR locally and add the following to the bottom of the `wp-parsely.php` file:

```
add_filter( 'gettext', '__return_empty_string', PHP_INT_MAX );
add_filter( 'gettext_with_context', '__return_empty_string', PHP_INT_MAX );
add_filter( 'ngettext', '__return_empty_string', PHP_INT_MAX );
add_filter( 'ngettext_with_context', '__return_empty_string', PHP_INT_MAX );
```
This will replace all internationalized strings with an empty string - so any strings remaining are not internationalised. The screenshot below shows this in action (the Important string is added via JS, not PHP, and is outside the scope of this PR):

Screenshot 2021-04-16 at 00 06 11

----

## 1. Internationalize strings in PHP

To allow strings to be displayed in other locales, we have to wrap them in i18n functions.

Also:
- Added text domains where they were missing (better to include strings under the plugin's text domain, than re-use strings from core's language packs.
- Fixed a couple of types / incorrect capitalizations.
- Adding some translator comments.

## 2. Fix concatenated i18n strings

The `WordPress.WP.I18n.NonSingularStringLiteralText` PHPCS sniff checks if the `$text_arg` from i18n functions contains a single string literal, which is required to properly extract the string.

This fixes the arguments that contained concatenated strings.

Strings literals can't be broken onto multiple lines, as while the display is fine (browsers normalize the line break into a space), the translatable string in the .pot file contains `\n` and a number of `\t` tab characters, which is undesirable for translation contributors to deal with.

## 3. Use numbered placeholders

Multiple placeholders should be ordered (`WordPress.WP.I18n.UnorderedPlaceholdersText`).

Some help texts contain multiple placeholders to handle anchor elements within the text (these don't need to be abstracted away - simple HTML is fine to include in translatable strings. Fixing this requires changing how the `$help_link` code works though, which falls outside the scope of these changes).

## 4. Add missing translator comments

These help translators to understand what the placeholders will be replaced with to give them more context when translating.

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -224,10 +224,7 @@ class Parsely {
 			self::MENU_SLUG
 		);
 
-		$h      = __( 'Your API secret is your secret code to %s%s%saccess our API.%s
-			It can be found at dash.parsely.com/yoursitedomain/settings/api
-		 ( replace yoursitedomain with your domain name, e.g. `mydomain.com` ) If you haven\'t purchased access to the API, and would
-		  like to do so, email your account manager or support@parsely.com!', 'wp-parsely' );
+		$h      = __( 'Your API secret is your secret code to %s%s%saccess our API.%s It can be found at dash.parsely.com/yoursitedomain/settings/api ( replace yoursitedomain with your domain name, e.g. `mydomain.com` ) If you haven\'t purchased access to the API, and would like to do so, email your account manager or support@parsely.com!', 'wp-parsely' );
 		$h_link = 'https://www.parse.ly/help/api/analytics/';
 
 		$field_args = array(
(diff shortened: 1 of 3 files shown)


wp-parsely.php
```diff
@@ -197,21 +195,21 @@ public function initialize_settings() {
...
+		$h = __( 'Your Site ID is your own site domain ( e.g. `mydomain.com` )', 'wp-parsely' );
 
 		$field_args = array(
 			'option_key' => 'apikey',
 			'help_text'  => $h,
 		);
 		add_settings_field(
 			'apikey',
-			'Parse.ly Site ID <div class="help-icons"></div>',
+			__( 'Parse.ly Site ID', 'wp-parsely' ),
```
jblz, line 212: Is the omission of these `<div class="help-icons"></div>` elements intentional? I don't see it detailed in the commit messages.

jblz (approved):
This is working well for me.

I love the idea of blanking out the strings via filter to ensure you've caught them all 👏 

I left a question about the icon divs just to make sure it's what you wanted to accomplish here.

wp-parsely.php
GaryJones: Good call - I meant to mention this on the notes and forgot.

Yes, it is intentional to remove them. WordPress used to have support for showing help icons in wp-admin, but that got removed (not very accessible), so the divs in these labels were redundant.

PR #230, opened 2021-04-16
GitHub Actions: Update correct PHP versions for integration tests

htdat:
### Issue 

I notice that with this fixed PHP version value, all of current tests are only run on PHP 7.4.x.

https://github.com/Parsely/wp-parsely/blob/1a398fac2ebc4935199299977522c5a817b62505/.github/workflows/integration-tests.yml#L37

### Description of this PR 

This PR is to change all fixed PHP version 7.4 to the dynamic version in the matrix configuration.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -31,10 +31,10 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@master
 
-      - name: Setup PHP 7.4
+      - name: Setup PHP ${{ matrix.php }}
         uses: shivammathur/setup-php@v2
         with:
-          php-version: '7.4'
+          php-version: ${{ matrix.php }}
           coverage: pcov
           # https://make.wordpress.org/hosting/handbook/handbook/server-environment/#php-extensions
           extensions: curl, dom, exif, fileinfo, hash, json, mbstring, mysqli, libsodium, openssl, pcre, imagick, xml, zip
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -20,6 +20,7 @@ jobs:
       WP_VERSION: latest
 
     strategy:
+      fail-fast: false
       matrix:
         php: [ '5.6', '7.0', '7.1', '7.2', '7.3', '7.4', '8.0' ]
         allowed_failure: [ false ]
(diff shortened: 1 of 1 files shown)


htdat:
### Update 

Currently, PHP < 7.4 is getting this error: 

> PHP Warning:  mysqli_real_connect(): The server requested authentication method unknown to the client [caching_sha2_password] in /tmp/wordpress/wp-includes/wp-db.php on line 1653

That's because our test machine is using the latest MySQL version 8.0.x and it switches to use `caching_sha2_password` authentication method from `mysql_native_password` by default. At the same time, only PHP >= 74. supports this new method. 

Reference: 

- https://dev.mysql.com/doc/refman/8.0/en/upgrading-from-previous-series.html#upgrade-caching-sha2-password 
- https://www.php.net/manual/en/mysqli.requirements.php

### Workaround 

There are a few workarounds: 

1. Set a specific old MySQL version (5.7.x) in a docker container. An example is [this repo](https://github.com/FortAwesome/wordpress-fontawesome/blob/1d73851141a4b3d991aa800a582f265643b6a9e5/.github/workflows/php.yml#L26).
2. Set up `default-authentication-plugin=mysql_native_password` in the MySQL config file. - [link](https://dev.mysql.com/doc/refman/8.0/en/upgrading-from-previous-series.html#upgrade-caching-sha2-password-compatibility-issues).
3. Set up this option `default-authentication-plugin=mysql_native_password` for a docker container. WP CLI is using this method to test various MySQL versions - [link](https://github.com/wp-cli/wp-cli/blob/75979b428d09347e3928d935692f0e76cf393b62/.github/workflows/testing.yml#L100).
4. Update MySQL user `root` to use the old authentication method - [link](https://dev.mysql.com/doc/refman/8.0/en/upgrading-from-previous-series.html#upgrade-caching-sha2-password-root-account)

### What I've done 

I tried (1) but somehow MySQL container could not trigger properly, then (3) can not work too. 
I tried (4) and it's much better but the same error still happens for PHP 5.6 and 7.0. https://github.com/Parsely/wp-parsely/actions/runs/754642532

**That's all I can have for now. Any idea to process this next? I think exploring option (1) or (3) is more feasible.**

mikeyarce:
What about using older versions of msyql for php < 7.0?

Something like this:

```
matrix:
  include:
    - {  php-version: 5.6, db-image: mysql_5.7, ...}
    - {  php-version: 7.0, db-image: mysql_5.7, ...}   
```

I saw that example here: https://github.community/t/improve-matrix-exclusion-feature/16340/5

GaryJones:
Closing in favour of #233.

PR #232, opened 2021-04-16
Better load Widget CSS and use plugin version for cache busting

### LINKED ISSUE #231: Widget: Only load required CSS if widget is active on page
Our main CSS file https://github.com/Parsely/wp-parsely/blob/develop/wp-parsely.css currently loads on the front end always, even though it only contains styles for the widget.  

We should change it so that it only loads if there is a widget present on the current page being loaded.

mikeyarce:
Move the `wp_enqueue_style` to within the public widget function so that its only enqueued if the widget is being loaded.

Bonus, change the version from `ilemtime( get_stylesheet_directory() )` to use the plugin version instead:  `Parsely::VERSION`

Fixes #231 

I'm also wondering if we should change the name of the file to something like `parsely-widget.css` instead?

### DIFF
--- a/class-parsely-recommended-widget.php
+++ b/class-parsely-recommended-widget.php
@@ -47,6 +47,8 @@ class Parsely_Recommended_Widget extends WP_Widget {
 		/** This filter is documented in wp-includes/widgets/class-wp-widget-pages.php */
 		$title = apply_filters( 'widget_title', $instance['title'] );
 
+		wp_enqueue_style( 'wp-parsely-style', plugins_url( 'wp-parsely.css', __FILE__ ), array(), Parsely::VERSION );
+
 		$allowed_tags = wp_kses_allowed_html( 'post' );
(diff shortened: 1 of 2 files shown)


GaryJones:
Can we take a slightly different approach here? 

Register the style in the usual hook, but only enqueue it (with just the handle) in the widget method. 

That gives others the ability to also enqueue it for other contexts if they wish, without having to register it from scratch.

mikeyarce:
> Register the style in the usual hook, but only enqueue it (with just the handle) in the widget method.
That gives others the ability to also enqueue it for other contexts if they wish, without having to register it from scratch.

I like that approach!  I will make the changes 😄

GaryJones (approved):
Tested with Parse.ly widget added to a widget area: CSS ref appears. 
Tested without Parse.ly widget - CSS ref does not appear.

👍🏻 All good.

PR #233, opened 2021-04-17
CI: Use matrix PHP version

GaryJones:
Previously multiple PHP-focused jobs were created from the matrix configuration, but 7.4 was hard-coded to be used in each instance anyway.

Props to @htdat for spotting this oversight and providing a fix.

The fix revealed an issue with the MySQL version (8) that comes with `ubuntu-latest`, which has an issue with older versions of PHP, so the CI now uses a previous OS version that sticks with MySQL 5.7

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -13,7 +13,9 @@ on:
 jobs:
   test:
     name: PHP ${{ matrix.php }}
-    runs-on: ubuntu-latest
+    # ubuntu-20.04 / ubuntu-latest includes MySQL 8, which has issues with older versions of PHP.
+    # ubuntu-16.04 includes PHP versions 5.6-8.0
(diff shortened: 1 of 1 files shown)


jblz (commented):
Looks like PHP 8.1 is fataling & even though we've specified `allowed_failure: true`, Github is blocking the merge:

`PHP Fatal error:  Cannot acquire reference to $GLOBALS in /home/runner/work/wp-parsely/wp-parsely/vendor/phpunit/phpunit/src/Util/Configuration.php on line 570`

Any ideas?

jblz:
Oh! Now that I'm looking at #234, it makes sense why this is failing on 8.1.

GaryJones:
I think that merging is only blocked because it hasn't got an approved review. If you can approve, then we can force it through if necessary.

jblz (approved):
🚀

PR #234, opened 2021-04-17
CI: Small improvements

GaryJones:
- Remove PHP 8.1 (nightly) test. Due to a syntax conflict with PHP 8.1 in PHPUnit 7.5, this job will never succeed.
- Add `fail-fast: false` to allow rare job failures to allow other jobs to complete and give more details.
- Add problem matcher for PHP.
- Consolidate near-duplicate Composer Install steps into one.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -14,19 +14,21 @@ jobs:
   test:
     name: PHP ${{ matrix.php }}
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


jblz (changes requested):
This looks good to me, but can we go ahead and merge #233 into this branch / PR?

GaryJones:
#233 should be able to be merged first, without needing to merge into this one - I specifically kept things separate.

jblz (approved):
sounds like a plan

PR #235, opened 2021-04-17
Use strict code coverage and add code coverage output

GaryJones:
Finish configuring PCOV to display code coverage for single site integration tests.

Currently focused on just two files with functions/methods on them - the main plugin file, and the widget class file.

If the PHP version is `8.0`, then the code coverage runs, otherwise it doesn't. There's no need to run it on every job.

Once #234 has been merged, this PR can be rebased and updated to set `coverage: none` and remove ext-pcov for all but PHP 8.0.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -77,9 +77,9 @@ jobs:
 
       - name: Run integration tests (single site)
         if: ${{ matrix.php != 8.0 }}
-        run: composer integration
+        run: composer test
       - name: Run integration tests (single site with code coverage)
         if: ${{ matrix.php == 8.0 }}
-        run: ./vendor/bin/phpunit --testsuite WP_Tests --coverage-text
+        run: composer coverage
       - name: Run integration tests (multisite)
-        run: composer integration-ms
+        run: composer test-ms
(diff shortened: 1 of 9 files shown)


GaryJones:
Rebased and updated, so now the code coverage extension and configuration is done only for PHP 8.0.

**PHP 7.3**: No coverage, 2s job time:
Screenshot 2021-04-28 at 10 04 18

**PHP 8.0**: PCOV extension added, in values set, coverage drive set:
Screenshot 2021-04-28 at 10 04 36

GaryJones:
Went ahead and enabled strict code coverage (#240) by updating the PHPUnit config, and adding annotations as needed to all tests.

GaryJones:
There are a couple of tests that count as Risky, but this is blocked by, and maybe be resolved by, #236. The one remaining noisy test is due to the same.

Screenshot 2021-04-30 at 12 52 09


Use:

- `composer test` to run tests without code coverage - you'll likely use this often when writing code.
- `composer test-ms` to run the same tests in Multisite mode - we have one test that is skipped unless tests are run in multisite mode.
- `composer coverage` to run tests and generate a local Clover XML (not too useful now, but GitHub Actions could send this to a third-party service to have it displayed and accessible to everyone)
- `composer coverage-local` to run tests and generate a local HTML directory that you can use to display code coverage metrics:
Screenshot 2021-04-30 at 12 48 34


Screenshot 2021-04-30 at 12 48 55

jblz (approved):
I'm really glad you're on our side 😄 

:shipit:

Issue #160, opened 2021-03-20
Internationalize all user-facing strings


GaryJones:
To support users who are using the plugin in a locale that is not English, we should let them display strings in their chosen locale. There are two parts to this - wrapping strings in internationalization functions, and providing translated strings. This Issue is concerned with the former.

mikeyarce:
Fixed in https://github.com/Parsely/wp-parsely/pull/226

GaryJones:
There are still JavaScript strings to add i18n functions to yet, but I hadn't tackled them as @jblz was refactoring the JS.

jblz:
The only JS string I see is this one which is now wrapped (as of #236):
https://github.com/Parsely/wp-parsely/blob/3cca62a4c228af6bd8893e744e710fb6b4269e94/src/js/lib/admin-page/index.js#L13-L20

I'm going to go ahead and close this.  Please let me know if I've missed any!

Issue #180, opened 2021-04-02
Stop tests from echoing


GaryJones:
It's not a great practice to have tests echo out strings, yet that is exactly what it does for many of the tests - see the [Run integration tests](https://github.com/Parsely/wp-parsely/pull/178/checks?check_run_id=2253518308).

The issue is that the tests call `insert_parsely_page()` ([example](https://github.com/Parsely/wp-parsely/blob/f9007743fefeea3128e9038a97416ce9f7af5c87/tests/all-test.php#L144)), and that includes an [`include 'parsely-parsely-page.php';`](https://github.com/Parsely/wp-parsely/blob/f9007743fefeea3128e9038a97416ce9f7af5c87/wp-parsely.php#L805) that contains markup.

The better approach would be to test the [`construct_parsely_metadata()` method](https://github.com/Parsely/wp-parsely/blob/f9007743fefeea3128e9038a97416ce9f7af5c87/wp-parsely.php#L817) which is [called](https://github.com/Parsely/wp-parsely/blob/f9007743fefeea3128e9038a97416ce9f7af5c87/wp-parsely.php#L804) right before the `include`, to check that the resulting metadata JSON is correct, and not test that PHP has correctly echo'd some content.

Other tests have other variations; giving output for JS, FBIA, AMP etc.

mikeyarce:
Tests to fix:

- [ ] test_parsely_tag
- [x] test_parsely_ppage_output
- [x] test_parsely_categories
- [x] test_parsely_tags_lowercase
- [x] test_parsely_cats_as_tags
- [x] test_custom_taxonomy_tags
- [x] test_use_top_level_cats
- [x] test_custom_taxonomy_as_section
- [x] test_top_level_taxonomy_as_section
- [x] test_http_canonicals
- [x] test_fbia_integration
- [x] test_amp_integration
- [x] test_parsely_page_filter
- [x] test_user_logged_in
- [x] test_user_logged_in_multisite

GaryJones:
#258 fixed most of the echoing tests, but one still remains, so re-opening.

jblz:
The `NOW IT BEGINS` echo previously in `test_parsely_tag` is no more as of #236.  Calling this one (re-)closed!

GaryJones:
<img width="810" alt="Showing non-noisy tests" src="https://user-images.githubusercontent.com/88371/117035638-75061100-acfc-11eb-8fb7-0b4933708f89.png">

🎉

PR #236, opened 2021-04-19
Pull Scripts out of PHP & Begin removing jQuery dependency

jblz:
This is the first pass at pulling the plugin JavaScript out of PHP files and into enqueued assets.

The [widget script](https://github.com/Parsely/wp-parsely/blob/14c342a9bca717850801fadb00aea7ae8d481761/class-parsely-recommended-widget.php#L111-L217) is not included in this change as this PR was already getting pretty large.  In addition, rewriting it with vanilla JavaScript is not a trivial effort and we're considering re-working it anyhow.

As such, this does not fully complete #134 or #158, but it's at least a step in that direction.

Additionally, we're able to use the WordPress i18n tooling to wrap a string in the Admin page JS for translation (see 3c62c8d). Again, it doesn't fully satisfy #160, but it paves the way for doing so.

### Automated Tests

* changes should be correct and comprehensive
* Automated tests should pass

### Building the Assets

* Follow instructions to [build the included assets](https://github.com/Parsely/wp-parsely/blob/remove/jquery-dependency/DEVELOP.md#building-included-assets)
* Make sure the instructions are clear and easy to follow

### Test the plugin

* The tracking pixel should be correctly enqueued
* The API initialization script should be correctly enqueued
* The Admin JS should be correctly enqueued
* The Admin CSS should be correctly applied
* The Recommended Content widget loading should be unaffected
* Recommended content should be appropriately displayed
* In general, there should be no changes that are noticeable to end users by virtue of this change (unless they open their dev tools panel 😄)


src/lib/init-api/index.js
```diff
@@ -0,0 +1,40 @@
+function uuidProfileCall( { apikey, uuid } ) {
+	const profileScript = document.createElement( 'script' );
+	profileScript.src = `https://api.parsely.com/v2/profile?apikey=${ encodeURIComponent(
+		apikey
+	) }&uuid=${ uuid || '' }&url=${ window.location.href }&callback=&_=${ +new Date() }`;
+	document.querySelector( 'head' ).appendChild( profileScript );
+}
```
jblz: This is the part I'm least confident about.  I'm attempting to replace this "jsonp" call:

https://github.com/Parsely/wp-parsely/blob/14c342a9bca717850801fadb00aea7ae8d481761/parsely-javascript.php#L24-L32

...which isn't obvious to me how it's used or what it does.  @hbbtstar -- can you shed any light on this?

build/admin-page.asset.php
```diff
@@ -0,0 +1 @@
+<?php return array('dependencies' => array('wp-polyfill'), 'version' => '6d8d3cc10bb4327a5d847f75b4f497e1');
```
GaryJones: Where does this hash come from?

class-parsely-recommended-widget.php
```diff
@@ -92,6 +92,7 @@ public function widget( $args, $instance ) {
 		$title = apply_filters( 'widget_title', $instance['title'] );
 
 		wp_enqueue_style( 'wp-parsely-style' );
+		wp_enqueue_script( 'jquery' );
```
GaryJones: Instead of enqueueing jQuery directly, can you add it as a dependency arg for the registration/enqueue of the now-external .js files?

package.json
```diff
@@ -6,9 +6,13 @@
   "directories": {
     "test": "tests"
   },
+  "engines": {
+    "node": "^14.16.1",
+    "npm": "^7.10.0"
```
GaryJones: These seem like really high dependency versions. What is it about these specific versions that are required to make the build work? What would the lowest versions of node and npm be that would still work?

src/lib/admin-page/index.js
```diff
@@ -0,0 +1,19 @@
...
+	const keyEl = document.querySelector( '#apikey' );
+	const requiresRecrawlNotice = document.querySelectorAll(
+		'div.parsely-form-controls[data-requires-recrawl="true"] .help-text'
+	);
+	if ( ! ( keyEl && requiresRecrawlNotice.length ) ) {
+		return;
+	}
+
+	const notice = sprintf(
+		'<p class="description"><strong style="color:red;">Important:</strong> changing this value on a site currently tracked with Parse.ly will require reprocessing of your Parse.ly data. Once you have changed this value, please contact <a href="mailto:support@parsely.com?subject=Please reprocess %s">support@parsely.com</a></p>',
```
GaryJones:
```suggestion
		'<p class="description"><strong style="color:#d63638;">Important:</strong> changing this value on a site currently tracked with Parse.ly will require reprocessing of your Parse.ly data. Once you have changed this value, please contact <a href="mailto:support@parsely.com?subject=Please reprocess %s">support@parsely.com</a></p>',
```

Just changes the red slightly to match the notice/form error red colour that is in the default admin colour scheme (aiming for consistency). Here's a comparison (new red is second instance):
Screenshot 2021-04-23 at 10 14 43

Alternatively, leave as `red` for now, and we can reassess this when it's refactored to remove the noise (since technically this "Important" isn't an error).

tests/all-test.php
```diff
@@ -103,30 +101,113 @@ public function setUp() {
 			'track_page_types'          => array( 'page' ),
 		);
 		update_option( 'parsely', $option_defaults );
-		self::$parsely_html = <<<PARSELYJS
-<script data-cfasync="false" id="parsely-cfg" data-parsely-site="blog.parsely.com" src="https://cdn.parsely.com/keys/blog.parsely.com/p.js"></script>
-PARSELYJS;
 	}
 
+	public function test_class_version() {
+		self::assertEquals( '2.4.1', \Parsely::VERSION );
```
GaryJones: This seems fragile, as in it will be another thing to update on each release.

The only thing it seems to be testing is that someone can write a 5-character string correctly in two places.

tests/all-test.php
```diff
@@ -103,30 +101,113 @@ public function setUp() {
...
+	public function test_parsely_default_scripts() {
 		ob_start();
 		$post_array = $this->create_test_post_array();
 		$post       = $this->factory->post->create( $post_array );
 		$this->go_to( '/?p=' . $post );
-		echo esc_html( self::$parsely->insert_parsely_javascript() );
+		echo self::$parsely->insert_parsely_javascript();
+
+		$intermediate_output = ob_get_contents();
+		self::assertEquals(
```
GaryJones:
```suggestion
		self::assertSame(
```

Does a type check (`===`) as well, instead of just the value check (`==`).

tests/all-test.php
```diff
@@ -103,30 +101,113 @@ public function setUp() {
...
+			wp_script_is( 'wp-parsely-tracker', 'enqueued' ),
+			'Failed to confirm tracker script was enqueued'
+		);
+
+		wp_print_scripts();
 		$output = ob_get_clean();
-		echo esc_html( $output );
-		self::assertContains( self::$parsely_html, $output );
+
+		self::assertEquals(
```
GaryJones:
```suggestion
		self::assertSame(
```

tests/all-test.php
```diff
@@ -103,30 +101,113 @@ public function setUp() {
...
+			"<script type='text/javascript' data-parsely-site=\"blog.parsely.com\" src='https://cdn.parsely.com/keys/http://blog.parsely.com/p.js?ver=" . PARSELY_VERSION . "' id=\"parsely-cfg\"></script>\n",
```
GaryJones: Aside: that type='text/javascript' is redundant for HTML5, since that's the default value.

tests/all-test.php
```diff
@@ -103,30 +101,113 @@ public function setUp() {
...
+		);
 	}
 
+	/**
+	 * Test the parsely script enqueues when an api secret is set.
+	 */
+	public function test_parsely_api_enabled_scripts() {
+		$options = get_option( 'parsely' );
+		$options['api_secret'] = 'hunter2';
+		update_option( 'parsely', $options );
```
GaryJones: Might want to consider moving this utility step into its own method:

```php
private set_api_secret( $api_secret ) {
	$options = get_option( 'parsely' );
	$options['api_secret'] = $api_secret;
	update_option( 'parsely', $options );
}
```

Then:
```suggestion
		$this->set_api_password( 'hunter2' );
```

tests/all-test.php
```diff
@@ -103,30 +101,113 @@ public function setUp() {
...
+
+		ob_start();
+		$post_array = $this->create_test_post_array();
+		$post       = $this->factory->post->create( $post_array );
+		$this->go_to( '/?p=' . $post );
+		echo self::$parsely->insert_parsely_javascript();
+
+		$intermediate_output = ob_get_contents();
+		self::assertEquals(
```
GaryJones:
```suggestion
		self::assertSame(
```

tests/all-test.php
```diff
@@ -103,30 +101,113 @@ public function setUp() {
...
+	public function test_parsely_api_enabled_scripts() {
+		$options = get_option( 'parsely' );
+		$options['api_secret'] = 'hunter2';
+		update_option( 'parsely', $options );
+
+		ob_start();
+		$post_array = $this->create_test_post_array();
+		$post       = $this->factory->post->create( $post_array );
+		$this->go_to( '/?p=' . $post );
+		echo self::$parsely->insert_parsely_javascript();
```
GaryJones: Might be beyond the scope of this PR, but instead of echoing out the value into an output buffer, can we adjust the function to just get a return and check against that instead? #180.

tests/all-test.php
GaryJones: Or if the other tests set other settings, then make it slightly more generic:

```php
private update_parsely_option( $key, $value ) {
	$options = get_option( 'parsely' );
	$options[ $key ] = $value;
	update_option( 'parsely', $options );
}

File: tests/all-test.php [hunk 13]
```diff
@@ -459,17 +575,79 @@ public function test_user_logged_in_multisite() {
...
+		self::assertTrue(
+			wp_script_is( 'wp-parsely-tracker', 'enqueued' ),
+			'Failed to confirm tracker script was enqueued'
+		);
+
+		wp_print_scripts();
 		$output = ob_get_clean();
-		self::assertContains( self::$parsely_html, $output );
+
+		self::assertEquals(
```
GaryJones:
```suggestion
		self::assertSame(
```

tests/all-test.php
```diff
@@ -459,17 +575,79 @@ public function test_user_logged_in_multisite() {
...
 		switch_to_blog( $second_blog );
 		self::assertEquals( get_current_blog_id(), $second_blog );
 		self::assertFalse( is_user_member_of_blog( $new_user, get_current_blog_id() ) );
 
 		ob_start();
-		echo esc_html( self::$parsely->insert_parsely_javascript() );
+		echo self::$parsely->insert_parsely_javascript();
+
+		$intermediate_output = ob_get_contents();
+		self::assertEquals(
```
GaryJones:
```suggestion
		self::assertSame(
```

tests/all-test.php
```diff
@@ -459,17 +575,79 @@ public function test_user_logged_in_multisite() {
...
+		self::assertFalse(
+			wp_script_is( 'wp-parsely-tracker', 'enqueued' ),
+			'Failed to confirm tracker script was not enqueued'
+		);
+
+		wp_print_scripts();
 		$output = ob_get_clean();
-		self::assertNotContains( self::$parsely_html, $output );
+
+		self::assertEquals(
```
GaryJones:
```suggestion
		self::assertSame(
```

tests/all-test.php
```diff
@@ -459,17 +575,79 @@ public function test_user_logged_in_multisite() {
 		self::assertFalse( is_user_member_of_blog( $new_user, $second_blog ) );
 
 		ob_start();
-		echo esc_html( self::$parsely->insert_parsely_javascript() );
+		echo self::$parsely->insert_parsely_javascript();
+
+		$intermediate_output = ob_get_contents();
+		self::assertEquals(
```
GaryJones:
```suggestion
		self::assertSame(
```

tests/all-test.php
```diff
@@ -419,10 +500,45 @@ public function test_user_logged_in() {
...
+		self::assertFalse(
+			wp_script_is( 'wp-parsely-tracker', 'enqueued' ),
+			'Failed to confirm tracker script was not enqueued'
+		);
+
+		wp_print_scripts();
 		$output = ob_get_clean();
-		self::assertNotContains( self::$parsely_html, $output );
+
+		self::assertEquals(
```
GaryJones:
```suggestion
		self::assertSame(
```

tests/all-test.php
```diff
@@ -419,10 +500,45 @@ public function test_user_logged_in() {
 		update_option( 'parsely', $options );
 		$new_user = $this->create_test_user( 'bill_brasky' );
 		wp_set_current_user( $new_user );
+
 		ob_start();
-		echo esc_html( self::$parsely->insert_parsely_javascript() );
+		echo self::$parsely->insert_parsely_javascript();
+
+		$intermediate_output = ob_get_contents();
+		self::assertEquals(
```
GaryJones:
```suggestion
		self::assertSame(
```

tests/all-test.php
```diff
@@ -103,30 +101,113 @@ public function setUp() {
...
+
+		self::assertTrue(
+			wp_script_is( 'wp-parsely-tracker', 'enqueued' ),
+			'Failed to confirm tracker script was enqueued'
+		);
+
+		wp_print_scripts();
+		$output = ob_get_clean();
+
+		self::assertEquals(
```
GaryJones:
```suggestion
		self::assertSame(
```

wp-parsely.php
```diff
@@ -104,12 +104,15 @@ public function __construct() {
 		add_action( 'save_post', array( $this, 'update_metadata_endpoint' ) );
 		add_action( 'instant_articles_compat_registry_analytics', array( $this, 'insert_parsely_tracking_fbia' ) );
 		add_action( 'template_redirect', array( $this, 'parsely_add_amp_actions' ) );
-		if ( ! defined( 'WP_PARSELY_TESTING' ) ) {
+		if ( ! self::is_test_env() ) {
```
GaryJones: Outside of the scope of this PR, but this really shouldn't be needed! Need to see why tests supposedly break when some styles are enqueued!

wp-parsely.php
```diff
@@ -1046,15 +1071,65 @@ public function insert_parsely_javascript() {
...
+				'wpParsely', // This globally-scoped object will hold our initialization variables
+				array(
+					'apikey' => $parsely_options['apikey'],
+				)
+			);
+		}
+
+		wp_enqueue_script(
+			'wp-parsely-tracker',
+			'https://cdn.parsely.com/keys/' . esc_url( $parsely_options['apikey'] ) . '/p.js',
```
GaryJones: Escaping of a URL is handled by WP when it outputs the HTML (late), so it's not needed here.

wp-parsely.php
```diff
@@ -1046,15 +1071,65 @@ public function insert_parsely_javascript() {
...
+	public function script_loader_tag( $tag, $handle, $src ) {
+		$parsely_options = $this->get_options();
+		if ( in_array( $handle, [ 'wp-parsely', 'wp-parsely-tracker' ] ) ) {
+			// Have ClouldFlare Rocket Loader ignore these scripts:
+			// https://support.cloudflare.com/hc/en-us/articles/200169436-How-can-I-have-Rocket-Loader-ignore-specific-JavaScripts-
+			$tag = preg_replace( '/^<script src=/', '<script data-cfasync="false" src=', $tag );
+		}
+
+		if ( $handle === 'wp-parsely-tracker' ) {
+			$tag = preg_replace( '/ id=(\"|\')wp-parsely-tracker-js\1/', ' id="parsely-cfg"', $tag );
```
GaryJones: Do we know why this change of ID is needed?

GaryJones (changes requested):
I love where this is going!

Added a few comments, including 3-4 items that we may want to spin up into separate issues/PRs, if you don't want to address them in this one.

build/admin-page.asset.php
jblz: This is a file that's generated by `wp-scripts build` process...

Looks like it's a way to dynamically include the dependencies on the WP side:
https://developer.wordpress.org/block-editor/reference-guides/packages/packages-dependency-extraction-webpack-plugin/#wordpress

I'll give it a try to see if it's worth the "magic" to avoid having to maintain the dependency list or not.

class-parsely-recommended-widget.php
jblz: > Instead of enqueueing jQuery directly, can you add it as a dependency arg for the registration/enqueue of the now-external .js files?

The Widget script is not yet extracted from the PHP, so it has to be explicitly enqueued here.

Once we do pull it out, we'll definitely want to add it as a dependency to script (unless we remove the dependency entirely, of course).

package.json
jblz: > These seem like really high dependency versions. What is it about these specific versions that are required to make the build work? What would the lowest versions of node and npm be that would still work?

These are the latest versions of the LTS node release & the accompanying `latest` tag of npm. While there's nothing tying it to these particular versions functionality-wise, the engine is only developer-facing and it seems reasonable to enforce compatibility with the current point releases.

src/lib/admin-page/index.js
jblz: I'm definitely good with this change, but I'd pull it out to a separate PR.

The color value is coming from here:
https://github.com/Parsely/wp-parsely/blob/1a398fac2ebc4935199299977522c5a817b62505/parsely-admin-header.php#L32

tests/all-test.php
jblz: The reason I added the explicit test of the value of the `\Parsely::VERSION` class constant is that it is being used in the cache buster by virtue of the `PARSELY_VERSION` defined global constant.

If it makes sense to cover the behavior around the functionality that sets the cache buster (& I contend that it is), it makes sense to go the extra mile with the other two values with the single hard-coded version string.

When we bump the version in the plugin, the test will fail if this line is missed & it's an easy change.

wp-parsely.php
jblz: I really don't know. This is another case of trying to not break anything without understanding why it is the way it is.

@hbbtstar -- any thoughts as to the necessity of having particular `id` attributes to the `script` tags?

build/admin-page.asset.php
jblz: ok, magic invoked in a2131bec664d422e3e9d578da4ab7c6c297b2001.

Pretty cool.  It lets us use `import` syntax and not have to manage dependencies inside the PHP side of the plugin.

package.json
GaryJones: Since it's the latest LTS and latest (minor) release, could it be:

```suggestion
    "node": "^14",
    "npm": "^7"
```

...instead? That gives a bit of wiggle room for contributors to start contributing and building, even if they aren't on the latest minor / patch version.

Right now, the latest npm is 7.11.1.

src/lib/admin-page/index.js
GaryJones: Opened https://github.com/Parsely/wp-parsely/issues/254.

tests/all-test.php
GaryJones: > If it makes sense to cover the behavior around the functionality that sets the cache buster (& I contend that it is), it makes sense to go the extra mile with the other two values with the single hard-coded version string.

For clarity, when we're talking about the cache-busting, we mean setting the arg in the `wp_register_script()` / `wp_enqueue_script()`?

Ensuring that arg is set in one of those function calls, is ours to test against. Checking that the value is used in the `<script>` output, is not as this is WP code that adds the querystring.

Happy to let it go in, but we'll want to review how it fits in when we audit and cleanup `all-tests.php`.

tests/all-test.php
GaryJones: Opened https://github.com/Parsely/wp-parsely/issues/255.

wp-parsely.php
GaryJones: Opened https://github.com/Parsely/wp-parsely/issues/256.

src/lib/init-api/index.js
GaryJones: As per the internal P2 post, CORS support in browsers is now enough, that JSONP isn't needed any more.

GaryJones:
Looks like the `parsely-admin-header.php` file can be removed as well, as the `include` to it is being removed.

tests/class-testutils.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Parsely\Tests;
+
+class TestUtils {
+	public static function set_options( $custom_options = array() ) {
```
GaryJones: We already have a custom TestCase that utility methods should go into.

GaryJones (changes requested):

src/class-parsely.php
```diff
@@ -1069,6 +1073,22 @@ function( $v ) {
...
+		if ( isset( $cache_buster ) ) {
+			return $cache_buster;
+		}
+
+		$cache_buster = defined( 'WP_DEBUG' ) && WP_DEBUG && empty( 'WP_TESTS_DOMAIN' ) ? wp_rand() : PARSELY_VERSION;
+
+		/**
+		 * Filters the cache buster value for linked scripts and styles.
+		 *
+		 * @param string $cache_buster Plugin version, unless WP_DEBUG is defined and truthy, and tests are not running.
```
```suggestion
		 * Filters the cache buster value for linked scripts and styles.
		 *
		 * @since 2.5.0
		 *
		 * @param string $cache_buster Plugin version, unless WP_DEBUG is defined and truthy, and tests are not running.
```

src/js/lib/admin-page/index.js
```diff
@@ -0,0 +1,24 @@
...
+	const keyEl = document.querySelector( '#apikey' );
+	const requiresRecrawlNotice = document.querySelectorAll(
+		'.parsely-form-controls[data-requires-recrawl="true"] .help-text'
+	);
+	if ( ! ( keyEl && requiresRecrawlNotice.length ) ) {
+		return;
+	}
+
+	const notice = sprintf(
+		// translators: %s: The API Key that will be used to request a "recrawl"
```
GaryJones:
```suggestion
		/* translators: %s: The API Key that will be used to request a recrawl */
```

GaryJones (changes requested):

GaryJones (approved):

PR #242, opened 2021-04-21
Support Custom Post Status Slugs

### LINKED ISSUE #218: Support custom post statuses
Originally posted at https://wordpress.org/support/topic/wp-parsely-status-limitation/ - repeated here verbatim for clarity:

----
## Support Forum Post

We ran into an issue with Parsely not showing up on posts with statuses other than publish. Kind of a weird situation, but we are making some custom statuses that are publicly available in order to accommodate an editorial workflow. Tracked the problem down to line 724 of wp-parsely.php:

```php
if ( in_array( get_post_type(), $parsely_options['track_post_types'], true ) && 'publish' === $post->post_status ) {
```

I fixed this by adding a filter on the posts statuses:

```php
$open_statuses = apply_filters('wp_parsely_open_statuses', array('publish'));
```

And then adjusting the conditional to allow for custom statuses:

```php
if ( in_array( get_post_type(), $parsely_options['track_post_types'], true ) && in_array($post->post_status, $open_statuses) ) {
```

Of course, altering a plugin’s code is not something I’m in the habit of doing, but I in this case it was necessary. I was hoping that this simple adjustment might make its way into the next release.

----
## Response

"line  724" is now [line 855](https://github.com/Parsely/wp-parsely/blob/1a398fac2ebc4935199299977522c5a817b62505/wp-parsely.php#L855) and [line 934](https://github.com/Parsely/wp-parsely/blob/1a398fac2ebc4935199299977522c5a817b62505/wp-parsely.php#L934), and there's also a `post_status` comparison against `publish` at [line 1063](https://github.com/Parsely/wp-parsely/blob/1a398fac2ebc4935199299977522c5a817b62505/wp-parsely.php#L1063).

We could create an open statuses filter, but we should also pass in the `get_post_type()` as an argument so that a filter can be applied that perhaps uses different lists of open statuses per post type.

jblz:
Fixes #218

Adds a new function `post_has_trackable_status` which accepts a single parameter: `$post`.

That function applies a new filter `wp_parsely_track_status_slugs` to get the allowed list of status slugs to track (It defaults to a list containing only `publish`).

Plugins can  override the list of status slugs that will enable tracking by returning a new list from a function hooked into the new filter.

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -784,6 +784,12 @@ class Parsely {
 	 * @return bool Should the post status be tracked for the provided post's post_type. By default, only 'publish' is allowed.
 	 */
 	public static function post_has_trackable_status( $post ) {
+		static $cache = array();
+		$post_id = is_int( $post ) ? $post : $post->ID;
+		if ( isset( $cache[ $post_id ] ) ) {
+			return $cache[ $post_id ];
+		}
+
 		/**
 		 * Filters the statuses that are permitted to be tracked.
 		 *
@@ -795,7 +801,8 @@ class Parsely {
 		 * @param int|WP_Post $post               Which post object or ID is being checked.
 		 */
 		$statuses = apply_filters( 'wp_parsely_trackable_statuses', array( 'publish' ), $post );
-		return in_array( get_post_status( $post ), $statuses, true );
+		$cache[ $post_id ] = in_array( get_post_status( $post ), $statuses, true );
+		return $cache[ $post_id ];
 	}
 
 	/**
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -791,11 +791,11 @@ class Parsely {
 		 *
 		 * @since 2.5.0
 		 *
-		 * @param string[]    $status_slugs The list of post_status slugs that are allowed to be tracked.
-		 * @param int|WP_Post $post         Which post object or id is being checked.
+		 * @param string[]    $trackable_statuses The list of post statuses that are allowed to be tracked.
+		 * @param int|WP_Post $post               Which post object or ID is being checked.
 		 */
-		$status_slugs = apply_filters( 'wp_parsely_track_status_slugs', array( 'publish' ), $post );
-		return in_array( get_post_status( $post ), $status_slugs, true );
+		$statuses = apply_filters( 'wp_parsely_trackable_statuses', array( 'publish' ), $post );
+		return in_array( get_post_status( $post ), $statuses, true );
 	}
 
 	/**
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -785,10 +785,9 @@ class Parsely {
 	 */
 	public static function post_has_trackable_status( $post ) {
 		static $cache = array();
-
-		$post_status = get_post_status( $post );
-		if ( isset( $cache[ $post_status ] ) ) {
-			return $cache[ $post_status ];
+		$post_id = is_int( $post ) ? $post : $post->ID;
+		if ( isset( $cache[ $post_id ] ) ) {
+			return $cache[ $post_id ];
 		}
 
 		/**
@@ -802,8 +801,8 @@ class Parsely {
 		 * @param int|WP_Post $post               Which post object or ID is being checked.
 		 */
 		$statuses = apply_filters( 'wp_parsely_trackable_statuses', array( 'publish' ), $post );
-		$cache[ $post_status ] = in_array( $post_status, $statuses, true );
-		return $cache[ $post_status ];
+		$cache[ $post_id ] = in_array( get_post_status( $post ), $statuses, true );
+		return $cache[ $post_id ];
 	}
 
 	/**
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -775,6 +775,28 @@ class Parsely {
 		return $parsely_page;
 	}
 
+	/**
+	 * Compare the post_status key against an allowed list (by default, only 'publish'ed content includes tracking data)
+	 *
+	 * @param  int|WP_Post $post  Which post object or id to check.
+	 * @return bool               Should the post status be tracked for the provided post's post_type. By default, only 'publish' is allowed.
+	 *
+	 * @see wp_parsely_track_status_slugs
+	 * @since 2.5.0
+	 */
+	public static function post_has_trackable_status( $post ) {
+		/**
+		 * Filters the status slugs that are permitted to be tracked.
+		 * By default, the only slug allowed is 'publish'. Override this list if your post type necessitates.
+		 *
+		 * @since 2.5.0
+		 *
+		 * @param string[]    $status_slugs The list of post_status slugs that are allowed to be tracked.
+		 * @param int|WP_Post $post         Which post object or id is being checked.
+		 */
+		$status_slugs = apply_filters( 'wp_parsely_track_status_slugs', array( 'publish' ), $post );
+		return in_array( get_post_status( $post ), $status_slugs, true );
+	}
 
 	/**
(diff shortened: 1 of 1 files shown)


wp-parsely.php
```diff
@@ -768,6 +768,28 @@ public function insert_parsely_page() {
...
+	 * Compare the post_status key against an allowed list (by default, only 'publish'ed content includes tracking data)
+	 *
+	 * @param  int|WP_Post $post  Which post object or id to check.
+	 * @return bool               Should the post status be tracked for the provided post's post_type. By default, only 'publish' is allowed.
+	 *
+	 * @see wp_parsely_track_status_slugs
+	 * @since 2.5.0
+	 */
+	public static function post_has_trackable_status( $post ) {
+		/**
```
GaryJones, line 793: Consider adding a static variable in here, so that the filter only runs if the static variable isn't populated.

wp-parsely.php
```diff
@@ -768,6 +768,28 @@ public function insert_parsely_page() {
...
 	}
 
+	/**
+	 * Compare the post_status key against an allowed list (by default, only 'publish'ed content includes tracking data)
+	 *
+	 * @param  int|WP_Post $post  Which post object or id to check.
+	 * @return bool               Should the post status be tracked for the provided post's post_type. By default, only 'publish' is allowed.
+	 *
+	 * @see wp_parsely_track_status_slugs
+	 * @since 2.5.0
```
GaryJones:
```suggestion
	 * Compare the post_status key against an allowed list (by default, only 'publish'ed content includes tracking data).
	 *
	 * @since 2.5.0
	 *
	 * @param int|WP_Post $post Which post object or ID to check.
	 * @return bool Should the post status be tracked for the provided post's post_type. By default, only 'publish' is allowed.
```

wp-parsely.php
```diff
@@ -768,6 +768,28 @@ public function insert_parsely_page() {
...
+	 */
+	public static function post_has_trackable_status( $post ) {
+		/**
+		 * Filters the status slugs that are permitted to be tracked.
+		 * By default, the only slug allowed is 'publish'. Override this list if your post type necessitates.
```
GaryJones:
```suggestion
		 * Filters the statuses that are permitted to be tracked.
		 *
		 * By default, the only status tracked is 'publish'. Use this filter if you have other published content that has a different (custom) status.
```

wp-parsely.php
```diff
@@ -768,6 +768,28 @@ public function insert_parsely_page() {
...
+		 *
+		 * @since 2.5.0
+		 *
+		 * @param string[]    $status_slugs The list of post_status slugs that are allowed to be tracked.
+		 * @param int|WP_Post $post         Which post object or id is being checked.
+		 */
+		$status_slugs = apply_filters( 'wp_parsely_track_status_slugs', array( 'publish' ), $post );
+		return in_array( get_post_status( $post ), $status_slugs, true );
```
GaryJones:
```suggestion
		 * @param string[]    $trackable_statuses The list of post statuses that are allowed to be tracked.
		 * @param int|WP_Post $post               Which post object or ID is being checked.
		 */
		$statuses = apply_filters( 'wp_parsely_trackable_statuses', array( 'publish' ), $post );
		return in_array( get_post_status( $post ), $statuses, true );
```

Despite there being a `get_post_stati()` function, I think `statuses` is more accessible to folks.

Removed references to "status slugs" as it's the terminology used in WP - there's no "slug" for it - the key is just is the "post status" (label is a separate thing).

GaryJones (changes requested):

wp-parsely.php
jblz: Added a cache in 6a6349c.  Since this function accepts a post and can be re-run with another, I keyed the static cache on post_id.  Let me know what you think, @GaryJones

GaryJones: @jblz Would it make more sense to have it keyed on the post_type instead?

Under what circumstances could a `foo` post type have different post statuses registered, for different instances/IDs of `foo`?

`register_post_status()` registered a status without assignment to any particular post type, but if https://core.trac.wordpress.org/ticket/12706 ever got completed, I think it would be associated with a post type, rather than individual instances of posts.

jblz: Good shout, @GaryJones -- updated to key the cache on post status instead.

GaryJones (commented):
Approvable, but I wanted to put a thought out:

Right now, the existing logic only looks at the post status (i.e. `publish`), to determine if the item is trackable or not. This is great, as that's what posts and pages both use, and all post types that use this status intend to have the same behaviour in terms of being tracked or not.

But if I, or a third-party plugin, register a custom status like `archive`, then I may want that to be trackable for one post type, and not for another (think, an archive of historical newsletters that we want to still get visitor traffic on, vs archived articles/site that should no longer be accessible).

With the `$post` passed to the filter, I can do conditionally change whether the post status gets marked as tracked or not, but with the `$cache` only looking at the post status, whichever post type got hit first would populate the cache and the other post types would be incorrectly tracked or not.

I kind of want to see it keyed as `$cache[ $post_type ][ $post_status ]`. This may just be excessive for one unlikely use case (a clash of custom statuses from a developer or plugins, that should have different behaviour), but then again, changing how data is saved in `$cache` isn't a big deal, and if it can avoid having to change this new filter in the future, it may be worth it now.

Compare:

```php
$cache = [
	'publish' => true,
	'private' => false,
	'archive' => true,
];
```
and

```php
$cache = [
	'newsletter' => [
		'publish' => true,
		'private' => false,
		'archive' => true,
	],
	'secret-stash' => [ // It's secret: don't track for any status.
		'publish' => false,
		'private' => false,
		'archive' => false,
	]
];
```

The offset would be in making the filter callback a little more complex.

Worth going the extra step now, or not?

jblz:
@GaryJones that sort of extensibility is why I initially had the cache keyed by post_id alone...

IMO, the main purpose of this static cache is to prevent needlessly re-running any added filter functions when called multiple during a single request cycle for a given post (as is the case when [this conditional](https://github.com/Parsely/wp-parsely/blob/cf64f16ca810af767d978a1758d16b87399e4bbd/wp-parsely.php#L937) hits, but the [previous one](https://github.com/Parsely/wp-parsely/blob/cf64f16ca810af767d978a1758d16b87399e4bbd/wp-parsely.php#L858) misses.

Keying the cache on post_id lets us push the complexity into the filtering functions if they need to do something as particular as the example you illustrated but maintains the performance considerations for the vast majority of cases...

What do you think?

GaryJones:
> What do you think?

For a single HTTP request, could there ever be more than one _ID_ requested that metadata needs to be constructed for?

Equally, for a single HTTP request, could there ever be more than one _post type_ requested that metadata needs to be constructed for?

I'm trying to think of a case where multiple metadata need to be composed - maybe some batch request or retrieval of posts in the REST API, for a decoupled situation? 🤔 

In each case, I think the recurring-ness of needing to construct metadata would be minimal (presumably it should only be once per HTTP response), so post ID would likely suffice, or maybe it doesn't even need keying?

jblz:
> I'm trying to think of a case where multiple metadata need to be composed - maybe some batch request or retrieval of posts in the REST API, for a decoupled situation? 🤔

Yeah, something like this.  If the function takes an argument allowing it to be called against distinct values for `$post`, we should honor that in the cache.

I see two simple ways forward:
* Reinstate the cache-by-post-id only
* Don't allow for passing a `$post` param to `post_has_trackable_status` and assume the "current" global post object (& do away with the cache key entirely)

I vote the former, but don't feel super strongly.

GaryJones:
> Reinstate the cache-by-post-id only

✅

GaryJones (approved):

PR #239, opened 2021-04-19
Try: React Components in the Admin Script

jblz:
WIP!

This is merely a demo of how we can use react components in the admin page.

PR #243, opened 2021-04-21
Try: REST API Endpoint for Settings

jblz:
Super quick and dirty REST API Endpoint to expose the plugin options / settings.

## To access

Make a request to `/wp-json/wp-parsely/v1/settings`

### DIFF
--- a/settings-api.php
+++ b/settings-api.php
@@ -1,0 +1,16 @@
+<?php
+
+function wp_parsely_settings_api_init() {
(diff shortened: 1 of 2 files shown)


jblz:
going ahead and closing this since it got pulled into #239

We'll still probably want to pull it back out into its own PR with tests once we're happy with it, but no sense in leaving this open in the meantime.

PR #251, opened 2021-04-23
Admin page: Improve page title

### LINKED ISSUE #222: Make settings page heading a H1 instead of H2
Better for accessibility, WordPress changed this for wp-admin pages quite a few versions ago.

Correct internationalization of the _contents_ are being handled separately.

GaryJones:
Use a H1 instead of H2 - it's better for accessibility, and WordPress changed this for wp-admin pages quite a few versions ago.

Also, move the Version span outside of the heading.

There is no visual style change, but it fixes the semantics.

Before:

Screenshot 2021-04-23 at 14 36 19

After first commit (switch to h1):

Screenshot 2021-04-23 at 14 36 27

After second commit (move version number out):

Screenshot 2021-04-23 at 14 40 12

After all commits:

Screenshot 2021-04-23 at 14 41 16

Fixes #222.

### DIFF
--- a/parsely-settings.php
+++ b/parsely-settings.php
@@ -14,7 +14,7 @@ $version_string = sprintf( __( 'Version %s', 'wp-parsely' ), $this::VERSION );
 ?>
 
 <div class="wrap">
-	<h2><?php echo esc_html( get_admin_page_title() ); ?> <span id="wp-parsely_version"><?php echo esc_html( $version_string ); ?></span></h2>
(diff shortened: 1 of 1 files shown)


jblz (approved):
Nice. Today I learned about `.wp-heading-inline`

GaryJones:
> Today I learned about `.wp-heading-inline`

It was added when there was an accessibility drive on admin page headings, and they needed to pull out the "Add New" links on the Posts and Pages list pages.

PR #252, opened 2021-04-23
Remove redundant plugin version meta tag

### LINKED ISSUE #244: Remove wp-parsely_version meta tag
```html
<meta name="wp-parsely_version" id="wp-parsely_version" content="2.4.1"/>
```

The meta tag only exists to tell Parse.ly support what version of the plugin is being used. The crawler doesn't use it.

The version number is also in the HTML comment that wraps the meta tag + structured data element anyway, so it is redundant.

We should also remove it from any documentation as well.

GaryJones:
The meta tag only existed to tell Parse.ly support what version of the plugin was being used. The crawler doesn't use it.

The version number is also in the HTML comment that wraps the meta tag + structured data element anyway, so it is redundant.

Fixes #244.

### DIFF
--- a/parsely-parsely-page.php
+++ b/parsely-parsely-page.php
@@ -11,7 +11,7 @@
 
 ?>
 
-<!-- BEGIN wp-parsely Plugin Version <?php echo esc_html( Parsely::VERSION ); ?> -->
+<!-- BEGIN Parse.ly <?php echo esc_html( Parsely::VERSION ); ?> -->
 <?php if ( ! empty( $parsely_page ) && isset( $parsely_page['headline'] ) ) : ?>
 	<?php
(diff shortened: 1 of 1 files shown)


jblz (approved):
LGTM. I still see the version number in the wrapping HTML comment &, per this discussion, it's not being used:

p1619099084048900-slack-C01MHPS2T4J

GaryJones:
I think it's worth keeping an easily accessible version number in the HTML comment until Support have a better way of identifying the plugin version being used.

What they didn't need was the version number mentioned more than once. I've added a second commit which tidies the wrapping HTML comment a little.

GaryJones:
Closing in favour of #257.

PR #253, opened 2021-04-23
Admin page: Improve accessibility of radio inputs

### LINKED ISSUE #228: Add fieldset and legend to radio settings, and make vertical
Rearrange the markup for radio button output to match WP core to improve accessibility and usability.

## Existing appearance and markup

Use of unsemantic divs, input-label-input-label with no fieldset grouping, and no alignment.

Screenshot 2021-04-16 at 00 27 47
Screenshot 2021-04-16 at 00 25 51

## Core setting appearance and markup

Core has a similar setting which uses radio buttons, but it groups it with a fieldset, adds a screen-reader-text legend for accessibility purposes, and stacks the options vertically, which makes it much more obvious which radio button goes with which label.

Screenshot 2021-04-16 at 00 23 25
Screenshot 2021-04-16 at 00 22 36

GaryJones:
- Groups related radio buttons in `fieldset` element instead of `div`.
- Adds `legend` element to give context to what the radio button values mean. Populated with same string as used in the visual appearance of faux-label.
- Fixes bug where only help text or requires recrawl data attributes were added, and consequently fixes recrawl notice from appearing when `requires_recrawl=false`.
- Displays radio buttons vertically, which makes it more obvious which radio button goes with which label.
- Switched the HTML-in-PHP-strings to be HTML-with-embedded-PHP, which helps with syntax highlighting and auto-completion in IDEs, and more focus on ensuring dynamic values are escaped.
- Fixed the recrawl message JS to not be so specific about which element the `.parsely-form-controls` class is on, because `div` is not always the best grouping element.

Fixes #228.

Other than displaying vertically, visually there's no difference. It's a big improvement for screen reader users though, as now the answers have an associated question, and it's cognitively easier to associate the label and the control.

Screenshot 2021-04-23 at 15 38 30

### DIFF
--- a/parsely-admin-header.php
+++ b/parsely-admin-header.php
@@ -39,7 +39,7 @@
 		recrawlRequiredMessage.append(supportLink);
 
 		recrawlRequiredMessage
-			.appendTo("div.parsely-form-controls[data-requires-recrawl='true'] .help-text");
+			.appendTo(".parsely-form-controls[data-requires-recrawl='true'] .help-text");
 	});
 })(jQuery);
 </script>
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -345,6 +345,7 @@ class Parsely {
 			self::MENU_SLUG,
 			'optional_settings',
 			array(
+				'title'            => __( 'Disable JavaScript', 'wp-parsely' ), // Passed for legend element.
 				'option_key'       => 'disable_javascript',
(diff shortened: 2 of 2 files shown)


jblz (approved):
lgtm.

Noting that this will affect the admin script portion of #236 -- I'll fix it up after this lands.

PR #257, opened 2021-04-24
Metadata: Move output into views

GaryJones:
The `parsely-parsely-page.php` in the root of the plugin was a mix of several views and conditional logic about what HTML to output. It wasn't easy to read.

This PR:

- Moves conditional logic from `parsely-parsely-page.php` into the main insert metadata method.
- Moves HTML output from `parsely-parsely-page.php` into three new view files.
  - I considered a `/views/front/` directory on account of there being a future `/views/admin/` directory, but as the admin pages are being re-built in React, then it can be avoided for now.
- Improves the alignment and arrangement of the output for JSON-LD and repeated metas.
- Fixes issues with author and keywords when they are unset (construct metadata method needs better validation) for repeated metadata. They gave PHP Warnings when viewing the homepage, for instance.
- Added spaces before closing `/>` on meta links, in case anyone was trying to use the XML-serialized output of HTML5, and as per WP coding standards.

Incorporates equivalent changes from #252.

Before (homepage, JSON-lD):
Screenshot 2021-04-24 at 17 25 49

After (homepage, JSON-lD):
Screenshot 2021-04-24 at 17 03 20

----

Before (homepage, repeated metas) (including PHP Warning stacktrace):
Screenshot 2021-04-24 at 17 23 31

After (homepage, repeated metas):
Screenshot 2021-04-24 at 17 02 32

----

Before (post, JSON-LD):
Screenshot 2021-04-24 at 17 25 25

After (post, JSON-LD):
Screenshot 2021-04-24 at 17 03 07

----

Before (post, repeated metas):
Screenshot 2021-04-24 at 17 22 31

After (post, repeated-metas):
Screenshot 2021-04-24 at 17 02 21

----

Further ideas for improvement:

 - use template part approach to allow themes and child themes to override these view files in the theme for further customization.
 - Don't output repeated metas that have empty strings, if the crawler doesn't need them (unsure - @hbbtstar?).


wp-parsely.php
```diff
@@ -764,7 +764,33 @@ public function insert_parsely_page() {
...
+		// Something went wrong - abort.
+		if ( empty( $parsely_page ) || ! isset( $parsely_page['headline'] ) ) {
+			return;
+		}
+
+		echo "\n" . '<!-- BEGIN Parse.ly ' . esc_html( Parsely::VERSION ) . ' -->' . "\n";
+
+		// Insert JSON-LD or repeated metas.
+		if ( 'json_ld' === $parsely_options['meta_type'] ) {
+			include __DIR__ . '/views/json-ld.php';
```
jblz, line 784:
```suggestion
			require __DIR__ . '/views/json-ld.php';
```

I usually tend toward `require` if it's a file that ships w/ my code that I expect to be present.  Is there a good reason to use `include` instead?

wp-parsely.php
```diff
@@ -764,7 +764,33 @@ public function insert_parsely_page() {
...
+		} else {
+			$parsely_post_type = 'NewsArticle' === $parsely_page['@type'] ? 'post' : 'sectionpage';
+			if ( is_array( $parsely_page['keywords'] ) ) {
+				$parsely_page['keywords'] = implode( ',', $parsely_page['keywords'] );
+			}
+
+			include __DIR__ . '/views/repeated-metas.php';
```
jblz, line 791:
```suggestion
			require __DIR__ . '/views/repeated-metas.php';
```

wp-parsely.php
```diff
@@ -764,7 +764,33 @@ public function insert_parsely_page() {
...
+		}
+
+		// Add any custom metadata.
+		if ( isset( $parsely_page['custom_metadata'] ) ) {
+			include __DIR__ . '/views/custom-metadata.php';
```
jblz, line 796:
```suggestion
			require __DIR__ . '/views/custom-metadata.php';
```

jblz (commented):
I love splitting this out into views, but I'm not super keen on adding more code and logic to the `insert_parsely_page` function / already-large class file.

What do you think about reinstating a top-level template file instead?

wp-parsely.php
GaryJones: `require` will cause a fatal if it's not present / accessible, while `include` just throws a warning. A lack of Parse.ly output (JSON-LD, repeated metas, custom meta), isn't a reason to kill the whole page/request IMO.

The files are expected to be present and accessible, but it's not critical to the success of the page.

GaryJones:
> I love splitting this out into views, but I'm not super keen on adding more code and logic to the `insert_parsely_page` function / already-large class file.

There will already need to be work to split this out into a new class, so this change consolidates the logic into a _cohesive_ block, and allows for more of a separation of concerns between PHP and HTML. We have other open PRs which add to the large file/class, and the work to start extracting into other files/classes hasn't started yet (and shouldn't for this PR, which is focused on the view files).

wp-parsely.php
jblz: I generally prefer to fail by design when sourced plugin components are not present -- it can make diagnosing the issue substantially more difficult in practice

That said, I understand the point of not wanting to break the front-end of a site in such cases, so let's leave `include` in place for these.

I definitely think we should default to `require` for any admin views, however, or at least manually handle cases where `include`d files are not present instead of failing silently.

jblz (approved):

wp-parsely.php
GaryJones: Agreed for admin views - the purpose of viewing an admin page is to see/access the content of it, so a missing view file there _is_ a critical issue.

PR #258, opened 2021-04-26
Prevent Unit Tests from echoing

### LINKED ISSUE #180: Stop tests from echoing

mikeyarce:
Fixes https://github.com/Parsely/wp-parsely/issues/180

Reworks the following unit tests:
-  test_parsely_ppage_output
-  test_parsely_categories
-  test_parsely_tags_lowercase
-  test_parsely_cats_as_tags
-  test_custom_taxonomy_tags
-  test_use_top_level_cats
-  test_custom_taxonomy_as_section
-  test_top_level_taxonomy_as_section
-  test_http_canonicals
-  test_fbia_integration
-  test_amp_integration
-  test_parsely_page_filter
-  test_user_logged_in
-  test_user_logged_in_multisite

The only unit test that still echo's output is `test_parsely_tag`.  The way the JS is going to be included is going to change, so this unit test needs to be reworked along with the other work in progress.

## Changes:
- Most unit tests no longer echo.
- Instead of using `insert_parsely_page` and checking the echo'd out strings, we're now using the `construct_parsely_metadata` function and checking the returned data from that function.
- For the `force_https_canonicals` test, we're now checking the scheme using `wp_parse_url()` instead of using `strpos`.

### DIFF
--- a/tests/all-test.php
+++ b/tests/all-test.php
@@ -248,7 +248,7 @@ PARSELYJS;
 	}
 
 	/**
-	 * Check out taxonomy tags.
+	 * Test custom taxonomy terms, categories, and tags in the metadata.
 	 *
 	 * @category   Function
 	 * @package    SampleTest
--- a/tests/all-test.php
+++ b/tests/all-test.php
@@ -221,7 +221,7 @@ PARSELYJS;
 	 * @category   Function
 	 * @package    SampleTest
 	 */
-	public function test_parsely_cats_as_tags() {
+	public function test_parsely_categories_as_tags() {
 		// Setup Parsley object.
 		$parsely         = new \Parsely();
 		$parsely_options = get_option( \Parsely::OPTIONS_KEY );
--- a/tests/all-test.php
+++ b/tests/all-test.php
@@ -135,14 +135,29 @@ PARSELYJS;
 	 * @package    SampleTest
 	 */
 	public function test_parsely_ppage_output() {
+		// Setup Parsley object.
+		$parsely         = new \Parsely();
+		$parsely_options = get_option( \Parsely::OPTIONS_KEY );
+
+		// Create a single post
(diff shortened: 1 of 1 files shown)


GaryJones (changes requested):
As discussed in Slack, there is still lots to address with these tests overall, but not having them be so noisy is a great thing!

I would like to see that final JS-related test addressed as well, even if there are other PRs that touch this test, as then the [PHPUnit config can be updated to be strict about output (new commit)](https://phpunit.readthedocs.io/en/7.5/risky-tests.html#output-during-test-execution). Once that's done, this PR is good to go.

GaryJones (approved):
Going to approve this and get this in since what's done is a good step forward, and there's nothing wrong with it.

Issue #174, opened 2021-03-30
Use correct protocol for canonical URLs


GaryJones:
There is currently a [Force HTTPS Canonicals setting](https://github.com/Parsely/wp-parsely/blob/c061f0b92b17d62f32570b007de07628977b7055/wp-parsely.php#L494-L505), that changes the protocol for URLs in the structured metadata, independently of whether the home_url() uses `https` or not.

By default, this setting is set to false - that is, `http` is used - even when the site uses `https` in the home URL.

This has the effect that a WP instance at `https://example.com` will get URLs that start with `http://example.com` which therefore makes it incorrect as being the "canonical URL", since canonical URLs need to take into account the protocol as well.

A customer support ticket implies that the Yoast SEO plugin seems to want to use the correct protocol (as would make sense), but this needs confirming.

Existing customers will have their data indexed under `http` (unless they've already changed the setting), so implementing behaviour that checks the `home_url()` and adjusts the protocol used in canonical URLs, will mean having to do a re-index.

As such, any change in behaviour here should only be for new installs (check for a version number saved in the options, and only change the behaviour if the version is present and equal to or greater than the release in which the change is implemented).

Alternatively, this change is made for everyone, but all of the relevant sites will need a reindex (can that be notified / triggered remotely?). On a site with 400-500k URLs, this could take around a day.

When this new behaviour is implemented (canonical URL protocols match the home_url() protocol), then the Force HTTPS Canonicals setting can be hidden. It could even be considered deprecated and marked for removal in a future major release.

One additional thing would to then add an admin notice for when the home URL is changed on the Settings->General screen, to advise that a Parse.ly reindex is needed.

htdat:
@GaryJones - I am proposing another idea. What do you think?

> This has the effect that a WP instance at https://example.com will get URLs that start with http://example.com which therefore makes it incorrect as being the "canonical URL", since canonical URLs need to take into account the protocol as well.

What's about if we set the correct option of `Force HTTPS Canonicals setting` for new installs? E.g: 
- If `home_url()` uses HTTPS, set `Force HTTPS Canonicals setting` to `true`.
- Otherwise, set `Force HTTPS Canonicals setting` to `false`.

With that, we do not need to handle different implementations for different versions? 

> When this new behaviour is implemented (canonical URL protocols match the home_url() protocol), then the Force HTTPS Canonicals setting can be hidden. It could even be considered deprecated and marked for removal in a future major release.
> 
[...]

PR #260, opened 2021-04-29
Fix: Inconsistent ports in canonical URLs

### LINKED ISSUE #198: Inconsistent ports in canonical URLs
When using the plugin on an archive like author or category on a domain like `http://localhost:10024/`, then the canonical URLs show the port twice. Example output:

```json
{"@context":"http:\/\/schema.org","@type":"WebPage","headline":"Author - admin","url":"http:\/\/localhost:10024:10024\/author\/admin\/"}	
```

On a single Post, then the port doesn't appear:

```json
{"@context":"http:\/\/schema.org","@type":"NewsArticle","mainEntityOfPage":{"@type":"WebPage","@id":"http:\/\/localhost\/post-2\/"},"headline":"Post 2","url":"http:\/\/localhost\/post-2\/","thumbnailUrl":"","image":{"@type":"ImageObject","url":""},"dateCreated":"2021-04-06T23:27:26Z","datePublished":"2021-04-06T23:27:26Z","dateModified":"2021-04-06T23:27:26Z","articleSection":"Uncategorized","author":[{"@type":"Person","name":"admin"}],"creator":["admin"],"publisher":{"@type":"Organization","name":"wp-parsely","logo":""},"keywords":[]}
```

On the homepage, it shows it correctly:

```json
{"@context":"http:\/\/schema.org","@type":"WebPage","headline":"wp-parsely","url":"http:\/\/localhost:10024"}
```

This suggests that the `get_current_url()` may need looking at, and tests added.

htdat:
Fix #198 

## Description 

This PR addresses inconsistent port issues for get_current_url() and a few other things: 

- Simplify the way we construct the current URL and respect the `force_https_canonicals` option. This is just a string replacement rather than take care of checking and concatenating scheme, host, URI, and port. 
- Change this function from `private` to `public` so that it can be tested with PHPUnit. 
- Add tests for this function.

Though this seems quite a lot of changes, I think it's worthwhile as this PR simplifies things and has tests. 

#### For the main issue #198 with the port number:

1. A simple way I found is to use this npm package https://developer.wordpress.org/block-editor/reference-guides/packages/packages-env/#quick-tldr-instructions
2. After that, visit the homepage, a post page, a few other pages (category, tag, author), and ensure that URLs are correct in the meta value `@context`.

#### For normal sites (without port number):

Basically, follow up step 2 above.


tests/all-test.php
```diff
@@ -591,4 +591,87 @@ public function test_user_logged_in_multisite() {
...
+		$options                           = get_option( \Parsely::OPTIONS_KEY );
+		$options['force_https_canonicals'] = $force_https;
+		update_option( \Parsely::OPTIONS_KEY, $options );
+
+		update_option( 'siteurl', $url );
+
+		// Test homepage
+		$this->go_to( '/' );
+		$res = self::$parsely->get_current_url();
+		self::assertSame( strpos( $res, $expected, 0 ), 0 );
```
GaryJones: Use a more specific assertion - https://phpunit.readthedocs.io/en/7.5/assertions.html#assertstringstartswith

tests/all-test.php
```diff
@@ -591,4 +591,87 @@ public function test_user_logged_in_multisite() {
...
+
+		// Test a specific post
+		$post_array = $this->create_test_post_array();
+		$post       = $this->factory->post->create( $post_array );
+		$this->go_to( '/?p=' . $post );
+		$res = self::$parsely->get_current_url();
```
GaryJones: You're testing against a post, but the default first argument of the method is `nonpost`. Seems like this should be `post`, along with the `$post` ID (rename variable as $post_id for clarity). Without `post` as the first arg value, there's a whole section of the `get_current_url()` method that isn't being covered in the tests.

Example:

```suggestion
		$post_id    = $this->factory->post->create( $post_array );
		$this->go_to( '/?p=' . $post_id );
		$res = self::$parsely->get_current_url( 'post', $post_id );
```

wp-parsely.php
```diff
@@ -1649,29 +1648,19 @@ private function get_current_url( $post = 'nonpost', $post_id = 0 ) {
...
-				$canonical = $scheme . sanitize_text_field( wp_unslash( $_SERVER['HTTP_HOST'] ) ) . sanitize_text_field( wp_unslash( $_SERVER['REQUEST_URI'] ) ); // Input var okay.
-			}
+			$url = apply_filters( 'wp_parsely_permalink', $permalink, $post, $post_id );
+		} else {
+			$request_uri = isset( $_SERVER['REQUEST_URI'] )
+					? sanitize_text_field( wp_unslash( $_SERVER['REQUEST_URI'] ) )
+					: '';
 
-			return $canonical;
+			$url = site_url( $request_uri ); // Input var okay.
```
GaryJones: That comment isn't doing anything.

wp-parsely.php
```diff
@@ -1649,29 +1648,19 @@ private function get_current_url( $post = 'nonpost', $post_id = 0 ) {
...
 			 */
-			$permalink        = apply_filters( 'wp_parsely_permalink', $permalink, $post, $post_id );
-			$parsed_canonical = wp_parse_url( $permalink );
-			// handle issue if wp_parse_url doesn't return good host & path data, fallback to page url as a last resort.
-			if ( isset( $parsed_canonical['host'], $parsed_canonical['path'] ) ) {
-				$canonical = $scheme . $parsed_canonical['host'] . $parsed_canonical['path'];
-			} elseif ( isset( $_SERVER['HTTP_HOST'], $_SERVER['REQUEST_URI'] ) ) { // Input var okay.
-				$canonical = $scheme . sanitize_text_field( wp_unslash( $_SERVER['HTTP_HOST'] ) ) . sanitize_text_field( wp_unslash( $_SERVER['REQUEST_URI'] ) ); // Input var okay.
-			}
+			$url = apply_filters( 'wp_parsely_permalink', $permalink, $post, $post_id );
```
GaryJones, line 1649: Can this filter be made available to the other branch as well? Does it make to do that?

wp-parsely.php
```diff
@@ -1631,12 +1631,11 @@ private function get_settings_url() {
 	 * @param int    $post_id id of the post you want to get the url for. Optional.
 	 * @return string|void
 	 */
-	private function get_current_url( $post = 'nonpost', $post_id = 0 ) {
-		$options = $this->get_options();
-		$scheme  = ( $options['force_https_canonicals'] ? 'https://' : 'http://' );
+	public function get_current_url( $post = 'nonpost', $post_id = 0 ) {
+		$url = '';
```
GaryJones: What does this assignment add?

tests/all-test.php
```diff
@@ -591,4 +591,87 @@ public function test_user_logged_in_multisite() {
...
+			array(
+				false,
+				'http://example.com:1234',
+				'http://example.com:1234',
+			),
+			array(
+				false,
+				'https://example.com:1234',
+				'http://example.com:1234',
+			),
```
GaryJones: How about mixing some of these up to test other strings, like `http://localhost:999` or `https://my-domain.local:1000/foo/bar`, since you're updating the site URL anyway?

GaryJones (changes requested):

wp-parsely.php
htdat: I keep it as-is like what the original code is doing.
I've looked at this filter again. It is passing `$post_id`, which is not available in another conditional branch so we should not make it available outside of the current scope.

wp-parsely.php
htdat: I do not have any strong opinion here. I just think it makes reading easier with a declaration somewhere. However, PHPStorm is also marking it's not necessary, then I am removing it.

wp-parsely.php
```diff
@@ -1649,29 +1646,19 @@ private function get_current_url( $post = 'nonpost', $post_id = 0 ) {
...
+			$url = apply_filters( 'wp_parsely_permalink', $permalink, $post, $post_id );
+		} else {
+			$request_uri = isset( $_SERVER['REQUEST_URI'] )
+					? sanitize_text_field( wp_unslash( $_SERVER['REQUEST_URI'] ) )
+					: '';
 
-			return $canonical;
+			$url = home_url( $request_uri );
 		}
-		$page_url = site_url( null, $scheme );
```
htdat, line 1663: Previously, this uses `site_url`. However, since we're dealing with the canonical links (front-end links), not WordPress file links, the correct usage should be `home_url`. 

I am also updating the relevant line in the test file to reflect this change too.

htdat:
Thanks @GaryJones :D 
I've updated this PR to address your valuable feedback. In this iteration, the most major change is from `site_url` to `home_url` as I explained above.

htdat:
I've just noticed we're having this issue too #174. It's directly related to this PR. 
I will continue researching and taking this issue into account.

GaryJones (approved):
Tests look good, changes look good, manual inspection of applied PR on local environment shows URLs displaying correctly 👍🏻

PR #261, opened 2021-04-29
Test Files: Update code style

jblz:
* Prepend class files with `class-`
* Lower-case the `TestCase.php` file
* Update the PHP class name in the "all-test" catch-all file to `All_Test` (as opposed to `SampleTest`)

This is a step at fixing the issues that the CI surfaced at https://github.com/Automattic/vip-go-mu-plugins/pull/2084

## To Test

* Confirm changes make sense in the context of the coding standards
* Automated tests should continue to pass and should match the number of tests and assertions in the upstream branch.


GaryJones (approved):
I'm personally not a fan of these changes (it'll fall apart when Interfaces are added, for instance), but I can see why it's being done. Approving with reluctance :-)

jblz:
> I'm personally not a fan of these changes

+1  We can certainly add some overrides when we get to that point

Issue #96, opened 2018-05-12
Split main plugin file into smaller files


joshuarrrr:
From 2018-05-02 VIP review:

> the main plugin file and class Parsely is 1300 lines long, which is rather large. For future reviews it would reduce churn to break this into smaller files.

PR #262, opened 2021-04-29
Rename the main class file to class-parsely.php & source it in the plugin entry file

jblz:
## Description

Akin to #261, this is an effort to fix the style issues surfaced in Automattic/vip-go-mu-plugins#2084

I wanted to split this one out from that change so we can discuss it independently if desired.

We've discussed doing a lot more to split the main class out, but I think keeping this change small and focused on the naming convention & addressing #96 with focused changes is probably for the best.

### The Change

Since this diff is a bit hard to grok in the web UI, I'll describe what I did:

* `git mv wp-parsely.php class-parsely.php`
* `touch wp-parsely.php`
* Cut the plugin header from the class file
* Paste the plugin header to the new main file
* Add a `require` statement at the bottom of the new main file
* Stage the changes, commit, & push

## To Test

This should really change nothing at all in the way the plugin works. It should register, activate, operate as before. Automated tests should function as before.


class-parsely.php
```diff
@@ -0,0 +1,1813 @@
+<?php
+/**
+ * This is the main class for Parsely
+ *
+ * @category   Class
+ * @package    Parsely
+ */
```
GaryJones: The file-level DocBlock, and the class-level DocBlock should be different things.

wp-parsely.php
```diff
@@ -22,1815 +22,4 @@
...
-
-require 'class-parsely-recommended-widget.php';
-
-add_action( 'widgets_init', 'parsely_recommended_widget_register' );
-/**
- * Register the Parse.ly Recommended widget.
- */
-function parsely_recommended_widget_register() {
-	register_widget( 'Parsely_Recommended_Widget' );
-}
```
GaryJones: This code here is outside of the class, so it should stay in the plugin root file. The new class-parsely.php file should only contain the class, so it can be included without side-effects happening.

GaryJones (changes requested):
Certainly +1 to extracting the [God] class into its own file, but a couple of quibbles to address.

wp-parsely.php
jblz: I'll move it over, but the temptation to refactor at the same time is pretty high :D

class-parsely.php
GaryJones: See https://github.com/GaryJones/plugin-boilerplate/blob/master/src/Plugin.php as an example of the difference.

GaryJones:
A further request - to put `class-parsely.php` into `/src` directory. The `/src` directory is the standard location for PHP source files, as seen at https://github.com/php-pds/skeleton (and the research it links through to).

Now, that means it will mix in with JS files (currently at least some of which will be in `src/lib`), so either:

- the PHP lives in `/src` and the JS moves to `/lib` or `/js`.
- the JS lives in `/src` and the PHP lives in `/php`.
- we mix PHP and JS together (urgh!).

Since the main language of the plugin is PHP, I know which I would prefer :-) 

The more we can reduce the number of active files in the root of the plugin, the better.

class-parsely.php
jblz: Not sure I really follow... I think that means we need to drop the `@category` annotation...?

jblz:
@GaryJones I went ahead and moved the PHP files into the `src/` directory.

I agree we should separate out the JS files, but I'd like to do that in a separate change if that's ok.

GaryJones:
As noted in https://github.com/Parsely/wp-parsely/pull/262/files (scroll to the bottom), there's a plugin root file reference in the tests bootstrap file to change as well.

Screenshot of the GitHub Actions workflow in action!:

Screenshot 2021-04-29 at 17 14 11

jblz:
Thanks for taking another look, @GaryJones ... I messed up the `git mv` and had to go into a demo before I could fix it.  All better now.

Let me know what you'd like to see in the docblock please.

src/class-parsely.php
```diff
@@ -0,0 +1,1795 @@
+<?php
+/**
+ * This is the main class for Parsely
+ *
+ * @category   Class
+ * @package    Parsely
+ */
+class Parsely {
```
GaryJones: The [file-level DocBlock summary](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#6-file-headers) is the **what** (no full stop). `@since` is the version when the file was created.

The [class-level DocBlock summary](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#2-classes) describes the **purpose** of the class (with a full stop). `@since` is the version the class was created, and in this case, an extra line to explain that it moved files.

`@category` is redundant and can be removed. Namespaces can serve a better purpose of programmatically organising code.

```suggestion
<?php
/**
 * Parsely class
 *
 * @package Parsely
 * @since 2.5.0
 */

/**
 * Holds most of the logic for the plugin.
 *
 * @internal This needs splitting up in the future.
 *
 * @since 1.0.0
 * @since 2.5.0 Moved from plugin root file to this file.
 */
class Parsely {
```

GaryJones (approved):

PR #266, opened 2021-04-30
Update the Plugin Version to 2.5.0-alpha & generalize the version test

jblz:
Change the version to `2.5.0-alpha` so it's easier to see in the UI that you're working against the `develop` branch (or some derivative thereof).

Additionally, this invert the canonical source of truth for plugin version to be the `PARSELY_VERSION` PHP constant and mirrors the value to the `Parsely::VERSION` class constant making it easier to keep up-to-date in the top-level file.

Tests are adjusted for the above and to no longer look for a specific version number -- just that the versioning is semver-compliant.


GaryJones:
SemVer recognises the terms `alpha`, `beta`, `rc` (they have higher weighting than custom strings IIRC).

WordPress jumps from a release version to an `alpha` ([example](https://core.trac.wordpress.org/changeset/50427/)).

[Software versioning identifies these](https://en.wikipedia.org/wiki/Software_versioning#Sequence-based_identifiers). Although `-develop` would be permitted under SemVer, it doesn't appear to be "standard".

Do we want to consider a release cycle process that can incorporate these alpha, beta, and RC, rather than a direct 2.4.1 -> 2.5.0-develop -> 2.5.0 -> 2.x.y-develop -> 2.x.y -> ...?

jblz:
> Do we want to consider a release cycle process that can incorporate these alpha, beta, and RC, rather than a direct 2.4.1 -> 2.5.0-develop -> 2.5.0 -> 2.x.y-develop -> 2.x.y -> ...?

I definitely think it makes sense to use terms like "beta" and "RC" for pre-release publications.

If we were to create a plugin package for use by a particular audience, I think we'll need to be intentional about the naming of such.

For example, I would contend we should specify ["build metadata"](https://semver.org/#spec-item-10) as a suffix to the "beta" label (I don't think it very likey we'll package up something we'd call "alpha" for this plugin).

---

"Versions" of the code with a `develop` suffix are not intended to be thought of as a "release."

The intention of this change is to signify that we're running code that's in the `develop` branch of the repository _rather than_ a packaged release / pre-release.

GaryJones:
In which case, at what point would you propose that `-develop` change to `-alpha`?

I don't know when that would be, but the change to `-beta` would be move obvious. 

Calling something `alpha` doesn't necessarily mean it's a releasable package. It means it's not feature-complete - i.e. a work in progress, which is exactly what anything on the `develop` branch is.

jblz:
Updated to `2.5.0-alpha`

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "wp-parsely",
-  "version": "2.4.1",
+  "version": "2.5.0-alpha",
```
GaryJones: Per https://docs.npmjs.com/cli/v7/configuring-npm/package-json#version `version` field is optional if there are no plans to publish it.

Can we consider adding the [private field](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#private), to stop it being published, and then remove this `version` field, so there is one less place to update the version on each release?

GaryJones (approved):
👍🏻 Approved, with one suggestion which may or may not be suitable to be done in this PR.

PR #268, opened 2021-05-01
i18n: Load plugin text domain

GaryJones:
The plugin supports earlier than WP 4.6, so we have to explicitly load the plugin text domain.

Since there are no plans to include language files within the plugin, as the plugin is available on WordPress.org along with the translations, then no fallback directory is provided for where to look for the language files, other than the default of the WP_LANG_DIR . /plugins/ directory, and the root of the plugin.

See #225.


jblz (approved):
lgtm

PR #269, opened 2021-05-01
Composer: Normalize and improve composer.json

GaryJones:
Note: As there's a change of indentation in this PR, then it will be easier to review the individual commits.

Using https://github.com/ergebnis/composer-normalize, it creates a standard order of items.

In this case, only the scripts were not in alphabetical order.

Also: The standard for WordPress, as confirmed by the .editorconfig for this plugin, is for JSON files to be indented with 4-space tabs, and not the two spaces that it had.

This will make it less frustrating to edit in compatible editors.

Also: Add script descriptions, for when commands are listed with `composer list`. Also renamed a couple of commands to better indicate that they only existed for the CI.


jblz (approved):
👍

PR #270, opened 2021-05-03
Add filter wp_parsely_post_type for NewsArticle

### LINKED ISSUE #221: Allow NewsArticle schema type to be changed with a filter
From https://wordpress.org/support/topic/ldjson-issue/:

> We’ve been using this plugin for content which is not news content.
>
> So, in my copy of the plugin, I’ve changed all instances of NewsArticle to Article.
>
> This… probably means a future problem for me, when the plugin gets updated. But, for now, I thought I should at least put some visibility on the issue…

There are a [few instances of `NewsArticle`](https://github.com/Parsely/wp-parsely/search?q=NewsArticle) that are hard-coded into the plugin.

While `NewsArticle` makes sense as the default, offering a filter here with a post ID and post type would allow a customer to amend what schema `@type` is used on a granular basis. The `after_set_parsely_page` filter is only be able to tackle some instances.

htdat:
Fix #221

## Description 

Adding a filter to `NewsArticle` is easy. However, the challenging part is to make sure that if any value different from `NewsArticle` is assigned to `@type`, it would be handled correctly. 

Therefore, this is what I've done: 

- Declare which types that Parsely supports based on this info https://www.parse.ly/help/integration/jsonld#distinguishing-between-posts-and-pages
- Handle the conversion from JSON-LD types to respective Parsely page types
- Add the `wp_parsely_post_type` filter. 
- Validate the new @type and fall back it to `NewsAritcle`
- Add tests for this filter.

Side-effect change: 
- I added a commit to update the correct path for `views` files. See my comment below. 

## Test 1 

1. Add this filter to your site. Instead of `Article`, you can use any type mentioned in https://www.parse.ly/help/integration/jsonld#distinguishing-between-posts-and-pages

```php
		add_filter('wp_parsely_post_type', function() {
			return 'Article';
		});
``` 

2. Visit a post, and check its HTML source.
3. Verify this value `"@type":"Article"` in the HTML source. 

## Test 2 

1. Add this filter

```php
		add_filter('wp_parsely_post_type', function() {
			return 'Wrong_Type';
		});
``` 
2. Visit a post, and check its HTML source.
3. Verify this value `"@type":"NewsArticle"` in the HTML source. 
4. There is an error message in the log regarding this wrong type.

## Update 

_Last update: 2020-May-04 05.00 UTC_
To move this PR forward, I am waiting for 3 changes applied in the `develop` branch: 

- [x] #276 - Fix paths of view files.
- [ ] #272 - Separate test cases
- [ ] Gary's work on "PHPUnit 9 items even while using PHPUnit 7.5" so that I don't need to use the old way to test PHP error.


tests/class-all-test.php
```diff
@@ -48,6 +49,7 @@ public function setUp() {
 			'lowercase_tags'            => true,
 			'track_post_types'          => array( 'post' ),
 			'track_page_types'          => array( 'page' ),
+			'logo'                      => '',
```
htdat: Without this value, there will be a warning regarding this line. 
https://github.com/Parsely/wp-parsely/blob/ffc769f1858e886d9aded9e1a36c691d614add61/src/class-parsely.php#L1009-L1009

I think that's because the default value here is not the same as the main class
https://github.com/Parsely/wp-parsely/blob/ffc769f1858e886d9aded9e1a36c691d614add61/src/class-parsely.php#L33

tests/class-all-test.php
```diff
@@ -813,4 +815,43 @@ public function test_get_current_url( $force_https, $url, $expected ) {
...
+		});
+
+		/**
+		 * Ideally we use two methods expectWarning and expectWarningMessageMatches to test this error
+		 * But they're not available until PHPUnit 8.4 while here we're still running PHPUnit 7.x
+		 * @see 7.5 https://phpunit.readthedocs.io/en/7.5/writing-tests-for-phpunit.html#testing-php-errors
+		 * @see 8.4 https://phpunit.readthedocs.io/en/8.4/writing-tests-for-phpunit.html#testing-php-errors-warnings-and-notices
+		 */
+		self::expectException( PHPUnit_Warning::class );
+		self::$parsely->construct_parsely_metadata( $options, $post_obj );
```
htdat: I just think it's important to test this case. 
I am not sure if I am making this over-complicated? Or should I just simple test against `NewsArticle` as it's the fall-back value for this case: 

https://github.com/Parsely/wp-parsely/blob/ffc769f1858e886d9aded9e1a36c691d614add61/src/class-parsely.php#L971-L972

src/class-parsely.php
```diff
@@ -767,19 +800,19 @@ public function insert_parsely_page() {
 
 		// Insert JSON-LD or repeated metas.
 		if ( 'json_ld' === $parsely_options['meta_type'] ) {
-			include __DIR__ . '/views/json-ld.php';
+			include plugin_dir_path( __DIR__ ) . '/views/json-ld.php';
```
htdat: I've updated this as it seems a regression of #262

jblz:
> Side-effect change:     I added a commit to update the correct path for views files. See my comment below.

Addressed separately in #276

src/class-parsely.php
```diff
@@ -915,7 +948,31 @@ public function construct_parsely_metadata( array $parsely_options, $post ) {
...
+			 *
+			 * @param array   $jsonld_type  JSON-LD @type value, default is NewsArticle
+			 * @param integer $id           Post ID
+			 * @param string  $post_type    Post type in WordPress
+			 */
+			$type          = apply_filters( 'wp_parsely_post_type', 'NewsArticle', $post->ID, $post->post_type );
+			$allowed_types = array_merge( $this->allowed_jsonld_post_types, $this->allowed_jsonld_non_post_types );
+
+			// Validate type before passing it further
+			// as an valid type will not be recognized by Parsely
```
```suggestion
			// Validate type before passing it further as an invalid type will not be recognized by Parse.ly.
```

src/class-parsely.php
```diff
@@ -915,7 +948,31 @@ public function construct_parsely_metadata( array $parsely_options, $post ) {
...
 
-			$parsely_page['@type']            = 'NewsArticle';
+			/**
+			 * Filters the JSON-LD @type.
+			 *
+			 * @since 2.5.0
+			 *
+			 * @param array   $jsonld_type  JSON-LD @type value, default is NewsArticle
+			 * @param integer $id           Post ID
+			 * @param string  $post_type    Post type in WordPress
```
```suggestion
			 * @param array   $jsonld_type  JSON-LD @type value, default is NewsArticle.
			 * @param integer $id           Post ID.
			 * @param string  $post_type    Post type in WordPress.
```

src/class-parsely.php
```diff
@@ -915,7 +948,31 @@ public function construct_parsely_metadata( array $parsely_options, $post ) {
...
+			 */
+			$type          = apply_filters( 'wp_parsely_post_type', 'NewsArticle', $post->ID, $post->post_type );
+			$allowed_types = array_merge( $this->allowed_jsonld_post_types, $this->allowed_jsonld_non_post_types );
+
+			// Validate type before passing it further
+			// as an valid type will not be recognized by Parsely
+			if ( ! in_array( $type, $allowed_types ) ) {
+				$error = sprintf(
+					__( '@type %1$s is not supported by Parsely. Please use a type mentioned in %2$s', 'wp-parsely' ),
```
```suggestion
					__( '@type %1$s is not supported by Parse.ly. Please use a type mentioned in %2$s', 'wp-parsely' ),
```

src/class-parsely.php
```diff
@@ -1788,4 +1841,26 @@ public function parsely_is_user_logged_in() {
...
+	public function convert_jsonld_to_parsely_type( $type ) {
+		if ( in_array( $type, $this->allowed_jsonld_post_types ) ) {
+			return 'post';
+		}
+
+		if ( in_array( $type, $this->allowed_jsonld_non_post_types ) ) {
+			return 'index';
+		}
+
+		// Fall-back when
```
GaryJones: Missing end of the sentence?

src/class-parsely.php
```diff
@@ -1788,4 +1841,26 @@ public function parsely_is_user_logged_in() {
...
+	 */
+	public function convert_jsonld_to_parsely_type( $type ) {
+		if ( in_array( $type, $this->allowed_jsonld_post_types ) ) {
+			return 'post';
+		}
+
+		if ( in_array( $type, $this->allowed_jsonld_non_post_types ) ) {
+			return 'index';
+		}
+
```
GaryJones: If the fallback is `index`, then it seems like this conditional could be removed completely.

```suggestion
```

tests/class-all-test.php
```diff
@@ -813,4 +815,43 @@ public function test_get_current_url( $force_https, $url, $expected ) {
 		$res = self::$parsely->get_current_url();
 		self::assertStringStartsWith( $expected, $res );
 	}
+
```
GaryJones, line 984: See where #272 is heading, and move these tests to their own file. You can then avoid / set sensible defaults in the `setUp()` or per test method.

tests/class-all-test.php
```diff
@@ -813,4 +815,43 @@ public function test_get_current_url( $force_https, $url, $expected ) {
...
+	public function test_filter_wp_parsely_post_type() {
+		$options = get_option( \Parsely::OPTIONS_KEY );
+
+		$post_array = $this->create_test_post_array();
+		$post_id    = $this->factory->post->create( $post_array );
+		$post_obj   = get_post( $post_id );
+		$this->go_to( '/?p=' . $post_id );
+
+		// Try to change the post type to an allowed value - BlogPosting
+		add_filter('wp_parsely_post_type', function() {
```
GaryJones: Run these changes through PHPCS.

tests/class-all-test.php
```diff
@@ -813,4 +815,43 @@ public function test_get_current_url( $force_https, $url, $expected ) {
...
+		// Try to change the post type to a non-allowed value - Not_Allowed
+		add_filter('wp_parsely_post_type', function() {
+			return 'Not_Allowed_Type';
+		});
+
+		/**
+		 * Ideally we use two methods expectWarning and expectWarningMessageMatches to test this error
+		 * But they're not available until PHPUnit 8.4 while here we're still running PHPUnit 7.x
+		 * @see 7.5 https://phpunit.readthedocs.io/en/7.5/writing-tests-for-phpunit.html#testing-php-errors
+		 * @see 8.4 https://phpunit.readthedocs.io/en/8.4/writing-tests-for-phpunit.html#testing-php-errors-warnings-and-notices
```
GaryJones: I'm currently working on making changes so that we can use PHPUnit 9 items even while using PHPUnit 7.5.

GaryJones (changes requested):

tests/class-all-test.php
htdat: Yup, thanks Gary!

tests/class-all-test.php
htdat: @GaryJones That's great! Do you think I should wait for that or just commit "my way" now?

htdat:
I've added the _Update_ section in the description of this PR to clarify what I am waiting in the `develop` branch.

tests/class-all-test.php
GaryJones: Go for your way now - we can re-review after.

GaryJones (approved):

GaryJones:
Needs rebasing and merge conflicts addressed.

htdat:
> Needs rebasing and merge conflicts addressed.

@GaryJones - this is done. Would you mind having a quick look before I merge this?

src/class-parsely.php
```diff
@@ -50,6 +50,39 @@ class Parsely {
 		'parsely_wipe_metadata_cache' => false,
 	);
 
+	/**
+	 * Declare post types that Parsely will process as posts
+	 *
+	 * @see https://www.parse.ly/help/integration/jsonld#distinguishing-between-posts-and-pages
+	 * @since 2.5.0
+	 * @var string[]
```
```suggestion
	 * Declare post types that Parse.ly will process as "posts".
	 *
	 * @link https://www.parse.ly/help/integration/jsonld#distinguishing-between-posts-and-pages
	 *
	 * @since 2.5.0
	 * @var string[]
```

src/class-parsely.php
```diff
@@ -50,6 +50,39 @@ class Parsely {
...
+		'Review',
+		'CreativeWork',
+	);
+
+	/**
+	 * Declare post types that Parsely will process as non-post pages
+	 *
+	 * @see https://www.parse.ly/help/integration/jsonld#distinguishing-between-posts-and-pages
+	 * @since 2.5.0
+	 * @var string[]
```
```suggestion
	 * Declare post types that Parse.ly will process as "non-posts".
	 *
	 * @link https://www.parse.ly/help/integration/jsonld#distinguishing-between-posts-and-pages
	 *
	 * @since 2.5.0
	 * @var string[]
```

src/class-parsely.php
```diff
@@ -918,7 +951,30 @@ public function construct_parsely_metadata( array $parsely_options, $post ) {
...
+			/**
+			 * Filters the JSON-LD @type.
+			 *
+			 * @since 2.5.0
+			 *
+			 * @param array   $jsonld_type  JSON-LD @type value, default is NewsArticle.
+			 * @param integer $id           Post ID.
+			 * @param string  $post_type    Post type in WordPress.
+			 */
+			$type          = apply_filters( 'wp_parsely_post_type', 'NewsArticle', $post->ID, $post->post_type );
```
```suggestion
			$type          = (string) apply_filters( 'wp_parsely_post_type', 'NewsArticle', $post->ID, $post->post_type );
```

Let's ensure that we're dealing with the type we think we're dealing with.

src/class-parsely.php
```diff
@@ -1860,4 +1912,22 @@ public function parsely_is_user_logged_in() {
...
 	}
+
+	/**
+	 * Convert JSON-LD types to respective Parse.ly page types
+	 *
+	 * @see https://www.parse.ly/help/integration/metatags#field-description
+	 * @since 2.5.0
+	 *
+	 * @param  $type string
+	 * @return string
```
```suggestion
	 * Convert JSON-LD type to respective Parse.ly page type.
	 *
	 * If the JSON-LD type is one of the types Parse.ly supports as a "post", then "post" will be returned.
	 * Otherwise, for "non-posts" and unknown types, "index" is returned.
	 *
	 * @since 2.5.0
	 *
	 * @see https://www.parse.ly/help/integration/metatags#field-description
	 *
	 * @param $type string JSON-LD type.
	 * @return string "post" or "index".
```

src/class-parsely.php
```diff
@@ -1860,4 +1912,22 @@ public function parsely_is_user_logged_in() {
...
+	 */
+	public function convert_jsonld_to_parsely_type( $type ) {
+		if ( in_array( $type, $this->allowed_jsonld_post_types ) ) {
+			return 'post';
+		}
+
+		// Fall back to this value for all other types including ones defined $this->allowed_jsonld_non_post_types
+		return 'index';
```
```suggestion
		return in_array( $type, $this->allowed_jsonld_post_types ) ? 'post' : 'index';
```

src/class-parsely.php
```diff
@@ -50,6 +50,39 @@ class Parsely {
...
 	);
 
+	/**
+	 * Declare post types that Parsely will process as posts
+	 *
+	 * @see https://www.parse.ly/help/integration/jsonld#distinguishing-between-posts-and-pages
+	 * @since 2.5.0
+	 * @var string[]
+	 */
+	private $allowed_jsonld_post_types = array(
```
GaryJones: Any post type is "allowed" (allowed by whom?) - what we're focusing on here is the post types that are supported / recognised by Parse.ly.

Consider updating the variable to be `$supported_jsonld_post_types` (likewise for the other variable). Avoid `$recognised_jsonld_post_types` as that changes spelling between UK and US English.

GaryJones (changes requested):

htdat:
@GaryJones Always great feedback. Thank you. I've added your change suggestions.

GaryJones (approved):

PR #271, opened 2021-05-03
Change $post to $parsely_type

### LINKED ISSUE #245: Change $post to $parsely_page_type
Follow on from https://github.com/Parsely/wp-parsely/pull/204#pullrequestreview-640058179

https://github.com/Parsely/wp-parsely/blob/d7a06bd696a7af3e31e49ee98f893ff8a1d07755/wp-parsely.php#L1572

The `$post` parameter in this private method would be better as `$parsely_page_type` to distinguish it from the more regular usage of the `$post` variable name within the WordPress ecosystem, which typically means an instance of WP_Post (or an integer for a post ID).

Updating this parameter is safe to do (not a breaking change), but we should also look for where this method is called, and find other instances of this `post` / `non-post` string and aim to change those references accordingly as well (including tests and potentially documentation).

htdat:
Fix #245

There is no major change in this PR. 

Instead, I reviewed the code: 

- [x] This method `get_current_url` is being used in the same class and only in this method `construct_parsely_metadata`. - [ref line](http://github.com/Parsely/wp-parsely/blob/73b0726dc6f2cdd178e9f348514a23ab8ab74253/src/class-parsely.php#L827-L827)
- [x] It is used in `class-all-test.php` file to test itself. I checked and there is no need to change anything also - [ref line](http://github.com/Parsely/wp-parsely/blob/73b0726dc6f2cdd178e9f348514a23ab8ab74253/tests/class-all-test.php#L792-L792).  
- [x] Unit tests are passed.


src/class-parsely.php
```diff
@@ -1613,26 +1613,26 @@ private function get_settings_url() {
 	 * Get the URL of the current PHP script.
 	 * A fall-back implementation to determine permalink
 	 *
-	 * @param string $post The post object you're interested in.
-	 * @param int    $post_id id of the post you want to get the url for. Optional.
+	 * @param string $parsely_type Parsely post type you're interested in, either "post" or "nonpost"
+	 * @param int    $post_id      id of the post you want to get the url for. Optional.
```
```suggestion
	 * @param int    $post_id      Optional. ID of the post you want to get the URL for. Default is 0, which means the global `$post` is used.
```

src/class-parsely.php
```diff
@@ -1613,26 +1613,26 @@ private function get_settings_url() {
 	 * Get the URL of the current PHP script.
 	 * A fall-back implementation to determine permalink
 	 *
-	 * @param string $post The post object you're interested in.
-	 * @param int    $post_id id of the post you want to get the url for. Optional.
+	 * @param string $parsely_type Parsely post type you're interested in, either "post" or "nonpost"
```
```suggestion
	 * @param string $parsely_type Optional. Parse.ly post type you're interested in, either 'post' or 'nonpost'. Default is 'nonpost'.
```

src/class-parsely.php
```diff
@@ -1613,26 +1613,26 @@ private function get_settings_url() {
...
 			 *
 			 * @since 1.14.0
 			 * @since 2.5.0  Added $post_id.
 			 *
-			 * @param string $permalink The permalink URL or false if post does not exist.
-			 * @param string $post      Post object type group ("post" or "nonpost").
-			 * @param int    $post_id   ID of the post you want to get the URL for. May be 0, so $permalink will be
-			 *                          for the global $post.
+			 * @param string $permalink         The permalink URL or false if post does not exist.
+			 * @param string $parsely_type      Parsely type ("post" or "nonpost").
```
```suggestion
			 * @param string $parsely_type      Parse.ly type ("post" or "nonpost").
```

GaryJones (changes requested):

GaryJones (approved):

htdat:
Thanks @GaryJones for your review and update.

PR #272, opened 2021-05-03
Tests: Reorganise tests into contextual files

GaryJones:
`class-all-tests.php` had some overlap with the new test files, and it was hard to see what tests there were, and what were missing.

This PR extracts and reorganises tests into more contextual files. I've also improved some of the tests, and made some small changes to main logic to facilitate this. It changes from 41 tests, 137 assertions, to 42 tests, 136 assertions.

Before:
Screenshot 2021-05-03 at 12 03 45

After:
Screenshot 2021-05-03 at 11 50 53

While this will cause conflicts with other open PRs, it will mean fewer conflicts in the future.

Now that metadata tests are split up, it's easier to see what contextual tests are missing - for example:

- Search page
- 404 page
- Year/month/day/hour/minute date archives
- Tag archive
- Singular CPT post
- Singular CPT non-post
- Custom Taxonomy Term archive

See the individual commits for more info.

There are other improvements to tests that are possible, which are outside of the scope of this PR.


GaryJones:
As well as splitting up the tests into contextual files, I'd also:

 - rewritten to improve the AMP integration test, which splitting out some logic into a new `is_amp_request()` method, and then mocking that. But, mocking doesn't with the PHP 8 tests, and so https://github.com/Yoast/wp-test-utils/ was needed.
   However, I'd hit an [issue](https://github.com/Yoast/wp-test-utils/pull/3) with that as it was never designed to work with the default WP integration test scaffolding tests structure of directories. [This PR](https://github.com/Yoast/wp-test-utils/pull/4) will fix it though, so my latest commits use the PR branch.
 - I'd also rewritten the FBIA integration test, and tried to use an assertion only added in a later version than what we use for a version of PHP. The wp-test-utils package includes https://github.com/Yoast/PHPUnit-Polyfills/ backfills these assertions, but I hit an [issue](https://github.com/Yoast/PHPUnit-Polyfills/issues/17) for PHP 5.6 and PHPUnit 5.7, that was down to me forgetting to set a value. I've fixed the test, so now all the checks pass.

What I need to do next, is to fix the merge conflicts with all of the other tests that have been added since I started this, and keep an eye on whether the PR for wp-test-utils will be merged and maybe released so that I can use a tagged version of that package.

GaryJones:
Adding WP Test Utils is now covered by https://github.com/Parsely/wp-parsely/pull/333.

GaryJones:
PHPUnit 9.1 (needed to be able to test against PHP 8.1) essentially deprecates not using PSR-4 (naming of class and files), so we should move to that.

See https://core.trac.wordpress.org/ticket/53010 and https://github.com/WordPress/WordPress-Coding-Standards/issues/1995.

Suggestion: 
- File paths: `tests/[SubFolder/]*Class_Under_Test/FunctionUnderTest[OptionalSubsetIndicator]Test.php`
- Namespace: `Parsely\Tests\[SubFolder\]*Class_Under_Test`
- Classname: `FunctionUnderTest[OptionalSubsetIndicator]Test`

GaryJones:
Screenshot 2021-06-15 at 00 24 47

Now with PSR-4 goodness, and refreshed from the latest `develop` which included some extra tests.

There would still be more of `class-all-test.php` that can be extracted into new test classes, but I wanted to get this added in since it's non-stale.

tests/StructuredData/BlogArchiveTest.php
```diff
@@ -0,0 +1,73 @@
...
+
+namespace Parsely\Tests\StructuredData;
+
+/**
+ * Structured Data Tests for the blog page (archive).
+ *
+ * @see https://www.parse.ly/help/integration/jsonld
+ * @covers \Parsely::construct_parsely_metadata
+ */
+final class BlogArchiveTest extends NonPostTestCase {
```
jblz, line 16: Just curious -- what's the rationale behind declaring test classes like these as `final`?

jblz (approved):
Thanks for all the tidying up!

tests/StructuredData/BlogArchiveTest.php
GaryJones: Test classes shouldn't be extended - there's just no need. Adding `final` can then highlight in IDEs and static analysers if someone does try to do that.

Generally it's good to default to `final` classes, and then maybe relax into non-final if absolutely needed. https://ocramius.github.io/blog/when-to-declare-classes-final/ and https://matthiasnoback.nl/2018/09/final-classes-by-default-why/ outline some reasons why.

For the tests perspective, there's less benefit, but it's also in the PHPUnit docs (https://phpunit.readthedocs.io/en/9.5/writing-tests-for-phpunit.html).

So...why _not_ make them `final`? :-)

PR #273, opened 2021-05-03
Convert Profile API call to simple XHR - no more JSONP

jblz:
(Note: this is targeting the branch in #236, not `develop`)

The API has CORS enabled, so there's no need for a JSONP workaround in #236.

The [`/profile` call](https://www.parse.ly/help/api/recommendations#get-profile) does not have to consume the response, so it's just "fire and forget."

This also fixes a couple of bugs in #236 with the rewrite -- sometimes it was not including the data in the `uuidProfileCall` call.  This fixes that by pulling those data in the function itself.


jblz:
Gonna go ahead and land this in the parent branch so we can evaluate the change as a whole.

PR #276, opened 2021-05-03
Fix Path to PHP Views

jblz:
## Description

EDIT: This was "fixed" in #236.  This adds a couple of improvements that make this more robust:

1. Remove an unneeded slash (the constant has a trailing slash)
1. Prevent a fatal if the constant is already defined.

Prior to merging in #236, this was the description and change:

Similarly to #274, the inclusion of the files in the `views` directory are failing since the main class file was moved in #262:

Screen Shot 2021-05-03 at 10 41 42 AM

This change:

* Moves the views directory inside the src directory with the rest of the PHP
* Defines a constant: `PARSELY_PLUGIN_DIR` in the top-level file
* References `PARSELY_PLUGIN_DIR` instead of `__DIR__` when sourcing the views.

### Confirm Current Behavior

* Run the `develop` branch
* Browse to a page that loads one of the views
* Notice the PHP Warning & lack of inclusion of the view script

### Confirm Fix

* Run this branch
* Browse to a page that loads one of the views
* Confirm no PHP Warning & inclusion of the view script


GaryJones (dismissed):
The constant can be defined, and used, without moving the view files.

The point of a views directory is that it is contains HTML with very little PHP (testable) logic. It should not be part of `/src/`.

wp-parsely.php
```diff
@@ -26,7 +26,9 @@
 
 if ( class_exists( 'Parsely' ) ) {
 	define( 'PARSELY_VERSION', Parsely::VERSION );
-	define( 'PARSELY_PLUGIN_DIR', plugin_dir_path( __FILE__ ) );
+	if ( ! defined( 'PARSELY_PLUGIN_DIR' ) ) {
+		define( 'PARSELY_PLUGIN_DIR', plugin_dir_path( __FILE__ ) );
+	}
```
GaryJones, line 31: Why does it need a `defined()` check? What benefit is there for someone to define this constant to any other value?

Also, why does it need to only be defined if the Parsely class exists? We know the class exists, as we just required the file that defines it.

GaryJones (approved):
Approved, but I think there's a small tweak to be made in this root file.

wp-parsely.php
jblz: > Why does it need a defined() check?

If we try to define an already defined constant, it fails and prints a notice.  This is to prevent the notice for folks who otherwise have this constant set.

I actually thought it would fatal, but it looks like not.

jblz: > why does it need to only be defined if the Parsely class exists?

It really doesn't. I was just trying to keep all the `define` calls together and the `PARSELY_VERSION` one was already wrapped in that conditional.  I'll go ahead and pull them all out in a new PR.

GaryJones: > This is to prevent the notice for folks who otherwise have this constant set.

No-one will have this set - we've only just started using it. And the chance of someone using it for some other purpose is highly mitigated by the fact it has a prefix that matches the plugin name.

In this case, the constant is a convenience thing for us - we should take ownership of it.

PR #279, opened 2021-05-04
Pull the Recommended Content Widget script into a JS file

### LINKED ISSUE #158: Widget: Move JavaScript to JS file

jblz:
## Description

Move the front end code from the PHP file to a new JavaScript file in the src directory & include it in the build step.

## Motivation and Context

Part of the larger goal to pull JS out of PHP:
Fixes #158.

## How Has This Been Tested?

* Add a known Parse.ly site URL to hosts file
* Configure your server to answer requests to that domain (for example, in [Local](https://localwp.com/), set the `Site Domain`)
* Configure the `Parse.ly Site ID` to match the site URL above
* Add & configure a widget 
* Browse to a page that loads the widget
  * Confirm the experience is equivalent on this branch and `origin/develop`

e.g.

Screen Shot 2021-05-05 at 10 47 59 PM
(before & after should be indistinguishable)


src/class-parsely-recommended-widget.php
```diff
@@ -108,115 +108,31 @@ public function widget( $args, $instance ) {
...
+				'personalized' => wp_json_encode( boolval( $instance['personalize_results'] ) ),
+				'widgetId' => esc_attr( $this->id ),
+			)
+		);
 
-		</script>
-		<?php
+		wp_enqueue_script( 'wp-parsely-recommended-widget' );
 
 		echo wp_kses( $args['after_widget'], $allowed_tags );
```
jblz, line 137: Any reason this can't just be:

```suggestion
		echo wp_kses_post( $args['after_widget'] );
```

src/js/recommended-widget/index.js
```diff
@@ -0,0 +1,93 @@
...
+	fullUrl,
+	imgSrc,
+	permalink,
+	personalized,
+	jQuery,
+	widgetId,
+} ) {
+	let uuid = false;
+
+	// TODO: If this code requires or merits attribution, add it.  If not, take this comment out.
```
jblz: TODO

src/class-parsely-recommended-widget.php
```diff
@@ -108,115 +108,31 @@ public function widget( $args, $instance ) {
...
+			true
+		);
 
+		wp_localize_script(
+			'wp-parsely-recommended-widget',
+			'wpParselyRecommended',
+			array(
+				'displayAuthor' => isset( $instance['display_author'] ) ? wp_json_encode( boolval( $instance['display_author'] ) ) : false,
+				'displayDirection' => isset( $instance['display_direction'] ) ? esc_js( $instance['display_direction'] ) : null,
+				'fullUrl' => esc_js( esc_url_raw( $full_url ) ),
```
GaryJones:
```suggestion
				'fullUrl' => esc_url_raw( $full_url ),
```

src/class-parsely-recommended-widget.php
```diff
@@ -108,115 +108,31 @@ public function widget( $args, $instance ) {
...
+			$reco_widget_script_asset[ 'dependencies' ],
+			PARSELY::get_asset_cache_buster(),
+			true
+		);
 
+		wp_localize_script(
+			'wp-parsely-recommended-widget',
+			'wpParselyRecommended',
+			array(
+				'displayAuthor' => isset( $instance['display_author'] ) ? wp_json_encode( boolval( $instance['display_author'] ) ) : false,
```
GaryJones, line 125: What does `wp_json_encode()` add here when it is already cast as a bool? The `wp_localize_script()` should be able to handle making a bool safe for JS.

As a proof point, the `false` fallback doesn't have the `wp_json_encode()` wrapper.

src/class-parsely-recommended-widget.php
```diff
@@ -108,115 +108,31 @@ public function widget( $args, $instance ) {
...
+				'displayDirection' => isset( $instance['display_direction'] ) ? esc_js( $instance['display_direction'] ) : null,
```
GaryJones:
```suggestion
				'displayDirection' => isset( $instance['display_direction'] ) ? $instance['display_direction'] : null,
```

`esc_js()` should only be used for DOM (HTML) attributes.

src/class-parsely-recommended-widget.php
```diff
@@ -108,115 +108,31 @@ public function widget( $args, $instance ) {
...
+				'fullUrl' => esc_js( esc_url_raw( $full_url ) ),
+				'imgSrc' => isset( $instance['img_src'] ) ? esc_js( $instance['img_src'] ) : null,
```
GaryJones:
```suggestion
				'imgSrc' => isset( $instance['img_src'] ) ? $instance['img_src'] : null,
```

src/class-parsely-recommended-widget.php
```diff
@@ -108,115 +108,31 @@ public function widget( $args, $instance ) {
...
+				'permalink' => wp_json_encode( esc_url_raw( get_permalink() ) ),
```
GaryJones:
```suggestion
				'permalink' => esc_url_raw( get_permalink() ),
```

The object of HTML-decoded strings are all encoded with `wp_json_encode()` as part of `wp_localize_script()` - no need to encode each value that we pass to it.

src/class-parsely-recommended-widget.php
```diff
@@ -108,115 +108,31 @@ public function widget( $args, $instance ) {
...
+				'personalized' => wp_json_encode( boolval( $instance['personalize_results'] ) ),
```
GaryJones:
```suggestion
				'personalized' => boolval( $instance['personalize_results'] ),
```

src/class-parsely-recommended-widget.php
```diff
@@ -108,115 +108,31 @@ public function widget( $args, $instance ) {
...
+				'widgetId' => esc_attr( $this->id ),
```
GaryJones:
```suggestion
				'widgetId' => $this->id,
```

src/class-parsely-recommended-widget.php
GaryJones: Considering line 97 is `$allowed_tags = wp_kses_allowed_html( 'post' );`, then it would seem that `wp_kses_post()` is sufficient.

Check line 99 as well.

GaryJones (changes requested):

src/class-parsely-recommended-widget.php
jblz: Changed it over in #290 since the change isn't really related to the rest of the spirit of this PR.

src/class-parsely-recommended-widget.php
jblz: These were all straight copied from the PHP file so I could pipe them through to the new JS file -- I hadn't had time to tweak them.  I _think_ the idea with `wp_json_encode` was to cast a boolean `true` to a string `"true"` (else the following would have been `var display_author = "1";`:

https://github.com/Parsely/wp-parsely/blob/develop/src/class-parsely-recommended-widget.php#L137

Anyhow, it's moot now.

GaryJones: Right now, with the `wp_json_encode()` still in place, this will appear to display a string `"true"`, or a boolean `false`. It's inconsistent.

src/js/lib/cookies.js
```diff
@@ -0,0 +1,15 @@
...
+}
+
+/**
+ * Get the value of a particular cookie
+ *
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie
+ * @param {string} key Which cookie value to get
+ * @return {string | undefined} The value of the specified key, or `undefined` if it's not set.
+ */
+export const getCookieValue = ( key ) => Cookies.get()
```
GaryJones, line 12: How "namespaced" is this constant? Does it add a `getCookieValue` to the global space that could clash with other plugins or core?

GaryJones (approved):
Approved with one small question.

src/js/lib/cookies.js
jblz: Everything in our built scripts is scope-limited to the module.  Nothing will pollute the global scope unless it's attached to `global` or `window`, etc.

PR #281, opened 2021-05-04
Refactor JavaScript Registration & Loading. Add Consistency to Filter Names

### LINKED ISSUE #169: Make hook prefixes consistent
The current non-core filters (there are no action hooks) in the plugin are:

 - `wp_parsely_post_tags`
 - `wp_parsely_post_category`
 - `wp_parsely_pre_authors`
 - `wp_parsely_post_authors`
 - `wp_parsely_permalink`
 - `after_set_parsely_page`
 - `parsely_filter_insert_javascript`

For version 2.5.0, the bottom two should have a new filter hook with the consistent `wp_parsely` prefix, and change the inconsistent ones from `apply_filters()` to [`apply_filters_deprecated()`](https://developer.wordpress.org/reference/functions/apply_filters_deprecated/).

In 3.0.0, the old filters can be removed. Any customer using the old filter will need to update their code to reference the new filter name instead. Once a PR is merged for this issue, then a new issue / PR for the removal should be created.

We can protect against a re-introduction of incorrectly prefixed hook names by setting a [prefix for all globals](https://github.com/WordPress/WordPress-Coding-Standards/wiki/Customizable-sniff-properties#naming-conventions-prefix-everything-in-the-global-namespace) in a PHPCS config file.

jblz:
## Description

* Deprecate `after_set_parsely_page` filter & add `wp_parsely_metadata` (including test coverage for latter)
* Deprecate `parsely_filter_insert_javascript` filter & add `wp_parsely_load_js_tracker` (including test coverage for both)
* Register front-end scripts independently from enqueuing them (in `init`)
* Decouple the tracking script from the API init script & enqueue them independently of one another. The tracking script registration is no longer a dependency of the API script
* Update automated testing to reflect the above

Fixes #169

## Motivation and Context

* We want all our hook names specific to this plugin to be prefixed with `wp_parsely_`.
* We want to deprecate old hook names so users can migrate to the new ones before we remove support.
* We want our filters to be named appropriately for the purpose they serve.
* We want to split out script registration/enqueues & decouple the scripts to provide more flexibility (e.g. if a site has implemented [dynamic tracking](https://www.parse.ly/help/integration/dynamic)). Further, decoupling them makes testing for each more straightforward.

## How Has This Been Tested?

* Automated tests should pass, changes to such should make sense
* When deprecated hook names are used, they should "still work" & deprecation notice in the logs should be correct.
* The tracking script loads on the front end while logged out
* The tracking script respects the `Track Logged-in Users` setting on the front end while logged in
* The init-api script is correctly enqueued and sourced when an API Secret is set (`PARSELY.onload` is set to include the `uuidProfileCall` function (or a wrapper if one is already set)
* The init-api script is neither enqueued nor sourced when an API Secret is not set (`PARSELY.onload` is not altered)

## Types of changes

Refactors.  There should be no functional change in behavior due to this.


src/class-parsely.php
```diff
@@ -964,13 +964,31 @@ public function construct_parsely_metadata( array $parsely_options, $post ) {
 		/**
 		 * Filters the structured metadata.
 		 *
+		 * @deprecated deprecated since version 2.5.0
```
```suggestion
		 * @deprecated 2.5.0 Use `wp_parsely_metadata` filter instead.
```

See https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc-tags.md#54-deprecated for recommended format of the `@deprecated` tag.

src/class-parsely.php
```diff
@@ -1119,13 +1137,32 @@ public function insert_parsely_javascript() {
 		 * If true, the JavaScript files are sourced.
 		 *
 		 * @since 2.2.0
+		 * @deprecated deprecated since version 2.5.0
```
```suggestion
		 * @deprecated 2.5.0 Use `wp_parsely_insert_javascript` filter instead.
```

src/class-parsely.php
```diff
@@ -1119,13 +1137,32 @@ public function insert_parsely_javascript() {
...
+			'2.5.0',
+			'wp_parsely_insert_javascript'
+		) ) {
 			return;
 		}
 
+		/**
+		 * Filters whether to include the Parsely JavaScript file.
+		 *
+		 * If true, the JavaScript files are sourced.
```
GaryJones: Since the conditions aren't about requiring a JS file or not, but more about whether some JS files are enqueued or not, can this description be updated?

Is "insert javascript" still the right phrase to convey what really happens after this point? Maybe something like "Insert tracker", "Apply tracking", "Track response" or similar? Yes, we're registering and enqueueing JS, but what feature is it that we're actually manipulating here from a user's perspective?

GaryJones (changes requested):
A couple of minor DocBlocks items, but then also a query on the naming of the new filter.

Also, there's at least [one place in the readme](https://github.com/Parsely/wp-parsely#how-can-i-edit-the-values-passed-to-the-json-ld-metadata) where the deprecated filter name is referenced that should be updated.

src/class-parsely.php
GaryJones: #294 is following this up.

GaryJones (approved):

PR #282, opened 2021-05-04
Fix The Plugin Action Link 

jblz:
## Description

Moving the `Parsely` class file out of the main entry PHP file (#262) caused a regression in that the "Settings" link on the Plugin listing page was no longer printed.  This reinstates it by adding a constant for the basename in the top-level file and referencing it in the class file.

|Before|After|
|---|---|
|Screen Shot 2021-05-04 at 1 27 44 PM|Screen Shot 2021-05-04 at 1 27 58 PM|

### Confirm current behavior

* Browse to `/wp-admin/plugins.php` with this plugin active
* Notice no "Settings" link on this plugin's entry

### Confirm fix

* Browse to `/wp-admin/plugins.php` with this plugin active
* Confirm "Settings" link on this plugin's entry is present and links to our settings page

### DIFF
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -76,9 +76,8 @@ class Parsely {
 		// display warning when plugin hasn't been configured.
 		add_action( 'admin_footer', array( $this, 'display_admin_warning' ) );
 
-		$basename = plugin_basename( __FILE__ );
 		add_filter(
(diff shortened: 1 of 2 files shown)


GaryJones (approved):
👍🏻

PR #284, opened 2021-05-04
Remove src/parsely-javascript.php

jblz:
## Description

Remove `src/parsely-javascript.php`.

## Motivation and Context
I was pretty certain this was removed in #236 (the change that obviates it), but apparently not.

This file is no longer sourced and its functionality has been relocated elsewhere.

## How Has This Been Tested?

* Automated tests are still passing with `WP_MULTISITE=1 ./vendor/bin/phpunit`
* Script code "still works"

na

## Types of changes
Maintenance


GaryJones (approved):
👍🏻 

Aside: I would say that this is a Maintenance issue, rather than a Bug (something affecting the users).

PR #285, opened 2021-05-05
Improve handling the "Force HTTPS canonicals" option

### LINKED ISSUE #174: Use correct protocol for canonical URLs

htdat:
Close #174 

## Description

There are two changes on this PR:

1, The default value for `Force HTTPS canonicals` option is set based on the protocol of `home_url()` value: 
- true for https
- false for HTTP

2, The help text for `Force HTTPS canonicals` is changed but generally, it does not intensify that users will need to change this value. They just need to do that if necessary. 

**NOTE: there is a slight change for (2) comparing with this comment https://github.com/Parsely/wp-parsely/issues/174#issuecomment-831762629 as I do not want to let all existing users contact Parse.ly support rep to run re-indexing. That might create unnecessary requests for Parse.ly support team.**

## Motivation and Context

See #174. 

#### Test 1: A new site with HTTP address

1. Delete `parsely` option via wp-cli: `wp option delete parsely` or directly in the database.
2. Visit `wp-admin/options-general.php?page=parsely`, `Force HTTPS canonicals` is set to `No`.
3. Change ``Force HTTPS canonicals` to `Yes` or change the site address to HTTPS. 
4. See this message `This option is not being consistent with the protocol of your site address. You can keep it as-is without affecting your Parse.ly data. However, if you're forced to change this value, please reach out to Parse.ly support`

#### Test 2: Test a new site with HTTPS address

1. Similar step 1 in test 1.
2. Visit `wp-admin/options-general.php?page=parsely`, `Force HTTPS canonicals` is set to `Yes`.
3. Change `Force HTTPS canonicals` to `No` or change the site address to HTTP. 
4. Similar to step 4 in test 1. 

#### Test 3: Test with a site upgrading from the previous version to this version. 

1. Use Parsely 2.4.1 - the current latest stable version.
2. Visit `wp-admin/options-general.php?page=parsely`, choose whatever value for `Force HTTPS canonicals`.
3. Change the plugin to this PR. 
4. Visit `wp-admin/options-general.php?page=parsely`, the value chosen in step 2 has not been changed.

## Types of changes

- [x] Improvement

Note: I added Improvement because all 3 previous options seem not to make sense for this PR.

## Consideration 

IMHO, this is pretty subtle and we do not need tests. But in case, any reviewer thinks tests are necessary, I can add them in new commits.


src/class-parsely.php
```diff
@@ -438,7 +438,13 @@ public function initialize_settings() {
 			)
 		);
 
-		$h = __( 'The plugin uses http canonical URLs by default. If this needs to be forced to use https, set this option to true. Note: the default is fine for almost all publishers, it\'s unlikely you\'ll have to change this unless directed to do so by a Parse.ly support rep.', 'wp-parsely' );
+		$parsely_options = $this->get_options();
+		if ( $parsely_options['force_https_canonicals'] === $this->is_home_url_with_https() ) {
+			$h = __( 'This option is being consistent with the protocol of your site address. It\'s unlikely you\'ll have to change this unless directed to do so by a Parse.ly support rep.', 'wp-parsely' );
```
mikeyarce, line 480: What if we changed this to:

> We've detected that your site uses https so we've set this option for you automatically!
> If you need to change this option, please speak to a Parse.ly customer support representative so we can properly index your site along with the change.

What do you think?

src/class-parsely.php
```diff
@@ -438,7 +438,13 @@ public function initialize_settings() {
...
+		} else {
+			$h = __( 'This option is not being consistent with the protocol of your site address. You can keep it as-is without affecting your Parse.ly data. However, if you\'re forced to change this value, please reach out to Parse.ly support', 'wp-parsely' );
```
mikeyarce, line 482: What about:

> Your site's default protocol appears to be https, but Parse.ly is currently crawling http links.  Don't worry, you can keep it as-is without affecting your Parse.ly data.  If you do decide to make a change, please reach out to a Parse.ly customer support representative so we can properly index your site after you've made this change.

src/class-parsely.php
```diff
@@ -1860,4 +1874,14 @@ public function parsely_is_user_logged_in() {
 	public function return_personalized_json() {
 
 	}
+
+	/**
+	 * Check if home_url is using the HTTPS protocol
+	 *
+	 * @since 2.5.0
```
mikeyarce: We should update the `@since` to 2.6.0 here and above.

mikeyarce (commented):
This looks great, @htdat !  

I tested this as well, thanks for the detailed testing instructions.
I think the logic is good, I just left some notes on the wording of the message and that we need to update the version number.  Let me know what you think.

src/class-parsely.php
htdat: > We've detected that your site uses https so we've set this option for you automatically!

This statement is not really correct for all cases. 
It's correct only if: 

```php
$parsely_options['force_https_canonicals'] === $this->is_home_url_with_https() === true
``` 

 but it's not correct for the other case: 
```php
$parsely_options['force_https_canonicals'] === $this->is_home_url_with_https() === false
``` 
Therefore, I think it would be better to stick with the "consistent" word.

Also, in terms of timeframe and dealing with the old versions, saying `we've set` is not technically correct for all cases. 

> If you need to change this option, please speak to a Parse.ly customer support representative so we can properly index your site along with the change.

Actually, I prefer my current wording (also the previous wording before this PR) as it indicates clearly everything still works properly and customers do not need to do anything.

src/class-parsely.php
htdat: @mikeyarce Similar to above regrading the "consistent" word.

Also, IMHO, this change is a significant change to their data so I'd like to make sure they reach out to Parse.ly support first rather than let them change and reach out to Parse.ly support later. 

What do you think?

src/class-parsely.php
htdat: Otherwise, in case, we need to state which option it's being right now (HTTP or HTTPS) in the help text, I can add more code to handle that. However, the downside is that the code may be verbose and not necessary in terms of telling what's next for users. 

What do you think? I do not have a strong opinion regarding this, just want to balance between "detailed info" and "simple code".

htdat:
Yay, thanks @mikeyarce. I've replied to your comment, merged `develop` branch to the current branch here, and updated the version to 2.6.0.

src/class-parsely.php
mikeyarce: Thanks for your thoughts @htdat!  I don't have any strong opinions, but curious to see what @GaryJones thinks about the wording for these messages.

GaryJones:
Ideally, we should be moving to getting rid of this setting, and provide a filter that can be used, even if we also provide and enact the callbacks as well.

The cases we need to consider:

1. Site URL is HTTP, and a site is not forced to HTTPS = ✅ canonical is correct.
2. Site URL is HTTP, and a site is forced to HTTPS = ❌ canonical is not correct.
3. Site URL is HTTPS, and a site is forced to HTTPS = ✅ canonical is correct.
4. Site URL is HTTPS, and a site is forced to HTTPS = ❌ canonical is not correct.

For the ❌ incorrect canonical cases, we should fix (always match the site URL), and trigger a re-index in Parse.ly Support.

This PR adds a new default value (match the site URL), which is a start for new installs, and there should be some documentation and tests changes around that, than start changing messages.

htdat:
@GaryJones Thanks for your feedback. 

> Ideally, we should be moving to getting rid of this setting, and provide a filter that can be used, even if we also provide and enact the callbacks as well.

Agreed this is ideal. But I don't think of a solution that we can get rid of this option and make it compatible with the old versions of plugins, especially for sites with the incorrect canonical option.

> For the ❌ incorrect canonical cases, we should fix (always match the site URL), and trigger a re-index in Parse.ly Support.

If we follow this path, a few things need to be considered: 

- There is no API to trigger the re-index automatically. 
- Site owners still need to contact Parse.ly Support to re-index the data.
- Due to this change (without telling site owners in advance), Parse.ly data will not be recorded properly, which may cause unnecessary load for Parse.ly Support? 

My PR here is not to change the current behavior of this "Force HTTPS canonicals" option. Instead, it will only handle the default option value for new installs. 

> This PR adds a new default value (match the site URL), which is a start for new installs, and there should be some documentation and tests changes around that, than start changing messages.

I can add test cases for this option. However, adding a default value for new installs lets the current message for this option no longer correct, and then the message should be changed in conjunction with the default value.

pauarge:
This PR has become stale and it would need to be reworked to match the current plugin state. We are closing it, but it can be used as a reference when we pick the task again in the future.

PR #286, opened 2021-05-05
Add data-cfasync="false" to script tags

jblz:
## Description

The current version of the plugin adds a `data-cfasync="false"` attribute pair to the script tags:

https://github.com/Parsely/wp-parsely/blob/de02bbd0decc7f9c880748c5774eba02bee26760/parsely-javascript.php#L22-L24

https://github.com/Parsely/wp-parsely/blob/de02bbd0decc7f9c880748c5774eba02bee26760/parsely-javascript.php#L57

https://github.com/Parsely/wp-parsely/blob/de02bbd0decc7f9c880748c5774eba02bee26760/class-parsely-recommended-widget.php#L67

This change:

* Adds the API script to the list of handles that gets the treatment
* Fixes the `preg_replace` to actually work
* (minor) changes a short array syntax to long to match existing code style

## Motivation and Context

It was intended to replicate this behavior in #236, but it did not make it into the change.

I didn't find a whole lot of context around the need for this, but there's #129. @hbbtstar -- can you please take a look and let me know if this is the right approach?

Some more context: this attribute looks like it's used to have [Cloudflare's Rocket Loader "ignore specific JavaScripts](https://support.cloudflare.com/hc/en-us/articles/200169436-How-can-I-have-Rocket-Loader-ignore-specific-JavaScripts-)."

## How Has This Been Tested?

* Automated tests added to confirm the attribute is correctly printed in this branch
* Source inspected to do the same

Screen Shot 2021-05-05 at 10 13 53 AM

## Types of changes


GaryJones:
As a side note, I would like to find a way to make these optional - that is, to have a filter that can conditionally add these integrations in, since they are redundant for the majority of sites that are not behind Cloudflare.

GaryJones (approved):

jblz:
Agreed.  Tracked in #293

PR #289, opened 2021-05-05
Rename `master` to `trunk`

mikeyarce:
Rename the `master` branch to `trunk`.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -8,7 +8,7 @@ jobs:
     name: New tag
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@master
+      - uses: actions/checkout@trunk
       - name: Build # Remove or modify this step as needed
         run: |
           npm install
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -8,7 +8,7 @@ jobs:
     name: New tag
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@trunk
+      - uses: actions/checkout@v2
       - name: Build # Remove or modify this step as needed
         run: |
           npm install
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -36,7 +36,7 @@ jobs:
       fail-fast: false
     steps:
       - name: Checkout code
-        uses: actions/checkout@trunk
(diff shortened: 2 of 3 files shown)


.github/workflows/deploy.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     name: New tag
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@master
+      - uses: actions/checkout@trunk
```
jblz: EDIT: Scratch that!

```suggestion
      - uses: actions/checkout@v2
        with:
          ref: trunk
```

...maybe?

https://github.com/actions/checkout/tree/25a956c84d5dd820d28caab9f86b8d183aeeff3d#checkout-a-different-branch

.github/workflows/integration-tests.yml
```diff
@@ -36,7 +36,7 @@ jobs:
       fail-fast: false
     steps:
       - name: Checkout code
-        uses: actions/checkout@master
+        uses: actions/checkout@trunk
```
jblz, line 39:
```suggestion
        uses: actions/checkout@v2
        with:
          ref: trunk
```

trying...

jblz:
```suggestion
        uses: actions/checkout@v2
```

.github/workflows/deploy.yml
jblz: Actually, I think it should just use the ref from the tag automatically if we use `actions/checkout@v2`

```suggestion
      - uses: actions/checkout@v2
```

jblz:
Switched to `actions/checkout@v2`. See: https://github.com/marketplace/actions/checkout

We don't need to specify a branch for this...it checks out the hash relating to the event that triggered the action run.

jblz (approved):
I'm not sure how to test the action changes & we could probably do without them, but it seems likely to work.

.github/workflows/deploy.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     name: New tag
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@master
+      - uses: actions/checkout@v2
```
GaryJones, line 11: Not a big issue, but this `master` was related to the version of the `actions/checkout` code, not the wp-parsely master branch.

PR #291, opened 2021-05-06
Add "Quick Action Link" to the Dashboard in the Post List

jblz:
## Description

Include a link to the statistics page for an article in the wp-admin Posts List. This uses the [Find in Parse.ly API Endpoint](https://www.parse.ly/help/integration/other-tools#find-in-parsely-api-endpoint).

Our filtering function is hooked into both the `post_row_actions` & the `page_row_actions` so it will do its thing on hierarchical and non lists.

## Motivation and Context

To make it easier to browse to the article page in the dashboard from within the WordPress interface.

## How Has This Been Tested?

* Browse to `/wp-admin/edit.php`
* There should be links to the Parse.ly Dashboard for each published post
* Browse to `/wp-admin/edit.php?post_type=page`
* There should be links to the Parse.ly Dashboard for each published page
* Browse to the wp-admin list for a "publicly-queryable" custom post type
* There should be links to the Parse.ly Dashboard for each published page
* Browse to the wp-admin list for a custom post type that is NOT "publicly-queryable"
* There should **NOT** be links to the Parse.ly Dashboard for each published page

|Link to `/find` endpoint|No Link on Private Posts|No Link on Drafts|
|---|---|---|
|Screen Shot 2021-05-21 at 4 27 10 PM|Screen Shot 2021-05-24 at 2 22 42 PM|Screen Shot 2021-05-24 at 2 22 49 PM|

## Types of changes


src/class-parsely.php
```diff
@@ -92,6 +93,24 @@ public function __construct() {
 		add_action( 'wp_enqueue_scripts', [ $this, 'wp_parsely_style_init' ] );
 	}
 
+	function add_parsely_link( $actions, $post ) {
+		if ( ! self::post_has_trackable_status( $post ) ) {
+			return $actions;
+		}
```
GaryJones, line 151: While a post may not currently have a trackable status, it may have had one previously - and so there may be data in Parse.ly dashboard about it.

Is this an edge case that we should ignore for now?

src/class-parsely.php
```diff
@@ -92,6 +93,24 @@ public function __construct() {
 		add_action( 'wp_enqueue_scripts', [ $this, 'wp_parsely_style_init' ] );
 	}
 
+	function add_parsely_link( $actions, $post ) {
```
```suggestion
	public function add_parsely_link( $actions, $post ) {
```

Also needs a DocBlock.

src/class-parsely.php
```diff
@@ -92,6 +93,24 @@ public function __construct() {
...
+		if ( ! self::post_has_trackable_status( $post ) ) {
+			return $actions;
+		}
+		$options = $this->get_options();
+		$apikey = $options['apikey'];
+		$post_url = get_permalink( $post );
+		$parsely_url = 'https://dash.parsely.com/' . $apikey . '/find?url=' . $post_url;
```
GaryJones: The `$post_url` is going to need encoding.

src/class-parsely.php
```diff
@@ -92,6 +93,24 @@ public function __construct() {
...
+		$actions = array_merge( $actions, array(
+			'find_in_parsely' => sprintf(
+				'<img style="padding-right: 4px; transform: translateY(2px);" width="16px" height="16px" src="https://www.parse.ly/favicon-32x32.png?v=f82f85b2b9b3e85c26fc141c6c44b638"/><a href="%1$s">%2$s</a>',
```
```suggestion
				'<img style="padding-right: 4px; transform: translateY(2px);" width="16px" height="16px" src="https://www.parse.ly/favicon-32x32.png?v=f82f85b2b9b3e85c26fc141c6c44b638" alt="" /><a href="%1$s">%2$s</a>',
```

GaryJones (changes requested):

src/class-parsely.php
jblz: Yes, I think this is an ok edge case to ignore for now. If it becomes obvious how we can know if an article has been crawled, we can revisit.  I think we're going to want to do so if / when we make this indication more dynamic.

jblz:
I noticed a suboptimal experience when browsing to one of these links while not logged into the dashboard.

Logged here: https://github.com/Parsely/dash/issues/1038

src/class-parsely.php
```diff
@@ -129,6 +131,52 @@ public function __construct() {
...
+	 * @see https://developer.wordpress.org/reference/hooks/post_row_actions/
+	 * @param [string[]] $actions  The list of actions to which this function will append
+	 * @param [WP_Post]  $post     The individual post object in the list
+	 * @return $actions            The filtered list of actions
+	 */
+	public function row_actions_add_parsely_link( $actions, $post ) {
+		if ( ! is_array( $actions ) ) {
+			return $actions;
+		}
+
```
GaryJones, line 148: I think we should also check for the post type being `public` as well - if it's not accessible from the front-end, then it can't be in Parse.ly. A post could be non-public post type and still have a `publish` post status (WPCOM Legacy Redirector may go that route, to distinguish active from non-active redirects).

src/class-parsely.php
```diff
@@ -129,6 +131,52 @@ public function __construct() {
...
 	}
 
+	/**
+	 * Include a link to the statistics page for an article in the wp-admin Posts List.
+	 * If the post object is the "front page," this will include the main dashboard link instead.
+	 *
+	 * @see https://developer.wordpress.org/reference/hooks/page_row_actions/
+	 * @see https://developer.wordpress.org/reference/hooks/post_row_actions/
+	 * @param [string[]] $actions  The list of actions to which this function will append
+	 * @param [WP_Post]  $post     The individual post object in the list
```
```suggestion
	 * @param array<string, string> $actions The existing list of actions.
	 * @param WP_Post               $post    The individual post object the actions apply to.
```

`string[]` is only for indexed array, like
`[ 'foo', 'bar' ]`. 

What we have in actions, is an associative array, with a string key and a string value: 
`[ 'foo' => 'a link', 'parsely_dash' => 'another link' ]`

src/class-parsely.php
```diff
@@ -129,6 +131,52 @@ public function __construct() {
...
+	 * @return $actions            The filtered list of actions
```
```suggestion
	 * @return array<string, string> The amended list of actions.
```

We return a value, not a variable.

src/class-parsely.php
```diff
@@ -129,6 +131,52 @@ public function __construct() {
...
+			return $actions;
+		}
+
+		if ( ! self::post_has_trackable_status( $post ) ) {
+			return $actions;
+		}
+
+		$options = $this->get_options();
+		$apikey = $options['apikey'];
+		$parsely_url = trailingslashit( 'https://dash.parsely.com/' . $apikey );
```
GaryJones: Should we check that apikey has been set before assuming it has? Can't create a link if we don't know where we're meant to link to.

src/class-parsely.php
```diff
@@ -129,6 +131,52 @@ public function __construct() {
...
+			$actions[ 'parsely_dash' ] = sprintf(
+				'<a href="%1$s">%2$s</a>',
+				esc_url( $parsely_url ),
+				esc_html__( 'Parse.ly Dashboard', 'wp-parsely' )
+			);
+			return $actions;
+		}
+
+		$post_url = get_permalink( $post );
+		$parsely_url .=  'find?url=' . urlencode( $post_url );
```
GaryJones: Does Parse.ly require `urlencode()`'d querystring, or `rawurlencode()`'d querystring? They do give slightly different results.

src/class-parsely.php
```diff
@@ -129,6 +131,52 @@ public function __construct() {
...
+		if (
+			'page' === get_option( 'show_on_front' ) &&
+			(int) get_option( 'page_on_front' ) === $post->ID
+		) {
+			// This is the "front page." Link to the dashboard.
+			$actions[ 'parsely_dash' ] = sprintf(
+				'<a href="%1$s">%2$s</a>',
+				esc_url( $parsely_url ),
+				esc_html__( 'Parse.ly Dashboard', 'wp-parsely' )
+			);
```
GaryJones: Why does this need to be different?

The front page still has it's own stats page on the Parse.ly dashboard i.e. `dash.parsely.com/wpvip.com/find?url=https://wpvip.com` still works.

Having a list of pages pointing to "Parse.ly Stats", and then one in the middle pointing to "Parse.ly Dashboard" seems odd. Let's keep it consistent.

GaryJones (changes requested):

src/class-parsely.php
jblz: Good suggestion.

It would be nice to just use [`is_post_publicly_viewable`](https://developer.wordpress.org/reference/functions/is_post_publicly_viewable/), but it was added in WP 5.7, so too new for our minimum.

Even [`is_post_type_viewable`](https://developer.wordpress.org/reference/functions/is_post_type_viewable/) requires WP >= 4.4.

What do you think we should check here?

src/class-parsely.php
jblz: I think we may want to switch to [`rawurlencode`](https://www.php.net/manual/en/function.rawurlencode.php) for the full [RFC3986](http://www.faqs.org/rfcs/rfc3986.html) treatment.  See also:
https://stackoverflow.com/a/996161

@hbbtstar -- do you have any insight into what the service expects?

src/class-parsely.php
jblz: I went ahead and took out the variation.  Agree that having it consistent makes more sense after more testing.

src/class-parsely.php
jblz: I went ahead and switched to `rawurlencode`. I think that's probably preferable, but we can change it back if needed.

src/class-parsely.php
GaryJones: We could always either create our own function/check, or bump the minimum to WP 4.4 (or maybe later if we're going to move it anyway).

jblz: 5e4885659b9870a9379e40eb8a1f5e54f8ee6406

jblz:
As of now, this includes the link on password-protected posts.

GaryJones (approved):

src/class-parsely.php
```diff
@@ -129,6 +131,53 @@ public function __construct() {
...
+				'url' => rawurlencode( get_permalink( $post ) ),
+			),
+			trailingslashit( 'https://dash.parsely.com/' . $options['apikey'] ) . 'find'
+		);
+
+		$actions['find_in_parsely'] = sprintf(
+			'<a href="%1$s">%2$s</a>',
+			esc_url( $parsely_url ),
+			esc_html__( 'Parse.ly Stats', 'wp-parsely' )
+		);
```
GaryJones: To have the same link text, pointing to different URLs, is a bad thing for accessibility. Links can be taken out of context, so they need to be more explainable.

Let's look at what WP core does for the existing Edit, Quick Edit, Trash, and View actions links:

Screenshot 2021-05-26 at 11 23 01

They use an `aria-label` attribute on the anchor element, that includes the (quoted) name of the post in some way. We should do the same for this Parse.ly Stats link.

GaryJones (changes requested):
We have an opportunity to maintain the level of accessibility that other row actions have.

src/class-parsely.php
GaryJones: Addressed in c6032d9.

GaryJones (approved):

PR #292, opened 2021-05-06
Fix Dependabot Reviewers

jblz:
## Description

fix for #287

a string is not sufficient...apparently it has to be an array

see: https://github.com/Parsely/wp-parsely/pull/266/checks?check_run_id=2514953476


jblz:
Going ahead and merging this since it's breaking test results

PR #295, opened 2021-05-06
Fix text domain (wp-parsley -> wp-parsely)

jblz:
## Description
Update a text domain slug in an i18n function.

## Motivation and Context

`wp-parsley` is not correct, `wp-parsely` is. 

I wonder if a generalized lint rule is in order :)

## How Has This Been Tested?

n/a

## Types of changes

### DIFF
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -704,7 +704,7 @@ class Parsely {
 
 		$message = sprintf(
 				/* translators: %s: Plugin settings page URL */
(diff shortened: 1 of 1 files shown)


GaryJones:
> I wonder if a generalized lint rule is in order :)

We already have it: https://github.com/Parsely/wp-parsely/blob/154c2e687fc4fe9031b01eb3e22fbce1d7ca1160/.phpcs.xml.dist#L56-L62

And it shows up in `composer cs`:

Screenshot 2021-05-06 at 15 22 29

It also shows up with `./vendor/bin/phpcs --sniffs=WordPress.WP.I18n --ignore=vendor .` as well.

GaryJones (approved):

GaryJones (approved):

PR #299, opened 2021-05-07
New Github Action to test js on push

jblz:
## Description

Hook up jest tests added in #297 to run on pushes and PRs.

## Motivation and Context

Quality++

## How Has This Been Tested?

This PR

Screenshot_20210507-080507

## Types of changes

DevEx


.github/workflows/node.js.yml
```diff
@@ -0,0 +1,30 @@
+# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
+
+name: Node.js CI
+
+on:
+  push:
+    branches: [ develop ]
+  pull_request:
+    branches: [ develop ]
```
GaryJones: [This code](https://github.com/Parsely/wp-parsely/blob/develop/.github/workflows/integration-tests.yml#L7-L8) aims to _not_ run the PHP tests when the commit only contains Markdown files.

Would it be possible to limit this Node.js Test workflow to only run if a (non-built) `.js` was updated? Could a change in a PHP file make JS tests fail?

GaryJones: By only testing `push` when it's the `develop` branch, then Node.js tests won't be run on feature branch pushes; they will only run when the PR is opened (as we don't `push` directly to `develop`. Is that the intent? Or should `push` be opened for all branches?

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,30 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v2
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm ci
+    - run: npm run build
+    - run: npm test
```
GaryJones: Adding a `name` for each step makes it a little easier to follow.

GaryJones (dismissed):
Looking at https://github.com/Parsely/wp-parsely/runs/2527060443?check_suite_focus=true, the longest step was the `npm ci`.

Could some caching help here?

jblz:
Thanks for the suggestions, @GaryJones 

I updated it to:
* limit to a set of path patterns on PRs (but not branches) on PRs
* limit branches on push to `[develop, trunk, release/**]`

Rationale: I noticed that the workflows are running twice when pushing to a branch that is attached to a PR. Once for the `push` event and another for the `pull_request.synchronize` event.

This way, it's only running the workflow once in such a case & should still run whenever we merge to a one of those "special" branches or tag a release.

I _think_ the path patterns are comprehensive while striving to minimize maintenance.  🤞 🤞

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,59 @@
...
+          cache-name: cache-node-modules
+        with:
+          path: ~/.npm
+          key: "${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}"
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.cache-name }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
+
+      - name: 'Install js dependencies'
```
GaryJones:
```suggestion
      - name: Install JavaScript dependencies
```

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,59 @@
...
+
+on:
+  pull_request:
+    paths:
+      - .github/workflows/*.js.yml
+      - src/js/**/*[tj]sx?
+      - tests/js/**/*.[tj]s
+      - package.json
+      - '*.js'
+  push:
```
GaryJones, line 14: I still think that this list of branches mean that if I push a JS file change to a feature branch, it's not going to run the tests. 

Only when the PR is made, will they run, and at that time I thought it was all working fine, but it might not have been.

GaryJones (commented):
Rest of the improvements look good. 

I did a quick check to see if there was a combined npm install + Cache action like we have for composer, but didn't see one.

.github/workflows/node.js.yml
jblz: If you push to a branch that's attached to a PR it will run by virtue of the `pull_request.synchronize` event. If you push to a branch that has not been PR'd, it won't run. Seems ok to me.

GaryJones (approved):

PR #306, opened 2021-05-10
DO NOT MERGE - this is a test branch to demo gh actions

jblz:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #308, opened 2021-05-10
Improve "cs-lint" & "integration test" Github Action Event Handling

jblz:
## Description

* Limit cs-lint & integration tests from running twice on push
    * Add an "allowed list" of branches to respond to `push` events
* Add `src/js/**` to the paths-ignore list, so changes limited to files there don't trigger these tests

## Motivation and Context

Run tests as few times as possible. See previous discussion in #299

## How Has This Been Tested?

#306 is based on this branch, #307 is based on `develop`

* Pushing to the branch in 306 (based on this one) prior to attaching it to a PR **did not** trigger these tests. Doing so to the branch in 307 **did**.
* Creating the Pull Request **did** run these tests on both branches.
* Pushing to the branch in 306 (based on this one) after it was PR'd ran these tests only once (for the `pull_request.synchronize` action type). Pushing to the branch in 307 (based on develop) after it was PR'd ran these tests twice.

### Pushing to a branch attached to a PR:

|Before|After|
|---|---|
|Screen Shot 2021-05-10 at 3 09 20 PM|Screen Shot 2021-05-10 at 3 09 34 PM|

Note it says "8 successful checks" vs 16.

## Types of changes
Dev tools

### DIFF
--- a/.github/workflows/cs-lint.yml
+++ b/.github/workflows/cs-lint.yml
@@ -4,9 +4,13 @@ on:
   # Run on all pushes and on all pull requests.
   # Prevent the "push" build from running when there are only irrelevant changes.
   push:
+    branches:
+      - develop
+      - trunk
+      - release/**
+  pull_request:
     paths-ignore:
       - '**.md'
-  pull_request:
   # Allow manually triggering the workflow.
   workflow_dispatch:
 
--- a/.github/workflows/cs-lint.yml
+++ b/.github/workflows/cs-lint.yml
@@ -11,6 +11,7 @@ on:
   pull_request:
     paths-ignore:
       - '**.md'
+      - src/js/**
   # Allow manually triggering the workflow.
   workflow_dispatch:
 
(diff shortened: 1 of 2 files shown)


jblz:
The other option I see here is to remove the `pull_request` event and only act on `push`. I think I prefer this, though.

GaryJones:
Questions:
- How would this affect branches that are pushed, but not PR opened (maybe an experimental branch that the author would like to see how the tests and build are handled)?
- How would these changes affect branches created on forks of this repo (i.e. an external contributor who doesn't have write access)? Do they get to see tests run, before opening a PR?

Pushes can be done to feature branches, without the "noise" (Slack notifications, GitHub notifications, etc.) of opening up a PR.

jblz:
Both of those cases (pushes to feature branches w/o a PR / forks w/o a PR) would not initiate the test runs.

If we want those, we should probably only act on `push`. I'll give it a go and see how it acts.

jblz:
I tried out only using `push` in #309 -- it solves for experimental branches pushed here, but not for forks (or even pull requests generated by them).

I did a bit of digging and it [looks like](https://github.community/t/duplicate-checks-on-push-and-pull-request-simultaneous-event/18012/4) the job runs are subtly different since the ref differs (`push` events refer to the latest commit on the instant branch while `pull_request` events refer to the `/refs/pull/:prNumber/merge` branch (so pulling any fast-forwardable changes from `develop`)).

I suppose let's just run both since they're not exactly "duplicates".

PR #318, opened 2021-05-13
Related "Recommended" Content Block

jblz:
## Description

This adds support for a `Parse.ly Related` block designed to showcase links to content on your site as provided by the [Parsely `/related` API endpoint](https://www.parse.ly/help/api/recommendations#get-related).

The block is not loaded at all unless a "feature flag" filter is explicitly added to the theme (or custom plugin) code. See the included README for details.

## Motivation and Context

This is a modern alternative to the ["Recommended Widget"](https://github.com/Parsely/wp-parsely/blob/2.5.0/src/class-parsely-recommended-widget.php) that allows flexible placement via the WordPress Block Editor on posts and pages as well as in site-design elements (such as a footer / sidebar / etc.) via full site editing.

See the [README](https://github.com/Parsely/wp-parsely/blob/5a1d5dccd66aa50e2bac2bb3027f8d52e8b97ebd/src/blocks/recommendations/README.md) included in this PR for details.

## How Has This Been Tested?

Make sure to opt in to block code loading by adding the feature flag filter -- see the included README for details.

Added the block in various contexts by the block editor, changed settings in the block controls and block inspector controls, verified desirable results in the front end of the site.

Edited and updated block attributes & verified desirable results in the front end of the site.

https://user-images.githubusercontent.com/1587282/124648099-d1082580-de64-11eb-9eaf-2e18efb0fada.mov

## Types of changes


src/blocks/recommendations/recommendations-store.js
```diff
@@ -0,0 +1,55 @@
+/**
+ * External dependencies
+ */
+import { createContext, useContext, useReducer } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import { RECO_BLOCK_ERROR, RECO_BLOCK_LOADED, RECO_BLOCK_RECOMMENDATIONS } from './constants';
+// TODO: consider using createReduxStore https://github.com/WordPress/gutenberg/tree/a7e2895829c16ecd77a5ba22d84f1dee1cfb0977/packages/data#registering-a-store
```
jblz, line 10: TODO: consider and move this over or remove this TODO

src/blocks/recommendations/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+Default: `score`
+
+### `tag`
+
+Default: none
+
+### `title`
+
+Default: `Related Content`
```
jblz, line 96: TODO:

* Document the shared data store and actions creators, etc.
* Document undocumented attributes

jblz:
I still have a couple of loose ends to tie up, but I'd like to get some additional 👀 on this at this point.

src/blocks/recommendations/README.md
```diff
@@ -0,0 +1,86 @@
+# Parsely Recommendations Block
+
+This block is designed to showcase links to content on your site as provided by the [Parsely `/related` API endpoint](https://www.parse.ly/help/api/recommendations#get-related).
+
+You can use the Parsely Recommendations Block wherever you use the WordPress Block Editor ("Gutenberg") to build your site. That means that not only can leverage this feature in a single `Post`'s content (or `Page` or nearly any `Custom Post Type`), you can also leverage it in the Full Site Editor's template editor [slated to debut with WordPress version 5.8](https://make.wordpress.org/core/2021/06/16/introducing-the-template-editor-in-wordpress-5-8/) to add it to various layout elements (e.g. a Header, Footer, or Sidebar).
+
+## How to Use
+
+- Inside the Block Editor, add the `Parse.ly Related` block via the [standard block controls](https://wordpress.com/support/wordpress-editor/#add-a-block).
```
GaryJones: There should be a wordpress.org link for this, rather than a wordpress.com link.

src/blocks/recommendations/README.md
```diff
@@ -0,0 +1,86 @@
+# Parsely Recommendations Block
+
+This block is designed to showcase links to content on your site as provided by the [Parsely `/related` API endpoint](https://www.parse.ly/help/api/recommendations#get-related).
+
+You can use the Parsely Recommendations Block wherever you use the WordPress Block Editor ("Gutenberg") to build your site. That means that not only can leverage this feature in a single `Post`'s content (or `Page` or nearly any `Custom Post Type`), you can also leverage it in the Full Site Editor's template editor [slated to debut with WordPress version 5.8](https://make.wordpress.org/core/2021/06/16/introducing-the-template-editor-in-wordpress-5-8/) to add it to various layout elements (e.g. a Header, Footer, or Sidebar).
```
GaryJones:
```suggestion
# Parse.ly Recommendations Block

This block is designed to showcase links to content on your site as provided by the [Parse.ly `/related` API endpoint](https://www.parse.ly/help/api/recommendations#get-related).

You can use the Parse.ly Recommendations Block wherever you use the WordPress Block Editor ("Gutenberg") to build your site. That means that not only can leverage this feature in single `Post`'s content (or `Page` or nearly any custom post type), you can also leverage it in the Full Site Editor's template editor [slated to debut with WordPress version 5.8](https://make.wordpress.org/core/2021/06/16/introducing-the-template-editor-in-wordpress-5-8/) to add it to various layout elements (e.g. a Header, Footer, or Sidebar).
```

src/blocks/recommendations/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+#### Options
+
+- `grid`
+
+  Display the results in a grid of cards that fills the available space horizontally and adds rows as needed.
+
+- `list`
+
+  Display the results in a vertically-oriented list
```
GaryJones:
```suggestion
  Display the results in a vertically-oriented list.
```

src/blocks/recommendations/index.php
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * @package Parsely
+ */
+
+/**
+ * Wrapper function to kick off the plugin. Sources & hooks initialization functions for the block & API classes.
+ *
+ * @return void
+ */
+function wp_parsely_recommendations_block_init() {
```
GaryJones: What's the benefit to putting this in a function and then immediately calling the function? There are no locally scoped variables - in fact the only variable used is a global one.

src/class-parsely-recommended-content.php
```diff
@@ -0,0 +1,49 @@
...
+	/**
+	 * Get the URL for the Recommendation API (GET /related).
+	 *
+	 * @see https://www.parse.ly/help/api/recommendations#get-related
+	 *
+	 * @since 2.6.0
+	 *
+	 * @param string $api_key          Publisher Site ID (API key).
+	 * @param int    $published_within Publication filter start date; see https://www.parse.ly/help/api/time for
+	 *                                 formatting details. No restriction by default.
```
GaryJones, line 23: Needs something in here to indicate that this is the number of **days**.

tests/recommended-api-test.php
```diff
@@ -77,6 +78,6 @@ public function data_recommended_api_url() {
 	public function test_recommended_api_url( $api_key, $published_within, $sort, $boost, $return_limit, $url ) {
 		$recommended_widget = new Parsely_Recommended_Widget();
```
GaryJones:
```suggestion
```

No longer needed it seems.

GaryJones (commented):
I've skimmed the code and left a few tiny bits.

All the new methods (at least for PHP), should have `@since 2.6.0` annotations added.

I've not yet tested this at runtime.

src/blocks/recommendations/README.md
jblz: Good catch!  I pasted the wrong one!

jblz:
```suggestion
- Inside the Block Editor, add the `Parse.ly Related` block via the [standard block controls](https://wordpress.org/support/article/adding-a-new-block/).
```

src/blocks/recommendations/index.php
jblz: It's just a habit to put everything in a function. Functions are cheap :)

I'll see how it feels pulling it out.

mbierly:
I did a bit of testing of this block using Andy Rhinehart's Polk Sports site. I recorded a screenshare of some my first impressions as a user. Some of the things I noticed might be virtues of the way Polk Sports is set up. I also don't know how much of what I'm seeing is because of settings in the Parse.ly API vs the block. I'm unaware of the technical limitations we're facing in building this - but that's the same case as our users.

**Video of first impressions:** https://www.loom.com/share/6e2997d34afb447980ea8fbc6c90a08f

* The Parse.ly thumbnail option doesn't appear to work
* In the dropdown list under boost there's a typo - Average is misspelled as "Avergae"
* In some cases, the same post is shown twice in the recommendation results
* Sometimes the post that I'm on is included in the recommendation results
* I kept waiting for the results within the editor to change while I was making changes to Sort or Boost. The results in the editor don't match the published results. I found it difficult while I was in the editor to trust that the block was working because it felt like when I changed those options, the results should update just like the other options (grid, images, etc) do.
* Is there a way for us to inherit the style of the site, or is that not possible within the block? The fonts/styling etc don't seem to match with the rest of the site, and I think that might be a sticking point for users. (I guess they can edit the CSS, right?)
* I'm assuming that by putting this in a block, it will be exposed to users who have never read our API documentation. They probably don't know what "Boost" means, or that "Tag" means that we'll filter results by tag, or what "Save results to block" will mean. At the least, I think we might want to link to some documentation from within the block, but some helper text would probably help too. (I can help with product copy if you want!)
* Consider renaming the block to Parse.ly Recommendations - it's kind of unclear to me what I will get from the block when I just see it in the list.
* A question: are there user permissions on who can see this block? I'm imagining making this block available to everyone in a newsroom and editors and senior leadership not wanting journalists to be able to add recommendation blocks willy nilly.

src/blocks/recommendations/class-parsely-recommendations-block-api.php
```diff
@@ -0,0 +1,137 @@
...
+			Parsely_Recommended_Content::get_api_url(
+				$apikey,
+				$pub_start,
+				$sort_recs,
+				$boost,
+				$limit
+			)
+		);
+
+		$response = wp_safe_remote_get( $full_api_url, array() );
```
mikeyarce, line 122: Should we check or VIP and use `vip_safe_wp_remote_get()` here if available?

src/blocks/recommendations/components/parsely-recommendations-title.js
```diff
@@ -0,0 +1,5 @@
+const ParselyRecommendationsTitle = ( { title } ) => (
+	<>{ title && <p className="parsely-recommendations__list-title">{ title }</p> }</>
```
mikeyarce, line 2: I have never seen this before - what are `<>` and `</>`  for?

mikeyarce:
First run through and some general UX notes!

## Results slider
The results slider is set from 1-99 - which means that it's sort of evenly spread along the slider.  However, precision is tough if you want to select something like 10 or 15 because the slider is so small and it has so many possibilities.   Maybe we should consider making this just a number input?

parsely-block-results-slider

## Performance (caching)

Does the `fetchApi` have any built-in caching sort of like Apollo's `InMemoryCache`?  I think it might be nice to have a default of like 5 minutes to cache related content and then let it be filtered to a different amount of time.

Currently, on every load I see a `Loading..` and then the fetch happens and then the content loads.  I think because this might affect things like Cumulative Layout Shift, maybe caching could be helpful?   

It appears like it should support all `fetch` options, but since this API request sends a `cache-control: no-cache` header, it won't be cached unless we set `force-cache`.

I also noticed that `get_items` does includes object caching but only when the items are coming from the REST API endpoint `wp-json/wp-parsely/v1/recommendations` which is not triggered by the block.   Is that intentional or should the block be using this method as well?   I feel like that could be useful.

## Performance ( multiple requests )

If I add two blocks that make the same request, the request seems to happen twice - I think that should be ok if we implement caching, but just wanted to make a note of this.  Also if someone decides to not cache results it might be good to de-dupe if it's the same query twice?    

I could see this happening if there's a loop going on where every post has this block present, which could cause for some slowness.
 
## Theme Compatibility

It looks like on some themes the related content items display in wrong proportions:

Twenty Nineteen:
Screen Shot 2021-07-09 at 1 43 48 PM

Twenty Twenty:
Screen Shot 2021-07-09 at 1 44 32 PM

I think maybe one solution would be to change `grid-template-columns: repeat(repeat(auto-fill, 13ch));` into `grid-template-columns: repeat(auto-fill, minmax(150px, 1fr));` - since the thumbnails appear to always be 150x150, is that right?

## Thumbnails

> The Parse.ly thumbnail option doesn't appear to work

In my tests they do work, but they also do appear worse quality

Parsely:
Screen Shot 2021-07-09 at 1 56 35 PM

Non-Parsely:
Screen Shot 2021-07-09 at 1 56 41 PM

> I kept waiting for the results within the editor to change while I was making changes to Sort or Boost

I had the same problem - Boost seems to have no effect on the results?

## Not loading at times

There are some instances where the block won't load on the front end... not totally sure why but sometimes it not show and these errors would be in the console:

Screen Shot 2021-07-09 at 2 10 14 PM

There is one error that's related to `recommendations.js`:

```
recommendations.js?ver=2179fdd3f49d58a3403c6c051888db79:1 Uncaught TypeError: Cannot read property 'createContext' of undefined
    at Object.<anonymous> (recommendations.js?ver=2179fdd3f49d58a3403c6c051888db79:1)
    at r (recommendations.js?ver=2179fdd3f49d58a3403c6c051888db79:1)
    at Object.<anonymous> (recommendations.js?ver=2179fdd3f49d58a3403c6c051888db79:1)
    at r (recommendations.js?ver=2179fdd3f49d58a3403c6c051888db79:1)
    at Module.<anonymous> (recommendations.js?ver=2179fdd3f49d58a3403c6c051888db79:1)
    at r (recommendations.js?ver=2179fdd3f49d58a3403c6c051888db79:1)
    at recommendations.js?ver=2179fdd3f49d58a3403c6c051888db79:1
    at recommendations.js?ver=2179fdd3f49d58a3403c6c051888db79:1
```

## Save results to block is broken (I think only locally)

If I select `Save Results to block` - the block does not display on the front end. 

Markup:

```
<section class="wp-block-wp-parsely-recommendations" data-boost="views" data-layoutstyle="grid" data-imagestyle="original" data-limit="3" data-personalized="" data-saveresults="1" data-savedresults="[]" data-showimages="1" data-sort="score" data-title="Related Content"></section>
```
I think it's because it's going through `wp_validate_redirect` so because it's my test site the and the URLs don't match, it wont' work.   Maybe we need to have some ENV variables here for development sites?

## Minimum version of WP / Gutenberg

- Have you tested with `MINIMUM_WORDPRESS_VERSION` which is 5.6?  
- What about if someone is running the Gutenberg plugin?

mbierly:
I had another thought while I was making dinner last night. Will this block appear for wp-parsely users who do not have access to the Parse.ly API? What's the experience like in that case?

Related: https://github.com/Parsely/web/issues/11984#issuecomment-879283730

GaryJones:
## Boost

> Boost seems to have no effect on the results?

As per [the docs](https://github.com/Parsely/wp-parsely/pull/parse.ly/help/api/recommendations#get-related), the boost parameter is only usable when the `sort=score`. When it is `sort=pub_date`, then it is redundant, so there is no need to show that field.

## Save Results to block

I'm not sure what "Save Results to block" is - this either needs to be called something more descriptive, or it needs some help text somewhere.

## Title in widget output

The output from a Gutenberg plugin-based block widget screen, the legacy widget gives a widget title with markup of:

```html
<h2 class="widget-title">Legacy Widget</h2>
```

With the new block, it's:

```html
<p class="parsely-recommendations__list-title">Related Content</p>
```
It neither uses the widget heading level, not has the very common `widget-title` class.

The highlighted line shows that core widgets also have the h2 with the widget-title class.

Screenshot 2021-07-13 at 23 32 13

## Featured image

As Mikey mentioned, the featured image in the new block doesn't seem right - here's a comparison of the legacy widget on the left, with the block-as-a-widget on the right:

Screenshot 2021-07-13 at 23 24 27

Poor Tess!

## Block Settings

Here are the block settings for a pull quote:

Screenshot 2021-07-13 at 23 39 43

Note how there's a Styles, Color, Text Color, and then the Advanced (closed by default) sections?

Here's the current view of the Parse.ly block settings:

Screenshot 2021-07-13 at 23 39 53

It's all in a less-then-helpfully titled Settings section, followed by the default Advanced section. The number of settings makes it a bit trickier to go through, compared to the collapsible sections.

I would suggest splitting up the Parse.ly Recommendations block settings into sections i.e. grouping the appearance settings together, and the behaviour settings together, etc. 

## Visual Style Selector

Following on, here are two Styles sections from image block and pull quote block:

Screenshot 2021-07-13 at 23 37 55
Screenshot 2021-07-13 at 23 37 42

Can the Parse.ly Display Style setting (grid vs list) get a similar visual selector, instead of the radio buttons?

## Item Count UI

+1 for changing the UI of the number of items selector - I also managed to struggle to use the slider to go from 3 to 6 items, and the picker buttons go up in groups of 10 by default, unless CMD is held down - quite unintuitive (but maybe that's a core component). Would be fine if this was used for width or some other large number.

## CSS

For the CSS, I would suggest keeping the styles minimal - no white background, no box-shadow, no text-align: center for the grid, etc. 

You could always add a Colors section to the settings if you wanted users to be able to edit the appearance.

jblz:
I'm going to go ahead and close this for now.

The rebase was a mess, so I squashed everything together into a single commit in a new branch and we can pick it up over there: #597

PR #328, opened 2021-05-27
Add conditional for CPT archives and CPT term archives

### LINKED ISSUE #327: Metadata:  Custom Post type category pages don't display the right metadata
## Describe the bug

When you have a custom post type with a category, that category page doesn't display the correct metadata.

## To Reproduce

1. Activate the Parse.ly plugin on a site (I'm using 2.6.0-alpha) 
2. Create a custom post type that has categories
3. Add a category to that post type
4. Add a few posts to the post type and assign them that category
5. Navigate to the category page and look at the `application/ld+json` data from Parse.ly

What you will see is this:

```
{"@context":"http:\/\/schema.org","@type":"NewsArticle","mainEntityOfPage":{"@type":"WebPage","@id":"http:\/\/plugin-dev.test\/recipe\/2\/"},"headline":"2","url":"http:\/\/plugin-dev.test\/recipe\/2\/","thumbnailUrl":"","image":{"@type":"ImageObject","url":""},"dateCreated":"2021-05-27T21:25:03Z","datePublished":"2021-05-27T21:25:03Z","dateModified":"2021-05-27T21:25:03Z","articleSection":"Uncategorized","author":[{"@type":"Person","name":"mikey"}],"creator":["mikey"],"publisher":{"@type":"Organization","name":"Plugins","logo":""},"keywords":[]}
```
It's bringing up the metadata from the first post. 

## Expected behaviour
It should be bringing in the metadata for the main category, not for the first post in the loop.

I think the problem is we only have an `is_category()` check, but that won't apply for a custom post type.  What we need is to add another check like `is_post_type_archive()` to that logic to catch this.

mikeyarce:
## Description
When you register a custom post type with a taxonomy, the archive pages don't have the right metadata, they pick up the metadata from the first item in the loop instead of the main archive page.

What I'm trying here is adding the `is_post_type_archive()` conditional and switching to getting the queried object instead of just calling `get_the_category()` as that doesn't work in all cases.

Edit:  I found that there's another case that I wanted to target: custom post type taxonomy archives.  I've addressed that and added unit tests for that as well. 

Fixes #327 

## Motivation and Context
Fixing this for custom post type archives

## How Has This Been Tested?
1. On your test site, make sure you have a CPT with a category
2. Also make sure you have a regular post type with a category
3. Check this PR out to your test site
4. Go to the Archive pages for both the CPT and the regular post category 
5. Check to see if the right metadata is present for them both.

## Types of changes

Question:  Should we also check if the post type is currently added to the Parse.ly options?


GaryJones:
This could do with some tests.

tests/structured-data/class-archive-post-test.php
```diff
@@ -233,6 +233,143 @@ public function test_term_archive() {
...
+		// Set permalinks, as Parsely currently strips ?page_id=... from the URL property.
+		// See https://github.com/Parsely/wp-parsely/issues/151
+		$this->set_permalink_structure( '/%postname%/' );
+
+		// Setup Parsley object.
+		$parsely         = new \Parsely();
+		$parsely_options = get_option( \Parsely::OPTIONS_KEY );
+
+		// Register custom post type and custom taxonomy for that post type.
+		register_post_type( 'custom_post_type', array( 'public' => true ) );
```
GaryJones: Is the fact that that custom taxonomy is associated with a CPT, relevant here, or just a distraction?

Could you have created a CTax and associated it with a `post`, and been able to test the same thing?

tests/structured-data/class-archive-post-test.php
```diff
@@ -233,6 +233,143 @@ public function test_term_archive() {
...
+		// Setup Parsley object.
+		$parsely         = new \Parsely();
+		$parsely_options = get_option( \Parsely::OPTIONS_KEY );
+
+		// Register Post Type with specific archive URL.
+		register_post_type(
+			'custom_post_type',
+			array(
+				'public'      => true,
+				'has_archive' => 'cpt-tax-archive',
```
GaryJones: If you've got a CPT, why give it an archive of `cpt-tax-archive`? The `-tax` bit seems irrelevant since there's no Custom Taxonomy aspect of this test.

tests/structured-data/class-archive-post-test.php
```diff
@@ -233,6 +233,143 @@ public function test_term_archive() {
...
+		flush_rewrite_rules();
+
+		// Go to the term archive page.
+		$this->go_to( $term_link );
+
+		// The query should be for a taxonomy archive.
+		self::assertQueryTrue( 'is_archive', 'is_tax' );
+
+		// Create the structured data for that category.
+		// The category metadata doesn't use the post data, but the construction method requires it for now.
```
GaryJones:
```suggestion
		// Create the structured data for that term archive.
		// The term archive metadata doesn't use the post data, but the construction method requires it for now.
```

tests/structured-data/class-archive-post-test.php
```diff
@@ -233,6 +233,143 @@ public function test_term_archive() {
...
+		$structured_data = $parsely->construct_parsely_metadata( $parsely_options, get_post() );
+
+		// Check the required properties exist.
+		$this->assert_data_has_required_properties( $structured_data );
+
+		// The headline should be the category name.
```
GaryJones:
```suggestion
		// The headline should be the term name.
```

tests/structured-data/class-archive-post-test.php
```diff
@@ -233,6 +233,143 @@ public function test_term_archive() {
...
+		// Register custom post type and custom taxonomy for that post type.
+		register_post_type( 'custom_post_type', array( 'public' => true ) );
+		register_taxonomy( 'custom_tax', array( 'custom_post_type' ) );
+
+		// Insert a single term, and a Post in that custom post type with the custom term.
+		$term    = self::factory()->term->create(
+			array(
+				'taxonomy' => 'custom_tax',
+				'slug'     => 'term',
+				'name'     => 'Custom Taxonomy',
```
GaryJones:
```suggestion
				'name'     => 'Custom Taxonomy Term',
```

It's easy to get confused between the name of the taxonomy, and the name of a term in that taxonomy, so keep the distinction clear.

The assertion on this name will also need updating.

src/class-parsely.php
```diff
@@ -885,9 +885,8 @@ public function construct_parsely_metadata( array $parsely_options, $post ) {
 			$author                   = ( get_query_var( 'author_name' ) ) ? get_user_by( 'slug', get_query_var( 'author_name' ) ) : get_userdata( get_query_var( 'author' ) );
 			$parsely_page['headline'] = $this->get_clean_parsely_page_value( 'Author - ' . $author->data->display_name );
 			$parsely_page['url']      = $current_url;
-		} elseif ( is_category() ) {
-			$category                 = get_the_category();
-			$category                 = $category[0];
+		} elseif ( is_category() || is_archive() ) {
```
GaryJones: `is_archive()` is true for many types of archives, like date or tag archives, which are below this conditional, and so requests to a date archive, for instance, will be captured here, instead of below, which will change the behaviour; there may not have been unit tests yet to flag this.

For now, use `is_post_type_archive()` as the most narrowest scope conditional for what you're trying to address here.

GaryJones (changes requested):

tests/structured-data/class-archive-post-test.php
mikeyarce: Good question!   I was testing this way because a bug report came in with this specific scenario, but it's really the same regardless of Custom Post Type or not. 

I think for simplicity I'll remove the CPT out of the equation here.

src/class-parsely.php
mikeyarce: `is_post_type_archive()` will only be true for the custom post type archive, so what if we did:

`(  is_category() || is_post_type_archive() || is_tax() ) `

GaryJones: Yep, that should work 👍🏻

tests/structured-data/class-archive-post-test.php
```diff
@@ -318,7 +318,7 @@ public function test_archive_custom_post_type() {
 	 * @uses \Parsely::update_metadata_endpoint
 	 * @group metadata
 	 */
-	public function test_term_archive_custom_post_type() {
+	public function test_custom_term_archive() {
```
GaryJones:
```suggestion
	public function test_custom_taxonomy_term_archive() {
```

All _terms_ are custom ;-) - it's the _taxonomy_ that's custom that is the focus here.

GaryJones (changes requested):
Fix the condition as you suggested, tweak the test method name as I've now added, and then this is good to land. Thank you Mikey 👍🏻

GaryJones (approved):
:shipit:

pauarge:
I have merged `develop` into this branch. I kept the changes, but I had to split the tests into two separate files to follow the pattern that we have adopted since we did this PR.

pauarge (approved):
LGTM

tests/StructuredData/CustomPostTypeArchiveTest.php
```diff
@@ -0,0 +1,83 @@
...
+		self::assertQueryTrue( 'is_archive', 'is_post_type_archive' );
+
+		// Create the structured data for that category.
+		// The category metadata doesn't use the post data, but the construction method requires it for now.
+		$structured_data = $parsely->construct_parsely_metadata( $parsely_options, get_post() );
+
+		// Check the required properties exist.
+		$this->assert_data_has_required_properties( $structured_data );
+
+		// The headline should be the category name.
```
GaryJones:
```suggestion
		// The headline should be the CPT name.
```

tests/StructuredData/CustomPostTypeArchiveTest.php
```diff
@@ -0,0 +1,83 @@
...
+		// phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.flush_rewrite_rules_flush_rewrite_rules
+		flush_rewrite_rules();
+
+		// Go to the custom post type archive page.
+		$this->go_to( home_url( '/cpt-archive' ) );
+
+		// The query should be for a custom post type archive.
+		self::assertQueryTrue( 'is_archive', 'is_post_type_archive' );
+
+		// Create the structured data for that category.
```
GaryJones:
```suggestion
		// Create the structured data for that CPT.
```

tests/StructuredData/CustomPostTypeArchiveTest.php
```diff
@@ -0,0 +1,83 @@
...
+		// The category metadata doesn't use the post data, but the construction method requires it for now.
```
GaryJones:
```suggestion
		// The CPT archive metadata doesn't use the post data, but the construction method requires it for now.
```

tests/StructuredData/CustomTaxonomyTermArchiveTest.php
```diff
@@ -0,0 +1,83 @@
...
+				'name'     => 'Custom Taxonomy Term',
+			)
+		);
+		$post_id = self::factory()->post->create();
+
+		wp_set_post_terms( $post_id, $term, 'custom_tax' );
+
+		$term_link = get_term_link( $term );
+
+		// Flush rewrite rules after creating new post type with archive.
```
GaryJones:
```suggestion
		// Flush rewrite rules after creating new taxonomy type.
```

tests/StructuredData/CustomPostTypeArchiveTest.php
```diff
@@ -0,0 +1,83 @@
...
+	 * @uses \Parsely::get_coauthor_names
+	 * @uses \Parsely::get_current_url
+	 * @uses \Parsely::get_first_image
+	 * @uses \Parsely::get_options
+	 * @uses \Parsely::get_tags
+	 * @uses \Parsely::post_has_trackable_status
+	 * @uses \Parsely::update_metadata_endpoint
+	 * @group metadata
+	 */
+	public function test_archive_custom_post_type() {
```
GaryJones:
```suggestion
	public function test_metadata_is_correctly_constructed_for_custom_post_type_archive() {
```

tests/StructuredData/CustomTaxonomyTermArchiveTest.php
```diff
@@ -0,0 +1,83 @@
...
+	 * @uses \Parsely::get_coauthor_names
+	 * @uses \Parsely::get_current_url
+	 * @uses \Parsely::get_first_image
+	 * @uses \Parsely::get_options
+	 * @uses \Parsely::get_tags
+	 * @uses \Parsely::post_has_trackable_status
+	 * @uses \Parsely::update_metadata_endpoint
+	 * @group metadata
+	 */
+	public function test_custom_taxonomy_term_archive() {
```
GaryJones:
```suggestion
	public function test_metadata_is_correctly_constructed_for_custom_taxonomy_term_archive() {
```

GaryJones (changes requested):
Minor tweaks.

GaryJones (approved):

PR #331, opened 2021-06-02
Lint fixes

### LINKED ISSUE #330: Unused Variable: Default Logo
Noticed while stepping through phpcs issues:

At:
https://github.com/Parsely/wp-parsely/blob/e1c3e43d98ce7e45d4c990421a690a677abdf829/src/class-parsely.php#L386

This variable is set, but never referenced:
`$option_defaults['logo'] = $this->get_logo_default();`

It looks like the intent is to populate the private member variable declared here:
https://github.com/Parsely/wp-parsely/blob/e1c3e43d98ce7e45d4c990421a690a677abdf829/src/class-parsely.php#L48

The `get_logo_default` method is also called in the `validate_options` method here:
https://github.com/Parsely/wp-parsely/blob/e1c3e43d98ce7e45d4c990421a690a677abdf829/src/class-parsely.php#L647-L649

...so I think it's ok to just remove L#386, but we should make sure the default logo is being set appropriately.

jblz:
## Description

Some things that PHP CodeSniffer hipped us to:
* incorrect indention levels
* multi-line function call styling
* string array member spacing
* ternary op spacing
* spacing before EOL semicolon
* Unused variable (Fixes #330)
* spacing issues in built assets (this change excludes them)

## Motivation and Context

Clear out some warnings so we can concentrate on more serious issues.

## How Has This Been Tested?

TBD

## Types of changes

### DIFF
--- a/.phpcs.xml.dist
+++ b/.phpcs.xml.dist
@@ -10,6 +10,7 @@
 	<!-- Ignoring Files and Folders:
 		https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#ignoring-files-and-folders -->
 	<exclude-pattern>/vendor/</exclude-pattern>
+	<exclude-pattern>/build/*.asset.php</exclude-pattern>
 
 	<!-- How to scan -->
 	<!-- Usage instructions: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Usage -->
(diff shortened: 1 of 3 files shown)


jblz:
Superseded by #337 -- we should still pull out 1730ea0 to #330 in a new PR

PR #332, opened 2021-06-03
API Improvements for REST-powered Admin page

jblz:
## Description

Targeting `remove/jquery-dependency--try-react` branch in #239.  This is intended to be merged in there.

## Motivation and Context

So we can update settings over the REST API.

## How Has This Been Tested?

* Check out branch
* `npm i` to pull in the new dependency
* `npm run start` to build the updated client
* browse to settings page
  * settings should still pull from the API and populate the state tree
* click the submit button
  * settings should be sent to the new `POST` endpoint and interpreted there

The write endpoint will likely not succeed as-is...we need to make some changes to the client code to send the data appropriately to pass the validation.


src/settings-api.php
```diff
@@ -1,16 +1,68 @@
...
+ *
+ * @param WP_REST_Request $request Full data about the request.
+ * @return WP_REST_Response
+ */
+function wp_parsely_settings_api_write_settings( WP_REST_Request $request ) {
+	global $parsely, $wp_settings_errors;
+
+	// `$parsely->validate_options` uses `add_settings_error`
+	// TODO: Decouple!
+	require_once ABSPATH . 'wp-admin/includes/template.php';
```
jblz, line 35: This is pretty gross.  We should refactor the validation function:
https://github.com/Parsely/wp-parsely/blob/e1c3e43d98ce7e45d4c990421a690a677abdf829/src/class-parsely.php#L620
...to not call the Settings API functionality.

Let's do it in a follow-up, though, so this unblocks development on the updating side of things.

Issue #179, opened 2021-04-02
Split tests into unit and integration (WP) tests


GaryJones:
See https://github.com/Automattic/WPCOM-Legacy-Redirector/pull/89 and https://github.com/Automattic/WPCOM-Legacy-Redirector/tree/develop/tests for some prior art on this.

Unit tests can be run on a local host machine. They are quick, as they mock WP functions (via [BrainMonkey](https://brain-wp.github.io/BrainMonkey/)) and therefore don’t need to load WordPress to work. These tests typically focus on individual public methods and functions, and should check against the happy path, unhappy path, and how types are handled, since still supporting PHP 5.6 means the plugin lacks some strongly typed type declarations for parameters and returns.

WP integration tests need a local server (or GitHub Actions) to be able to load WordPress and use the database. The existing tests are these type of tests, though some may be able to be moved into unit tests. Integration tests take longer to run.

[WP-Test-Utils](https://github.com/Yoast/wp-test-utils/) is a cross-version compatibility package for PHPUnit that allows for PHPUnit 9 syntax and methods to be used in unit tests. WordPress has a hard cap of PHPUnit 7.5 though, so the WP integration tests need to use that.

GaryJones:
We only have one unit test (that extends a `Yoast\WPTestUtils\BrainMonkey\TestCase`), but this needs moving out to avoid an issue where the wrong Mockery trait is loaded due to the presence of the WordPress test library declaring a class alias.

While we could have `tests/Unit` and `tests/Integrations` directories, looking at https://github.com/Yoast/wpseo-woocommerce and similar, then a `tests/` and `integration-tests/` would be possible. Alternatively, we call it `tests/` and `wp-tests/` etc.

PR #333, opened 2021-06-04
Tests: Add WP Test Utils package

GaryJones:
## Description
The [WP Test Utils](https://github.com/Yoast/wp-test-utils) package is a PHPUnit cross-version compatibility layer for testing plugins and themes build for WordPress.

This PR adds it as a dependency, and makes changes to integrate it.

## Motivation and Context

The package has several useful features, but the features we'll use to start with, are:

- Cross-version compatibility with PHPUnit 5.7 - 9.x via the [PHPUnit Polyfills](https://packagist.org/packages/yoast/phpunit-polyfills) package.
  _Note: WordPress Core limit tests to running on PHPUnit 7.5 max. However, using these polyfills we can already start using the up-to-date PHPUnit 9.x syntax, even though the tests don't use PHPUnit 9 yet._
- A custom autoloader which will selectively autoload the WP copies of the PHPUnit 9 native MockBuilder files for PHPUnit < 9 when run on PHP 8 to get round the use of the new reserved keyword `match` as was used in older versions of these files. Without this, we can't use the MockBuilder in PHPUnit.

In the future, we can split out unit tests from WP (integration) tests (#179), and WP Test Utils provides testcases that have [BrainMonkey](https://brain-wp.github.io/BrainMonkey/) (test utility package with specific mocking features for WordPress) support as well as pre-defining some WP-specific constants that our code might be using.

The version of wp-test-utils is pinned to a commit-ref that addresses a [particular issue](https://github.com/Yoast/wp-test-utils/issues/3) when using the standard WP-CLI tests scaffolding directory structure.

The `cs-lint.yml` change is to stop the action failing on the warning that comes with using a commit-ref.

The bootstrap file changes ensure the custom autoloader works correctly, and mocks can be created.

## How Has This Been Tested?
I've written integration tests that attempt to use mocking, and use assertion methods from newer versions of PHPUnit. These tests will be added in a later PR, so this PR can be seen as laying the groundwork. 

See Actions results to see that the existing tests are still passing.

Screenshot 2021-06-04 at 23 41 10

## Types of changes

- [x] ~New feature~ Maintenance (non-breaking change which adds functionality)

### DIFF
--- a/.github/workflows/cs-lint.yml
+++ b/.github/workflows/cs-lint.yml
@@ -42,7 +42,7 @@ jobs:
       # Validate the composer.json file.
       # @link https://getcomposer.org/doc/03-cli.md#validate
       - name: Validate Composer installation
-        run: composer validate --no-check-all --strict
+        run: composer validate --no-check-all
 
       # Install dependencies and handle caching in one go.
       # @link https://github.com/marketplace/actions/install-composer-dependencies
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,8 @@
 		"phpcompatibility/phpcompatibility-wp": "^2.1",
 		"phpunit/phpunit": "^4 || ^5 || ^6 || ^7",
 		"squizlabs/php_codesniffer": "^3.5",
-		"wp-coding-standards/wpcs": "^2.3.0"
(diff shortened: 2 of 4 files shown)


jblz (approved):
💥

PR #334, opened 2021-06-05
Declare setting name, pass it back with onChange, leverage WP Control Components

jblz:
## Description

Targeting branch in #239. this is intended to merge into that PR.

## Motivation and Context

Using built-in "control" components gives us standardized accessibility gains and simplifies our app code.
Pulling the setting name out to its own prop simplifies the Settings component.
Referencing the setting name instead of the type of the value makes for more consistent handling of the setting in the face of corrupted options, etc.


jase-d-ace:
This looks so good. Thanks for taking the time to make sure we're in line with accessibility standards! I've never done it before, so this is a big learning experience for me. Should I merge this or do you want to do it?

jblz:
@jase-d-ace merge when ready

PR #337, opened 2021-06-16
Fix all Coding Standards violations

GaryJones:
## Description
Fix many of the code standards violations. The only outstanding ones are in the tests/ directory, and they will be addressed once #272 is merged.

## Motivation and Context
By fixing the violations, we can eventually turn on PHPCS checks as part of the CI. Following consistent code standards helps us to avoid bugs, avoid performance and security issues, and use the same style of code among different authors.

## How Has This Been Tested?
`composer cs` - and also `phpcs --report-summary` if you just want to see what files have violations in them.

## Types of changes

- [X] Maintenance (improves the code in some way that isn't to fix a bug)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #339, opened 2021-06-18
CI: Enable CS checks on CI

GaryJones:
## Description
Enable code standards checks (via PHPCS) on all pushes and PRs, and also lint the PHPCS config file, similar to how the PHPUnit config file was already being linted.

## Motivation and Context
Now that #337 addressed all of the remaining CS issues, we can now enable this check here to stop future violations from being merged into the codebase.

## How Has This Been Tested?
See https://github.com/Parsely/wp-parsely/runs/2856943521?check_suite_focus=true that shows the Action from the updated workflow for the push to the branch of this PR, passing the two new steps successfully.

## Types of changes

- [X] Maintenance


jblz (approved):
lovely, thx

PR #341, opened 2021-06-18
Tests: Improve Other Tests (fka all-tests)

### LINKED ISSUE #340: Some tests don't have correct code coverage annotations
## Describe the bug
As per the "integration tests with code coverage" step of https://github.com/Parsely/wp-parsely/runs/2856943725?check_suite_focus=true some tests don't have the correct `@covers` and `@uses` annotations, so are not being tracked in code coverage metrics correctly.

## To Reproduce
See https://github.com/Parsely/wp-parsely/runs/2856943725?check_suite_focus=true

## Expected behaviour
No risky tests due to incorrect or missing coverage annotations.

## Environment

GitHub Action, or locally if run with `composer coverage`.

GaryJones:
## Description
- Rename from `class-all-tests.php` to `OtherTests.php`, since many tests are now in separate files. Over time, I expect the tests in this file to also be moved to other files, but the change makes sense for now.
- Renamed the class to match the PSR-4 naming requirement.
- Updated code coverage configuration to run with WP multisite, so that multisite-specific tests are not skipped.
- Fixed `@covers`, `@coversNothing` and `@uses` annotations to stop tests being marked as risky for strict code coverage.
- Marked class a final, like other test classes.
- Removed redundant TestCase alias.
- Used a regexp-specific assertion.
- Avoided a `strlen()` function call by just comparing against an empty string.
- Used a PHPUnit 8 assertion where it made sense to (polyfilled for PHP 7.5), which then also allowed the assertion to run on PHP 5.6.

## Motivation and Context
Fixes #340 - coverage annotations.
See #272 - PSR-4 test-file naming.
Other small inspection fixes.

## How Has This Been Tested?
Ran `composer coverage` locally.

## Types of changes

- [X] Maintenance

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,6 +32,7 @@
 	},
 	"scripts": {
 		"coverage": [
+			"@putenv WP_MULTISITE=1",
 			"@php ./vendor/bin/phpunit --coverage-html ./build/coverage-html"
 		],
 		"coverage-ci": [
--- a/tests/OtherTest.php
+++ b/tests/OtherTest.php
@@ -7,18 +7,11 @@
 
 namespace Parsely\Tests;
 
-use Parsely\Tests\TestCase as ParselyTestCase;
-use PHPUnit\Framework\Error\Warning as PHPUnit_Warning;
-
 /**
  * Catch-all class for testing.
- * TODO: Break this into multiple targeted files
- *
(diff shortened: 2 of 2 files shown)


jblz (approved):
lgtm

PR #345, opened 2021-06-24
Integrations: Extract into separate classes

GaryJones:
## Description
- Extract code for AMP and Facebook Instant Article integrations into new classes in separate files, out of the `Parsely` class.
- Improve and add tests that better check for the intended behaviour - adds 6 new tests and 19 new assertions.

It's the same behaviour for the end-user and site.

See #96.

## Motivation and Context
The Parsely class does far too much, and needs to be split out into more logical classes to make it easier to maintain.

One area where this can be done is the extra code that makes the plugin work with other plugins i.e. integrations. 

By creating a class for each integration, we logically group related code together, making it easier to see what integrations there are. The use of a collection and an interface (though this interface isn't yet checked for when registering an integration to the collection) provides a structure for built-in and other plugins to create their own integrations between Parse.ly and another plugin.

### AMP

1. With the AMP plugin not active, the `amp_*` filters won't run, so there are no negative side effects.
2. Install and activate the [AMP plugin](https://wordpress.org/plugins/amp/).
3. Visit any post and add `/amp` to the end of the URL.
4. Inside the source of the resulting page, look for a `<amp-analytics` opening tag that contains references to Parse.ly.

Screenshot 2021-09-18 at 16 28 33

### Facebook Instant Articles

1. With the FBIA plugin not active, the `instant_articles_*` filter won't run, so there are no negative side effects.
2. Install and activate the [Facebook Instant Articles plugin](https://wordpress.org/plugins/fb-instant-articles/). Visit the Settings->Permalinks page to flush the rewrite rules.
3. Visit the `/wp-admin/admin.php?page=instant-articles-wizard` admin page. If only one field is visible, enter any random string and save to then see many more settings.
4. Look for the Analytics section, and enable the Parse.ly Analytics checkbox.
5. Visit `/feed/instant-articles` and you should see XML. Look for the Parsely embed code from `Parsely\Integrations\Facebook_Instant_Articles::get_embed_code()`.

Screenshot 2021-09-18 at 16 31 29
 
Screenshot 2021-09-18 at 16 29 56

Coverage metrics for the new Integrations directory:

Screenshot 2021-09-18 at 17 06 03

Integrations directory coverage compared to the rest of the plugin:
Screenshot 2021-09-18 at 17 05 58

Integrations-related tests:

Screenshot 2021-09-18 at 17 18 13

## Types of changes

- [X] Maintenance (non-breaking change which fixes an issue)


src/Integrations/class-facebook-instant-articles.php
```diff
@@ -0,0 +1,77 @@
...
+				onload: function() {
+					PARSELY.beacon.trackPageView({
+						urlref: \'http://facebook.com/instantarticles\'
+					});
+					return true;
+				}
+			}
+		</script>
+		<script data-cfasync="false" id="parsely-cfg" data-parsely-site="' . esc_attr( $api_key ) . '" src="//cdn.parsely.com/keys/' . esc_attr( $api_key ) . '/p.js"></script>
+		<!-- END Parse.ly Include: Standard -->';
```
pauarge: I am trying to find where the `<!-- START Parse.ly Include: Standard -->'` tag is in the repository, but I don't manage to find it. Is it possible that it got lost in some other PR?

pauarge (changes requested):
This is awesome, it really improves the plugin structure! I left a minor question

src/Integrations/class-facebook-instant-articles.php
GaryJones: Did it ever exist? Does the end comment actually add any value?

pauarge: I'm not sure if it ever existed (for sure it does not in `trunk` currently). I think we're safe by getting rid of it.

GaryJones: Resolved in the latest change.

src/Integrations/class-facebook-instant-articles.php
```diff
@@ -0,0 +1,76 @@
...
+	 * Get the payload / embed code.
+	 *
+	 * @since 2.6.0
+	 *
+	 * @param string $api_key API key.
+	 *
+	 * @return string
+	 */
+	public function get_embed_code( $api_key ) {
+		$embed_code = '<script>
```
pauarge: Could we return `$embed_code` directly without having to declare the variable?

GaryJones: Yes, done :-)

pauarge (approved):

PR #355, opened 2021-08-03
Load the api init script before the tracker so values are populated

jblz:
## Description

The function that enqueues the tracker script from the CDN is currently firing prior to the one that enqueues the API init script.

This change reverses that and makes explicit via the [`add_action`'s `$priority` param](https://developer.wordpress.org/reference/functions/add_action/#parameters) that it should run in that order.

## Motivation and Context

The `global.PARSELY.onload` function has to be populated prior to the tracker being loaded so it can actually be called and the `/profile` API call conditionally fired.

See: https://github.com/Parsely/wp-parsely/blob/12976fb9feba317caa3a51236a11fdfe385b133b/src/js/lib/init-api/index.js#L18-L30

## How Has This Been Tested?

On a sandboxed version of blog.parse.ly

## Types of changes

### DIFF
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -123,8 +123,8 @@ class Parsely {
 		// inserting parsely code.
 		add_action( 'wp_head', array( $this, 'insert_parsely_page' ) );
 		add_action( 'init', array( $this, 'register_js' ) );
+		add_action( 'wp_footer', array( $this, 'load_js_api' ), 9 );
 		add_action( 'wp_footer', array( $this, 'load_js_tracker' ) );
-		add_action( 'wp_footer', array( $this, 'load_js_api' ) );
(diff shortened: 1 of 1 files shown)


GaryJones:
What advantage does changing it to priority 9 make? The move to change the order in which it is registered should be enough.

jblz:
> What advantage does changing it to priority 9 make? The move to change the order in which it is registered should be enough.

It makes no functional difference in this change -- it just calls out that it should occur in that order as a visual hint. 

If the lines were to be inadvertently reversed e.g in a future refactoring, it would not break the load order.

GaryJones:
So, add a code comment, rather than change the behaviour to something less standard?

jblz:
done

GaryJones (approved):

PR #364, opened 2021-08-30
Fix cookie parsing -- values can contain equal signs

jblz:
## Description

Pull in a library to parse out the value for the `_parsely_visitor` cookie instead of rolling our own.

Additionally, pass the `uuid` through `encodeURIComponet` for good measure.

## Motivation and Context

Unfortunately, when the recommended content widget was refactored in #279, the lib that we wrote to parse out cookies neglected to take into account cookie values that contain an equal sign (`=`). I noticed on a site that has personalization enabled, the cookie containing the uuid was not getting parsed. This means that personalization of results in the widget has been broken since version 2.5.

The [js-cookie library](https://www.npmjs.com/package/js-cookie) is active, mature, well-tested, and relatively small. Leveraging it allows us to rely on its battery of testing. Even so, I've adapted the testing for our previous lib to cover the previously covered use cases as well as the issue this PR intends to fix all the same.

### Automated testing

`npm run test`

### Manual testing

* Run this branch on a site with API access
* Configure the widget to display (in a sidebar, footer, etc.)
* Browse to a page with the widget on it
    * Confirm that personalized results are requested / received
    * Namely, you should see a populated `uuid` query argument in the call to `https://api.parsely.com/v2/related` as per [this line](https://github.com/Parsely/wp-parsely/blob/f465d3db4536bddb22b5ffb9388e405a87e240be/src/js/widgets/recommended/index.js#L31).

## Types of changes


jblz:
It looks like the uncompressed bundle size difference is about 1.1K. Seems acceptable to me.

GaryJones (changes requested):
Why do we have tests for a third-party component? This isn't our code to test. Nothing about these tests is about the _integration_ of js-cookie into our code; it's just testing the raw js-cookie behaviour.

In terms of the confidence that `=` signs in cookie values are handled correctly by js-cookie, then https://github.com/js-cookie/js-cookie/blob/0ba77141dd215782cc7770347a457906908c66ff/test/encoding.js#L315-L334 (at a very quick skim) would seem to cover that.

Instead, what I'd rather see are tests in our plugin for `widgetLoad()` that grab a cookie value with a `=` and ensure the fullUrl that is created is correct - and if that isn't easy to write a test for, then `widgetLoad()` needs to be broken down into smaller functions so that the logical parts of it can be tested.

Edit: Of course, writing and committing _that_ integration test first, and see it fail with the existing code, and then adding in js-cookie and see it pass, would be the best workflow.

src/js/widgets/recommended/index.js
```diff
@@ -20,19 +18,11 @@ function widgetLoad( {
...
-			uuid = JSON.parse( unescape( cookieVal ) ).id;
-		} catch ( e ) {}
-	}
-
 	let fullUrl = apiUrl;
+	const uuid = personalized ? getUuidFromVisitorCookie() : undefined;
 
-	if ( personalized && uuid ) {
-		fullUrl += `&uuid=${ uuid }`;
+	if ( uuid ) {
```
jblz, line 24: This change is so we don't waste the i/o looking at cookies if there's no chance of using it. `uuid` is only referenced when `personalized` is truthy.

tests/js/lib/personalization.test.js
```diff
@@ -0,0 +1,143 @@
...
+	getVisitorCookieRaw,
+} from '../../../src/js/lib/personalization';
+
+describe( 'lib/personalization', () => {
+	beforeEach( () => {
+		Object.keys( getCookie() ).forEach( ( cookieName ) => removeCookie( cookieName ) );
+	} );
+
+	describe( 'getVisitorCookieRaw', () => {
+		test( 'no set cookies should return undefined', () => {
```
GaryJones: The wording of this seems ambiguous - I read that as "no cookies which are set, should return undefined" followed by a test that expects the raw cookie value to be undefined.

Try:
```suggestion
		test( 'no cookies are set, so should return undefined', () => {
```

tests/js/lib/personalization.test.js
```diff
@@ -0,0 +1,143 @@
...
+
+		test( 'visitor cookie with empty value should return empty string', () => {
+			global.document.cookie = '_parsely_visitor=';
+			global.document.cookie = 'genre=Science Fiction';
+			expect( getVisitorCookieRaw() ).toBe( '' );
+		} );
+
+		test( 'set visitor cookie should return value', () => {
+			global.document.cookie = 'genre=Science Fiction';
+			global.document.cookie = '_parsely_visitor=Species 8472';
```
GaryJones, line 38: That's...niche, even amongst the rest of the geekness 😁

GaryJones (approved):
One minor suggestion on a test name, but other LGTM.

PR #370, opened 2021-09-11
Fix repeated metas notices

### LINKED ISSUE #365: Empty meta values should not be printed
## Describe the bug

When the [`meta_type` is set to `repeated_metas`](https://github.com/Parsely/wp-parsely/blob/e196413774b645423fe7022dc4b0e2742620a534/src/class-parsely.php#L873-L880
), we print a series of `<meta />` tags in the page source:

https://github.com/Parsely/wp-parsely/blob/e196413774b645423fe7022dc4b0e2742620a534/views/repeated-metas.php#L12-L22

We print the tag even if the value is empty -- essentially a waste of bytes over the wire.

## To Reproduce

1. Browse to `/wp-admin/options-general.php?page=parsely`
1. Set the "Metadata Format" to `repeated_metas`
1. Browse to the front end
1. View page source
1. See various `<meta name="parsely-...` tags with empty values

## Expected behaviour

Only populated meta values to be printed. Empty values should omit the tag altogether.

## Screenshots

Screen Shot 2021-09-06 at 11 11 51 AM

jblz:
## Description

If configured to print repeating metas, any keys that don't exist in the hash map throw PHP notices. This change ensures that values are set for a given index prior to referencing them so as to not trigger the notices.

Further, this prevents printing meta tags that are unpopulated.
fixes #365

## Motivation and Context

If debugging is enabled, these are pretty noisy and not really actionable.

### Verify the issue

* Run the `develop` branch
* Browse to Settings | Parse.ly
* Set `Metadata Format` to `repeated_metas`
* Set `define( 'WP_DEBUG', true );` in your wp-config.php
* Browse to the front-end of the site on a route that does not have all the printed meta values -- a sure way to do this is to browse to a blog post listing archive page.
    * Confirm you see `Undefined Index` PHP notice(s)
* View the page source
    * Confirm you see `<meta name="parsely-...` HTML tags with empty values

### Verify the fix

* Switch to this branch
* Reload the page
    * No notices should be printed
* View the page source
    * Confirm you see no `<meta name="parsely-...` HTML tags with empty values
    * Confirm you see `<meta name="parsely-...` HTML tags with populated values for all those that exists in the meta data

Screen Shot 2021-09-10 at 10 02 54 PM

## Types of changes

Bug fix

### DIFF
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -894,7 +894,7 @@ class Parsely {
 			$parsely_metas = array_filter( $parsely_metas, array( $this, 'filter_empty_and_not_string_from_array' ) );
 
 			if ( isset( $parsely_page['author'] ) ) {
-				$parsely_page_authors = array_map( array( $this, 'map_extract_author_name' ), $parsely_page['author'] );
+				$parsely_page_authors = wp_list_pluck( $parsely_page['author'], 'name' );
 				$parsely_page_authors = array_filter( $parsely_page_authors, array( $this, 'filter_empty_and_not_string_from_array' ) );
 			}
 
@@ -909,16 +909,6 @@ class Parsely {
 		echo '<!-- END Parse.ly -->' . "\n\n";
 
 		return $parsely_page;
-	}
-
-	/**
-	 * Getter function to be used with `array_filter` to fetch the `name` property
-	 *
-	 * @param mixed $var Value to filter from the array.
-	 * @return mixed Value of the name property
-	 */
-	private static function map_extract_author_name( $var ) {
-		return $var['name'];
 	}
 
 	/**
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -909,7 +909,7 @@ class Parsely {
 	/**
 	 * Function to be used in `array_filter` to clean up repeated metas
 	 *
-	 * @param $var mixed Value to filter from the array
+	 * @param mixed $var Value to filter from the array.
 	 * @return bool Returns true if the variable is not empty, and it's a string
 	 */
 	private static function filter_empty_and_not_string_from_array( $var ) {
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -877,6 +877,16 @@ class Parsely {
 				$parsely_page['keywords'] = implode( ',', $parsely_page['keywords'] );
 			}
 
+			$metas = array(
+				"title" => isset( $parsely_page['headline'] ) ? $parsely_page['headline'] : null,
+				"link" => isset( $parsely_page['url'] ) ? $parsely_page['url'] : null,
+				"type" => $parsely_post_type,
+				"image-url" => isset( $parsely_page['thumbnailUrl'] ) ? $parsely_page['thumbnailUrl'] : null,
+				"pub-date" => isset( $parsely_page['datePublished'] ) ? $parsely_page['datePublished'] : null,
+				"section" => isset( $parsely_page['articleSection'] ) ? $parsely_page['articleSection'] : null,
+				"tags" => isset( $parsely_page['keywords'] ) ? $parsely_page['keywords'] : null,
+			);
+
 			include PARSELY_PLUGIN_DIR . 'views/repeated-metas.php';
 		}
 
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -877,14 +877,14 @@ class Parsely {
 				$parsely_page['keywords'] = implode( ',', $parsely_page['keywords'] );
 			}
 
-			$metas = array(
-				"title" => isset( $parsely_page['headline'] ) ? $parsely_page['headline'] : null,
-				"link" => isset( $parsely_page['url'] ) ? $parsely_page['url'] : null,
-				"type" => $parsely_post_type,
-				"image-url" => isset( $parsely_page['thumbnailUrl'] ) ? $parsely_page['thumbnailUrl'] : null,
-				"pub-date" => isset( $parsely_page['datePublished'] ) ? $parsely_page['datePublished'] : null,
-				"section" => isset( $parsely_page['articleSection'] ) ? $parsely_page['articleSection'] : null,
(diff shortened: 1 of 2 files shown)


GaryJones (commented):
I wonder if it would be cleaner to create an array of meta field name => value (if not empty) in the `class-parsely.php`, and just have the view loop over that array and print the HTML?

src/class-parsely.php
```diff
@@ -871,11 +871,22 @@ public function insert_parsely_page() {
...
 		} else {
+			// Assume `meta_type` is `repeated_metas`.
 			$parsely_post_type = $this->convert_jsonld_to_parsely_type( $parsely_page['@type'] );
-			if ( is_array( $parsely_page['keywords'] ) ) {
+			if ( isset( $parsely_page['keywords'] ) && is_array( $parsely_page['keywords'] ) ) {
 				$parsely_page['keywords'] = implode( ',', $parsely_page['keywords'] );
 			}
 
+			$parsely_metas = array(
+				'title'     => isset( $parsely_page['headline'] ) ? $parsely_page['headline'] : null,
```
pauarge, line 885: I've used this syntax instead of `??` because I don't want to break compatibility with PHP 5.x.

src/class-parsely.php
```diff
@@ -875,11 +875,22 @@ public function insert_parsely_page() {
...
+				'link'      => isset( $parsely_page['url'] ) ? $parsely_page['url'] : null,
+				'type'      => $parsely_post_type,
+				'image-url' => isset( $parsely_page['thumbnailUrl'] ) ? $parsely_page['thumbnailUrl'] : null,
+				'pub-date'  => isset( $parsely_page['datePublished'] ) ? $parsely_page['datePublished'] : null,
+				'section'   => isset( $parsely_page['articleSection'] ) ? $parsely_page['articleSection'] : null,
+				'tags'      => isset( $parsely_page['keywords'] ) ? $parsely_page['keywords'] : null,
+			);
+
```
GaryJones, line 900: I would suggest that calling `array_filter()` here ([example](https://github.com/Parsely/wp-parsely/issues/365)), to remove keys that have empty values, would be cleaner than the empty + is_string conditions in the view. If the logic is "Don't show metas with empty values" (per https://github.com/Parsely/wp-parsely/issues/365) then let's keep that logic here, instead of in the view.

views/repeated-metas.php
```diff
@@ -8,15 +8,19 @@
...
-<meta name="parsely-link" content="<?php echo esc_attr( $parsely_page['url'] ); ?>" />
-<meta name="parsely-type" content="<?php echo esc_attr( $parsely_post_type ); ?>" />
-<meta name="parsely-image-url" content="<?php echo esc_attr( $parsely_page['thumbnailUrl'] ); ?>" />
-<meta name="parsely-pub-date" content="<?php echo esc_attr( $parsely_page['datePublished'] ); ?>" />
-<meta name="parsely-section" content="<?php echo esc_attr( $parsely_page['articleSection'] ); ?>" />
-<meta name="parsely-tags" content="<?php echo esc_attr( $parsely_page['keywords'] ); ?>" />
-<?php foreach ( (array) $parsely_page['author'] as $parsely_author ) { ?>
+foreach ( $parsely_metas as $parsely_meta_key => $parsely_meta_val ) {
+	if ( ! empty( $parsely_meta_val ) && is_string( $parsely_meta_val ) ) {
+		echo '<meta name="parsely-' . esc_attr( $parsely_meta_key ) . '" content="' . esc_attr( $parsely_meta_val ) . '" />' . "\n";
```
GaryJones: Rather than a concatenated string of 6 pieces, it's probably easier to see the format of the HTML being added, and that the values are correctly escaped, if something like the following is used (untested):

```suggestion
		printf(
			'<meta name="%s" content="%s" />',
			esc_attr( 'parsely-' . $parsely_meta_key ),
			esc_attr( $parsely_meta_val )
		) . "\n";
```

GaryJones (changes requested):

views/repeated-metas.php
```diff
@@ -8,15 +8,22 @@
...
+		'<meta name="%s" content="%s" />%s',
+		esc_attr( 'parsely-' . $parsely_meta_key ),
+		esc_attr( $parsely_meta_val ),
+		PHP_EOL
+	);
+}
+
+if ( isset( $parsely_page['author'] ) ) {
+	foreach ( (array) $parsely_page['author'] as $parsely_author ) {
+		if ( empty( $parsely_author['name'] ) || ! is_string( $parsely_author['name'] ) ) {
```
GaryJones: Can the conditional logic in this line be moved out of the view, so that the data we provide to the view is already usable?

src/class-parsely.php
```diff
@@ -875,11 +875,29 @@ public function insert_parsely_page() {
...
+				'image-url' => isset( $parsely_page['thumbnailUrl'] ) ? $parsely_page['thumbnailUrl'] : null,
+				'pub-date'  => isset( $parsely_page['datePublished'] ) ? $parsely_page['datePublished'] : null,
+				'section'   => isset( $parsely_page['articleSection'] ) ? $parsely_page['articleSection'] : null,
+				'tags'      => isset( $parsely_page['keywords'] ) ? $parsely_page['keywords'] : null,
+				'author'    => isset( $parsely_page['author'] ),
+			);
+			$parsely_metas = array_filter( $parsely_metas, array( $this, 'filter_empty_and_not_string_from_array' ) );
+
+			if ( isset( $parsely_page['author'] ) ) {
+				$parsely_page_authors = array_map( array( $this, 'map_extract_author_name' ), $parsely_page['author'] );
```
GaryJones: Instead of an array_map with a custom callback, can `wp_list_pluck()` be used to pull out the `name` instead?

GaryJones (approved):
Ideally we would have some unit tests for repeated metas in https://github.com/Parsely/wp-parsely/tree/develop/tests/StructuredData so that we can check that the logic of not showing empty fields is correct, but there aren't any right now.

pauarge:
@GaryJones Will add tests in a separate PR

PR #373, opened 2021-09-11
Fix cookie import

jblz:
## Description

When checking out #366, I saw a warning:

```
WARNING in ./src/js/lib/personalization.js 7:9-18
export 'get' (imported as 'getCookie') was not found in 'js-cookie' (possible exports: default)
 @ ./src/js/widgets/recommended/index.js 15:0-69 27:28-52
```

...even though our tests still passed... 😕 

This syntax makes the warning go away.

## Motivation and Context

I'm not sure why, but it seems like webpack 5 (or the WP wrapper) didn't work w/ the existing style import. The "new" way follows the [directions on the tin](https://www.npmjs.com/package/js-cookie), so this looks like the way to go.

## How Has This Been Tested?

See #364

### Automated testing

`npm run test`

### Manual testing

* Run this branch on a site with API access
* Configure the widget to display (in a sidebar, footer, etc.)
* Browse to a page with the widget on it
    * Confirm that personalized results are requested / received
    * Namely, you should see a populated `uuid` query argument in the call to `https://api.parsely.com/v2/related` as per [this line](https://github.com/Parsely/wp-parsely/blob/f465d3db4536bddb22b5ffb9388e405a87e240be/src/js/widgets/recommended/index.js#L31).

## Types of changes


GaryJones (approved):
How can we improve our unit tests so that the old way would fail?

Or is this just a warning but the import code still worked anyway?

pauarge:
@GaryJones Been taking a look at your suggestion. `js.cookie.mjs` exports by default a function (`api`) that creates an object with the `get` property. That is why the code would work fine but trigger the warning. By importing it the way Jeff does in this PR, we are importing the default export hence no warning.

I'm haven't found a solution for improving the unit tests, so I guess we're good by merging the code as is.

PR #374, opened 2021-09-12
Add a step to the node CI test to confirm built scripts are included in the change

jblz:
## Description

After CI runs `npm run build`, flag the branch a failure if any uncommitted changes are detected.

## Motivation and Context

Currently, there's no enforcement that built JS files are included when the source changes (or if build tooling alters the build). This seeks to provide some protection that build files are kept up to date with those changes.

## How Has This Been Tested?

See the commit history & screen shots below.

|Screen Shot 2021-09-12 at 2 13 04 PM|
|:--:|
|Fail when files are missing in the build dir after build|

|Screen Shot 2021-09-12 at 2 16 40 PM|
|:--:|
|Fail when changes to src are not present in build|

|Screen Shot 2021-09-12 at 2 18 50 PM|
|:--:|
|Success when no uncommitted changes are present|

## Types of changes

Dev Tooling


GaryJones:
The screenshots indicate that it's trying to protect against two cases, but the error message only points towards one?

.github/workflows/node.js.yml
```diff
@@ -53,5 +53,16 @@ jobs:
...
 
+      - name: Confirm the build is up-to-date
+        run: |
+          REPOSTATUS=`git status --porcelain 2>/dev/null`;
+          if [[ -n "$REPOSTATUS" ]]; then
+            echo "Repo is not clean. Did you forget to commit changes to the /build dir?";
+            echo "$REPOSTATUS";
+            exit 1;
+          else
+            echo The repo file system is clean.;
```
GaryJones: While true, it isn't indicative of what was actually being checked - that built files are present or not. Right now, those unfamiliar may not be able to join the dots between whether the file system is clean (is that good or bad?) and the real potential problem.

pauarge: I've added a clarification message regarding this.

pauarge (approved):

PR #378, opened 2021-09-16
Release 2.5.2

pauarge:
## Description

Minor bug fix release that includes (as per the Changelog):

### Changed

- Specify that browserslist should use the defaults setting when building bundles. #363
- Wrapping post list links in an opt-in filter. #369

### Fixed

- Fix notices that would appear if the plugin was set up to print repeating metas but those wouldn't exist. #370
- Fix cookie parsing. In some edge cases, a cookie that contained special characters would not be parsed correctly. #364

## Motivation and Context

Pull in a couple of recent fixes so customers can benefit.

## How Has This Been Tested?

All PRs have been reviewed and tested individually. The bundle has been tested locally as well. 

## Types of changes


CHANGELOG.md
```diff
@@ -5,6 +5,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.5.2] - 2021-09-17
```
GaryJones, line 8: Missing a line at the bottom of the file with the compare link.

phpunit.xml.dist
```diff
@@ -16,6 +16,7 @@
 	</php>
 	<testsuites>
 		<testsuite name="WP_Tests">
+			<directory>./tests/</directory>
```
GaryJones, line 19: This seems like an odd entry, considering the one below it as well.

GaryJones (changes requested):
Just need the changelog fixing.

phpunit.xml.dist
GaryJones: The [default value of `suffix`](https://github.com/sebastianbergmann/phpunit/blob/master/phpunit.xsd#L134) is `Test.php`, so the new line is definitely needed.

The actual issue would be the lack of removal of the line below (with `suffix="-test.php"`).

I think I left it in because the `recommended-api-test.php` test file never got renamed into `*Test.php` because Jeff was still working on it via another big PR, so I didn’t want to break that.

Not a blocker for 2.5.2 release, but a good reminder to see if the `recommended-api-test.php`-related PR (https://github.com/Parsely/wp-parsely/pull/318?) got merged, then the second `directory` element can be removed from the config.

GaryJones (approved):

phpunit.xml.dist
GaryJones: Opened #379 to address.

PR #383, opened 2021-09-17
Tests: Improve the get_current_url data provider

GaryJones:
## Description

Data providers actually provide an iterator rather than just an array, so we can build up the data using `yield` instead.

As per https://peakd.com/hive-168588/@crell/fun-with-phpunit-data-providers (hat tip @crell) this has several advantages:

1. It's easier to read, making it better for future you to understand the code.
1. It's less indented, which gives you more room for longer lines for each value.
1. Because each test set is syntactically independent, it's easier to add new ones, reorder them, or temporarily comment out a single test set.
1. If a given test set needs more setup than just a static list, it's simple to put that setup before the `yield` statement for which it is relevant.

As per the follow-up article https://peakd.com/hive-168588/@crell/advanced-phpunit-shenanigans, I've also included a `@testdox` annotation on the test method. When this contains at least one method parameter, it completely overrides the sentence that appears when the `--testdox` flag is called when running PHPUnit.

## Motivation and Context

Here's what the tests looked like with `--testdox` before the change (not descriptive):

Screenshot 2021-09-17 at 10 06 29
(The names of some of the other test methods could be improved to be more descriptive about what is actually being tested, but that's outside the scope of this PR.)

----

Here's the view when the `yield` keys (same as what associative arrays would look like) only are used (more descriptive, but clunky):
Screenshot 2021-09-17 at 10 12 52

----

And here's what they look like with the complete change, including `@testdox` with parameter names:

Screenshot 2021-09-17 at 10 52 48

## How Has This Been Tested?
Run the test suite before and after - all green, no change.

## Types of changes

- [x] Maintenance

### DIFF
--- a/tests/GetCurrentUrlTest.php
+++ b/tests/GetCurrentUrlTest.php
@@ -16,78 +16,87 @@ final class GetCurrentUrlTest extends TestCase {
 	/**
 	 * Data provider for test_get_current_url
 	 *
-	 * @return array[]
+	 * @return iterable
 	 */
 	public function data_for_test_get_current_url() {
-		return array(
-			// Start cases with 'force_https_canonicals' = true.
-			array(
-				true,
-				'http://example.com',
-				'https://example.com',
-			),
-			array(
-				true,
-				'https://example.com',
-				'https://example.com',
-			),
-			array(
-				true,
-				'http://example.com:1234',
-				'https://example.com:1234',
-			),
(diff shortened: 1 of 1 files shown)


pauarge (approved):
Great improvement!

Crell:
Awesome!  I love seeing that people actually read my posts. :smile:

Issue #325, opened 2021-05-18
Drop Support for IE11 Alongside WordPress 5.8


jblz:
WordPress core is in the process of [dropping support for Internet Explorer 11](https://make.wordpress.org/core/2021/04/22/ie-11-support-phase-out-plan/) and steps to do so have already begun [making their way into the front-end build tooling](https://github.com/WordPress/gutenberg/pull/31110).

Our trusty dependabot has opened the following PRs against our dev dependences to that end:
#319
#320
#322

Do we want to go forward with this change for 2.6?

GaryJones:
WP 5.8 is due at ~the end of the year~ July 20 - Parse.ly 2.6 would be out before then, so maybe aim for Parse.ly 2.7 or 2.8?

jblz:
Marked those PRs as "ignore this major version"

We'll reopen / revisit once WP 5.8 ships.

PR #385, opened 2021-09-20
Remove jQuery from Recommended Widget

pauarge:
## Description

This is the last PR on the workstream to remove jQuery dependencies on the plugin. With this work, we've replaced all jQuery-implemented elements with vanilla JS.

In this case, we're removing jQuery from the Recommended Widget. It's one single Javascript file plus the enqueue of jQuery itself on the PHP side. We have rebuilt the assets with

```
npm run build
```

## Motivation and Context

https://github.com/Parsely/wp-parsely/issues/134

## How Has This Been Tested?

This has been tested by generating the new assets on a local environment plus a sandbox and visually checking that the contents of the recommended widget are the same as with the old implementation in different browsers.

### Known issues

Neither the new or the old implementation work in Firefox (they do in Safari and Chrome). This will be addressed in a separate issue.

Screen Shot 2021-09-20 at 11 40 05 AM

## Types of changes

Bug fix (non-breaking change which fixes an issue


src/js/widgets/recommended/index.js
```diff
@@ -26,66 +25,83 @@ function widgetLoad( {
...
-		jQuery.each( data.data, function( key, value ) {
-			const widgetEntry = jQuery( '<li>' )
-				.addClass( 'parsely-recommended-widget-entry' )
-				.attr( 'id', 'parsely-recommended-widget-item' + key );
+	const outerList = document.createElement( 'ul' );
+	outerList.className = 'parsely-recommended-widget';
+	outerDiv.appendChild( outerList );
 
-			const textDiv = jQuery( '<div>' ).addClass( 'parsely-text-wrapper' );
+	fetch( fullUrl )
```
GaryJones, line 53: The [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) doesn't support IE11, but our [`defaults` for browserslist](https://github.com/Parsely/wp-parsely/blob/fcc8b363b0ea20a840f1ec7c52cd12aa1688f819/package.json#L32-L34) includes IE11 (run `npx browserslist` in the plugin directory to get the full list).

Since the plugin still supports WP 4.0, then we either need to make a call and document about the widgets not supporting IE11, or not use / provide a workaround for the Fetch API.

GaryJones (commented):
While the proposal looks good, we'll be breaking BC for browser support on the front-end, and that needs some discussion with the Parse.ly folks as that's a business decision.

src/js/widgets/recommended/index.js
pauarge: If we want to provide a workaround, we could use a Polyfill: https://github.com/github/fetch

However, I'm happy to stop supporting IE11. We can clearly state it in the widget configuration.

GaryJones: Internal stats suggest that IE11 requests account for around 0.1% of requests.

We should also check the behaviour of what the page looks like (is it empty? Does it have a heading but no content? I think there's an issue for this already), but in terms of using `fetch()`, ✅.

GaryJones (approved):

src/js/widgets/recommended/index.js
pauarge: Actually, there is an open issue about dropping IE 11 in the upcoming release: https://github.com/Parsely/wp-parsely/issues/325

GaryJones: > We should also check the behaviour of what the page looks like (is it empty? Does it have a heading but no content? I think there's an issue for this already)

https://github.com/Parsely/wp-parsely/pull/193

Issue #326, opened 2021-05-27
Multisite: Remove error message from Network Admin


mikeyarce:
## Describe the bug
When you activate the Parse.ly plugin on a multisite, this admin notice on Network Admin:
`The Parse.ly plugin is not active. You need to provide your Parse.ly Dash Site ID before things get cooking.`

<img width="1266" alt="Screen Shot 2021-05-27 at 10 52 54 AM" src="https://user-images.githubusercontent.com/3220162/119873761-b80d6b80-bed9-11eb-9f1e-4c760f7315fe.png">

Clicking on "provide your Parse.ly Dash Site ID" takes you to the main root site for the multisite.  This could lead to thinking that if you set it up there maybe that's all you need to do.

I think a better notice would be to prompt the user to set Parse.ly up on every subsite properly OR not do a network activation but only activate on a per-site basis.  

If you Network Activate, every site will receive that message until the Parse.ly settings are updated.

## To Reproduce
1. Install the Parse.ly plugin on a Multisite
2. Network Activate
3. You'll see the admin notice

PR #392, opened 2021-09-21
Hide admin warning on network admin

pauarge:
## Description

The plugin displays a warning message on `wp-admin` if it's activated but it's not active (no Parse.ly API key has been provided). This message would appear in the network admin of a multisite WordPress installation. If a user follows the provided link, they would go to the network's root site and only activate the plugin for that site, not the whole network.

To avoid confusion, we are now not displaying the warning in the network admin (i.e. `/wp-admin/network/`) and only doing so in root `wp-admin`.

## Motivation and Context

https://github.com/Parsely/wp-parsely/issues/326

## How Has This Been Tested?

Tested that it hides the message in a multisite installation and it does nothing on a non-multisite installation.

**Before**

Screen Shot 2021-09-21 at 4 09 49 PM

**After**

Screen Shot 2021-09-21 at 4 10 17 PM

## Types of changes


src/class-parsely.php
```diff
@@ -26,7 +26,7 @@ class Parsely {
 	const CAPABILITY  = 'manage_options';      // The capability required for the user to administer settings.
 
 	/**
-	 * Declare some class propeties
+	 * Declare some class properties
```
pauarge, line 29: This is unrelated with the PR, but I found it while working on it and it's not worth a separate PR.

GaryJones (approved):

tests/OtherTest.php
```diff
@@ -582,4 +582,63 @@ function() {
...
+
+	/**
+	 * Test that test_display_admin_warning action returns a warning when there is no key
+	 *
+	 * @covers \Parsely::display_admin_warning
+	 */
+	public function test_display_admin_warning_without_key() {
+		$options           = get_option( \Parsely::OPTIONS_KEY );
+		$options['apikey'] = '';
+		$this->set_options( $options );
```
GaryJones:
```suggestion
		$this->set_options( array( 'apikey' => '' ) );
```

tests/OtherTest.php
```diff
@@ -582,4 +582,63 @@ function() {
...
+
+		ob_start();
+		self::$parsely->display_admin_warning();
+		$output = trim( ob_get_contents() );
+		ob_end_clean();
```
GaryJones: If we're having to capture output to see if it contained a string or not, then that's a code smell that we really should be testing against a boolean function that handles the logic of whether it should display or not i.e. a new `should_display_admin_warning()`. 

Then we can have a unit test to test the various conditions of that, and an optional second one that tests the contents of the message (that's probably redundant in this case though as you'd be testing a hard-coded string against a hard-coded expected string).

tests/OtherTest.php
GaryJones: Similar for the next test.

GaryJones (changes requested):
Thanks for adding the tests - in this case, it's highlighted an opportunity to refine the code, to separate out the decision about whether to show the notice, and the echoing of the notice itself.

src/class-parsely.php
```diff
@@ -818,6 +810,27 @@ public function display_admin_warning() {
...
+	 *
+	 * @return bool True if the admin warning should be displayed
+	 */
+	private function should_display_admin_warning() {
+		if ( is_network_admin() ) {
+			return false;
+		}
+
+		$options = $this->get_options();
+		if ( isset( $options['apikey'] ) && ! empty( $options['apikey'] ) ) {
```
GaryJones: I think the `isset()` here can be dropped, as `empty()` already checks for that.

Then we're left with a negative condition that could be reversed and `return true;` instead, or simplify the whole condition down to `return empty( $options['apikey'] );`

I think the "get options and check if apikey is not empty populated" could be extracted to its own helper method, so `return ! $this->has_saved_api_key();` or `return $this->api_key_not_set();` or similar.

pauarge: I like the idea of simplification, but I think making a method just for `has_saved_api_key` might be a bit too much for the scope of this PR. I'm committing the first change you are suggesting.

GaryJones (approved):

src/class-parsely.php
GaryJones: Opened https://github.com/Parsely/wp-parsely/issues/395 as checking for a saved API key is indeed a bigger issue.

PR #393, opened 2021-09-22
Add telemetry backend

pauarge:
## Description

This code was initially written by @jblz. I've simply taken it and done some minor tweaks.

We are introducing telemetry into the Parse.ly plugin to better understand how the plugin is used and how it can be improved. This functionality is opt-in via a filter (it is disabled by default).

This PR adds a telemetry framework that connects with Automattic's Tracks system. It is prepared to be extended with other telemetry back-ends. It currently tracks a small set of events happening in `wp-admin` with respect to the Parse.ly plugin, but we expect this events to grow in the future.

## Motivation and Context

We want to gather insights on how the plugin is being used so we can better assess the next features we'll be building.

## How Has This Been Tested?

We will be adding unit tests in a subsequent pull request.

## Types of changes


src/telemetry/class-parsely-a8c-tracks-event.php
```diff
@@ -0,0 +1,135 @@
...
+ * @package Parsely
+ * @since 3.0.0
+ */
+
+declare(strict_types=1);
+
+/**
+ * Instances of this class are fit for recording to the Automattic Tracks system (unless an error occurs during instantiation).
+ */
+class Parsely_A8c_Tracks_Event {
```
GaryJones: Let's use a `Parsely\Telemetry` namespace with an unprefixed class name like `A8c_Tracks_Event` (and similar), for these new classes.

src/telemetry/class-parsely-telemetry.php
```diff
@@ -0,0 +1,160 @@
...
+	 *
+	 * @var Parsely_A8c_Tracks
+	 */
+	private $tracks;
+
+	/**
+	 * Parsely_Telemetry constructor.
+	 */
+	public function __construct() {
+		$this->tracks = new Parsely_A8c_Tracks();
```
GaryJones: This highly couples the telemetry system to using the Tracks implementation - that means other systems (including mocks for tests) can't be used.

Use constructor dependency injection, type hinted at an interface, here instead.

Roughly (but you can think of better names):

```php
interface Telemetry_System {
	public function setup();
	public function record_event();
}
```

```php
class A8c_Tracks implements Telemetry_System {
	...
}
```

```php
class Telemetry {
	private $telemetry_system;

	public function __construct( Telemetry_System $telemetry_system ) {
		$this->telemetry_system = $telemetry_system;
	}
	...
}
```

```php
// At the point of instantiating Telemetry(), pass in an instance of the desired telemetry system (which at the moment, is only A8c_Tracks):
if ( apply_filters( 'wp_parsely_enable_telemetry_backend', false ) ) {
	$telemetry_system = apply_filters( 'wp_parsely_telemetry_system', new A8c_Tracks() );
	$this->telemetry = new \Parsely\Telemetry\Telemetry( $telemetry_system );
}
```

Now, folks have the ability to use a different system that extend our interface, and we can insert mocks to test the `Telemetry` class without assuming that the implementation is the 
A8c_Tracks.

src/telemetry/class-parsely-telemetry.php
```diff
@@ -0,0 +1,160 @@
...
+
+	/**
+	 * Hook functions into WordPress actions and / filters for which we want to record events.
+	 *
+	 * @todo Move this to a dedicated file or some other organization scheme.
+	 * @todo Use named functions so they can be (conceivably) unhooked if desired.
+	 *
+	 * @return void
+	 */
+	protected function add_event_tracking(): void {
```
GaryJones: It feels like this list of events should be held outside of this class, and registered to the class with a `register_event()` or `add_event()` method.

That would give more flexibility in being able to add extra events for testing, and allow others to add new events to be tracked, and better follows the Single Responsibility Principle.


Simplified: 
```php
// Create new telemetry dispatcher with an instance of the system to use:
$telemetry = new Telemetry( new A8c_Tracks() );

// Now register action and filter hooks and their callbacks.
$telemetry->register_event(
	new Telemetry_Event_Action (
		'my_action_hook',
		'my_callable'
	)
);
$telemetry->register_event(
	new Telemetry_Event_Filter (
		'my_filter_hook',
		'my_callable',
		'my_priority',
		'my_args_count'
	)
);
```

A `$events` property inside Telemetry now contains an array/collection of Telemetry_Event_* data objects.

GaryJones (changes requested):
I think some different architecture here will be beneficial. Think about if we had to use a different telemetry system (or consider a `NullTelemetry` class, which would allow events to be registered and recorded, even if it wasn't actually sent anywhere) - what changes would need to be made to the other classes to support this?

pauarge:
Closing this PR because we are implementing the tracking on the VIP side. https://github.com/Automattic/vip-go-mu-plugins/pull/2500

Issue #206, opened 2021-04-11
Widget: Duplicate API calls made when multiple instances of the widget are present


GaryJones:
Related: #158.

As the JavaScript for the widget is loaded in the `widget()` method output, then there is an API call for each instance of the widget that is on the page i.e. two Parse.ly Recommended widgets means two API calls.

If the settings in each widget instance are different (maybe one personalized, maybe one not, or maybe sorted or boosted by different values), then it makes sense to do a request per widget.

When it's the same settings though, then the same results should be returned for all calls, which makes the duplicate requests redundant.

mikeyarce:
I think for this, one approach might be to separate out the JS from the Widget PHP, and then use `wp_localize_script` to send URLs that need to be fetched - then with a list of URLs that need to be fetched we could de-dupe requests there.
Related: https://github.com/Parsely/wp-parsely/issues/158, https://github.com/Parsely/wp-parsely/issues/134

PR #397, opened 2021-09-23
Fix rendering errors when rendering multiple recommendation widgets

pauarge:
## Description

If multiple recommendation widgets were set up in a page, only the last one would render. That was caused because the script responsible of rendering the widget was enqueued multiple times, but it only the last enqueue would survive.

This PR changes the logic slightly: the script is still enqueued multiple times, but we take advantage of the fact that it's actually called only once. To render the appropriate widgets, we render an empty div from the PHP side and pass the data to javascript using `data` attributes. Then, the script reads all the `div` with that attributes and renders the actual widget.

## Motivation and Context

Not exactly this issue, but related: https://github.com/Parsely/wp-parsely/issues/206

## How Has This Been Tested?

Rendered multiple widgets with the same and different configurations in the page.

**Before**

Screen Shot 2021-09-23 at 3 45 27 PM

**After**

Screen Shot 2021-09-23 at 3 45 53 PM

## Types of changes


wp-parsely.css
```diff
@@ -15,7 +15,6 @@
 .parsely-recommended-widget-entry img {
 	float: left;
 	margin-right: 15px;
-	max-width: 50%;
```
GaryJones, line 18: How does this change fit in?

GaryJones (commented):
I can see how different widget configs would not conflict with each other. ✅ 

What happens when the JS it blocked / not available / fails? Does the div that is now hard-coded in the HTML break any layouts with the default styles?

wp-parsely.css
pauarge: Styles weren't applied correctly in the past and this would result in the image looking too small with the changes in this PR.

pauarge:
@GaryJones If Javascript is disabled, the contents don't load. The only thing that gets rendered is an empty div that doesn't affect the rest of the page.

src/class-parsely-recommended-widget.php
```diff
@@ -115,6 +115,20 @@ public function widget( $args, $instance ) {
 
 		$reco_widget_script_asset = require PARSELY_PLUGIN_DIR . 'build/admin-page.asset.php';
 
+		?>
+
+		<div class="parsely-recommended-widget"
+			data-parsely-widget-display-author="<?php echo esc_attr( isset( $instance['display_author'] ) ? wp_json_encode( boolval( $instance['display_author'] ) ) : false ); ?>"
+			data-parsely-widget-display-direction="<?php echo esc_attr( isset( $instance['display_direction'] ) ? $instance['display_direction'] : null ); ?>"
```
GaryJones: Instead of adding `null` string to the attribute, could an empty string be used instead? Just seems cleaner.

GaryJones (approved):
✅

Issue #151, opened 2021-03-07
Incorrect metadata URL when using Plain permalinks structure


GaryJones:
`Parsely::get_current_url()` strips querystrings, which means it doesn't produce the correct URL when pretty permalinks are not used.

All URLs are the home URL (i.e. `https://example.com` instead of `https://example.com/?p=1` or `https://example.com/?page_id=2` etc.).

hbbtstar:
I suspect this was actually by design, once upon a time: Parsely's backend strips querystring arguments completely, so this implementation followed suit. Given Wordpress' use of the p querystring, I think in addition to fixing this we also have to ask Parsely to explicitly allow the p querystring in its backend (since WP uses it)

jblz:
@hbbtstar Any new thoughts here? Is this something we want to / can support on the service-side?

jblz:
Should we bump this to a later milestone?

hbbtstar:
ah, sorry- I forgot to update this. Yeah, we can make this change: it'll definitely be possible to support these without too much of a lift.

hbbtstar:
oh I missed the second comment: this change won't happen in the Parsely backend in the near term, so it's definitely a candidate for bumping imo!

abelsonlive:
Is there someway within the context of `wp-parsely` for the site admin to indicate that they don't use pretty permalinks? If that information could be passed somehow to the backend upon installation, it would allow us to maintain the `p` or `page_id` parameters without causing issues with other, non-wordpress sites.

pauarge:
Yes, we can tell that from WordPress. However, we can assume that all WordPress sites have it enabled. Even though a site uses _pretty_ permalinks, `?p=` will still be supported.

PR #398, opened 2021-09-24
Improving reliability of get_current_url tests

pauarge:
## Description

This PR improves the tests on the `get_current_url` function. Concretely, we're changing how we check the results, from `assertStringStartsWith` to the stricter `assertEquals`.

With this PR, we have tests to prove that if a query parameter, such as `?p=1` gets added to the URL, the function will correctly catch it. This closes an issue (linked below) that pointed to this function not working correctly.

## Motivation and Context

https://github.com/Parsely/wp-parsely/issues/151

## How Has This Been Tested?

Ran PHPUnit tests locally and in CI.

## Types of changes

### DIFF
--- a/tests/GetCurrentUrlTest.php
+++ b/tests/GetCurrentUrlTest.php
@@ -22,74 +22,74 @@ final class GetCurrentUrlTest extends TestCase {
 		yield 'Home is http with force HTTPS true' => array(
 			'force_https' => true,
 			'home'        => 'http://example.com',
-			'expected'    => 'https://example.com',
+			'expected'    => 'https://example.com/',
 		);
 
 		yield 'Home is https with force HTTPS true' => array(
 			'force_https' => true,
 			'home'        => 'https://example.com',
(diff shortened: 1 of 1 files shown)


GaryJones:
The focus of these tests is on the parsing issues related to the domain, port, and potential path, and this is represented in the data provider.

However, it's good to be able to confirm that going to the homepage, random URL and a post via its ID, also work well for each of the combinations in the first parts of the URL.

Since the setup is the same for each of the tests, we can just merge these three tests into one, and have it test the homepage, a specific post, and a random URL, and add the message attribute to the assertions to make it clear which part might have failed.

If you're happy with my changes @pauarge, then I can rebase into a single commit or it can be merged with a squash, to keep the history cleaner.

GaryJones (approved):

pauarge:
@GaryJones Your changes look good to me, they make the tests easier to read. It's a 👍 from my side.

Issue #402, opened 2021-09-25
e2e tests give errors when run locally.


GaryJones:
## Describe the bug

When running the e2e tests locally, per the [instructions](https://github.com/Parsely/wp-parsely/blob/develop/tests/e2e/README.md), then I get a constant error for each of them. When the tests are run in the CI, then it passes correctly ([example](https://github.com/Parsely/wp-parsely/runs/3708536507?check_suite_focus=true)).

## To Reproduce
```
$ npm run test:e2e

> test:e2e
> wp-scripts test-e2e

Chromium is already in /Users/gary/Sites/wp-parsely/app/public/wp-content/plugins/wp-parsely/node_modules/puppeteer-core/.local-chromium/mac-901912; skipping download.
 FAIL  tests/e2e/specs/plugin-action-link.spec.js
  ● Plugin action link › Should link to plugin settings page

    thrown: Response {
      "size": 0,
      "timeout": 0,
      Symbol(Body internals): Object {
        "body": PassThrough {
          "_events": [Object],
          "_eventsCount": 2,
          "_maxListeners": undefined,
          "_readableState": [ReadableState],
          "_writableState": [WritableState],
          "allowHalfOpen": true,
          Symbol(kCapture): false,
          Symbol(kTransformState): [Object],
        },
        "disturbed": false,
        "error": null,
      },
      Symbol(Response internals): Object {
        "counter": 0,
        "headers": Headers {
          Symbol(map): [Object],
        },
        "status": 400,
        "statusText": "Bad Request",
        "url": "http://localhost:8889/wp-admin/admin-ajax.php?action=rest-nonce",
      },
    }

      at _getError (node_modules/jest-circus/build/utils.js:442:18)
          at Array.map (<anonymous>)

 FAIL  tests/e2e/specs/activation-flow.spec.js
  ● Activation flow › Should progress as intended

    thrown: Response {
      "size": 0,
      "timeout": 0,
      Symbol(Body internals): Object {
        "body": PassThrough {
          "_events": [Object],
          "_eventsCount": 2,
          "_maxListeners": undefined,
          "_readableState": [ReadableState],
          "_writableState": [WritableState],
          "allowHalfOpen": true,
          Symbol(kCapture): false,
          Symbol(kTransformState): [Object],
        },
        "disturbed": false,
        "error": null,
      },
      Symbol(Response internals): Object {
        "counter": 0,
        "headers": Headers {
          Symbol(map): [Object],
        },
        "status": 400,
[...]

PR #401, opened 2021-09-25
Split out plugin actions links logic

GaryJones:
## Description
Moves the logic that adds the plugin action link to its own class.

Adds integration tests.

## Motivation and Context
The main `Parsely` god class is too big, so moving distinct bits of logic to their own class helps to cut this class down.

See #96.

## How Has This Been Tested?
Unit tests and visually. 

(e2e tests also written, but coming in a separate PR until I know it works in the CI)

## Screenshots (if appropriate):

Screenshot 2021-09-25 at 12 01 02
Screenshot 2021-09-25 at 23 28 10

PR #403, opened 2021-09-25
Add e2e test for the plugin action link

GaryJones:
## Description
Adds an e2e test for the plugin action link.

This based on `feature/extract-plugin-action-link` branch (and so contains the same file changes as it as well as the e2e test and readme file changes). Pulled against the feature branch, as the feature branch adds in a class that the e2e test uses to target an element. PR #401 should be merged first, and then this PR can be rebased on to `develop` as per usual, and only the two relevant changed files will be left.

The test logs in, activates the plugin, goes to the plugin page, looks for and clicks the Settings link in the Parse.ly plugin row, and then checks if the Settings page is loaded by checking for the Version string we have there.

I also updated the e2e test readme, since `docker-composer` is now `docker compose`.

## Motivation and Context
Ideally, all UI-related code should have an e2e test to check the functionality and behaviour is present.

I'll note that we may not consider the ability to click a Settings link on the Plugins page to be a critical user journey, but whilst we're still very light on e2e tests, I think it's worthwhile to get in, even if it's removed later on.

## How Has This Been Tested?
Ran e2e test locally (but had failures - see https://github.com/Parsely/wp-parsely/issues/402) but verified it was actually correct by watching the test interactively and seeing the pass here in the CI.

### DIFF
--- a/tests/e2e/README.md
+++ b/tests/e2e/README.md
@@ -16,12 +16,12 @@ See this post for more infomation: https://make.wordpress.org/core/2019/06/27/in
 
 - Provision the back end
 
-  - Make sure [`docker-compose` is installed](https://docs.docker.com/compose/install/).
-  - From the top-level directory, run:
+  - Make sure [`docker` is installed](https://docs.docker.com/engine/install/).
+  - From the `wp-content/plugins/wp-parsely` directory, run:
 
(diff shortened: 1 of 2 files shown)


pauarge (approved):
LGTM

PR #404, opened 2021-09-25
API Key: Add utility method for checking it is set

### LINKED ISSUE #395: Create utility method for checking for saved API key
There are multiple occurrences in the codebase where the behaviour changes depending on whether the API key setting has been populated or not. Each of these are directly coupled to retrieving options and checking if a value is empty or not falsy (and a string, etc.). It's inconsistent.

The logic would be cleaner if we had utility methods like `$this->api_key_is_empty()` and `$this->api_key_is_set()` that returned a boolean, so these are then more descriptive and abstracted away from the idea that options are from an array pulled from the database.

The alternative it to leave it like so (these are all examples in the current codebase):
```php
$options = get_option( \Parsely::OPTIONS_KEY );

if ( empty( $options['apikey'] ) ) { 
// or
if ( ! ( $options['apikey'] ) ) {
// or
if ( ! array_key_exists( 'apikey', $options ) || '' === $options['apikey'] ) {
// or
if ( ! (
	isset( $options['apikey'] ) &&
	is_string( $options['apikey'] ) &&
	strlen( $options['apikey'] ) > 0
) ) {
```

## Additional context
Unit tests can still use the `set_options()` test method to set the value of the saved option as needed, without having to mock the new methods (though that is also possible if needed).

GaryJones:
## Description
Add utility method for checking the API key is saved in the `parsely` options, and a second utility method for checking if it is missing.

Adds a third utility method for retrieving the API key.

Includes unit tests.

In the future, these methods may be moved to a Parsely_Options or similar class.

## Motivation and Context
These methods can be used by various other classes and methods to see if this key trigger of behaviour changes is available, and provides the key as needed.

Closes #395.

## How Has This Been Tested?
Integration tests pass, and work on #406 which utilises the new methods.

Screenshot 2021-09-25 at 22 29 04

Screenshot 2021-09-26 at 00 02 26

### DIFF
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -2085,4 +2085,17 @@ class Parsely {
 	public function api_key_is_missing() {
 		return ! $this->api_key_is_set();
 	}
+
+	/**
+	 * Get the API key if set.
+	 *
+	 * @since 2.6.0
+	 *
+	 * @return string API key if set, or empty string if not.
+	 */
+	public function get_api_key() {
+		$options = $this->get_options();
+
+		return $this->api_key_is_set() ? $options['apikey'] : '';
+	}
 }
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -2057,4 +2057,32 @@ class Parsely {
 	public function convert_jsonld_to_parsely_type( $type ) {
 		return in_array( $type, $this->supported_jsonld_post_types ) ? 'post' : 'index';
 	}
+
+	/**
+	 * Determine if an API key is saved in the options.
+	 *
+	 * @since 2.6.0
+	 *
+	 * @return bool True is API key is set, false if it is missing.
+	 */
+	public function api_key_is_set() {
+		$options = $this->get_options();
+
+		return (
+				isset( $options['apikey'] ) &&
+				is_string( $options['apikey'] ) &&
+				'' !== $options['apikey']
+		);
+	}
+
+	/**
+	 * Determine if an API key is not saved in the options.
+	 *
+	 * @since 2.6.0
+	 *
+	 * @return bool True if API key is missing, false if it is set.
+	 */
+	public function api_key_is_missing() {
+		return ! $this->api_key_is_set();
+	}
 }
--- a/tests/OtherTest.php
+++ b/tests/OtherTest.php
@@ -622,4 +622,22 @@ var wpParsely = {\"apikey\":\"blog.parsely.com\"};
 		$response = $should_display_admin_warning->invoke( self::$parsely );
 		self::assertFalse( $response );
 	}
+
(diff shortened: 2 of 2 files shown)


src/class-parsely.php
```diff
@@ -2057,4 +2057,45 @@ public function parsely_is_user_logged_in() {
...
+	 *
+	 * @return bool True is API key is set, false if it is missing.
+	 */
+	public function api_key_is_set() {
+		$options = $this->get_options();
+
+		return (
+				isset( $options['apikey'] ) &&
+				is_string( $options['apikey'] ) &&
+				'' !== $options['apikey']
```
pauarge, line 2056: Is there any reason not to use `empty()` here?

src/class-parsely.php
```diff
@@ -2057,4 +2057,45 @@ public function parsely_is_user_logged_in() {
...
+	}
+
+	/**
+	 * Determine if an API key is not saved in the options.
+	 *
+	 * @since 2.6.0
+	 *
+	 * @return bool True if API key is missing, false if it is set.
+	 */
+	public function api_key_is_missing() {
```
pauarge, line 2067: I'm not sure about this function. I think it's fine to call in our code `! $this->api_key_is_set()` and it's still very clear what it is doing.

pauarge (commented):
Changes look good. Should we find places where we're checking for the API key in our code and replace it with this functions in this PR or on a subsequent one?

src/class-parsely.php
GaryJones: What's the downside of including it?

It reads more cleanly in English ("if API key is missing", as opposed to "if not API is set".

GaryJones:
> Should we find places where we're checking for the API key in our code and replace it with this functions in this PR or on a subsequent one?

That can be a subsequent PR.

src/class-parsely.php
GaryJones: `empty()` is a function call, which is more expensive than a comparison as I understand it.

Comparing against `''` is also more descriptive - since `empty()` checks against multiple things - for instance, if `$options['apikey` was a literal string of `'0'`, then it's set, it's a string, and `empty( '0' )` would be true, but `'' !== '0'` would be correctly false.

See https://beberlei.de/2021/02/19/when_to_use_empty_in_php_i_say_never.html

src/class-parsely.php
pauarge: > What's the downside of including it?

I tend to like less functions and less lines of code unless strictly necessary, and in this case we're just negating the output of a function. 

> It reads more cleanly in English ("if API key is missing", as opposed to "if not API is set".

That is true, we can keep it.

pauarge (approved):

PR #405, opened 2021-09-25
Testcase: Allow getMethod() to use different class

GaryJones:
## Description
The existing `Testcase::getMethod()` method made the assumption that only non-public methods in the `Parsely` class would need to be made available for testing with Reflection.

This updates the test utility method to allow non-public methods in other source classes to be reflected as well.

## Motivation and Context
As we move more logic out of the `Parsely` class (#96), then we need to be able to use test utility methods with other source classes as well.

## How Has This Been Tested?
Tested with PR #406 for Row_Actions. All existing (and new tests in the new PR) pass correctly.

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -127,12 +127,13 @@ abstract class TestCase extends WPIntegrationTestCase {
 	 * Get a method from the Parsely class. This should be used when trying to access a private method for testing.
 	 *
 	 * @param string $method_name Name of the method to get.
+	 * @param string $class       Name of the class the method is in. Can be passed as Foo::class.
 	 *
 	 * @return \ReflectionMethod
 	 * @throws \ReflectionException The method does not exist in the class.
 	 */
-	public static function getMethod( $method_name ) {
-		$class  = new \ReflectionClass( 'Parsely' );
+	public static function getMethod( $method_name, $class = 'Parsely' ) {
+		$class  = new \ReflectionClass( $class );
 		$method = $class->getMethod( $method_name );
 		$method->setAccessible( true );
 		return $method;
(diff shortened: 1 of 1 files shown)


tests/TestCase.php
```diff
@@ -127,12 +127,13 @@ public function create_test_taxonomy( $taxonomy_key, $term_name ) {
...
 	 * @param string $method_name Name of the method to get.
+	 * @param string $class       Name of the class the method is in. Can be passed as Foo::class.
 	 *
 	 * @return \ReflectionMethod
 	 * @throws \ReflectionException The method does not exist in the class.
 	 */
-	public static function getMethod( $method_name ) {
-		$class  = new \ReflectionClass( 'Parsely' );
+	public static function getMethod( $method_name, $class = 'Parsely' ) {
+		$class  = new \ReflectionClass( $class );
```
pauarge:
```suggestion
	public static function getMethod( $method_name, $class_name = 'Parsely' ) {
		$class  = new \ReflectionClass( $class_name );
```

tests/TestCase.php
```diff
@@ -127,12 +127,13 @@ public function create_test_taxonomy( $taxonomy_key, $term_name ) {
 	 * Get a method from the Parsely class. This should be used when trying to access a private method for testing.
 	 *
 	 * @param string $method_name Name of the method to get.
+	 * @param string $class       Name of the class the method is in. Can be passed as Foo::class.
```
pauarge:
```suggestion
	 * @param string $class_name  Name of the class the method is in. Can be passed as Foo::class.
```

pauarge (changes requested):
The changes make sense. I left a suggestion regarding naming.

pauarge (approved):

PR #406, opened 2021-09-25
Split out row action link logic

GaryJones:
## Description
Splits out the logic that handles the row action logic that adds the "Parse.ly Stats" link on the list of posts and pages screens.

Builds upon #405 (which should also be merged before this one) which improves a test utility class to make it more flexible, which is then used in this PR.

Also builds upon the `feature/api_key_is_set_or_missing` branch which introduces some utility methods around the API key, which are used in this PR.

As such, #404 and #405 should both be merged first (either order of those two is fine), and when this PR is rebased on to `develop`, only the relevant code changes will be shown.

## Motivation and Context
The `Parsely` god class is too big, so moving logic to its own class makes it easier to maintain. The logic in the new class can be made into more granular methods and tests applied more easily.

See #96.

## How Has This Been Tested?
Visual checks and integration tests.

Screenshot 2021-09-25 at 22 15 24
Screenshot 2021-09-25 at 23 26 33

### DIFF
--- a/src/UI/class-row-actions.php
+++ b/src/UI/class-row-actions.php
@@ -1,0 +1,150 @@
+<?php
+/**
+ * Parsely row actions class
+ *
+ * @package Parsely
+ * @since 2.6.0
+ */
+
+namespace Parsely\UI;
+
+use Parsely;
+use WP_Post;
+
+/**
+ * Handle the post/page row actions.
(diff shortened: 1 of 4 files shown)


pauarge (approved):
🚢 🚢 🚢

Issue #390, opened 2021-09-21
Document new minimums for 3.0.0


GaryJones:
## Is your feature request related to a problem?
v3.0.0 of the plugin will have higher minimum supported version of PHP and WP. This unlocks new syntax (like type declarations which can reduce bugs), reduces the cost of CI tests, and reduces the likelihood of behaviour variances due to using old versions of WordPress.

## Describe the solution you'd like
- [x] Decide on the new minimums. https://github.com/Parsely/wp-parsely/issues/390#issuecomment-926515494
- [x] Document them in the readme before the next release. #400
- [x] Add a dismissable note to the Parse.ly plugin settings page for administrators for any releases between now and 3.0.0 to inform them of the change. https://github.com/Parsely/wp-parsely/pull/408

## Additional context
For the 3.0.0 release, we can also increase the "Requires:" and "Requires PHP:" fields in the plugin headers, which will stop any auto-updates from happening if the requirements aren't met.

pauarge:
Following some internal discussion, we have agreed on bumping the minimum required versions on the `3.0.0` release to:

- **PHP 7.1** (from PHP 5.6).  We want to jump to PHP 7 because we want to take advantage of type declarations and [some other syntactic sugar](https://github.com/Parsely/wp-parsely/pull/370#discussion_r708276211). This changes will make development faster and more reliable. As per the minor version, we go with 7.1 instead of 7.0 because it contains pretty much all the types we need.
- **WordPress 5.0** (from WordPress 4.0). Aside from tackling an [open issue that we already have](https://github.com/Parsely/wp-parsely/issues/381), WordPress 5.0 brought Gutenberg. We are working on supporting blocks in the plugin, so this will ease the development moving forward.

PR #408, opened 2021-09-27
Add admin warning for minimum requirements in 3.0

pauarge:
## Description

This PR adds a dismissible admin warning for users who have a PHP version lower than 7.1 or a WordPress version lower than 5.0. 

This PR should be reverted once 3.0 gets released.

## Motivation and Context

https://github.com/Parsely/wp-parsely/issues/390

Screen Shot 2021-09-27 at 1 38 37 PM

## Types of changes

Breaking change (fix or feature that would cause existing functionality to change)


src/class-parsely.php
```diff
@@ -793,6 +796,29 @@ public function display_admin_warning() {
...
+	/**
+	 * Display a dismissible admin warning if the current WordPress or PHP versions are below the required minimum for the 3.0 release of wp-parsely
+	 * We should get rid of this warning when we release 3.0
+	 *
+	 * @category Function
+	 * @package Parsely
+	 */
+	public function display_admin_upgrade_warning() {
+		global $wp_version;
+		if ( version_compare( phpversion(), '7.1', '>=' ) && version_compare( $wp_version, '5.0', '>=' ) ) {
```
```suggestion
		if ( version_compare( PHP_VERSION, '7.1.0', '>=' ) && version_compare( $wp_version, '5.0', '>=' ) ) {
```

A constant is more performant than a function call.

src/class-parsely.php
```diff
@@ -793,6 +796,29 @@ public function display_admin_warning() {
...
+			return;
+		}
+
+		$message = sprintf(
+		/* translators: %s: Plugin settings page URL */
```
```suggestion
			/* translators: %s: Plugin settings page URL */
```

src/class-parsely.php
```diff
@@ -793,6 +796,29 @@ public function display_admin_warning() {
...
+			__( '<strong>The next version of the Parse.ly plugin will not work with the current setup.</strong> WordPress 5.0 and PHP 7.1 will be the new required minimum versions. You can read more <a href="%s">here</a>.', 'wp-parsely' ),
```
```suggestion
			__( '<strong>The next version of the Parse.ly plugin will not work with the current setup.</strong> WordPress 5.0 and PHP 7.1 will be the <a href="%s">new required minimum versions</a>.', 'wp-parsely' ),
```

We don't need a whole sentence just to add a "click here" (bad for accessibility).

GaryJones (changes requested):

src/class-parsely.php
```diff
@@ -793,6 +796,29 @@ public function display_admin_warning() {
 		<?php
 	}
 
+	/**
+	 * Display a dismissible admin warning if the current WordPress or PHP versions are below the required minimum for the 3.0 release of wp-parsely
+	 * We should get rid of this warning when we release 3.0
+	 *
+	 * @category Function
+	 * @package Parsely
```
```suggestion
	 * 
	 * @since 2.6.0
```

GaryJones (approved):
:shipit:

PR #409, opened 2021-09-27
Using utility function to check for API keys

pauarge:
## Description

This PR replaces all checks for an API key on the `class-parsely.php` file with the helper methods introduced in https://github.com/Parsely/wp-parsely/pull/404. 

We should do similar changes on the rest of the project files in a separate PR.

## Motivation and Context

https://github.com/Parsely/wp-parsely/pull/404

## Types of changes


src/class-parsely.php
```diff
@@ -840,7 +836,7 @@ public function insert_parsely_page() {
 
 		if (
 			// No API key.
```
```suggestion
```

src/class-parsely.php
```diff
@@ -1305,7 +1301,7 @@ public function register_js() {
 		$parsely_options = $this->get_options();
 
 		// If we don't have an API key, there's no need to proceed.
```
```suggestion
```

src/class-parsely.php
```diff
@@ -1343,7 +1339,7 @@ public function register_js() {
 	public function load_js_tracker() {
 		$parsely_options = $this->get_options();
 		// If we don't have an API key, there's no need to proceed.
```
```suggestion
```

GaryJones (approved):

PR #411, opened 2021-09-27
Feature/improve test environment

### LINKED ISSUE #179: Split tests into unit and integration (WP) tests

GaryJones:
Apologies in advance - this has been a battle, so the commit granularity and scope of work is not up the desired standard, but the end result is worth it. I've left the mess of commits in the PR, but I **strongly** suggest that the PR be squashed during the merge so that it becomes a single commit.

## Description

- Uses WP trunk (soon to be 5.9) for integration tests. As per the [dev-note](https://make.wordpress.org/core/2021/09/27/changes-to-the-wordpress-core-php-test-suite/), the WP tests library has been updated.
- The update includes removing the restriction on PHPUnit 7.5 and supporting the latest PHPUnit (9.5).
- This in turns means that PHP 8.0 and later can now be safely tested against.
- The plugin was already using [PHPUnit-Polyfills](https://github.com/Yoast/PHPUnit-Polyfills/) via the [WP Test Utils](https://github.com/Yoast/wp-test-utils) package.
- An issue was discovered with testing against PHP 7.2, which meant the unit tests (based on `Yoast\WPTestUtils\BrainMonkey\TestCase`) needed to be separated out from the integration tests so that it could be run without requiring WordPress. This was done and fixes #179.
- The unit and integration tests both must also pass in PHP 8.1, and are allowed to fail (but still pass) in PHP 8.2 (nightly).
- Since WordPress isn't available for unit tests, then constants defined in the root plugin file can't use functions from WordPress (e.g. `plugin_path_dir()`, etc.), so the constants have been simplified in this and the files where they were being used.
- `convertDeprecationsToExceptions="true"` has been added to the PHPUnit configs which ensures it is even more stricter when it comes to PHP native deprecation notices (counts as a test fail), and lots of those were added to PHP 8.1. All the items highlighted by this have been fixed, but the small code coverage means it's likely that more will be found as new tests are added which call other parts of the plugin source files.
- The PHPUnit output now defaults to TestDox, which means that each test method is converted to English, making for some Agile documentation.
- Missing code coverage annotations from recent merges have been added.
- Assertions that are marked as deprecated for removal in future versions of PHPUnit have been replaced with their non-deprecated equivalents.
- A hack that was needed for running PHPUnit 7.5 with PHP 8 has been removed, as PHP 8 will now use PHPUnit 9.5.

## Motivation and Context
Being able to run tests on PHP 8.0 is essential, and to have them running on PHP 8.1 in preparation for the release is really useful. Not being restricted to PHPUnit 7.5 is wonderful.

## How Has This Been Tested?
Running local and CI tests.

We only have two unit tests, but now the structure is there to use BrainMonkey and mock PHP and WP functions, then we can try to add more unit tests that are quicker to run because they don't require an environment with a database and WordPress to be installed:
Screenshot 2021-09-27 at 23 47 52

Integration tests are run with code coverage (pcov) under PHP 8.0. Also note the use of testdox format which makes it more descriptive instead of a row of Pacman's favourite food...
Screenshot 2021-09-27 at 23 48 43

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -44,6 +44,8 @@ jobs:
             ini-values: pcov.directory=., "pcov.exclude=\"~(vendor|tests)~\""
             coverage: pcov
           - php: '8.1'
+            experimental: false
+          - php: '8.2'
             experimental: true
       fail-fast: false
     continue-on-error: ${{ matrix.experimental }}
@@ -65,10 +67,15 @@ jobs:
       - name: Setup Problem Matchers for PHPUnit
         run: echo "::add-matcher::${{ runner.tool_cache }}/phpunit.json"
 
-      - name: Install Composer dependencies
+      - name: Install Composer dependencies for PHP < 8.2
+        if: ${{ matrix.php < 8.2 }}
+        uses: ramsey/composer-install@v1
+
+      - name: Install Composer dependencies for PHP >= 8.2
+        if: ${{ matrix.php >= 8.2 }}
         uses: ramsey/composer-install@v1
         with:
-          composer-options: '${{ matrix.composer-options }}'
+          composer-options: --ignore-platform-reqs
 
       - name: Start MySQL Service
         run: sudo systemctl start mysql.service
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -30,19 +30,23 @@ jobs:
       # PHP 7.4 uses PHPUnit 9.5.10
       # PHP 8.0 uses PHPUnit 9.5.10
       # PHP 8.1 uses PHPUnit 9.5.10
+      # Keys:
+      # - coverage: Whether to run the tests with code coverage.
+      # - experimental: Whether the build is "allowed to fail".
       matrix:
         php: ['5.6', '7.0', '7.1', '7.2', '7.3', '7.4']
+        experimental: [false]
         include:
+          # Separate out PHP 8.0 so it can run with code coverage.
           - php: '8.0'
-            # Ignore platform requirements, so that PHPUnit 7.5 can be installed on PHP 8.0 (and above).
-#            composer-options: '--ignore-platform-reqs'
+            experimental: false
             extensions: pcov
             ini-values: pcov.directory=., "pcov.exclude=\"~(vendor|tests)~\""
             coverage: pcov
-          # There is no PHP nightly. Due to https://github.com/sebastianbergmann/phpunit/issues/4575 this is never
-          # going to succeed as WordPress is hard-coded to only support PHPUnit 7.5, and then fix only appears
-          # in PHPUnit 8.5.14.
+          - php: '8.1'
+            experimental: true
       fail-fast: false
+    continue-on-error: ${{ matrix.experimental }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -58,15 +58,6 @@ jobs:
       - name: Setup problem matchers for PHP
         run: echo "::add-matcher::${{ runner.tool_cache }}/php.json"
 
-      # Setup PCOV since we're using PHPUnit < 8 which has it integrated. Requires PHP 7.1.
-      # Ignore platform reqs to make it install on PHP 8.
-      # https://github.com/krakjoe/pcov-clobber
-#      - name: Setup PCOV
-#        if: ${{ matrix.php == 8.0 }}
-#        run: |
-#          composer require pcov/clobber --ignore-platform-reqs
-##          vendor/bin/pcov clobber
-
       - name: Setup Problem Matchers for PHPUnit
         run: echo "::add-matcher::${{ runner.tool_cache }}/phpunit.json"
 
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -86,8 +86,6 @@ jobs:
       - name: Run integration tests (single site)
         if: ${{ matrix.php != 8.0 }}
         run: composer testwp
-      - name: Run integration tests (single site with code coverage)
+      - name: Run integration tests (multisite site with code coverage)
         if: ${{ matrix.php == 8.0 }}
         run: composer coveragewp-ci
-      - name: Run integration tests (multisite)
-        run: composer testwp-ms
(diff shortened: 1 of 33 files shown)


.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,81 @@
...
+  # Allow manually triggering the workflow.
+  workflow_dispatch:
+
+jobs:
+  test:
+    name: PHP ${{ matrix.php }}
+    # ubuntu-20.04 / ubuntu-latest includes MySQL 8, which has issues with older versions of PHP.
+    # ubuntu-18.04 includes PHP versions 7.1-8.0, but 5.6-7.1 are cached, so setup is about 5 seconds.
+    # See https://setup-php.com/i/452
+    runs-on: ubuntu-18.04
```
pauarge, line 19: What about running it on ubuntu-20.04? Do we still have issues with older versions of PHP? Same question for the integration tests.

composer.json
```diff
@@ -33,10 +40,19 @@
 	"scripts": {
 		"coverage": [
 			"@putenv WP_MULTISITE=1",
-			"@php ./vendor/bin/phpunit --coverage-html ./build/coverage-html"
+			"@php ./vendor/bin/phpunit --coverage-html ./build/coverage-html/unit -v"
+		],
+		"coveragewp": [
```
pauarge, line 45: Maybe let's rename this to `coverage-wp` for consistency with other commands?

composer.json
```diff
@@ -33,10 +40,19 @@
...
+			"@putenv WP_MULTISITE=1",
+			"@php ./vendor/bin/phpunit --coverage-html ./build/coverage-html/integration -c phpunit-integration.xml.dist -v"
 		],
 		"coverage-ci": [
-			"@php ./vendor/bin/phpunit"
+			"@putenv WP_MULTISITE=1",
+			"@php ./vendor/bin/phpunit -v"
+		],
+		"coveragewp-ci": [
```
pauarge, line 53: And this one to `coverage-wp-ci`

composer.json
```diff
@@ -48,14 +64,21 @@
...
+			"bash bin/install-wp-tests.sh wordpress_test root root localhost trunk"
 		],
 		"test": [
 			"@php ./vendor/bin/phpunit --no-coverage"
 		],
 		"test-ms": [
 			"@putenv WP_MULTISITE=1",
 			"@composer test"
+		],
+		"testwp": [
```
pauarge, line 76: This one to `test-wp`

composer.json
```diff
@@ -48,14 +64,21 @@
...
+			"@php ./vendor/bin/phpunit --no-coverage -c phpunit-integration.xml.dist"
+		],
+		"testwp-ms": [
```
pauarge, line 79: `test-wp-ms`

tests/Integration/bootstrap.php
```diff
@@ -52,3 +44,6 @@ class PHP_Token_NAME_FULLY_QUALIFIED extends PHP_Token {
 // Include the Parsely custom test cases.
 require_once __DIR__ . '/TestCase.php';
 require_once __DIR__ . '/StructuredData/NonPostTestCase.php';
+
+define( 'PARSELY_VERSION', '123456.78.9' );
```
pauarge, line 48: Does this break our checks to test the correct version?

pauarge (commented):
This changes are a great improvement. Awesome work!

.github/workflows/integration-tests.yml
```diff
@@ -75,9 +85,7 @@ jobs:
 
       - name: Run integration tests (single site)
         if: ${{ matrix.php != 8.0 }}
-        run: composer test
-      - name: Run integration tests (single site with code coverage)
+        run: composer testwp
+      - name: Run integration tests (multisite site with code coverage)
```
pauarge, line 89: Would it be possible to test for multi *and* single site?

.github/workflows/unit-tests.yml
GaryJones: The issue was with MySQL 8 - I can't remember the specifics, and right now the hassle isn't worth it, as 18.04 provides everything we need.

Once we move to a minimum of PHP 7.1, _then_ we can look again at bumping to ubuntu-20.04.

composer.json
GaryJones: I was keeping the phrase un-hypenated, as it's not between different states (single site vs -ms for multisite), but completely different tests (unit vs integration).

The fewer places to remember where the hyphens are for these commands, the better, but I'm open to changes if you have a strong feeling on this.

Should probably add the descriptions for the new commands as well.

tests/Integration/bootstrap.php
GaryJones: No, because that test was removed - it was redundant, as it wasn't really testing anything.

If anything, being able to define a random version, and still have the tests pass is a good thing, as it means we don't have logic hard-coded to expect specific version numbers.

.github/workflows/integration-tests.yml
GaryJones: It would, but right now, there's only one test that runs on multisite and skipped on single, so it's really not worth running 99.5% of the same tests twice.

.github/workflows/unit-tests.yml
pauarge: That makes sense. I've opened an issue for when we release 3.0. https://github.com/Parsely/wp-parsely/issues/413

composer.json
pauarge: No strong feelings about it. We can leave it as is.

pauarge (approved):

PR #416, opened 2021-09-29
Increase minimum PHP and WP versions to 7.1 and 5.0

GaryJones:
## Description
We're increasing the minimum supported versions of PHP and WordPress. Version 3.0.0 of this plugin will require PHP 7.1 to work. It will also only work with WordPress 5.0.0 and above.

Other changes, like removal of admin notice, documentation about the increase, and syntax changes that are now permitted, will come in other PRs.

## Motivation and Context
See #390. Since this is a major version release, then backwards-incompatible changes around the expected environment are fine.

## How Has This Been Tested?
No change in CI result, other than fewer jobs now running.

## Types of changes

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -13,8 +13,8 @@ on:
 jobs:
   test:
     name: PHP ${{ matrix.php }}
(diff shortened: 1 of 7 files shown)


pauarge (approved):

jblz:
This is awesome!  Thanks.

PR #419, opened 2021-09-30
CI: Specify coverage: none where it is not needed

### LINKED ISSUE #418: CI: Set "coverage: none" when not explicitly needing coverage
The CI config doesn't specify whether code coverage tooling is enabled or not, other than for the specific case when it should be.

As per https://github.com/shivammathur/setup-php/issues/254 this means that the presence of Xdebug depends on the underlying OS. 

Add `coverage: none` where it specifically is not needed (most instances of PHP running in CI jobs).

Do nothing, but then potentially have tests running slower than needed.

GaryJones:
## Description
Include `coverage: none` where code coverage drivers are not needed for CI runs. This ensures that Xdebug and pcov are both disabled, and isn't reliant on the default of the underlying OS.

Re-ordered a couple of the keys to put them in a consistent order.

## Motivation and Context
Closes #418.

## How Has This Been Tested?
Ran the CI jobs and saw they still passed.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -33,17 +33,20 @@ jobs:
       # - experimental: Whether the build is "allowed to fail".
       matrix:
         php: ['7.1', '7.2', '7.3', '7.4']
+        coverage: [none]
         experimental: [false]
         include:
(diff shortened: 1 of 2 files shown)


pauarge (approved):
LGTM

Issue #389, opened 2021-09-21
PHP 7: Add parameter and return type declarations and declare strict_types=1


GaryJones:
## Is your feature request related to a problem?
PHP 8 introduces stricter controls around type coercion for built-in PHP functions. To build on top of that, we can do the same for our userland code as well.

## Describe the solution you'd like
With PHP 7.1 as the minimum support version, we can add strict type declarations to each PHP file, and (potentially nullable) scalar types to parameters and return types. We can be more confident about what data types are being pushed around, and therefore reduce bugs, as well as facilitating better static analysis and IDE autocompletion.

## Describe alternatives you've considered
We could do nothing, and not add these, but that would be a missed opportunity to reduce the chances of finding or addressing bugs.

## Additional context
Milestone of 3.0.0, since this major release can incorporate a bump in the minimum PHP version required to run the code. PHP 7.1.0 was released in Dec 2016, and the last update was Oct 2019, and is out of Security Support from around the same time - hopefully users are already on newer versions of PHP anyway.

PR #420, opened 2021-09-30
Add strict_types=1 declaration

GaryJones:
## Description
On its own this doesn't do much, but when more type definitions are added, particularly scalar types, then PHP will handle the types without juggling them.

For choice of positioning below the file-level DocBlock, see https://www.php-fig.org/psr/psr-12/.
For lack of spaces inside the parentheses and around the `=`, see https://make.wordpress.org/core/2020/03/20/updating-the-coding-standards-for-modern-php/.

## Motivation and Context
See #389.

## How Has This Been Tested?
Ran CI tests and quick visual inspection.


GaryJones:
Well, that caught a bug around types already! A boolean being passed to a `str_replace()` native PHP function, which produced a TypeError on PHP 8 and above.

I'll get this one fixed.

pauarge:
Looking good! Does it make sense to update our `CONTRIBUTING` or `README` files to let people know that we expect `strict_types=1` on all files?

pauarge (approved):

PR #427, opened 2021-09-30
Tests: Stop using deprecated setMethods method

### LINKED ISSUE #426: Tests: Stop using deprecated setMethods().
As per https://github.com/sebastianbergmann/phpunit/pull/3687 `setMethods()` is deprecated and will be removed in PHPUnit 10.

The option is to use `onlyMethods()` instead, but this needs checking to see if PHPUnit-Polyfills handles this for PHPUnit 7, which we still need to test PHP 7.1.

We only have a single instance of `setMethods()`, so we may be able to fashion a workaround if it's not polyfilled.

GaryJones:
## Description
Use a different MockBuilder function for mocking particular SUT methods.

## Motivation and Context
`PHPUnit\Framework\MockObject\MockBuilder::setMethods()` is deprecated from PHPUnit 8, and will be removed in PHPUnit 10. The replacement is `onlyMethods()`, which was added in PHPUnit 8.

Ideally, PHPUnit-Pollyfills would handle this for PHPUnit 7, but right now it's not supported, so this PR includes a simple workaround.

Closes #426.

## How Has This Been Tested?
Ran the CI tests. 

## Types of changes

### DIFF
--- a/tests/Unit/Integrations/IntegrationsTest.php
+++ b/tests/Unit/Integrations/IntegrationsTest.php
@@ -52,7 +52,13 @@ final class IntegrationsTest extends TestCase {
 	 * @uses \Parsely\Integrations\Integrations::register
 	 */
 	public function test_registered_integrations_have_their_integrate_method_called(): void {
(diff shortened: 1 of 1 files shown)


pauarge (approved):
LGTM

PR #429, opened 2021-09-30
Add return type declarations

GaryJones:
## Description
Add return types to functions and methods, and add/update `@return` DocBlocks as needed.

## Motivation and Context
It allows us to work in a more strongly-typed codebase, which reduces the chances of bugs that happen through silent type coercion. See #389.

## How Has This Been Tested?
Run the CI tests.


pauarge (approved):
LGTM

Issue #396, opened 2021-09-23
Add remaining namespaces


GaryJones:
## Is your feature request related to a problem?
Most src/ and tests/ files have got a PHP namespace, but a couple of old ones don't. 

## Describe the solution you'd like
Add namespaces, and rename the classes as necessary.

## Describe alternatives you've considered
Not changing, but then consistency of namespaces is better.

PR #430, opened 2021-09-30
Add namespaces to files

### LINKED ISSUE #425: Move some missed tests into tests/Integration
`tests/StructuredData/*.php` (x2) and `tests/UI/*.php` (x2, but one already present at `tests/Integration/UI/*.php`) somehow got missed in the refactor to split unit tests and integration tests.

GaryJones:
## Description
Add namespaces to several files, and adjust references. See #396.

This PR also moves the missed integration tests around - fixes #425.

This PR does NOT include a namespace on the `Parsely` class file, as the effects are too-wide reaching (think `@covers` and `@uses` annotations in tests). This will come in a future PR.

The PR also does NOT move the widget class into its own directory and namespace. I was able to add the widget when I had done, but was getting errors, even after running `npm run build`. Can be handled in a future PR.

## Motivation and Context
Namespaces allow long class names to be shortened as classes are logically grouped together. Some files already had namespaces, so this adds them to all but the `Parsely` class file and the view files.

## How Has This Been Tested?
Run the CI, and visually checked.

BREAKING CHANGE.


src/class-parsely-recommended-widget.php
```diff
@@ -12,10 +12,7 @@
 declare(strict_types=1);
 
 /**
- * This is the class for the recommended widget
- *
- * @category   Class
- * @package    Parsely_Recommended_Widget
```
pauarge, line 18: Why are we removing this kind of annotations?

tests/UI/PluginsActionsTest.php
```diff
@@ -1,47 +0,0 @@
-<?php
```
pauarge, line 1: Are we sure to delete this file?

src/class-parsely-recommended-widget.php
GaryJones: Because they aren't valid for class and function/method DocBlocks. 

`@category` is deprecated / removed from the draft list of [PHP DocBlock tags](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md), as `@package` can accept multiple levels (`@subcategory` is also deprecated / removed for the same reason).

@package is only typically useful until namespaces are used anyway, so we'd keep a correct `@package` in the file-level DocBlock (it's wrong in this file), and that's the only single `@package` tag that should be in a file.

tests/UI/PluginsActionsTest.php
GaryJones: Yes, because it already exists at [`tests/Integrations/UI/PluginsActionsTest.php`](https://github.com/Parsely/wp-parsely/blob/develop/tests/Integration/UI/PluginsActionsTest.php).

I'd added it as a separate PR recently because I thought it was missing after I messed up a rebase, but I just wasn't looking for it in the right place.

pauarge:
👋 @GaryJones Could you check this PR again? Tests are failing.

pauarge (approved):

PR #432, opened 2021-10-04
Adding filter to toggle registration of scripts

pauarge:
## Description

This PR releases a new version of the plugin, `2.6.1`, that contains an additional hook to control the plugin's behavior:

- `wp_parsely_should_insert_parsely_page`. Whether the Parse.ly metadata should be inserted in the page's HTML.

By default, the meta tags are rendered (the filters return `true`). 

## Motivation and Context

Some of the plugin's users are using it in a slightly custom way, where they don't always want to display the scripts that Parse.ly adds. 

## How Has This Been Tested?

Tested that both hooks indeed toggle the rendering of the scripts.

## Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -335,7 +335,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add support for parsely-page and JavaScript on home page and published pages and posts as well as archive pages (date/author/category/tag).
 
 [2.6.1]: https://github.com/Parsely/wp-parsely/compare/2.6.0...2.6.1
-[2.6.0]: https://github.com/Parsely/wp-parsely/compare/2.5.2...2.6.0_
+[2.6.0]: https://github.com/Parsely/wp-parsely/compare/2.5.2...2.6.0
 [2.5.2]: https://github.com/Parsely/wp-parsely/compare/2.5.1...2.5.2
 [2.5.1]: https://github.com/Parsely/wp-parsely/compare/2.5.0...2.5.1
 [2.5.0]: https://github.com/Parsely/wp-parsely/compare/2.4.1...2.5.0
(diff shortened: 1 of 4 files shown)


src/class-parsely.php
```diff
@@ -798,6 +799,17 @@ public function print_dynamic_tracking_note() {
 	 * Actually inserts the code for the <meta name='parsely-page'> parameter within the <head></head> tag.
 	 */
 	public function insert_parsely_page() {
+		/**
+		 * Filters whether the Parse.ly meta tags should be inserted in the page.
+		 *
+		 * By default, the tags are inserted.
+		 *
+		 * @since 2.6.1
+		 */
```
GaryJones, line 808:
```suggestion
		 * @since 2.6.1
		 *
		 * @param bool $insert_parsely_page
		 */
```

FWIW, I think this "parsely_page" identity isn't great - we should be talking about "meta tags" or "structured data". We need a better collective name for these two metadata formats, that can then be applied to filter names etc. e.g. `wp_parsely_should_insert_metadata`.

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.6.1] - 2021-10-05
+
+### Added
+
+- Adding filters to toggle registration of scripts #432
```
GaryJones, line 12:
```suggestion
- Add filter to toggle registration of scripts #432
```

GaryJones (changes requested):
I’m strongly against the logic change + version bump/changelog being in the same PR.

pauarge:
Closing this PR since we'll include this changes in 3.0. https://github.com/Parsely/wp-parsely/pull/440

PR #433, opened 2021-10-05
Displaying Open on Parse.ly links by default

pauarge:
## Description

This PR changes the default behavior of the filter that controls whether or not to display the _Open on Parse.ly_ links on posts and pages in `wp-admin`. With this changes, the links will display by default. We are keeping the filter in case the functionality needs to be disabled or displayed only in certain cases.

## Motivation and Context

Making the feature widely available.

## How Has This Been Tested?

Running the plugin's default configuration and by manually setting the filter to `false`.

Screen Shot 2021-10-05 at 4 14 19 PM

## Types of changes

### DIFF
--- a/src/UI/class-row-actions.php
+++ b/src/UI/class-row-actions.php
@@ -48,7 +48,7 @@ final class Row_Actions {
 		 *
 		 * @param bool $enabled True if enabled, false if not.
 		 */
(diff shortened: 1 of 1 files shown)


GaryJones (approved):
✅

amontalenti:
cc @kbourgoin here since it's related to what we were talking about in ProdStrategy (about making sure folks are logged in before linking up their stats etc.)

PR #434, opened 2021-10-05
Adding E2E test to check for available recommended widget

pauarge:
## Description

This PR adds an end to end test that checks if the recommended widget exists an can be added in the site. We have also extracted some common functions into an utils file.

## Motivation and Context

Improve overall testing on the plugin.

## How Has This Been Tested?

Tests pass locally and in GitHub Actions

## Types of changes

### DIFF
--- a/tests/e2e/specs/recommended-widget.spec.js
+++ b/tests/e2e/specs/recommended-widget.spec.js
@@ -1,0 +1,34 @@
+import {
+	activatePlugin,
+	loginUser,
+	visitAdminPage,
+} from '@wordpress/e2e-test-utils';
+
+const waitForWpAdmin = () => page.waitForSelector( 'body.wp-admin' );
+
+describe( 'Recommended widget', () => {
+	jest.setTimeout( 30000 );
+	it( 'Widget should be available', async () => {
+		await loginUser();
+		await activatePlugin( 'wp-parsely' );
+		await visitAdminPage( '/widgets.php', '' );
+
+		await waitForWpAdmin();
+
+		await page.keyboard.press( 'Escape' );
+
+		await page.click( '.block-list-appender' );
+		await page.focus( '#block-editor-inserter__search-0' );
+		await page.keyboard.type( 'parse.ly recommended widget' );
+
+		const [ button ] = await page.$x( "//button[contains(., 'Parse.ly Recommended Widget')]" );
+		await button.click();
+
+		await page.waitForSelector( '.wp-block-legacy-widget__edit-form-title', {
+			visible: true,
+		} );
+
+		const [ h3 ] = await page.$x( "//h3[contains(., 'Parse.ly Recommended Widget')]" );
+		expect( h3 ).toBeTruthy();
+	} );
+} );
--- a/tests/e2e/specs/recommended-widget.spec.js
+++ b/tests/e2e/specs/recommended-widget.spec.js
@@ -20,6 +20,11 @@ const searchForParselyWidget = async () => {
 	await page.keyboard.type( 'parse.ly recommended widget' );
 };
 
+const selectParselyWidgetFromWidgetSearch = async () => {
+	const [ button ] = await page.$x( "//button[contains(., 'Parse.ly Recommended Widget')]" );
+	await button.click();
+};
+
 describe( 'Recommended widget', () => {
 	jest.setTimeout( 30000 );
 	it( 'Widget should be available', async () => {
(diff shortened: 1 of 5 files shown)


tests/e2e/utils.js
```diff
@@ -0,0 +1 @@
+export const waitForWpAdmin = () => page.waitForSelector( 'body.wp-admin' );
```
GaryJones, line 1: Can you mark this utility behaviour in the e2e README, so that other contributors know it is available to use?

tests/e2e/specs/recommended-widget.spec.js
```diff
@@ -0,0 +1,40 @@
...
+describe( 'Recommended widget', () => {
+	jest.setTimeout( 30000 );
+	it( 'Widget should be available', async () => {
+		await loginUser();
+		await activatePlugin( 'wp-parsely' );
+		await visitAdminPage( '/widgets.php', '' );
+
+		await waitForWpAdmin();
+
+		await page.keyboard.press( 'Escape' );
```
GaryJones: Why is this `Escape` needed? If it's to close a modal pop-up, or something else, can we make this more descriptive - prepareWidgetScreen, closeNewWidgetScreenModal, etc.?

tests/e2e/specs/recommended-widget.spec.js
```diff
@@ -0,0 +1,40 @@
...
+
+		await page.click( '.block-list-appender' );
+		await page.focus( '#block-editor-inserter__search-0' );
+		await page.keyboard.type( 'parse.ly recommended widget' );
```
GaryJones: Similarly, could these be grouped into a new bit of code that is called `searchForParselyWidget` or similar? 

Could even then bundle that into a set of actions called `searchForAndAddParselyWidget` etc. Building up a list of human-describable actions, rather than programmatic steps, will make them more re-usable for new e2e tests, and move away from the declarative code of looking for specific classes on elements in the main bulk of tests.

GaryJones (approved):
Approved, but a couple of suggestions that would make the code easier to read and re-use.

tests/e2e/specs/recommended-widget.spec.js
pauarge: That does make sense. We will add more tests into this spec in the near future, for instance adding the widget and checking that it has some options, that it renders on the front-end, etc. It will be pretty useful to be able to reuse this actions.

tests/e2e/specs/recommended-widget.spec.js
```diff
@@ -0,0 +1,45 @@
...
+		await activatePlugin( 'wp-parsely' );
+		await visitAdminPage( '/widgets.php', '' );
+
+		await waitForWpAdmin();
+
+		await closeWidgetScreenModal();
+		await searchForParselyWidget();
+
+		const [ button ] = await page.$x( "//button[contains(., 'Parse.ly Recommended Widget')]" );
+		await button.click();
```
GaryJones: This could be a `selectParselyWidgetFromWidgetSearch`.

tests/e2e/specs/recommended-widget.spec.js
```diff
@@ -0,0 +1,45 @@
...
+
+		await page.waitForSelector( '.wp-block-legacy-widget__edit-form-title', {
+			visible: true,
+		} );
+
+		const [ h3 ] = await page.$x( "//h3[contains(., 'Parse.ly Recommended Widget')]" );
+		expect( h3 ).toBeTruthy();
```
GaryJones, line 46: This could be `parselyWidgetIsPresent`.

tests/e2e/specs/recommended-widget.spec.js
```diff
@@ -0,0 +1,45 @@
...
+
+		await closeWidgetScreenModal();
+		await searchForParselyWidget();
+
+		const [ button ] = await page.$x( "//button[contains(., 'Parse.ly Recommended Widget')]" );
+		await button.click();
+
+		await page.waitForSelector( '.wp-block-legacy-widget__edit-form-title', {
+			visible: true,
+		} );
```
GaryJones, line 44: I'm not sure what the best name for this one would be, but anything has got to be better than that cryptic class name!

GaryJones (approved):
Approved, with a few more suggestions.

Codeception has the idea of [Helpers](https://codeception.com/docs/06-ModulesAndHelpers#Helpers) - in there, they define some custom _actions_ (like you've started to do here), and also some assertions which they starts with `see` or `dontsee` (which works well with their `$I->...` code).

PR #440, opened 2021-10-08
Add wp_parsely_should_insert_metadata filter

pauarge:
## Description

This PR contains an additional hook to control the plugin's behavior:

- `wp_parsely_should_insert_parsely_page`. Whether the Parse.ly metadata should be inserted in the page's HTML.

By default, the meta tags are rendered (the filters return `true`). 

## Motivation and Context

Some of the plugin's users are using it in a slightly custom way, where they don't always want to display the scripts that Parse.ly adds. 

## How Has This Been Tested?

Tested that both hooks indeed toggle the rendering of the scripts.

## Types of changes


src/class-parsely.php
```diff
@@ -772,6 +773,17 @@ public function print_dynamic_tracking_note(): void {
 	 * @return string|null|array
 	 */
 	public function insert_parsely_page() {
+		/**
+		 * Filters whether the Parse.ly meta tags should be inserted in the page.
+		 *
+		 * By default, the tags are inserted.
+		 *
+		 * @since 3.0.0
+		 */
```
```suggestion
		 * @since 3.0.0
		 *
		 * @param bool $insert_metadata True to insert the metadata, false otherwise.
		 */
```

GaryJones (changes requested):

GaryJones (approved):
✅

PR #441, opened 2021-10-08
Removing old init Python script

pauarge:
## Description

This PR removes an old Python script that was used to initialize the repository through SVN. This is not used anymore and it's not mentioned anywhere in the repository, so it's safe to delete.

## Motivation and Context

General cleanup.

## Types of changes


GaryJones (approved):
✅

Issue #423, opened 2021-09-30
Refactor Parsely::insert_parsely_page()


GaryJones:
Right now, this tries to return a string, null, or an array, which makes the consumption of it tricky.

Splitting it into multiple functions so that each has a consistent return type, means that code that calls it can be sure what type it will return.

PR #443, opened 2021-10-11
Constant return type on insert_parsely_page

pauarge:
## Description

This PR refactors the `insert_parsely_page` method to have a constant return type, `void` (it used to return `null`, an empty string, or an unused array depending on the case). The main purpose of this method is to `echo` the Parse.ly tags into the page.

## Motivation and Context

https://github.com/Parsely/wp-parsely/issues/423

## How Has This Been Tested?

All tests suites passing and Parse.ly tags rendering in the header.

## Types of changes

### DIFF
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -856,7 +856,10 @@ class Parsely {
 	}
 
 	/**
-	 * Actually inserts the code for the <meta name='parsely-page'> parameter within the <head></head> tag.
+	 * Deprecated. Echoes the metadata into the page, and returns the inserted values.
+	 *
+	 * To just echo the metadata, use the `insert_page_header_metadata()` method.
+	 * To get the metadata to be inserted, use the `construct_parsely_metadata()` method.
 	 *
 	 * @deprecated 3.0.0
 	 * @see construct_parsely_metadata()
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -121,7 +121,7 @@ class Parsely {
 		add_action( 'parsely_bulk_metas_update', array( $this, 'bulk_update_posts' ) );
 
 		// inserting parsely code.
-		add_action( 'wp_head', array( $this, 'insert_parsely_page' ) );
+		add_action( 'wp_head', array( $this, 'insert_page_header_metadata' ) );
 		add_action( 'init', array( $this, 'register_js' ) );
 
 		// load_js_api should be called prior to load_js_tracker so the relevant scripts are enqueued in order.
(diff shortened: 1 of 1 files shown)


src/class-parsely.php
```diff
@@ -851,8 +851,6 @@ public function insert_parsely_page() {
 		}
 
 		echo '<!-- END Parse.ly -->' . "\n\n";
-
-		return $parsely_page;
```
GaryJones, line 855: This is a breaking change - some customers may be expecting the values to be returned here.

Instead of adjusting this function, could we create a new class / set of functions that handle the distinct behaviour, and then formally deprecate this method, perhaps for removal in 4.0?

pauarge: That is a good point. I have added a deprecation warning, but instead of redirecting users to the new method, I point them to the `construct_parsely_metadata`. That method does return an array instead of printing something, hence, it is more reliable to interact with that data and probably the method customers should have used in the first place.

src/class-parsely.php
```diff
@@ -851,8 +853,20 @@ public function insert_parsely_page() {
...
+	 *
+	 * @deprecated 3.0.0
+	 * @see construct_parsely_metadata()
+	 *
+	 * @return array
+	 */
+	public function insert_parsely_page(): array {
+		_deprecated_function( __FUNCTION__, '3.0', 'construct_parsely_metadata()' );
+		$this->insert_page_header_metadata();
+		return $this->get_options();
```
GaryJones: Is this returning the right thing? We previously returned the metadata, and this is returning the options?

GaryJones (changes requested):

src/class-parsely.php
pauarge: You are right, this is now fixed.

src/class-parsely.php
```diff
@@ -851,8 +853,22 @@ public function insert_parsely_page() {
 		}
 
 		echo '<!-- END Parse.ly -->' . "\n\n";
+	}
 
-		return $parsely_page;
+	/**
+	 * Actually inserts the code for the <meta name='parsely-page'> parameter within the <head></head> tag.
```
```suggestion
	 * Deprecated. Echoes the metadata into the page, and returns the inserted values.
	 *
	 * To just echo the metadata, use the `insert_page_header_metadata()` method.
	 * To get the metadata to be inserted, use the `construct_parsely_metadata()` method.
```

GaryJones (approved):
I think the DocBlock could be improved, but fine otherwise.

Definitely one to highlight as a breaking change in the changelog.

PR #445, opened 2021-10-12
Upgrading Ubuntu CI to 20.04

### LINKED ISSUE #413: Move CI images to Ubuntu 20.04

We are currently using `ubuntu-18.04` images in our GitHub Actions, due to MySQL 8 and old PHP compatibility issues. We should use more up to date system images.

Update those images to use `ubuntu-20.04`. We'll be able to upgrade once the minimum PHP required version is 7.1.

pauarge:
## Description

This PR upgrades the Ubuntu version in which our CI runs, from 18.04 to 20.04. Some of the pipelines were running in `ubuntu-latest`,  but we're changing it to 20.04 explicitly to prevent CI breaking in the future due to an external change.

The reason we were holding off in 18.04 was because PHP 7.3 and lower would break because if would not be compatible with a change introduced in MySQL 8, namely, changing the default authentication to `caching_sha2_password`. This PR changes that back to `mysql_native_password` for those versions.

## Motivation and Context

Closes #413 

## How Has This Been Tested?

CI runs correctly in all PHP versions.


.github/workflows/integration-tests.yml
```diff
@@ -7,16 +7,11 @@ on:
     paths-ignore:
       - '**.md'
   pull_request:
-  # Allow manually triggering the workflow.
-  workflow_dispatch:
```
GaryJones, line 11: Why was this removed?

GaryJones (approved):
✅

.github/workflows/integration-tests.yml
pauarge: I'm not a big fan of manually triggering workflows, because they might have unintended consequences. I find it way better and easier to track to tie executions to a push/pull request.

Issue #422, opened 2021-09-30
Investigate why update_meta_endpoint() returns an empty string


GaryJones:
Doesn't seem right to return a string from a guard clause in `\Parsely::update_metadata_endpoint()`, when there's no other `return`.

GaryJones:
This function also assigns some values to variables that are never used.

PR #446, opened 2021-10-12
Fixing return types of update_metadata_endpoint

pauarge:
## Description

The `update_metadata_endpoint` function was not returning consistently. Since its a hook and its result is not used anywhere, we're making it return `void`.

The function basically performs a request to Parse.ly and then updates the metadata of the updated post with a timestamp. This PR also adds a check to see if the request succeeded and only updates the metadata if that's the case.

## Motivation and Context

https://github.com/Parsely/wp-parsely/issues/422

## How Has This Been Tested?

Running the code in a local environment.

## Types of changes


GaryJones (approved):
✅

Issue #246, opened 2021-04-22
Update references of "nonpost" to "non-post"


GaryJones:
We have a default value of "nonpost" [here](https://github.com/Parsely/wp-parsely/blob/d7a06bd696a7af3e31e49ee98f893ff8a1d07755/wp-parsely.php#L1572), but the [Parse.ly documentation](https://www.parse.ly/help/integration/jsonld#distinguishing-between-posts-and-pages) calls it "non-post" (with the hyphen).

We should aim to be consistent and have references be "non-post". 

There may only be one hard-coded instance, but the value is passed to a filter hook, so changing this would count as a breaking change since a filter callback that was looking to act on "nonpost" would no longer work correctly.

kbourgoin:
Took a quick look in the crawl code and this will be just fine.  This is the relevant line: https://github.com/Parsely/engineering/blob/7e174978042b2bbfa6761c1ac6411139207e3642/engicore/engicore/crawl/parsely_spider.py#L317

In short, anything not labeled "post" will be considered a nonpost (or non-post as the case may be).  We're good with making this change.

PR #447, opened 2021-10-12
Change get_current_url default type to non-post

pauarge:
## Description

This PR changes the default argument of the `get_current_url` function from `nonpost` to `non-post`. This change is to keep in line with the naming of the rest of Parse.ly

## Motivation and Context

https://github.com/Parsely/wp-parsely/issues/246

## How Has This Been Tested?

Test that URL metadata in a post works fine and the URL is generated correctly.

## Types of changes


GaryJones (approved):
✅

PR #451, opened 2021-10-13
Fix recommended widget not following configuration

pauarge:
## Description

This PR contains a bug fix that prevented users from properly configuring whether the personalized widget would display the author name or not. 

This issue was caused because the parameters were passed as strings, so `"true"` and `"false"` would all evaluate `true`. We are now casting those values.

## Motivation and Context

Some customers could not select if the author name was displayed in the widget.

## How Has This Been Tested?

Set up two parallel widgets on a test site: one should display the author and the other shouldn't. Checked that this was the case.

**Before:**

Screen Shot 2021-10-13 at 9 08 56 AM

**After:**

Screen Shot 2021-10-13 at 9 08 26 AM

## Types of changes


GaryJones (approved):
✅

PR #455, opened 2021-10-18
Adding argument types to functions

pauarge:
## Description

This PR adds parameter types to all functions that need them in the codebase. As a notable exception, we haven't added a type annotation on functions that accept a union type, because that's not supported on PHP 7.1.

We have also updated some tests that were lacking some parameters and would pass before, but with stricter type definitions, they fail.

## Motivation and Context

## How Has This Been Tested?

Check that the plugin operates well and no tests fail.

## Types of changes


src/UI/class-row-actions.php
```diff
@@ -90,7 +90,7 @@ public function row_actions_add_parsely_link( $actions, WP_Post $post ): array {
 	 * @param WP_Post $post    Which post object or ID to check.
 	 * @return bool True if the link cannot be shown, false if the link can be shown.
 	 */
-	private function cannot_show_parsely_link( $actions, WP_Post $post ): bool {
+	private function cannot_show_parsely_link( array $actions, WP_Post $post ): bool {
 		return ! is_array( $actions ) ||
```
GaryJones: Since we're type declaring the parameter, we can remove this `! is_array()` check.

src/class-parsely-recommended-widget.php
```diff
@@ -51,7 +51,7 @@ public function __construct() {
 	 * @param int    $return_limit     Number of records to retrieve; defaults to "10".
 	 * @return string API URL.
 	 */
-	public function get_api_url( $api_key, $published_within, $sort, $boost, $return_limit ): string {
+	public function get_api_url( string $api_key, int $published_within, string $sort, string $boost, int $return_limit ): string {
```
GaryJones, line 54: The casting of `$published_within` to int on line 67 can be removed.

src/class-parsely.php
```diff
@@ -1323,10 +1323,10 @@ public function load_js_api(): void {
 	 *
 	 * @param string $tag    The `script` tag for the enqueued script.
 	 * @param string $handle The script's registered handle.
-	 * @param string $src    The script's source URL.
+	 * @param string $src    Unused? The script's source URL.
```
GaryJones, line 1328: Agreed that this seems unused.

src/class-parsely.php
```diff
@@ -1681,10 +1680,10 @@ private function get_bottom_level_term( $post_id, $taxonomy_name ): string {
 	 * Get all term values from custom taxonomies.
 	 *
 	 * @param WP_Post $post_obj The post object.
-	 * @param array   $parsely_options The pparsely options.
+	 * @param array   $parsely_options Unused? The parsely options.
```
GaryJones, line 1685: Agreed that this is also unused now.

src/class-parsely.php
```diff
@@ -1681,10 +1680,10 @@ private function get_bottom_level_term( $post_id, $taxonomy_name ): string {
...
 	 * @return array
 	 */
-	private function get_custom_taxonomy_values( $post_obj, $parsely_options ): array {
+	private function get_custom_taxonomy_values( WP_Post $post_obj, array $parsely_options ): array {
 		// filter out default WordPress taxonomies.
 		$all_taxonomies = array_diff( get_taxonomies(), array( 'post_tag', 'nav_menu', 'author', 'link_category', 'post_format' ) );
```
GaryJones, line 1690: Not for this PR, but this seems like a brittle list - could probably do a lookup of these built in taxonomies so that it keeps itself up to date.

src/class-parsely.php
```diff
@@ -1756,7 +1754,7 @@ private function get_coauthor_names( $post_id ): array {
 	 * @param WP_User $author The author of the post.
 	 * @return string
 	 */
-	private function get_author_name( $author ): string {
+	private function get_author_name( WP_User $author ): string {
 		// gracefully handle situation where no author is available.
 		if ( empty( $author ) || ! is_object( $author ) ) {
```
GaryJones, line 1761: Can one or both parts of this condition be removed now that we're declaring a non-nullable type declaration? 

Alternatively, should the type be `?WP_User`? Or a better way to account for when there apparently might be no author (maybe in where this code is called from)?

src/class-parsely.php
```diff
@@ -1825,13 +1823,12 @@ private function get_author_names( $post ): array {
 	 * @param string $val The content you'd like sanitized.
 	 * @return string
 	 */
-	public function get_clean_parsely_page_value( $val ): string {
+	public function get_clean_parsely_page_value( string $val ): string {
 		if ( is_string( $val ) ) {
```
GaryJones: This `is_string()` check is now redundant.

tests/Integration/RecommendedApiTest.php
```diff
@@ -62,21 +62,22 @@ public function data_recommended_api_url(): array {
 	 *
 	 * @dataProvider data_recommended_api_url
 	 * @covers \Parsely\Widgets\Recommended_Widget::get_api_url
-	 * @uses \Parsely\Widgets\Recommended_Widget::__construct
-	 * @group widgets
 	 *
 	 * @param string $api_key          Publisher Site ID (API key).
 	 * @param int    $published_within Publication filter start date; see https://www.parse.ly/help/api/time for
-	 *                                 formatting details. No restriction by default.
+	 *                                    formatting details. No restriction by default.
```
GaryJones: Why the indentation here?

tests/Integration/OtherTest.php
```diff
@@ -277,8 +277,9 @@ public function test_parsely_page_filter(): void {
 		$parsely_options = get_option( \Parsely::OPTIONS_KEY );
 
 		// Create a single post.
-		$post_id = $this->factory->post->create();
-		$post    = get_post( $post_id );
+		$post_array = $this->create_test_post_array();
+		$post_id    = $this->factory->post->create( $post_array );
```
GaryJones: What advantage does overriding the [default test post array values](https://core.trac.wordpress.org/browser/trunk/tests/phpunit/includes/factory/class-wp-unittest-factory-for-post.php#L17) here give?

tests/Integration/TestCase.php
```diff
@@ -37,17 +37,18 @@ abstract class TestCase extends WPIntegrationTestCase {
...
 
 	/**
 	 * Create a test post.
 	 *
 	 * @param string $post_type Optional. Post type. Default is 'post'.
+	 *
 	 * @return array An array of WP_Post fields.
 	 */
-	public function create_test_post_array( $post_type = 'post' ): array {
+	public function create_test_post_array( string $post_type = 'post' ): array {
```
GaryJones, line 51: What benefit does the method serve? Only one value is overrideable (and a post title saying "Post" wouldn't make sense when the post type is not `post`), so it gives less flexibility than using the [default test post values](https://core.trac.wordpress.org/browser/trunk/tests/phpunit/includes/factory/class-wp-unittest-factory-for-post.php#L17) with the factory `create()` and similar functions.

We may want to think about scrapping it.

tests/Integration/TestCase.php
```diff
@@ -106,9 +110,10 @@ public function create_test_blog( $domain, $user_id ) {
 	 *
 	 * @param string $taxonomy_key Taxonomy key, must not exceed 32 characters.
 	 * @param string $term_name    The term name to add.
+	 *
```
GaryJones: Technically the WPCS standard has no clear line between the `@param` and the `@return` lines.

tests/Integration/TestCase.php
```diff
@@ -106,9 +110,10 @@ public function create_test_blog( $domain, $user_id ) {
...
 	 * @return array|WP_Error An array containing the term_id and term_taxonomy_id, WP_Error otherwise.
```
GaryJones, line 110: How about adding in a try...catch to this method (and similar methods) so that any WP_Error is caught and handle, and letting this method only return the intended `array`?

tests/Integration/TestCase.php
```diff
@@ -135,7 +140,7 @@ public function create_test_taxonomy( $taxonomy_key, $term_name ) {
 	 * @return \ReflectionMethod
 	 * @throws \ReflectionException The method does not exist in the class.
 	 */
-	public static function getMethod( $method_name, $class_name = 'Parsely' ): \ReflectionMethod {
+	public static function getMethod( string $method_name, string $class_name = 'Parsely' ): \ReflectionMethod {
```
GaryJones, line 139: The naming of this method is camelCase, yet all of the others are snake_case.

Could even potentially rename it to something slightly more descriptive.

GaryJones: I'd also like to see the calls to this method updated to avoid the optional value on the second argument - that'll make the integration tests more understandable.

(And yes, I know I was the one who added the optional value recently!)

GaryJones (changes requested):
I will also note https://steemit.com/php/@crell/php-never-type-hint-on-arrays

src/class-parsely.php
pauarge: I will address that on a separate PR.

src/class-parsely.php
pauarge: Same as above

src/class-parsely.php
pauarge: I think we're safer here moving to `?WP_User`. I can see some scenarios where this can be nullable.

tests/Integration/RecommendedApiTest.php
pauarge: Added by mistake.

tests/Integration/OtherTest.php
pauarge: The default values create a post with a user that doesn't exist. This can be reverted now that we've changed the `get_author_name` to have a nullable argument.

tests/Integration/TestCase.php
pauarge: Agreed that we could get rid of it in the future.

tests/Integration/TestCase.php
pauarge: This is PHPStorm doing its things. I will revert.

tests/Integration/TestCase.php
pauarge: That makes sense, but it is out of the scope of this PR. Will address it separately.

tests/Integration/TestCase.php
pauarge: You have a point. Again, I will address this separately.

tests/Integration/OtherTest.php
GaryJones: > This can be reverted now that we've changed the `get_author_name` to have a nullable argument.

The removal of the newly added `$this->create_test_post_array();` across multiple tests seems to be outstanding.

tests/Integration/TestCase.php
GaryJones: > I will revert.

This one hasn't been reverted.

GaryJones (changes requested):
Two items that I read as you suggesting would be addressed in this PR and haven't been.

tests/Integration/OtherTest.php
pauarge: Actually, it looks like we can't really remove this changes https://github.com/Parsely/wp-parsely/pull/455/checks?check_run_id=3937955570

GaryJones: If we're having to provide extra data when creating a test, then it seems like our non-test code should be able to better cope with the data we're passing to `get_author_name()`.

Can we add some more defensive code to the `get_author_names()` (plural) function around line 1804 of `class-parsely.php`, so that we don't pass a `false` to the array_map callback?

This is exactly why we're adding strongly typed code, to catch issues like this. I'm not keen on the current approach of just setting an author, to gloss over this issue.

GaryJones (changes requested):

src/class-parsely.php
```diff
@@ -1788,7 +1788,10 @@ private function get_author_name( ?WP_User $author ): string {
 	private function get_author_names( WP_Post $post ): array {
 		$authors = $this->get_coauthor_names( $post->ID );
 		if ( empty( $authors ) ) {
-			$authors = array( get_user_by( 'id', $post->post_author ) );
+			$post_author = get_user_by( 'id', $post->post_author );
+			if ( $post_author ) {
+				$authors = array( get_user_by( 'id', $post->post_author ) );
```
```suggestion
				$authors = array( $post_author );
```

Does this change make sense and avoid repeating the `get_user_by()` call?

GaryJones (approved):
✅

PR #457, opened 2021-10-19
Removing Parse.ly load text domain

pauarge:
## Description

Removing a text domain set up that was required for WordPress versions up to 4.6. With the 3.0 release of the plugin, we're raising the minimum required WP version to 5.0.

## Motivation and Context

General plugin cleanup.

## How Has This Been Tested?

The admin interface loads correctly and labels are properly rendered.

## Types of changes

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -77,24 +77,6 @@ function parsely_recommended_widget_register(): void {
 	register_widget( Parsely_Recommended_Widget::class );
 }
 
-add_action( 'init', __NAMESPACE__ . '\\parsely_load_textdomain' );
-/**
- * Load plugin textdomain.
- *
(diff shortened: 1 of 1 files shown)


GaryJones:
> The admin interface loads correctly and labels are properly rendered.

Have you for a screenshot of this loading in non-English with this patch under WP 5.x, from the vip-go-mu-plugins directory? I'm curious if we need to provide documentation on how to add the language files for this plugin for VIP-hosted sites, so that just having them present means that they work when the user interface setting is not US English.

pauarge:
@GaryJones This does work well on VIP and non-VIP environments (all VIP sites are above 4.6).

GaryJones (approved):
✅

PR #459, opened 2021-10-19
Move Parse.ly settings to views folder

pauarge:
## Description

This PR moves the plugin's settings page (view) from the `src/` folder to the `views/` one. Views like it are located in that folder.

## Motivation and Context

Improved code structure.

## How Has This Been Tested?

The settings page loads correctly.

## Types of changes

### DIFF
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -206,7 +206,7 @@ class Parsely {
 			wp_die( esc_html__( 'You do not have sufficient permissions to access this page.', 'wp-parsely' ) );
 		}
 
-		include plugin_dir_path( PARSELY_FILE ) . 'src/parsely-settings.php';
+		require_once plugin_dir_path( PARSELY_FILE ) . 'views/parsely-settings.php';
 	}
 
 	/**
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -206,7 +206,7 @@ class Parsely {
 			wp_die( esc_html__( 'You do not have sufficient permissions to access this page.', 'wp-parsely' ) );
 		}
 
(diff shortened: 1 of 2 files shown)


src/class-parsely.php
```diff
@@ -206,7 +206,7 @@ public function display_settings(): void {
 			wp_die( esc_html__( 'You do not have sufficient permissions to access this page.', 'wp-parsely' ) );
 		}
 
-		include plugin_dir_path( PARSELY_FILE ) . 'src/parsely-settings.php';
+		require_once plugin_dir_path( PARSELY_FILE ) . 'views/parsely-settings.php';
```
```suggestion
		include plugin_dir_path( PARSELY_FILE ) . 'views/parsely-settings.php';
```

`require` will throw a fatal if the file doesn't exist, which isn't typically right for a view, since we want the user to be able to better understand what went wrong (i.e. still see the WP admin frame), and not WSOD them, so stick with `include`.

Since this is our view, and we control when it's required, and we definitely want it to render here, even if it might theoretically have been required elsewhere, then avoid the use of `_once`, since that wouldn't give any warning and leave the user confused / hide the issue.

GaryJones (changes requested):

src/class-parsely.php
pauarge: Good point. I actually changed to `require_once` be more strict (the `parsely-settings.php` file should always be there if the plugin is working correctly). However, I see your point so happy to revert.

GaryJones: All view files should always be there - but the strictness relates to the user experience if it's not there for some reason.

GaryJones (approved):
✅

Issue #399, opened 2021-09-24
Add testing on the recommended widget


pauarge:
## Is your feature request related to a problem?

The recommended widget fetches the data and renders completely on the client side through Javascript. That Javascript is self contained and built through Webpack, but it doesn't have any testing. 

We also don't have testing on the PHP side. Some integration tests around how data is managed and updated for the plugin options would be helpful.

## Describe the solution you'd like

We'd like to add Jest-based unit tests (similar to `personalization.test.js`) plus E2E tests (similar to `activation-flow.spec.js`) that cover the front-end part of the widget.

To cover the back-end part, we would want PHP integration tests.

## Additional context

We could also consider adding testing to cover this widget on the WordPress (PHP) side, basically regarding the admin panel.

jblz:
See #545 -- the widget was broken and this sort of test would have caught it.

acicovic:
We are planning on removing the Widget in wp-parsely 4.0.0. I'm unsure we would write tests for a feature that is to be removed, so we should see if this ticket is still relevant.

pauarge:
We are safe to close this. The recommended widget is in maintenance mode right now and it will be deprecated.

PR #460, opened 2021-10-20
Avoid making duplicate calls on Recommended Widget

### LINKED ISSUE #206: Widget: Duplicate API calls made when multiple instances of the widget are present

pauarge:
## Description

The Recommended Widget would make an API call for each instance of the widget in the page. That is suboptimal given the fact that the API call doesn't have parameters and the complete URL doesn't change much. For this reason, we have refactored the front-end code of the recommended widget to optimize the number of calls.

We have rewritten the code in such a way that we first generate the URL we should call for each widget and then we group the widgets with the same URL. Then, we do a call for each group and we then render every single widget, sharing the data.

## Motivation and Context

Closes https://github.com/Parsely/wp-parsely/issues/206

## How Has This Been Tested?

Place multiple widgets on the site and set them up so some use the "personalized" options and others don't. Open network tools and check that the number of requests gets reduced (especially if multiple widgets have similar configurations).

## Types of changes


GaryJones:
> Place multiple widgets on the site and set them up so some use the "personalized" options and others don't. 

While we probably can't easily count network connections in the e2e tests, can we at least have an e2e test added that shows the widgets before this patch, and after this patch, are consistently rendered?

pauarge:
@GaryJones Agreed that we should improve the E2E coverage on this front. I've added some tests that improve the current situation but don't really cover what we are changing here.

This is because we need to mock some network calls when we render the widget. I've been working on this for a bit and I will open a separate PR with the changes, given that they aren't straight forward.

src/class-parsely.php
```diff
@@ -907,7 +907,7 @@ public static function post_has_trackable_status( $post ): bool {
 	 * @param WP_Post $post object.
 	 * @return array
 	 */
-	public function construct_parsely_metadata( array $parsely_options, WP_Post $post ): array {
+	public function construct_parsely_metadata( array $parsely_options, ?WP_Post $post ): array {
```
GaryJones: Why was this nullable needed?

src/UI/class-recommended-widget.php
```diff
@@ -276,7 +276,7 @@ class="tiny-text" aria-labelledby="<?php echo esc_attr( $this->get_field_id( 'pu
 	public function update( $new_instance, $old_instance ): array {
 		$instance                        = $old_instance;
 		$instance['title']               = trim( wp_kses_post( $new_instance['title'] ) );
-		$instance['published_within']    = (int) trim( $new_instance['published_within'] );
+		$instance['published_within']    = is_int( $new_instance['published_within'] ) ? $new_instance['published_within'] : (int) trim( $new_instance['published_within'] );
```
GaryJones, line 279: Should there be a test that ensures we capture the logic being added here?

GaryJones (approved):
✅

src/UI/class-recommended-widget.php
pauarge: Yes, we'll cover it here: https://github.com/Parsely/wp-parsely/issues/399

PR #462, opened 2021-10-20
i18n: Fix JS string translation

GaryJones:
## Description
The textdomain string was added as an argument to the `sprintf()` instead of the `__()` call.

WordPress was not told to load the script translations for strings in JavaScript files.

This also configures the ESLint rules to define the correct textdomain.

## Motivation and Context
The plugin is mostly internationalized, but there was a string in the JS file that was not being picked up, and so appeared in English regardless of the used locale.

## How Has This Been Tested?
Got it working locally - with WordPress setup to display in French, the JS string that was displaying in English is now displaying in French.

Before:

Screenshot 2021-10-20 at 13 53 08

After:

Screenshot 2021-10-20 at 14 12 48

## Types of changes


pauarge (approved):
LGTM

PR #464, opened 2021-10-22
Adding labels to admin settings form

### LINKED ISSUE #229: Add labels for text input fields in settings page
Radio buttons and checkboxes have their labels right next to the input, but text input fields don't have a marked up label, unless the `label_for` argument is used with a value of the input `id` attribute, in the custom args parameter for the [`add_settings_field()` function](https://developer.wordpress.org/reference/functions/add_settings_field/) calls.

pauarge:
## Description

This PR adds an accessibility feature to the plugin's settings page: labels to `select` and `input type="text"` fields. The radio buttons and checkboxes already had the labels. We are adding labels by using `label_for` on WP's `add_settings_field`.

This PR changes those fields in the `class-parsely.php` file. While working with this code it has become quite obvious that we can split that logic out of the file. We're not doing this in this PR to clearly show what is being changed. After that, we'll move the admin settings out of the big class.

## Motivation and Context

Closes https://github.com/Parsely/wp-parsely/issues/229

## How Has This Been Tested?

The admin settings page has the form working and the text fields on the left column are correctly tagged with `label for="".

## Types of changes

### DIFF
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -315,17 +315,17 @@ class Parsely {
 		$h_link     = 'https://www.parse.ly/help/integration/jsonld/';
 		$field_id   = 'meta_type';
 		$field_args = array(
-			'option_key'         => $field_id,
-			'help_text'          => $h,
-			'help_link'          => $h_link,
+			'option_key'       => $field_id,
+			'help_text'        => $h,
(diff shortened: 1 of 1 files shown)


GaryJones (approved):
✅

PR #467, opened 2021-10-25
Extract settings page from Parsely class

pauarge:
## Description

This PR extracts all the methods related to the rendering of the `wp-admin` settings page to a separate class file. The code itself hasn't been modified, but there still is some room for improvement within this new class, which is considerably large.

## Motivation and Context

https://github.com/Parsely/wp-parsely/issues/96

## How Has This Been Tested?

The wp-admin settings page for the plugin renders and saves all the data as expected.

## Types of changes


wp-parsely.php
```diff
@@ -65,6 +66,15 @@ function(): void {
 	}
 );
 
+require __DIR__ . '/src/UI/class-settings-page.php';
+add_action(
+	'init',
```
GaryJones: Bearing in mind the hooks called in `run()`, could we limit this to `admin_init` hook, even if it's priority 9? We probably shouldn't need to load this Settings Page code for every front-end request.

pauarge: Agreed, we shouldn't hook this in the front-end (although the only result would be three additional hooks that would never be executed). 

I initially hooked it into `admin_init` but it wouldn't work. I've now hooked into `_admin_menu` and it does work.

GaryJones (approved):
✅ (Haven't had chance to pull down and run locally, but changes look sensible)

PR #473, opened 2021-10-28
Adding wp_parsely_enable_cfasync_tag filter to display Cloudflare links

### LINKED ISSUE #293: Make cfasync script attributes opt-in

As mentioned [here](https://github.com/Parsely/wp-parsely/pull/286#issuecomment-833474663), there's no need to have these attributes in place for sites that don't use Cloudflare.

It should, therefore, be opt in.  Since it's a breaking change, I'm adding it to the 3.0.0 milestone.

Either a checkbox in the "Advanced" settings section or a filter to opt-in to this behavior (I'd lean toward the latter).

pauarge:
## Description

A Cloudflare tag is being added in all scripts so it gets ignored by their Rocket Loader. Since most of the sites running this plugin are not behind Cloudflare, we are making that behavior optional and we're hiding it behind a filter. 

## Motivation and Context

Closes https://github.com/Parsely/wp-parsely/issues/293.

## How Has This Been Tested?

With the default configuration, the Parse.ly scripts should not contain the `data-cfasync` tag. Add the filter to activate:

```
add_filter( 'wp_parsely_enable_cfasync_tag', '__return_true' );
```

then the scripts should contain the tag.

## Types of changes


GaryJones:
This should include a README update as well.

pauarge:
@GaryJones We'll have a very comprehensive changelog for 3.0 as well, detailing changes like this.

README.md
```diff
@@ -86,6 +86,16 @@ Some common use-cases for dynamic tracking are slideshows or articles loaded via
 
 Tracking these events requires manually implementing additional JavaScript above [the standard Parse.ly include](http://www.parsely.com/help/integration/basic/) that the plugin injects into your page source. Please consult [the Parse.ly documentation on dynamic tracking](https://www.parsely.com/help/integration/dynamic/) for instructions on implementing dynamic tracking, or contact Parse.ly support for additional assistance.
 
+### CloudFlare support
+
+If the site is running behind a CloudFlare DNS, [their Rocket Loader technology will alter how Javascript files are loaded. [A Javascript file can be marked](https://support.cloudflare.com/hc/en-us/articles/200169436-How-can-I-have-Rocket-Loader-ignore-specific-JavaScripts) so it is not touched by using `data-cfasync="false"`.
```
GaryJones:
```suggestion
If the site is running behind a Cloudflare DNS, their Rocket Loader technology will alter how JavaScript files are loaded. [A JavaScript file can be ignored by Rocket Loader](https://support.cloudflare.com/hc/en-us/articles/200169436-How-can-I-have-Rocket-Loader-ignore-specific-JavaScripts) by using `data-cfasync="false"`.
```

README.md
```diff
@@ -86,6 +86,16 @@ Some common use-cases for dynamic tracking are slideshows or articles loaded via
...
+
+Previous versions of the plugin would mark all scripts with that tag by default. Starting in version 3, that behavior has become optional and scripts won't be annotated with `data-cfasync="false"`. The previous behavior can be restored by adding the following filter:
```
GaryJones:
```suggestion
Previous versions of this plugin would mark all scripts with that tag by default. Starting in version 3.0, that behavior has become optional and scripts won't be annotated with `data-cfasync="false"`. The previous behavior can be restored by adding the following filter:
```

src/class-parsely.php
```diff
@@ -1346,9 +1346,18 @@ public function script_loader_tag( string $tag, string $handle, string $src ): s
...
+			/**
+			 * Filter whether to include the CloudFlare Rocket Loader tag (`data-cfasync=false`) in the script. Only
+			 * needed if the site being served is behind Cloudflare. Should return false otherwise.
+			 * https://support.cloudflare.com/hc/en-us/articles/200169436-How-can-I-have-Rocket-Loader-ignore-specific-JavaScripts
+			 *
+			 * @since 3.0.0
+			 *
+			 * @param bool $enabled True if enabled, false if not.
+			 */
+			if ( apply_filters( 'wp_parsely_enable_cfasync_tag', false ) ) {
```
```suggestion
			if ( apply_filters( 'wp_parsely_enable_cfasync_tag', false, $handle ) ) {
```

If we make the handle available, then it allows folks to conditionally apply this to only some of the 4 Parse.ly scripts if they wish.

tests/Integration/OtherTest.php
```diff
@@ -148,6 +148,51 @@ public function test_load_js_tracker(): void {
...
+			'',
+			$intermediate_output,
+			'Failed to confirm scripts were not printed by load_js_tracker()'
+		);
+
+		self::assertTrue(
+			wp_script_is( 'wp-parsely-tracker', 'enqueued' ),
+			'Failed to confirm tracker script was enqueued'
+		);
+
```
GaryJones: Do these assertions provide any value? The ability to load the JS tracker at all, is not really what's being tested here; these checks are handled in the previous test method.

Let's aim to keep the assertions in this method more focused to be just the difference between the previous test method and this one (i.e. what happens when the Cloudflare flag is true).

tests/Integration/OtherTest.php
```diff
@@ -148,6 +148,51 @@ public function test_load_js_tracker(): void {
...
+	 * @uses \Parsely::api_key_is_set
+	 * @uses \Parsely::get_options
+	 * @uses \Parsely::post_has_trackable_status
+	 * @uses \Parsely::register_js
+	 * @uses \Parsely::script_loader_tag
+	 * @uses \Parsely::update_metadata_endpoint
+	 * @group insert-js
+	 */
+	public function test_load_js_tracker_with_cloudflare(): void {
+		add_filter( 'wp_parsely_enable_cfasync_tag', '__return_true' );
```
GaryJones:
```suggestion
		add_filter( 'wp_parsely_enable_cfasync_attribute', '__return_true' );
```

The `data-cfasync` is an attribute, not a tag.

But, instead of being so granular, perhaps we could call it `wp_parsely_enable_cloudflare_support` so that it's easier to remember, more self-descriptive, and allows the same filter to also power any future Cloudflare adaptations that are needed.

pauarge: Agreed on naming it `wp_parsely_enable_cfasync_attribute`. 

However, I'd keep that instead of something more Cloudflare generic. That is for two situations that might arise in the future: 

- Cloudflare releases a new feature and a user wants to have `cfasync` enabled but that feature disabled.
- A party other than Cloudflare starts using the `cfasync` attribute. 

If you agree, I'd apply your suggestion as is (and change it in other occurrences in this PR's code).

GaryJones: OK, keep it as `wp_parsely_enable_cfasync_attribute` for now then.

README.md
```diff
@@ -86,6 +86,16 @@ Some common use-cases for dynamic tracking are slideshows or articles loaded via
 
 Tracking these events requires manually implementing additional JavaScript above [the standard Parse.ly include](http://www.parsely.com/help/integration/basic/) that the plugin injects into your page source. Please consult [the Parse.ly documentation on dynamic tracking](https://www.parsely.com/help/integration/dynamic/) for instructions on implementing dynamic tracking, or contact Parse.ly support for additional assistance.
 
+### CloudFlare support
```
GaryJones:
```suggestion
### Cloudflare support
```

GaryJones (approved):
One typo, but otherwise ✅

PR #474, opened 2021-10-28
Fix type errors

pauarge:
## Description

Types in function arguments were added in https://github.com/Parsely/wp-parsely/pull/455. In some situation, types might not be enforced. This PR adds some explicit casts to make sure types are always correct.

## Motivation and Context

## How Has This Been Tested?

The Recommended Widget renders correctly, both in the front-end and in the widgets editor in `wp-admin`. 

## Types of changes


GaryJones (approved):
✅ 

Should there be tests to show that a `string` can now be passed in and it will be handled correctly (thinking about how PHP changed how it coerced number-like strings in PHP 8).

pauarge:
@GaryJones Agreed. We have an open issue to add tests to the Recommended Widget (there are none now). https://github.com/Parsely/wp-parsely/issues/399

PR #476, opened 2021-11-01
e2e tests: fix watch command

GaryJones:
## Description
Fix `npm run test:e2e:watch`. Previously it would error as `test-e2e` command didn't exist.

## Motivation and Context
Being able to run e2e tests and see it in action is useful.

## How Has This Been Tested?
Run it locally.


package.json
```diff
@@ -14,7 +14,7 @@
     "test:e2e": "wp-scripts test-e2e",
     "test:e2e:debug": "wp-scripts --inspect-brk test-e2e --puppeteer-devtools",
     "test:e2e:help": "wp-scripts test-e2e --help",
-    "test:e2e:watch": "npm run test-e2e -- --watch",
+    "test:e2e:watch": "npm run test:e2e -- --puppeteer-interactive",
```
pauarge: The `test:e2e` change makes sense, but do we want to change `--watch` to `--puppeteer-interactive`?

pauarge (changes requested):

package.json
GaryJones: I wasn't sure what the `--puppeteer-interactive` bit did exactly as `--watch` did make it do what I wanted it to. I thought that maybe they were synonymous.

pauarge: I think `--watch` changes the behavior of Jest so the tests run continuously every time you change a file. You'd get a screen like this:

Screen Shot 2021-11-02 at 10 20 08 AM

`--puppeteer-interactive` opens a Chromium window so you can see whatever is going on the test. In my opinion we should keep `test:e2e:watch` to `--watch` and if it is useful to have an alias for `--puppeteer-interactive`, we could add that.

pauarge:
```suggestion
    "test:e2e:watch": "npm run test:e2e -- --watch",
```

pauarge (approved):

PR #478, opened 2021-11-04
Move tracker logic to separate file

pauarge:
## Description

This PR moves the code responsible of inserting the tracking Javascript code from the `class-parsely.php` file to its own `class-tracker.php` file. Tests have been updated accordingly.

## Motivation and Context

See #96.

## How Has This Been Tested?

Tracking code (JS files on front-end) are correctly rendered.

## Types of changes


src/class-parsely.php
```diff
@@ -16,8 +16,6 @@
 /**
  * Holds most of the logic for the plugin.
  *
- * @internal This needs splitting up in the future.
```
GaryJones, line 19: There's still a lot of items that could be removed from this class; I'd like to think we could scrap the class completely. The methods left are not cohesive.

src/class-tracker.php
```diff
@@ -0,0 +1,201 @@
...
+	 * @param string $handle The script's registered handle.
+	 * @param string $src    Unused? The script's source URL.
+	 * @return string Amended `script` tag.
+	 */
+	public function script_loader_tag( string $tag, string $handle, string $src ): string {
+		$parsely_options = $this->parsely->get_options();
+		if ( in_array(
+			$handle,
+			array(
+				'wp-parsely',
```
GaryJones: Do we even still have a script registered with this handle? I don't think we do.

GaryJones (commented):
Now this is separated out, I wonder if:

 - `register_js` can be called `register_scripts`; no need for abbreviations, and matches up with the `wp_register_script()` function call name.
 - `load_js_*` can be called `enqueue_*`, so it matches to the language of the `wp_enqueue_script()` function call name.
 - Whether it makes sense to register the API script in a class separately to the Tracker (or if it makes sense to keep them bundled, rename the file to `class-scripts.php` or similar).
 - Whether a slightly different arrangement of files or methods are needed, because the presence of `wp-parsely-recommended-widget` in this class's `script_loader_tag()` (when the rest of the registering and enqueuing of that script is handled elsewhere) seems like a code smell.

src/class-parsely.php
pauarge: I agree with you that we need to keep iterating on the file and we might even end up removing it. However, I think it's best to track this intent in GitHub issues (like #96).

pauarge:
@GaryJones I agree with all your comments above. However, I tried to not change the code other than the strictly necessary parts for moving it. I think it's better if we apply the changes on a separate PR to better isolate changes.

pauarge:
I've opened a draft PR with the changes (I will complete it once we merge this one): https://github.com/Parsely/wp-parsely/pull/481 @GaryJones

src/class-scripts.php
```diff
@@ -0,0 +1,201 @@
...
+	public function register_js(): void {
+		$parsely_options = $this->parsely->get_options();
+
+		if ( $this->parsely->api_key_is_missing() ) {
+			return;
+		}
+
+		wp_register_script(
+			'wp-parsely-tracker',
+			'https://cdn.parsely.com/keys/' . $parsely_options['apikey'] . '/p.js',
```
sjinks: Are we sure that the API key will always be valid (meaning it won't contain any "forbidden" characters)? I am a bit paranoid about all kinds of injection attacks :-)

src/class-scripts.php
```diff
@@ -0,0 +1,201 @@
...
+		 *
+		 * @param bool $display True if the JavaScript file should be included. False if not.
+		 */
+		if ( ! apply_filters( 'wp_parsely_load_js_tracker', $display ) ) {
+			return;
+		}
+
+		if ( ! has_filter( 'script_loader_tag', array( $this, 'script_loader_tag' ) ) ) {
+			add_filter( 'script_loader_tag', array( $this, 'script_loader_tag' ), 10, 3 );
+		}
```
sjinks, line 132: I have noticed that this construction is used in at least two places… Does it make sense to extract it into a separate helper method?

src/class-scripts.php
pauarge: The parameter should always be clean, since it's validated in `validate_options` in `class-settings-page.php`. However, I've added an additional step to be sure that we're not enqueuing something bad (I agree with you that this is critical).

src/class-scripts.php
pauarge: Similar to the reply to @GaryJones above, I will do this on a separate PR, but it does make sense.

sjinks (approved):

src/class-scripts.php
```diff
@@ -0,0 +1,205 @@
...
+			 * Filter whether to include the CloudFlare Rocket Loader attribute (`data-cfasync=false`) in the script.
+			 * Only needed if the site being served is behind Cloudflare. Should return false otherwise.
+			 * https://support.cloudflare.com/hc/en-us/articles/200169436-How-can-I-have-Rocket-Loader-ignore-specific-JavaScripts
+			 *
+			 * @since 3.0.0
+			 *
+			 * @param bool $enabled True if enabled, false if not.
+			 * @param string $handle The script's registered handle.
+			 */
+			if ( apply_filters( 'wp_parsely_enable_cfasync_attribute', false, $handle ) ) {
```
hbbtstar, line 190: Not worth blocking on, but worth noting that for a lot of Wordpress users they have no / limited dev resources: it might be worth adding a setting in the UI for this as well as having it in a filter, so non-developer clients have an option for fixing this too.

hbbtstar (approved):
Looks good to me! :shipit:

src/class-scripts.php
pauarge: I see your point, but for this specific issue I'd say it's OK to have a filter. Having a UI option requires a database access and more complex logic, for a feature that we'll probably want to have disabled anyway. 

Given that Cloudflare is a developer-centric feature, I'd leave it as is.

PR #482, opened 2021-11-08
Use checkboxes for multiple selects in settings page

### LINKED ISSUE #196: Settings:  Improve select UI for post and page types
Currently:
Screen Shot 2021-04-09 at 1 31 37 PM

A few ideas:
1. Combine into a single "post type" area, we might not need two
2. Remove post types that we never want to track (revisions, wp_block, nav-menu-items, etc)
3. Instead of a `ctrl-click` way to select, look at maybe just Checkboxes or Select2/SelectWoo or another library

pauarge:
## Description

This PR changes the UI for selecting post and page types to track, from `<select multiple>` to multiple checkboxes. The changes leave the data structure intact, it's purely a UI change.

This is a first iteration of this new UI, which is clearer for users and more accessible. We'll be working in the future on handling edge cases (such as having big lists of post and page types).

## Motivation and Context

Closes https://github.com/Parsely/wp-parsely/issues/196

## How Has This Been Tested?

Go to the plugin's settings. Select/unselect some post and page types to track and press _save_. Check that the changes stick.

**Before**

Screen Shot 2021-11-08 at 4 07 24 PM

**After**

Screen Shot 2021-11-08 at 7 29 28 PM

## Types of changes


GaryJones:
- Can we hide non-public / queryable post types, like revision, custom_css,  customize_changeset, oembed_cache, wp_block, wp_template, etc?
- Can we use JS (and back-end validation) to disable checkboxes when it has already been marked in the other type?
- Alternatively, I wonder whether radio buttons would help enforce this better (shown as a table here, but really fieldset rows, one row for each viable post_type) since what we really want is for each type to be assigned to exactly one of either `post`, `page` or `none`:

|   | None | Post |  Page |
| ------------- | ------------- | ------------- | ------------- |
| post  | o | ⦻ | o |
| page  | o | o | ⦻ |
| attachment  | ⦻ | o | o |

pauarge:
@GaryJones I've addressed your first point since it's a quick win. 

Regarding the second one, it makes sense but I reckon that it will involve a bit of additional code, so I'd rather have it separately. I've created an issue for it: https://github.com/Parsely/wp-parsely/issues/483

src/UI/class-settings-page.php
```diff
@@ -757,6 +739,38 @@ public function print_binary_radio_tag( array $args ): void {
...
+		if ( isset( $args['help_text'] ) ) {
+			echo '<div class="parsely-form-controls" data-has-help-text="true">';
+		}
+		if ( isset( $args['requires_recrawl'] ) && true === $args['requires_recrawl'] ) {
+			echo '<div class="parsely-form-controls" data-requires-recrawl="true">';
+		}
+
+		foreach ( $select_options as $key => $val ) {
+			$selected = in_array( $val, $options[ $args['option_key'] ], true );
+			echo sprintf( "<p><input type='checkbox' name='%s[]' id='%s[]' value='%s' ", esc_attr( $name ), esc_attr( $name ), esc_attr( $key ) );
```
sjinks: The `id` attribute probably does not need square brackets

```suggestion
			echo sprintf( "<p><input type='checkbox' name='%s[]' id='%s' value='%s' ", esc_attr( $name ), esc_attr( $name ), esc_attr( $key ) );
```

src/UI/class-settings-page.php
```diff
@@ -757,6 +739,38 @@ public function print_binary_radio_tag( array $args ): void {
...
+			echo checked( true === $selected, true, false );
+			echo sprintf( " /> <label for='%s_%s'>%s</label></p>", esc_attr( $id ), esc_attr( $key ), esc_html( $val ) );
```
sjinks: I am a bit confused with this:
  * on line 763: `id='%s[]'` => `esc_attr( $name )`
  * here: the `for` attribute does not match the `input`'s ID or am I getting this wrong?

src/UI/class-settings-page.php
```diff
@@ -757,6 +739,38 @@ public function print_binary_radio_tag( array $args ): void {
...
+		}
+
+		if ( isset( $args['help_text'] ) ) {
+			echo '<div class="help-text"><p class="description">' . esc_html( $args['help_text'] ) . '</p></div>';
+		}
+		echo '</div>';
```
sjinks, line 771: If `$args['help_text']` is not set, or `args['requires_recrawl']` is empty, I suspect that this closing `</div>` might break the layout…

src/UI/class-settings-page.php
```diff
@@ -757,6 +739,38 @@ public function print_binary_radio_tag( array $args ): void {
...
+	public function print_multiple_checkboxes( array $args ): void {
+		$options        = $this->parsely->get_options();
+		$select_options = $args['select_options'];
+		$id             = esc_attr( $args['option_key'] );
+		$name           = Parsely::OPTIONS_KEY . "[$id]";
+
+		if ( isset( $args['help_text'] ) ) {
+			echo '<div class="parsely-form-controls" data-has-help-text="true">';
+		}
+		if ( isset( $args['requires_recrawl'] ) && true === $args['requires_recrawl'] ) {
```
sjinks, line 757: Could we have a situation when both parameters are set? If so, I guess we will have more `div`'s than necessary…

Line 771 has an unconditional closing `</div>`, so maybe we need to refactor this in something like this:

```php
$attributes = [];
if ( isset( $args['help_text'] ) ) {
    $attributes[] = 'data-has-help-text="true"';
}

if ( isset( $args['requires_recrawl'] ) && true === $args['requires_recrawl'] ) {
    $attributes[] = 'data-requires-recrawl="true"';
}

$attributes = $attributes ? ' ' . join( ' ', $attributes ) : '';
echo '<div class="parsely-form-controls"', $attributes, '>';
```

?

sjinks (changes requested):

src/UI/class-settings-page.php
pauarge: You got it right, I've now changed the `for` attribute.

src/UI/class-settings-page.php
pauarge: That's a good catch, and actually a problem with the rest of the fields in the class. You also point out a similar issue in your other comment. 

I think this is worth a wider refactor, I'm thinking using helper methods in the class (notice that the logic of displaying `help_text` is repeated multiple times). 

I've opened an issue to fix that: https://github.com/Parsely/wp-parsely/issues/485

sjinks (approved):

src/UI/class-settings-page.php
pauarge: This will be addressed in the issue I mentioned above.

hbbtstar (approved):
This looks good to me! :shipit:

PR #486, opened 2021-11-09
Making class members private

### LINKED ISSUE #428: Declare visibility for class-level constants
A class should not expose its contents, including constants, unless necessary. Right now, all of our class constants are public and available for other code to read.

Now, with PHP 7.1 as the minimum, we can default to a "non-public by default" approach, and change the visibility of constants as needed.

We could leave these all public (and some will no doubt stay public anyway), but then we're not following best practices regarding encapsulation.

## Additional context
One approach would be to set them all to `private` by default, and then run tests / the plugin, and see what breaks, and iteratively give some constants increased visibility as needed.

pauarge:
## Description

This PR makes some class constants and methods that were previously public, private. They were only being accessed from within the class, so we are now preventing incorrect uses of the plugin.

## Motivation and Context

Closes https://github.com/Parsely/wp-parsely/issues/428

## How Has This Been Tested?

The changed classes work as expected and pass the tests.

## Types of changes


sjinks (approved):

tests/Integration/Integrations/FacebookInstantArticlesTest.php
```diff
@@ -17,7 +17,9 @@
  * Test Facebook Instant Articles integration.
  */
 final class FacebookInstantArticlesTest extends TestCase {
-	public const REGISTRY_KEY = 'parsely-analytics-for-wordpress';
+	// This values should mirror the values in `class-facebook-instant-articles.php`.
+	private const REGISTRY_IDENTIFIER   = 'parsely-analytics-for-wordpress';
+	private const REGISTRY_DISPLAY_NAME = 'Parse.ly Analytics';
```
GaryJones: > This values should mirror the values in `class-facebook-instant-articles.php`

This is a code smell.

If no other application logic needs to access the constant, then let's make them private in `class-facebook-instant-articles.php`, scrap these test class constants, and use Reflection to grab the constants from the class under test instead.

- https://www.php.net/manual/en/reflectionclass.getreflectionconstants.php
- https://www.php.net/manual/en/reflectionclass.getconstants.php

GaryJones (changes requested):
I agree with making the class constant private if no code other than test files need it, but I also think we can improve how those constants are accessed within tests without having to duplicate constants in the test class as well.

tests/Integration/Integrations/FacebookInstantArticlesTest.php
```diff
@@ -17,35 +18,63 @@
...
+	 * Internal variable.
+	 *
+	 * @var Facebook_Instant_Articles $fbia Holds the Facebook_Instant_Articles object.
+	 */
+	private static $fbia;
+
+	/**
+	 * Internal variable.
+	 *
+	 * @var string $registry_identifier Hols the same value as the private constant in the class.
```
GaryJones:
```suggestion
	 * @var string $registry_identifier Holds the same value as the private constant in the class.
```

tests/Integration/Integrations/FacebookInstantArticlesTest.php
```diff
@@ -17,35 +18,63 @@
...
+	 */
+	private static $registry_display_name;
+
+	/**
+	 * The setUp run before each test
+	 */
+	public function set_up(): void {
+		parent::set_up();
+
+		self::$fbia = new Facebook_Instant_Articles();
```
GaryJones, line 48: Be careful with instantiating the class under test in a set_up() method. If the class under test has properties that are set in one test, then it can inadvertently effect the behaviour of other tests.

GaryJones (approved):
Approved, but note the comments.

Issue #487, opened 2021-11-11
Add support for decoupled WordPress setups


pauarge:
## Is your feature request related to a problem?

With a growing demand for decoupled WordPress architectures it's a good time to think about the support that the plugin provides to decoupled architectures.

The idea of the wp-parsely plugin is to provide its users a simple way to integrate Parse.ly in a WordPress website so they can quickly start getting data into the Parse.ly Dashboard. More specifically, the plugin inserts two things: 

- **Parse.ly tracking script.** Independent of the content of the individual page, its contents rarely change. It is the most basic thing needed to perform tracking from our side.

```
<script id='wp-parsely-api-js-extra'>
var wpParsely = {"apikey":"vip-prod-pauarge.go-vip.net"};
</script>
 
<script data-cfasync="false" data-parsely-site="vip-prod-pauarge.go-vip.net" src='https://cdn.parsely.com/keys/vip-prod-pauarge.go-vip.net/p.js?ver=2.6.1' id="parsely-cfg"></script>
```

- **Page metadata**. Not strictly necessary, this data is used by the Parse.ly crawler to offer rich insights on the dashboard. It is typically the thing in which customers who don’t use WordPress/the plugin have the most issues.

```
<!-- BEGIN Parse.ly 2.6.1 -->
<meta name="parsely-title" content="Before Astro, these were the robots people invited home" />
<meta name="parsely-link" content="http://vip-prod-pauarge.go-vip.net/blog/2021/10/01/before-astro-these-were-the-robots-people-invited-home/" />
<meta name="parsely-type" content="post" />
<meta name="parsely-pub-date" content="2021-10-01T11:01:07Z" />
<meta name="parsely-section" content="Uncategorized" />
<meta name="parsely-author" content="pauarge" />
<!-- END Parse.ly -->
```

It is quite straight forward to implement this additions as a WordPress plugin in a monolith site. What we are doing is hooking into `wp_head` and `wp_enqueue_script` to render these bits in the WordPress controlled front-end. Another advantage of the monolithical architecture is that developers can hook and load conditionally or modify the values of those pieces, using the provided filters. 

## Describe the solution you'd like

In a decoupled architecture, we cannot hook nor insert HTML in a front-end that we don’t control from the plugin’s perspective. For that reason, we cannot seamlessly enable Parse.ly on a decoupled site just with the plugin. Adding Parse.ly will inevitably involve developer work on the front-end, but that’s something that the site owner signed up for when they chose decoupled.

[...]

PR #489, opened 2021-11-12
Adding decoupled support

pauarge:
## Description

This is the initial iteration in order to add support for decoupled WordPress architectures in the plugin. In this PR, we're extending the `post` and `page` models in the WordPress REST API to include the `parsely-meta` field. This field returns a JSON object containing the meta tags using the same function that's being used to generate the data in the front-end, `construct_parsely_metadata`.

This behavior is controllable by the `wp_parsely_enable_rest_api_support` filter.

## Motivation and Context

See https://github.com/Parsely/wp-parsely/issues/487

## How Has This Been Tested?

In a WordPress installation with wp-parsely, navigate to `?rest_route=/wp/v2/posts/1` and you should expect to see the `parsely-meta` object.

Screen Shot 2021-11-18 at 4 29 46 PM

## Types of changes


sjinks (dismissed):

GaryJones (approved):
Works equally well for page:

Screenshot 2021-11-25 at 11 42 31

✅ 

Any reason this couldn't be pushed to 3.0? New useful feature for users (developers).

GaryJones (changes requested):
See https://github.com/Parsely/wp-parsely/pull/500 for some discussion about a different format.

src/class-rest.php
```diff
@@ -0,0 +1,78 @@
...
+		$callback = function( array $object ): array {
+			$post_id = $object['id'];
+			$options = $this->parsely->get_options();
+			$post    = WP_Post::get_instance( $post_id );
+			return $this->parsely->construct_parsely_metadata( $options, $post );
+		};
+
+		$args = array( 'get_callback' => $callback );
+		register_rest_field( 'post', self::PARSELY_META_REST_FIELD_NAME, $args );
+		register_rest_field( 'page', self::PARSELY_META_REST_FIELD_NAME, $args );
```
chriszarate: Could we:

1. Target all public post types that support `editor`? This would allow easy custom post type support.
2. Allow the post types to be filtered before registering the field?

chriszarate:
@pauarge I've created a draft PR of our decoupled plugin bundle against this branch, which extracts the data so that it can be exposed by GraphQL:

https://github.com/Automattic/vip-decoupled-bundle/pull/36

In order to extract the data we need, we must use `construct_parsely_metadata` from the `$parsely` global. Is this stable / safe?

chriszarate:
@pauarge Would you be open to conditionally registering WPGraphQL fields here as well? Lots of plugins take this approach, behind a simple `function_exists` check.

pauarge:
Thanks for your comments @chriszarate !

> Target all public post types that support editor? This would allow easy custom post type support.
    Allow the post types to be filtered before registering the field?

I've updated the code to use `track_post_types` and `track_page_types` from the plugin's settings in wp-admin. This way we let the user customize whatever they want to track and match the behavior of the WP FE code insertion.

> GraphQL

We've had some discussions with @GaryJones regarding the addition of GraphQL to the plugin. We thought about doing so in the next version. From our side, we're more than happy to integrate the code that you made to the main plugin.

> In order to extract the data we need, we must use construct_parsely_metadata from the $parsely global. Is this stable / safe?

Yes

src/class-rest.php
```diff
@@ -18,7 +18,8 @@
  * @since 3.1.0
  */
 class Rest {
-	private const PARSELY_META_REST_FIELD_NAME = 'parsely-meta';
+	private const PARSELY_REST_VERSION         = '1.0.0';
+	private const PARSELY_META_REST_FIELD_NAME = 'parsely';
```
GaryJones: The `META` part of this constant name could be removed, as it's the name for the whole field.

src/class-rest.php
```diff
@@ -68,11 +69,18 @@ public function register_parsely_meta(): void {
...
+				'version' => self::PARSELY_REST_VERSION,
+				'meta'    => $meta,
+			);
 		};
 
-		$args = array( 'get_callback' => $callback );
-		register_rest_field( 'post', self::PARSELY_META_REST_FIELD_NAME, $args );
-		register_rest_field( 'page', self::PARSELY_META_REST_FIELD_NAME, $args );
+		$options     = $this->parsely->get_options();
+		$object_type = array_merge( $options['track_post_types'], $options['track_page_types'] );
```
GaryJones: We could probably do with an `array_unique()` here, in case someone has registered the same type for both post and page (we don't block the checkboxes yet).

Also, should there be a filter here?

GaryJones (changes requested):

src/class-rest.php
pauarge: > We could probably do with an array_unique() here

Agreed, adding in a commit.

> Also, should there be a filter here?

We could, I imagine that you are suggesting this in case a user wants logic more complex than ON/OFF, which the checkboxes already provide? I would add the filter in a further iteration to keep it simple at the beginning.

GaryJones: More that I'm thinking an engineer may want to guarantee what REST API outputs are affected, regardless of what some colleague may fiddle with the checkboxes are accidentally break things.

src/class-rest.php
```diff
@@ -79,8 +79,17 @@ public function register_parsely_meta(): void {
...
-		$args        = array( 'get_callback' => $callback );
 
+		/**
+		 * Filters the list of author object types that the Parse.ly REST API is hooked into.
+		 *
+		 * @since 3.1.0
+		 *
+		 * @param string[] $object_type Array of strings containing the object types, i.e. `page`, `post`, `term`.
+		 */
+		$object_type = apply_filters( 'wp_parsely_rest_object_type', $object_type );
```
GaryJones: Let's make `$object_type` and the filter name plural, to indicate that we're expected an array or more than one type (at least by default).

pauarge: I did use singular because `register_rest_field` uses singular. Then again, in that case it can be a string or an array, in our case, if should always be an array.

chriszarate:
> We've had some discussions with @GaryJones regarding the addition of GraphQL to the plugin. We thought about doing so in the next version. From our side, we're more than happy to integrate the code that you made to the main plugin.

Sounds good. I'll work on formalizing our approach this week.

src/class-rest.php
```diff
@@ -0,0 +1,95 @@
...
+		$object_types = array_unique( array_merge( $options['track_post_types'], $options['track_page_types'] ) );
+
+		/**
+		 * Filters the list of author object types that the Parse.ly REST API is hooked into.
+		 *
+		 * @since 3.1.0
+		 *
+		 * @param string[] $object_types Array of strings containing the object types, i.e. `page`, `post`, `term`.
+		 */
+		$object_types = apply_filters( 'wp_parsely_rest_object_types', $object_types );
```
GaryJones, line 78: We need a test to demonstrate how this filter should and shouldn't be used.

sjinks (dismissed):

tests/Integration/RestTest.php
```diff
@@ -0,0 +1,106 @@
...
+	 * Test that the REST fields are registered to WordPress REST API.
+	 *
+	 * @covers \Parsely\Rest\register_parsely_meta
+	 */
+	public function test_register_parsely_meta_registers_fields() {
+		global $wp_rest_additional_fields;
+
+		self::$rest->register_parsely_meta();
+
+		$expected_fields = array( 'get_callback', 'update_callback', 'schema' );
```
GaryJones: Instead of checking that the array_keys of `$wp_rest_additional_fields['...']['parsely']` are these three values (which, they [always will be](https://github.com/WordPress/WordPress/blob/47c4ff7102291136473d1ae52302dd6ebfad3345/wp-includes/rest-api.php#L137) - just with different callback values, with a default of `null`), why not check that the parsely property exists for that post type in the global, and that the callback is not the default `null`?

```php
isset( $wp_rest_additional_fields['post']['parsely'] ) && ! is_null( $wp_rest_additional_fields['post']['parsely']['get_callback']
```

There's also an opportunity to check that `['get_callback']` is populated correctly - an array, with a `version` key being a string, and `meta` key with an array which is at least going to contain some expected fields for particular post types (post vs non-post).

tests/Integration/RestTest.php
```diff
@@ -0,0 +1,106 @@
...
+			function( $object_types ) {
+				return array( 'term' );
+			}
+		);
+
+		self::$rest->register_parsely_meta();
+
+		$expected_fields = array( 'get_callback', 'update_callback', 'schema' );
+
+		$this->assertEquals( $expected_fields, array_keys( $wp_rest_additional_fields['term']['parsely'] ) );
```
GaryJones: As well as changing what is tested for the `term`, I would also suggest checking for a lack of changes (`get_callback` is null), for a `post` or `page` type that would ordinarily get the changes.

GaryJones (changes requested):

tests/Integration/RestTest.php
```diff
@@ -0,0 +1,125 @@
...
+	 */
+	public function test_register_parsely_meta_registers_fields() {
+		global $wp_rest_additional_fields;
+
+		self::$rest->register_parsely_meta();
+
+		$expected_fields = array( 'get_callback', 'update_callback', 'schema' );
+
+		$this->assertEquals( $expected_fields, array_keys( $wp_rest_additional_fields['post']['parsely'] ) );
+		$this->assertEquals( $expected_fields, array_keys( $wp_rest_additional_fields['page']['parsely'] ) );
```
GaryJones: This is just testing core functionality of `register_rest_field()`. It serves no value for our tests.

src/class-rest.php
```diff
@@ -0,0 +1,95 @@
...
+	/**
+	 * Registers the `parsely-meta` field in the REST API.
+	 *
+	 * @since 3.1.0
+	 *
+	 * @return void
+	 */
+	public function register_parsely_meta(): void {
+		$callback = function( array $object ): array {
+			$post_id = $object['id'];
```
jblz: Should we check the status against [`Parsely::post_has_trackable_status`](https://github.com/Parsely/wp-parsely/blob/develop/src/class-parsely.php#L273) & return early if false?

src/class-rest.php
```diff
@@ -0,0 +1,95 @@
...
+
+	/**
+	 * Registers the `parsely-meta` field in the REST API.
+	 *
+	 * @since 3.1.0
+	 *
+	 * @return void
+	 */
+	public function register_parsely_meta(): void {
+		$callback = function( array $object ): array {
```
jblz: Is there a reason this shouldn't be a class member function instead of being declared inside this one?  If we pull it out, we can cover it with a unit test.

src/class-rest.php
pauarge: I'm not sure about this one. The checks already happen inside `construct_parsely_metadata` and I wanted to isolate as much logic as possible from the `Rest` class.

src/class-rest.php
```diff
@@ -66,7 +66,7 @@ public function run(): void {
 	 */
 	public function register_parsely_meta(): void {
 		$callback = function( array $object ): array {
-			$post_id = $object['id'];
+			$post_id = $object['ID'] ?? $object['id'];
```
GaryJones: I was finding some tests issues as `id` could not be found as a key (but `ID` could.

tests/Integration/RestTest.php
```diff
@@ -107,16 +119,19 @@ function( $object_types ) {
...
 		self::assertArrayHasKey( $post_type, $wp_rest_additional_fields );
 		self::assertArrayHasKey( 'parsely', $wp_rest_additional_fields[ $post_type ] );
 		self::assertIsArray( $wp_rest_additional_fields[ $post_type ]['parsely'] );
-		self::assertArrayHasKey( 'version', $wp_rest_additional_fields[ $post_type ]['parsely']['get_callback'] );
+		/*
+		self::assertArrayHasKey( 'version', $wp_rest_additional_fields[ $post_type ]['parsely']['get_callback']( $object ) );
 		self::assertIsString( $wp_rest_additional_fields[ $post_type ]['parsely']['get_callback']['version'] );
-		self::assertArrayHasKey( 'meta', $wp_rest_additional_fields[ $post_type ]['parsely']['get_callback'] );
+		self::assertArrayHasKey( 'meta', $wp_rest_additional_fields[ $post_type ]['parsely']['get_callback']( $object ) );
+		*/
```
GaryJones: With these assertions commented out, the tests should pass.

I wasn't able to get the callback to be correctly called with the right array of the WP_Post / WP_Term data, despite creating the post/page/term (category) in the main test methods.

GaryJones (changes requested):
Needs fixes, but sharing the latest work.

src/class-rest.php
jblz: Good call.

GaryJones:
Great! The `$object` arg of `assertParselyRestFieldIsConstructedCorrectly()` can be removed again, as can the code in each test method that existed to populate that arg.

tests/Integration/RestTest.php
```diff
@@ -115,7 +115,7 @@ function( $object_types ) {
 		$category = self::factory()->category->create( array( 'name' => 'Test Category' ) );
 		self::factory()->post->create( array( 'post_category' => array( $category ) ) );
 		$term = get_term( $category, 'post_category', 'ARRAY_A' );
```
GaryJones: This code can go.

tests/Integration/RestTest.php
```diff
@@ -77,15 +77,15 @@ public function test_register_parsely_meta_registers_fields(): void {
...
 		$post    = get_post( $post_id, 'ARRAY_A' );
-		$this->assertParselyRestFieldIsConstructedCorrectly( 'post', $post, $wp_rest_additional_fields );
+		$this->assertParselyRestFieldIsConstructedCorrectly( 'post', $wp_rest_additional_fields );
 
 		$page_id = self::factory()->post->create(
 			array(
 				'post_type' => 'page',
 			)
 		);
 		$page    = get_post( $page_id, 'ARRAY_A' );
```
GaryJones: This code can go.

tests/Integration/RestTest.php
```diff
@@ -77,15 +77,15 @@ public function test_register_parsely_meta_registers_fields(): void {
 
 		$post_id = self::factory()->post->create();
 		$post    = get_post( $post_id, 'ARRAY_A' );
```
GaryJones: This code can go.

GaryJones (changes requested):

tests/Integration/RestTest.php
```diff
@@ -0,0 +1,144 @@
...
+	 * Test that the REST fields are can be modified using the `wp_parsely_rest_object_types` filter.
+	 *
+	 * @covers \Parsely\Rest::register_parsely_meta
+	 */
+	public function test_register_parsely_meta_with_filter(): void {
+		global $wp_rest_additional_fields;
+
+		add_filter(
+			'wp_parsely_rest_object_types',
+			function( $object_types ) {
```
jblz:
```suggestion
			function() {
```

per: https://sonarcloud.io/project/issues?id=Parsely_wp-parsely&issues=AX2ftYPDuw8Yrhl6Lsoo&open=AX2ftYPDuw8Yrhl6Lsoo&pullRequest=489

src/class-rest.php
```diff
@@ -0,0 +1,103 @@
...
+namespace Parsely;
+
+use WP_Post;
+
+/**
+ * Injects Parse.ly Metadata to WordPress REST API
+ *
+ * @since 3.1.0
+ */
+class Rest {
```
jblz, line 20:
```suggestion
class Rest_Post_Metadata {
```

Given the high likelihood we'll empower additional REST API functionality with this plugin, I suggest we be more specific with class naming.

src/class-rest.php
```diff
@@ -0,0 +1,103 @@
...
+	private const PARSELY_REST_VERSION    = '1.0.0';
+	private const PARSELY_REST_FIELD_NAME = 'parsely';
```
jblz:
```suggestion
	private const REST_VERSION    = '1.0.0';
	private const REST_FIELD_NAME = 'parsely';
```

If we're more specific about what this class is doing, I don't think we need these prefixes.

src/class-rest.php
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	/**
+	 * Registers the `parsely-meta` field in the REST API.
+	 *
+	 * @since 3.1.0
+	 *
+	 * @return void
+	 */
+	public function register_parsely_meta(): void {
```
jblz:
```suggestion
	public function register_meta(): void {
```

src/class-rest.php
```diff
@@ -0,0 +1,103 @@
...
+		 *
+		 * @param bool $enabled True if enabled, false if not.
+		 */
+		if ( apply_filters( 'wp_parsely_enable_rest_api_support', true ) ) {
+			add_action( 'rest_api_init', array( $this, 'register_parsely_meta' ) );
+		}
+	}
+
+	/**
+	 * Registers the `parsely-meta` field in the REST API.
```
jblz:
```suggestion
	 * Registers the meta field on the appropriate resource types in the REST API.
```

This would need to be corrected to the proper value anyhow, but since we're sourcing it from a class constant, I think it's appropriate to leave the exact field name out of the doc block.

src/class-rest.php
```diff
@@ -0,0 +1,103 @@
...
+
+	/**
+	 * Function to get hooked into the `get_callback` property of the `parsely` REST API field. It generates
+	 * the `parsely` object in the REST API.
+	 *
+	 * @param array $object The WordPress object to extract to render the metadata for, usually a post or a page.
+	 * @return array The `parsely` object to be rendered in the REST API. Contains a version number describing the
+	 * response and the `meta` object containing the actual metadata.
+	 */
+	public function parsely_rest_get_callback( array $object ): array {
```
jblz: I'd just call this `<CLASSNAME>::get_callback`, so it can be more easily ported to additional integrations.  The context is in the namespace and class name.

```suggestion
	public function get_callback( array $object ): array {
```

jblz (changes requested):
This is working really well.  I left some comments about symbol naming that I think we should address.

src/class-rest.php
pauarge: I don't necessarily agree with this change. We will for sure have more REST functionality, which will be spawned from a base `Rest` class. That `Rest` class will then instantiate `Rest_Post_Metadata` or whatever other features we want to implement.

At this point, we don't have other features, so I feel separating those classes is overengineering. When we do implement a second feature, I agree that we should move the specifics of the page/post metadata to its own separate class.

jblz (approved):
Nice work <3

sjinks (approved):

GaryJones (approved):

PR #494, opened 2021-11-17
Remove unnecessary npm step in e2e

pauarge:
## Description

E2E tests are a bit flaky and sometimes we have to re-run them, which might be time consuming. While working in the solution, it is better for us to have the test take shorter to fail and to re-run.

This PR removes a step in the GitHub Action that runs the e2e tests: an `npm` update. This step is redundant because the GitHub Actions runner already provides a sufficiently up-to-date version of `npm`. 

## Motivation and Context

Make E2E pipeline quicker to run.

## How Has This Been Tested?

E2E tests keep working as they did.

## Types of changes


GaryJones (approved):
✅

.github/workflows/e2e-tests.yml
```diff
@@ -28,9 +28,6 @@ jobs:
           node-version: lts/*
           cache: 'npm'
 
-      - name: Update npm to latest
-        run: npm i -g npm
```
sjinks, line 32: Just in case this might affect tests — npm@6 and npm@7 handle `peerDependencies` differently; with npm@6 you have to install them manually; with `npm@7+`, they are installed automatically.

sjinks (approved):

.github/workflows/e2e-tests.yml
pauarge: Thanks for pointing that out. We're using the `lts` node version, which in this case is 16 and includes npm 7+.

.github/workflows/e2e-tests.yml
```diff
@@ -1,14 +1,8 @@
 name: End-to-end (e2e) Tests
 
 on:
-  # Run on all pushes and on all pull requests.
-  # Prevent the "push" build from running when there are only irrelevant changes.
-  push:
-    paths-ignore:
-      - '**.md'
```
sjinks, line 8: Just in case: `paths-ignore` is [valid](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#onpushpull_requestpaths) for `pull_request` as well

PR #496, opened 2021-11-18
Remove custom admin CSS

pauarge:
## Description

This PR removes the custom CSS code that was being `echo`-ed in the admin page. It is not ideal to echo code directly and we considered moving it to a separate CSS file. However, given the low value added by that CSS, we have decided to remove it completely and rely on WordPress's default styling.

## Motivation and Context

Code cleanup.

## How Has This Been Tested?

The UI in the admin settings page is correctly rendered.

**Before**

Screen Shot 2021-11-18 at 5 00 44 PM

Screen Shot 2021-11-18 at 5 00 53 PM

**After**

Screen Shot 2021-11-18 at 4 59 53 PM

Screen Shot 2021-11-18 at 5 00 16 PM

## Types of changes

### DIFF
--- a/src/UI/class-settings-page.php
+++ b/src/UI/class-settings-page.php
@@ -55,14 +55,6 @@ final class Settings_Page {
 	 * @return void
 	 */
 	public function add_admin_header(): void {
-		// TODO: Extract style to CSS file.
-		echo '
-<style>
-#wp-parsely_version { color: #777; font-size: 12px; margin-left: 1em; }
-.help-text { width: 75%; }
-</style>
-';
-
 		$admin_script_asset = require plugin_dir_path( PARSELY_FILE ) . 'build/admin-page.asset.php';
 		wp_enqueue_script(
 			'wp-parsely-admin',
--- a/views/parsely-settings.php
+++ b/views/parsely-settings.php
@@ -18,7 +18,8 @@ namespace Parsely;
 $parsely_version_string = sprintf( __( 'Version %s', 'wp-parsely' ), Parsely::VERSION );
 ?>
 <div class="wrap">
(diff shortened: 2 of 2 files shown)


sjinks (approved):

GaryJones:
I'm not necessarily a fan of this change, but I can disagree and commit. 

If the styles for the help-text class are being removed, can we remove the `<div class="help-text">` elements that wrap the description texts?

sjinks:
If I am not mistaken, .help-text comes from WP Admin CSS, probably it is worth double-checking...

pauarge:
@sjinks We're adding that class in a few places in `class-settings-page.php`: https://github.com/Parsely/wp-parsely/blob/d869beeb3e6b4fa76daf47ca82e240d4f147c88e/src/UI/class-settings-page.php#L656

GaryJones (changes requested):
WP core has:

```css
.form-wrap p, p.description {
    margin: 2px 0 5px;
    color: #646970;
}
.form-wrap p, p.description, p.help, span.description {
    font-size: 13px;
}
```

...so it handles the look of the description text already. All the .help-text did was to narrow the width of the text, but if we look at the core Settings page, then WP lets it run wide as well:
Screenshot of WP admin settings with long description text

Since the `.help-text` div's now serve no purpose, I'd like to see them removed from the output.

----

For the version string, I'd prefer to see it left as a `span`, since it still has the same impact is it does now, even with the CSS removed:

With CSS:
light grey version text to the right of the header
With no CSS:
less light grey version text, just to the right of the header

GaryJones (approved):
Removing the `.help-text` has knock-on effects for the `Important!` message showing, so that would be best to tidy up in another PR. So marking this as ✅.

PR #500, opened 2021-11-22
Add parsely-meta-string field to REST API

### LINKED ISSUE #487: Add support for decoupled WordPress setups

pauarge:
To be merged after https://github.com/Parsely/wp-parsely/pull/489.

## Description

This PR adds an additional field to the REST API, `parsely-meta-string`. This field contains essentially the same data that `parsely-meta` has, but in this case the data is returned as an HTML string. This is intended to be inserted as-is in a decoupled front-end.

The field mirrors the behavior of the WordPress front-end, that is, an admin can select from the Parse.ly settings which format they want to get the information -- `repeated-metas` or `json-ld`.

## Motivation and Context

Closes https://github.com/Parsely/wp-parsely/issues/487.

## How Has This Been Tested?

In a WordPress installation with wp-parsely, navigate to `?rest_route=/wp/v2/posts/1` and you should expect to see the `parsely-meta` object.

**Repeated metas**

Screen Shot 2021-11-23 at 12 22 30 PM

**JSON-LD**

Screen Shot 2021-11-23 at 12 26 21 PM

## Types of changes


sjinks (dismissed):

GaryJones (commented):
Screenshot 2021-11-25 at 11 53 05

Looking at some of the existing fields, I wonder if following the same format would be a better fit? Maybe a:
- parsely
  - rendered (HTML string of JSON or `meta` elements)
  - meta (array of individual fields)
    - `@context`
    - `@type`
    - `headline`.
    - `url`
    - etc.

If not, then I think `parsely-meta-html` would be more accurate and descriptive than `parsely-meta-string`.

GaryJones:
Having a top-level `parsely` object would also allow us to have a `parsely.version` value, that was indicative of a version number of the parsely REST API output, since we don't have a versioned route to play with instead. This would allow consumers to check the version and handle different outputs accordingly.

pauarge:
I like your approach. I will update #489 and then will pull it to this (as this would be "the second part"). If you agree, here's what we'll do:

- `parsely` 
  - `version`. This could be the plugin version (not really like it because it leaks plugin info) or a hardcoded string corresponding to some kind of changes (i.e. we add a new field, we bump the version).
  - `meta`. The array of fields introduced in #489.
  - `rendered`. What's introduced in this PR, i.e. dynamically return rendered HTML with meta tags or JSON-LD 

@GaryJones

GaryJones:
`version` should definitely be the API version, not related to the plugin version.

+1 on your suggestion.

tests/Integration/RestTest.php
```diff
@@ -126,6 +149,62 @@ public function test_get_callback(): void {
...
+		$post_id = self::factory()->post->create();
+
+		// phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
+		$post = get_post( $post_id );
+		$date = gmdate( 'Y-m-d\TH:i:s\Z', get_post_time( 'U', true, $post ) );
+
+		$meta_string = self::$rest->get_rendered_meta();
+		$expected    = '
+<meta name="parsely-title" content="Post title 0000055" />
+<meta name="parsely-link" content="http://example.org/?p=23" />
```
GaryJones: This assumes that the tests are run in the same order, which they aren't. So the failure is a flaw in the test logic, not the active code.

The post title can be set when creating the post with the factory, or retrieved from `$post_id`.
The `?p=` value can use `$post_id`.

GaryJones (changes requested):
Should be a minor fix to get the tests passing.

tests/Integration/RestTest.php
GaryJones: Fixed in [this commit](https://github.com/Parsely/wp-parsely/pull/500/commits/504fc6d5b13cbbfbed5af225c8cfa6c96df68329).

GaryJones:
This needs some documentation to explain the REST support, then it's good to go.

README.md
```diff
@@ -49,6 +50,49 @@ The plugin requires an active Parse.ly account. Parse.ly gives creators, markete
...
+	}
+);
+```
+
+The `parsely` field contains the following fields:
+ - `version`, which is a string identifying the version of the REST API output; this will be updated if changes are made to the output, so consuming applications can check against it.
+ - `meta`, which is an array of metadata for the specific post, page or other object type.
+ - `rendered`, which is the rendered HTML of the metadata for the post, page or other object type. This will be a JSON-LD `<script>` tag, or a set of `<meta>` tags, depending on the format selected in the plugin settings. The decoupled code can consume and use this directly, instead of building the values from the `meta` field values.
+
+The `rendered` field is a convenience field is only included if the `wp_parsely_enable_rest_rendered_support` filter returns `true`. To disable this, return `false` from the filter.
```
jblz:
```suggestion
The `rendered` field is a convenience field containing the HTML-formatted meta data which can be printed to a decoupled page as is.

This can be disabled by returning `false` from the `wp_parsely_enable_rest_rendered_support` filter.
```

jblz (changes requested):
This is excellent!  I spotted one line in the documentation that needs a tweak and offered a suggestion & the `$post` inclusion in the `apply_filters` call, but otherwise this looks good to go.

src/class-rest.php
```diff
@@ -93,11 +93,34 @@ public function get_callback( array $object ): array {
...
 		);
+
+		/**
+		 * Filter whether REST API support in rendered string format is enabled or not.
+		 *
+		 * @since 3.1.0
+		 *
+		 * @param bool $enabled True if enabled, false if not.
+		 */
+		if ( apply_filters( 'wp_parsely_enable_rest_rendered_support', true ) ) {
```
jblz: Since we have already inflated the `WP_Post` instance, it'd be nice to include in the available params to filtering functions:

```suggestion
		if ( apply_filters( 'wp_parsely_enable_rest_rendered_support', true, $post ) ) {
```

jblz (changes requested):

GaryJones (approved):

jblz (approved):

PR #508, opened 2021-11-25
Deploy on published release

GaryJones:
As per https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#release, the workflow will only fire when we publish a release on GitHub. The `released` type will only fire for stable releases,  and not pre-release releases. 

This allows tagging and release of alpha and beta versions, without it automatically pushing to WPORG SVN which then result in sites pulling those in via plugin auto-updates.


pauarge (approved):

sjinks:
Not related to this PR, but this needs to be fixed:

https://github.com/Parsely/wp-parsely/pull/508/files#diff-28802fbf11c83a2eee09623fb192785e7ca92a3f40602a517c011b947a1822d3R13

```
      - name: Build # Remove or modify this step as needed
        run: |
          npm install
          npm run build
```

Consider using `npm ci` instead of `npm install` to make sure that you release what you have tested. [`npm ci`](https://docs.npmjs.com/cli/v8/commands/npm-ci) installs exactly the same versions as specified in `package-lock.json` (and makes sure that `package-lock.json` matches `package.json`); `npm install` in some cases can update transitive dependencies.

PR #510, opened 2021-11-26
Making sure the coauthors plugin returns a WP_User

pauarge:
## Description

The `get_author_name` now expects an `$author` parameter that has type `?WP_User`. This wasn't the case when the Co-Authors Plus plugin was enabled and a Guest Author was being assigned to a post. It would throw a fatal error about that object not being a `WP_User`.

This PR casts the objects to WP_User to make sure they are compliant to the function signature.

## Motivation and Context

Fix an issue for some customers in 3.0.

## How Has This Been Tested?

With the Co-Authors Plus plugin installed, create a new Guest Author (Users->Guest Authors) in wp-admin. Then when editing a post, see the Authors section of the Post attributes, and start typing the slug / name, and add them.

image

image

The front-end should throw a fatal error without the patch, should work fine and render all co-authors as meta tags.

## Types of changes

### DIFF
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -788,7 +788,7 @@ class Parsely {
 						$post_author   = $coauthors_plus->get_coauthor_by( 'user_nicename', $coauthor_slug );
 						// In case the user has been deleted while plugin was deactivated.
 						if ( ! empty( $post_author ) ) {
-							$coauthors[] = $post_author;
+							$coauthors[] = new WP_User( $post_author );
 						}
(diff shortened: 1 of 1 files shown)


GaryJones (approved):
✅  Agreed that this fixes the fatal error.

Would like to see more defensive code and tests added, and refactored out as a `Parsely\Integration` (like other integration-related code) for wp-parsely 3.1.

GaryJones (approved):
✅

PR #511, opened 2021-11-26
Tests: Run in random order

GaryJones:
## Description
Run Unit and Integration (WP) tests in a random order.

## Motivation and Context
This ensures that we don't have tests written in a way that are dependent on other tests having been run first, which may have left the environment in a presumptive state.

## How Has This Been Tested?
Run locally several times, including with a `--debug` flag that shows the actual order of tests (the results are always shown in the regular order). No test fails or errors.


pauarge (approved):
This is great!

PR #518, opened 2021-11-28
Settings Page UI improvements

### LINKED ISSUE #8: Clean up the settings page
We're more than a little bit text heavy on the settings page.  Much of the information is good, but we don't need it in everyone's faces all the time and some of the copy is probably too verbose.  We should migrate to using something like https://jqueryui.com/tooltip/ to show information on demand maybe.

GaryJones:
## Description

- Remove `help-text` and `parsely-form-controls` divs. Since there is no longer any custom CSS applied after #496, there is no need for these divs.
- Re-order settings. Groups those needing a recrawl together.
- Amend settings sections - from Required (one field) and Optional, to Basic, Requires Recrawl, and Advanced.
- Move repeated "Important!" message from each setting that would require a recrawl, to the Requires Recrawl section callback.
- Remove admin-page JavaScript file and associated webpack config entry. This was only used to insert the repeated "Important!" message for those fields that required a recrawl.
- Remove now-redundnat recrawl_required field flags.
- Mark required field as required, in label and on the field itself.
- Mark form as `novalidate`, so that the `required` attribute doesn't actually block form submission, which is an inconsistent inaccessible experience across different browsers.
- Remove separated help_links, and allow translatable help text to contain HTML.
- Move logic for conditionally showing description (help text) to a new method for consolidation and use in each rendering method.
- Add screen help text. Only a few notes added to start, including the note about dynamic tracking.
- Refactor Types to Track checkboxes to use fieldset, and fix non-unique ID attributes which mis-linked labels (reproduced for `develop` with #517).

## Motivation and Context
Closes #8. There may be further ideas (like displaying the sections as Tabs, and moving the Wipe option into a Screen Option to make it harder to trigger accidentally), but there should be opened as new Issues.

## How Has This Been Tested?
Manual use.

View of the new sectioning, including showing the Requires Recrawl section and consolidated warning message:

localhost_10024_wp-admin_options-general php_page=parsely (4)

This could be progressively enhanced with CSS and JS into Tabs.

View of the contextual help:
Screenshot 2021-11-28 at 15 48 30


sjinks (approved):

pauarge (approved):
Great work!

PR #519, opened 2021-11-28
Hide Disable AMP setting when AMP plugin is not enabled

### LINKED ISSUE #516: Only show Disable AMP setting conditionally
When AMP isn't supported on the site, then the Disable AMP setting has no purpose. We should hide it until it has a purpose (i.e. AMP plugin is active).

Only show the setting when the AMP plugin is active. This code could even be moved into the AMP integration code to keep it grouped together correctly.

Could leave the setting as is, but that runs the risk of causing confusion to new users as to why this setting exists when they may not use AMP.

## Additional context
There's a default setting for `disable_amp` in the Parsely class that shouldn't be overlooked as well.

GaryJones:
## Description
Hide the Disable AMP setting if the AMP plugin is not enabled.

Handle the validation of the submitted values, but using the previously saved value as the incoming value if none was submitted for `disable_amp`.

## Motivation and Context
Fixes #516.

## How Has This Been Tested?
Requires #518. Once #518 has been merged into `develop`, then this PR should be rebased.

1. Activate the AMP plugin
2. Change Disable AMP to a different value, then save.
2. Deactivate the plugin, make a different change on the settings page, and save. Note no validation errors despite no disable_amp value being sent.
3. Reactivate the AMP plugin, and reload the Parse.ly settings page, and see how the value for disable_amp has been retained.
4. Repeat 2-3 with the other value for disable_amp, to show that it doesn't just stick to a default value.

### DIFF
--- a/src/UI/class-settings-page.php
+++ b/src/UI/class-settings-page.php
@@ -230,20 +230,22 @@ Once you have changed a value and and saved, please contact support@parsely.com 
 			)
 		);
 
-		// Disable AMP tracking.
-		$h = __( 'If you use a separate system for JavaScript tracking on AMP pages (Tealium / Segment / Google Tag Manager / other tag manager solution) you may want to use that instead of having the plugin load the tracker.', 'wp-parsely' );
-		add_settings_field(
-			'disable_amp',
-			__( 'Disable AMP Tracking', 'wp-parsely' ),
-			array( $this, 'print_binary_radio_tag' ),
-			Parsely::MENU_SLUG,
-			'basic_settings',
-			array(
-				'title'      => __( 'Disable AMP Tracking', 'wp-parsely' ), // Passed for legend element.
-				'option_key' => 'disable_amp',
-				'help_text'  => $h,
-			)
-		);
+		if ( defined( 'AMP__VERSION' ) ) {
+			// Disable AMP tracking.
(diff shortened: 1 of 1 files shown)


src/UI/class-settings-page.php
```diff
@@ -108,250 +121,274 @@ public function initialize_settings(): void {
...
+		// These are the Requires Recrawl Settings.
+		add_settings_section(
+			'requires_recrawl_settings',
+			__( 'Requires Recrawl Settings', 'wp-parsely' ),
+			function (): void {
+				echo '<strong>' . wp_kses_post( __( '<span style="color:#d63638">Important:</span> changing any of these values below on a site currently tracked with Parse.ly will require reprocessing of your Parse.ly data.', 'wp-parsely' ) ) . '</strong><br />';
+				printf(
+					/* translators: Mailto link  */
+					esc_html__( 'Once you have changed a value and and saved, please contact %s to request a recrawl.', 'wp-parsely' ),
+					wp_kses_post( '<a href="mailto:support@parsely.com?subject=Please reprocess ' . $this->parsely->get_api_key() . '">support@parsely.com</a>' )
```
sjinks: May I suggest `urlencode()`'ing the subject? I remember I once had an issue with an iPhone and an unencoded whitespace character in the link.

GaryJones: Thanks - I've handled this in #518, and then rebased this PR onto the updated #518.

GaryJones:
Rebased.

pauarge (approved):
Tested locally, works great!

PR #523, opened 2021-11-29
Removing PHP 8.2 from CI

pauarge:
## Description

The PHP 8.2 pipeline is failing in CI. Since that PHP version is not stable yet and due to change, we're disabling it temporarily and we'll enable it again when we're closer to its release date.

## Motivation and Context

Stable CI pipeline.

## How Has This Been Tested?

CI pipeline passes correctly.


GaryJones (approved):
✅

PR #528, opened 2021-12-02
Adding E2E test to verify if scripts are rendered in FE

pauarge:
## Description

This PR adds an end-to-end test spec that contains two tests. The tests check script tags that contain the Parse.ly tracker are inserted in the WordPress-controlled front-end.

## Motivation and Context

Critical functionality of the plugin that we must ensure is always working.

## How Has This Been Tested?

Tests run locally and in GitHub Actions.


GaryJones (approved):
✅

Issue #520, opened 2021-11-28
Idea: Add Screen Options to toggle visibility of Requires Recrawl and Advanced Settings


GaryJones:
## Is your feature request related to a problem?
Even with the refactored UI from #518, the settings page can still appear daunting, with a lot of options.

## Describe the solution you'd like
We could add a Screen Option that enables the display of the Requires Recrawl, and Advanced Sections. This not only makes it an extra step to do something potentially destructive (though fixable), it also can reduce the number of visible settings for those who don't need it.

## Describe alternatives you've considered
Displaying each section as a Tab, which provides a little more discoverability than a screen option.

## Additional context
We could even combine the Tabs and screen options ideas together.

acicovic:
This has been implemented in #531 so I think we can close this issue (if we consider tabs, we should probably open another issue).

Issue #485, opened 2021-11-09
Clean up settings page class


pauarge:
## Is your feature request related to a problem?

The settings page class (`class-settings-page.php`) contains some repeated logic for rendering fields, and it's missing some checks that might lead to a broken layout. The page works, but the code is smelly and some changes on it might break the page easily.

## Describe the solution you'd like

We should simplify the methods and extract common behavior into helper functions. See comments below.

## Describe alternatives you've considered

Don't touch anything as it's technically not in a broken state right now.

## Additional context

This has been discussed in https://github.com/Parsely/wp-parsely/pull/482#discussion_r745389969 and https://github.com/Parsely/wp-parsely/pull/482#discussion_r745393411.

GaryJones:
I actually started a little on this on the weekend but bit off too much, and started following https://carlalexander.ca/designing-class-wordpress-admin-page/ to at least move some markup-heavy code into views. Not saying this whole article is what we should aim for, but certainly there are ways to cut the class down in length.

acicovic:
Note to our future selves: In the above merged #531 PR, @GaryJones has noted some improvements to be made to the code. The respective comments can be found easily as they are the 3 comments that are not marked as outdated.

acicovic:
Notes from #531:
- https://github.com/Parsely/wp-parsely/pull/531#discussion_r796513470 (this would also apply to other parts of our code that use `empty()`. 
- https://github.com/Parsely/wp-parsely/pull/531#discussion_r796514163
- https://github.com/Parsely/wp-parsely/pull/531#discussion_r796515607

Notes from #766:
- https://github.com/Parsely/wp-parsely/pull/766#discussion_r848546763

mehmoodak:
It seems like we already have the needed helper functions in place like:

- `print_filter_text`
- `print_description_text`
- `print_text_tag`
- `print_checkbox_tag`
- `print_select_tag`
- `print_radio_tags`
- `print_media_single_image`

So this issue can be closed. Please do let me know if anyone have concerns otherwise will close it in next 2 days.

acicovic:
I took a quick look into the setting's page code.

It is a big file, but the code seems structured well-enough. There are always further opportunities for improvement, but we can open those as specific issues if needed. There are also other open issues that will probably result in further changes/restructuring to that file.

So @mehmoodak, feel free to close this.

danielabloch:
Closing this, per the last 2 comments.

PR #531, opened 2021-12-13
Add Screen Options to toggle settings sections

GaryJones:
## Motivation and Context
The settings page can be long so by using the Screen Options feature we can enable users to hide those settings that either require a recrawl or advanced settings that would wipe the data. We could even hide these rarely-used settings by default.

## Description

Working proof of concept for #520.

Note that for the $_POST data to be captured, the Settings_Page class is instantiated on the `init` hook instead of the (internal-only) `_admin_menu`. The `run()` method of the Settings_Page class hooks into specific admin-page related hooks, so there is no substantial impact to front-end requests.

One of the difficulties was our validation - if the submitted value was `null`, then it would reset it to be a default value. When a section is not shown, and settings are submitted, then the values for the hidden section would be `null`. The validation has been (clumsily) updated to use the previously saved database values instead. Tidying up the code in the validation is needed, but outside the scope of this PR.

## How Has This Been Tested?
So far, this has been tested via a lot of manual testing, but if we want to move forward with this, then e2e tests could be added to ensure that each of the settings in hideable sections is persisted correctly.

Screen Options are accessed via the "Screen Options" tabbed button in the top right (for LTR) of an admin screen.

Screenshot 2021-12-13 at 10 58 09

I've done a screen recording (with audio) to show it in action:

https://user-images.githubusercontent.com/88371/145798262-98848a0c-156b-4b47-bd8f-c2372874892e.mov

## Alternatives considered

I've left this as a draft PR for now, as while this makes use of a WP core feature, it could be that it's easy to miss this, which would potentially lead to more support interactions for not being able to find an option. Or maybe hiding it would lead to fewer support requests from folks who accidentally changed a Requires Recrawl setting and now feel that they need to get a recrawl done unnecessarily.

A different alternative would be to use a Tabs interface instead - that is, each section becomes its own tab. This would have the advantage that all settings would be submitted (`$_POST` would be complete making validation simpler again), and different section titles are more visible on screen than they are with Screen Options; better discoverability, while still making the settings page simpler and less scary.

Add to 3.1 milestone, but may be bumped to 3.2.


acicovic:
Playing a bit with the UI, I like it more than I thought I would. I still find it unconventional, but nobody can say that it doesn't provide a clean interface. 😄

When clicking the "Save Changes" button I get white screens with errors like:

```
Notice: Undefined index: disable_amp in /var/www/html/wp-content/plugins/wp-parsely/src/UI/class-settings-page.php on line 882 Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/wp-content/plugins/wp-parsely/src/UI/class-settings-page.php:882) in /var/www/html/wp-includes/pluggable.php on line 1340 Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/wp-content/plugins/wp-parsely/src/UI/class-settings-page.php:882) in /var/www/html/wp-includes/pluggable.php on line 1343
```

Note that the errors also depend on the sections that you've selected to show. Another example:
```
Notice: Undefined index: disable_amp in /var/www/html/wp-content/plugins/wp-parsely/src/UI/class-settings-page.php on line 882 Notice: Undefined index: track_post_types in /var/www/html/wp-content/plugins/wp-parsely/src/UI/class-settings-page.php on line 899 Notice: Undefined index: track_page_types in /var/www/html/wp-content/plugins/wp-parsely/src/UI/class-settings-page.php on line 907 Notice: Undefined index: meta_type in /var/www/html/wp-content/plugins/wp-parsely/src/UI/class-settings-page.php on line 942 Notice: Undefined index: content_id_prefix in /var/www/html/wp-content/plugins/wp-parsely/src/UI/class-settings-page.php on line 949 Notice: Undefined index: use_top_level_cats in /var/www/html/wp-content/plugins/wp-parsely/src/UI/class-settings-page.php on line 957 Notice: Undefined index: custom_taxonomy_section in /var/www/html/wp-content/plugins/wp-parsely/src/UI/class-settings-page.php on line 976 Notice: Undefined index: cats_as_tags in /var/www/html/wp-content/plugins/wp-parsely/src/UI/class-settings-page.php on line 984 Notice: Undefined index: lowercase_tags in /var/www/html/wp-content/plugins/wp-parsely/src/UI/class-settings-page.php on line 1004 Notice: Undefined index: force_https_canonicals in /var/www/html/wp-content/plugins/wp-parsely/src/UI/class-settings-page.php on line 1024 Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/wp-content/plugins/wp-parsely/src/UI/class-settings-page.php:882) in /var/www/html/wp-includes/pluggable.php on line 1340 Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/wp-content/plugins/wp-parsely/src/UI/class-settings-page.php:882) in /var/www/html/wp-includes/pluggable.php on line 1343 
```

acicovic:
Suggesting something that I would consider a usability improvement, and would also help those who don't know what/where screen options are:

In the text "More settings can be enabled via Screen Options.", turn the "Screen Options" part into a link, which when clicked rolls down the screen options.

pauarge:
I've done some minor changes and made the e2e tests pass. Overall it looks good to me. @GaryJones would you take a second look if my changes affect your initial code?

Having thought about it, I actually like the behavior as-is (sections hidden by default).

pauarge:
SonarCloud detects the following block of code as duplicate:

Screen Shot 2022-01-31 at 12 05 28 PM

We could extract it in a function, but that is outside of the scope of this PR so we can ignore it for the time being.

pauarge (dismissed):

src/UI/class-settings-page.php
```diff
@@ -76,16 +102,115 @@ public function enqueue_settings_scripts( string $hook_suffix ): void {
...
+	 * @since 3.2.0
+	 *
+	 * @param mixed  $screen_option The value to save instead of the option value. Default false (to skip saving the current option).
+	 * @param string $option        The option name.
+	 * @param mixed  $value         The option value.
+	 * @return mixed Updated option value.
+	 */
+	public function set_screen_option( $screen_option, string $option, $value ) {
+		if ( $this->screen_options_name === $option ) {
+			check_admin_referer( 'screen-options-nonce', 'screenoptionnonce' );
```
GaryJones: This checks the nonce, but line 126 said that this wasn't needed?

src/UI/class-settings-page.php
```diff
@@ -672,18 +847,84 @@ public function validate_options( array $input ): array {
...
-		// these can't be null, if somebody accidentally deselected them just reset to default.
-		if ( ! isset( $input['track_post_types'] ) ) {
-			$input['track_post_types'] = array( 'post' );
-		}
-		if ( ! isset( $input['track_page_types'] ) ) {
-			$input['track_page_types'] = array( 'page' );
+
+		$input['api_secret'] = sanitize_text_field( $input['api_secret'] );
+
+		if ( ! empty( $input['metadata_secret'] ) ) {
```
GaryJones, line 853: Can we use a better check than `empty()`?

src/UI/class-settings-page.php
```diff
@@ -672,18 +847,84 @@ public function validate_options( array $input ): array {
...
+			if ( strlen( $input['metadata_secret'] ) !== 10 ) {
```
GaryJones, line 854: This `10` seems like a magic number that should be made into a class constant to better self-document it.

src/UI/class-settings-page.php
```diff
@@ -672,18 +847,84 @@ public function validate_options( array $input ): array {
...
+				add_settings_error(
+					Parsely::OPTIONS_KEY,
+					'metadata_secret',
+					__( 'Metadata secret is incorrect. Please contact Parse.ly support!', 'wp-parsely' )
+				);
+			} elseif ( isset( $input['parsely_wipe_metadata_cache'] ) && 'true' === $input['parsely_wipe_metadata_cache'] ) {
+				delete_post_meta_by_key( 'parsely_metadata_last_updated' );
+
+				wp_schedule_event( time() + 100, 'everytenminutes', 'parsely_bulk_metas_update' );
```
GaryJones, line 863: Why 100 seconds? How is setting up a cron job related to adding Screen Options toggles?

src/UI/class-settings-page.php
GaryJones: I see that this is just moved code, not new code.

src/UI/class-settings-page.php
GaryJones: I see that this is just moved code, not new code.

src/UI/class-settings-page.php
GaryJones: I see that this is just moved code, not new code.

tests/e2e/specs/activation-flow.spec.js
```diff
@@ -34,4 +59,27 @@ describe( 'Activation flow', () => {
...
+		expect( await checkH2DoesNotExist( 'Requires Recrawl Settings' ) ).toBe( true );
+		expect( await checkH2DoesNotExist( 'Advanced Settings' ) ).toBe( true );
+
+		await selectScreenOptions( true, true );
+
+		await page.waitForXPath( '//h2[contains(text(), "Basic Settings")]' );
+		await page.waitForXPath( '//h2[contains(text(), "Requires Recrawl Settings")]' );
+		await page.waitForXPath( '//h2[contains(text(), "Advanced Settings")]' );
+
+		await selectScreenOptions( false, true );
```
GaryJones: What would you do if we had 6 or 10 toggleable sections, instead of 2? Probably better to treat this argument as a single associated array (section-name => true/false), rather than multiple boolean arguments.

views/parsely-settings.php
```diff
@@ -20,6 +20,7 @@
 <div class="wrap">
 	<h1 class="wp-heading-inline"><?php echo esc_html( get_admin_page_title() ); ?></h1>
 	<span id="wp-parsely_version"><?php echo esc_html( $parsely_version_string ); ?></span>
+	<p><em><?php esc_html_e( 'More settings can be enabled via Screen Options (button on the top right corner of the page).', 'wp-parsely' ); ?></em></p>
```
GaryJones: RTL languages can make this say "top left" as needed.

src/UI/class-settings-page.php
```diff
@@ -76,16 +102,115 @@ public function enqueue_settings_scripts( string $hook_suffix ): void {
...
+	 *
+	 * @since 3.2.0
+	 *
+	 * @param mixed  $screen_option The value to save instead of the option value. Default false (to skip saving the current option).
+	 * @param string $option        The option name.
+	 * @param mixed  $value         The option value.
+	 * @return mixed Updated option value.
+	 */
+	public function set_screen_option( $screen_option, string $option, $value ) {
+		if ( $this->screen_options_name === $option ) {
```
acicovic, line 136: Consider Yoda condition

acicovic (approved):

acicovic:
I had posted an erroneous review comment which I removed. I think this should be merged before #633.

acicovic:
Also, the suggestions made from @GaryJones regarding the moved code may be worth noting for future consideration.

pauarge:
> Also, the suggestions made from @GaryJones regarding the moved code may be worth noting for future consideration.

We do have an issue (https://github.com/Parsely/wp-parsely/issues/485) to work on the settings page. I think it is clear that it has grown too much and we should rethink it. All Gary's suggestions should be addressed on this work.

pauarge (approved):

PR #535, opened 2021-12-14
Adding WordPress to Composer for development

pauarge:
## Description

Some IDEs require a WordPress installation to offer WordPress code completion. 

This PR adds WordPress as a development dependency in Composer, so whenever a user runs `composer install`, they will get the WordPress codebase automatically in `vendor/`. This goal can be achieved in a variety of ways, but with this method no configuration is needed from the developer's point of view.

## Motivation and Context

WordPress-aware code completion on IDEs.

## How Has This Been Tested?

WordPress functions are picked up by all IDEs without external WordPress files present in the system.


GaryJones:
Is this really needed? Don't folks either have the plugin installed in a `wp-content/plugins/...` environment where the WP core files are already available, or would set their IDE to use a globally included directory of WP core files anyway?

acicovic:
> Is this really needed? Don't folks either have the plugin installed in a `wp-content/plugins/...` environment where the WP core files are already available, or would set their IDE to use a globally included directory of WP core files anyway?

I'm wondering the same thing, given that some IDEs/Editors can also accept WP stubs.

I don't think it is *really* needed, but that it could be useful in some cases. I guess a question would be if there's a con by introducing this, and if the pros outweigh the cons in that case.

pauarge:
Happy to evaluate this. In my personal setup, I used to have WordPress cloned in my machine and my IDE pointed to there. I don't have a local WP install where I run development from, since I use `npm run dev:start` for development.

This repository does not have a recommended way of doing so. My intention with this was to provide an easy alternative to onboard people in the project and IMO the simplest option is to run `composer install` and have it available.

What do you think @GaryJones @acicovic ?

acicovic:
My setup is identical to @pauarge's but I've got an extension that contains WP stubs so I'm covered regarding code completion.

If including this helps even one of our contributors or users without introducing any side effects, then I would probably go for including this.

This becomes more interesting if we're able to make a path mapping in our projects so we can step in WordPress code when debugging.

GaryJones:
I'm just not seeing the benefit on this, especially as:

- we don't have any external contributors.
- nobody has asked for this type of support
- nobody has said this missing is a blocker to them contributing. 

Let's not solve for a hypothetical problem.

PR #536, opened 2021-12-14
Fix Undefined index in class-settings-page.php

acicovic:
## Description
In my local environment, I was getting a "Notice: Undefined index: disable_amp in /src/UI/class-settings-page.php on line 712" resulting in a "Warning: Cannot modify header information - headers already sent" error when saving the plugin's settings.

I've added a check to see if the variable in question is set. The flow is exactly the same as if the variable was null.

## How Has This Been Tested?
Whenever I started with a fresh database, this error would occur. After my change, I've tested with a fresh database and didn't face any issues.


GaryJones (approved):

pauarge (approved):
Nice catch!

PR #538, opened 2021-12-14
Leverage Husky to enforce linters and coding standards prior to commit

### LINKED ISSUE #375: Consider pre-push hooks

Doing some checks locally prior to pushing to a remote branch will:
* provide a tighter development feedback loop for changes that are destined to be rejected by CI
* reduce load on CI for pushes that would be destined to fail

Implement [husky](https://github.com/typicode/husky) pre-push checks (or similar solution) for common cases of failing CI runs:
* uncommitted changes
* phpcs violations
* eslint violations
* etc.

:hulksmash:

jblz:
## Description

* Add [Husky](https://typicode.github.io/husky/) as a dev dependency
* Adds [concurrently](https://www.npmjs.com/package/concurrently) as a dev dependency
* Leverage the [`prepare` lifecycle script hook](https://docs.npmjs.com/cli/v8/using-npm/scripts#life-cycle-scripts) to run `husky install`
* Adds a custom npm script: `lint` which concurrently runs all the linters
* Adds a pre-commit hook which blocks the commit if either a linter and coding standard tool encounters a warning or error at "severity=1" for phpcs

To reach the new phpcs standard, there are a few minor code changes included in this PR.

## Motivation and Context

This picks up from #344 and is independent of the locally installed version of PHP.

Doing work, committing, pushing, and then waiting on Actions to complete to then find out there is a CS error, can be frustrating. 

It's possible to run `composer cs` and other linting tasks locally, but these may often be forgotten.

To bring the check earlier, we can use local git hooks to do the linting.

The hooks can also be used to improve the quality of the commit message, to make it easier to review work later on, however that work will happen in a follow up PR.

No other hooks (such as commit-msg, pre-push, etc.) are currently enabled, so this change is limited to pre-commit inspections for now.

Closes #375
Partially addresses #544

## How Has This Been Tested?

Commit as usual.  If no violations are encountered, the commit succeeds without warning.

#### Invalid JS

Screen Shot 2021-12-14 at 1 09 11 PM

#### Invalid PHP

Screen Shot 2021-12-14 at 1 35 14 PM

### If there are phpcs errors, the commit is blocked

Screen Shot 2021-12-16 at 9 02 06 AM

### If there are phpcs warnings, the commit is blocked

Screen Shot 2021-12-16 at 9 01 16 AM

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npm run lint
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -2,4 +2,4 @@
 . "$(dirname "$0")/_/husky.sh"
 
 npm run lint
-composer cs
+composer cs -- --severity=1
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -2,4 +2,4 @@
 . "$(dirname "$0")/_/husky.sh"
 
 npm run lint
-composer cs -- -n
+composer cs -- --runtime-set ignore_warnings_on_exit 1
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -2,3 +2,4 @@
 . "$(dirname "$0")/_/husky.sh"
 
 npm run lint
+composer cs
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -2,4 +2,4 @@
 . "$(dirname "$0")/_/husky.sh"
 
 npm run lint
-composer cs
+composer cs -- -n
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -2,4 +2,4 @@
 . "$(dirname "$0")/_/husky.sh"
 
 npm run lint
-composer cs -- --runtime-set ignore_warnings_on_exit 1
+composer cs
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
 	},
 	"scripts": {
 		"build": "wp-scripts build",
-	  	"dev:logs": "wp-env logs",
+		"dev:logs": "wp-env logs",
 		"dev:start": "wp-env start",
 		"dev:stop": "wp-env stop",
 		"lint": "concurrently \"npm:lint:*\"",
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
 		"@wordpress/env": "^4.1.3",
 		"@wordpress/eslint-plugin": "^9.3.0",
 		"@wordpress/scripts": "^19.2.2",
+		"husky": "^7.0.4",
 		"prettier": "^2.5.1"
 	},
 	"scripts": {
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -60,6 +64,7 @@
 		"test:unit": "wp-scripts test-unit-js",
 		"test:unit:debug": "wp-scripts --inspect-brk test-unit-js --runInBand --no-cache",
 		"test:unit:help": "wp-scripts test-unit-js --help",
-		"test:unit:watch": "wp-scripts test-unit-js --watch"
+		"test:unit:watch": "wp-scripts test-unit-js --watch",
+		"prepare": "husky install"
```
pauarge: Could we have the commands sorted alphabetically?

pauarge (changes requested):
This looks awesome and I like the approach. However, I have a couple of comments.

- Could we add some instructions (or at least some information) about what's being added in this PR in `CONTRIBUTING.md`?
- I would fail on PHPCS warnings and not let the commit go through, because warnings will make the GitHub Action fail. What's more, could we even try to automatically fix them (by running `phpcbf`)?

.husky/pre-commit
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npm run lint
+composer cs -- --runtime-set ignore_warnings_on_exit 1
```
GaryJones: Why ignore warnings? Letting them go through is then just noise for the next person.

I'd also like to see us add `--severity=1` (and fix up the issues that appear as well).

GaryJones (dismissed):

.husky/pre-commit
jblz: My rationale was: if the were not severe enough to be deemed an error, why should the be severe enough to block commits?

In the same way that a PHP warning alerts (if you're configured to see them) but continues executing.

> I'd also like to see us add --severity=1 (and fix up the issues that appear as well).

Does that make what are now phpcs warnings into errors?

jblz:
> could we even try to automatically fix them (by running phpcbf)?

IMO, auto fixes like that are better handled in the IDE on save than during the commit part of the dev cycle. I'm open to other opinions of course, but would like to push that sort of change to an issue and not address it in this PR.

sjinks (dismissed):

pauarge:
> IMO, auto fixes like that are better handled in the IDE on save than during the commit part of the dev cycle. I'm open to other opinions of course, but would like to push that sort of change to an issue and not address it in this PR.

Fair enough, an automation changing your code can be confusing.

.husky/pre-commit
pauarge: I stick with Gary on this one -- unless there's a very good reason (and in that case, you should be using a per-line ignore), I would treat warnings as errors. This would result in a more cohesive codebase. This is in line with `--severity=1`.

jblz: works for me

jblz:
I enforced the warning check & added the `--severity=1` flag and fixed up a few violations.

The code changes are all very low-stakes, so I'm good with including them in this PR.

pauarge (approved):
Awesome!

PR #540, opened 2021-12-14
Handle broken database option

pauarge:
## Description

It might be the case that the `parsely` option in the WordPress database is corrupted and is not an array. If that's the case, the site would fatal because the function `get_options` function is expected to always return an array. This PR checks that the value returned from the DB is indeed a an array and if it's not, it returns the default options.

## Motivation and Context

Preventing potential sites from fataling.

## How Has This Been Tested?

Go to the database and manually change the option value to something random, like `asdf`. With this PR the site should function normally.

### DIFF
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -648,7 +648,7 @@ class Parsely {
 	public function get_options(): array {
 		$options = get_option( self::OPTIONS_KEY, $this->option_defaults );
 
-		if ( "array" != gettype( $options ) ) {
+		if ( 'array' != gettype( $options ) ) {
 			return $this->option_defaults;
 		}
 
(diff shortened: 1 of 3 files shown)


jblz:
I confirmed the error case against the develop branch:

```
bash-5.1$ wp option set parsely 42
Success: Updated 'parsely' option.
bash-5.1$ wp option get parsely
Error: There has been a critical error on this website.Learn more about troubleshooting WordPress. There has been a critical error on this website.
```

This seems to fix the issue & return the default settings when the option is a scalar type.

Not sure what's up with that e2e test.... looking.

jblz (approved):
This tests well for me locally.  Should be gtg when #541 is incorporated

Issue #483, opened 2021-11-08
Settings page: Improve UI for post/page types to track


pauarge:
## Is your feature request related to a problem?

We don't want users to track posts and page types twice. Right now, the UI lets them do that and we'd like to not allow them.

## Describe the solution you'd like

@GaryJones describes a good solution here: https://github.com/Parsely/wp-parsely/pull/482#issuecomment-963287549

acicovic:
This is a comment is from @GaryJones and was posted [here](https://github.com/Parsely/wp-parsely/pull/482#issuecomment-963287549). I'm including the comment here for reference and ease of use:

> * Can we use JS (and back-end validation) to disable checkboxes when it has already been marked in the other type?
> 
> * Alternatively, I wonder whether radio buttons would help enforce this better (shown as a table here, but really fieldset rows, one row for each viable post_type) since what we really want is for each type to be assigned to exactly one of either `post`, `page` or `none`:
> 
> 
> |   | None | Post |  Page |
> | --- | :---: | :---: | :---: |
> | post  | o | ⦻ | o |
> | page  | o | o | ⦻ |
> | attachment  | ⦻ | o | o |

acicovic:
I would love to get feedback from the others. Here's my take:

> Can we use JS (and back-end validation) to disable checkboxes when it has already been marked in the other type?

Back-end validation is being worked on (see #543). JS validation implementation will depend on the solution we decide to implement.

> Alternatively, I wonder whether radio buttons would help enforce this better...

This makes sense to me personally. Else, in the case of checkboxes we could just implement an auto-uncheck JS function.

GaryJones:
> we could just implement an auto-uncheck JS function

That's assuming that JS is enabled - and we can get the same benefit by using native HTML.

pauarge:
+1 on implementing this without JS, just with native HTML.

acicovic:
Given that most of us seem to agree with the radio solution, I'll take note of this and look into it.

In the meantime, don't hesitate to provide any other feedback.

PR #543, opened 2021-12-15
Settings page: Show error message when duplicate tracking selected

acicovic:
## Description
In the settings page, if the user selects the same element in the "Post Types to Track" and "Page Types to Track" sections, an error message will be shown to them telling them that the element should be selected only in one of the sections.

## Motivation and Context
The reasoning behind this is that we should not allow duplicate tracking. This PR partly addresses the second point found in https://github.com/Parsely/wp-parsely/pull/482#issuecomment-963287549. Also related: #483.

## How Has This Been Tested?
- Tested manually using the Parse.ly plugin's options page,
- Added an integration test.

## Screenshots
Error that appears when items are selected in both "Post Types to Track" and "Page Types to Track":

image


pauarge (changes requested):
Great work, but I'm leaving a few notes:

- Could we add (at least) one integration test and an end to end test to double check this functionality is working correctly?
- It would be great if you added a screenshot of the error in the PR's description.
- If it doesn't complicate the code very much, could we display which item(s) is duplicated in the error message? That would make the issue easier to understand for an end user.

acicovic:
> Could we add (at least) one integration test and an end to end test to double check this functionality is working correctly?

Will look into this.
> It would be great if you added a screenshot of the error in the PR's description.

Done! If anyone has an idea for a better error message, please let me know.

> If it doesn't complicate the code very much, could we display which item(s) is duplicated in the error message? That would make the issue easier to understand for an end user.

This isn't complicated to do and I thought of it myself, but was thinking that it could give some trouble to localized strings. But I should probably look into this more.

GaryJones:
> localized strings

You'll need to use variables (and add `/* translator: ... */` comments), but that's all standard and [documented](https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#variables).

That said, I'd rather we addressed the root issue as per https://github.com/Parsely/wp-parsely/pull/482#issuecomment-963287549 than simply rely on error notifications being noticed and acted upon.

acicovic:
Thanks for the input @GaryJones, I'll make sure to check the documentation you sent.

Concerning the root issue, it's on our radar and (from my understanding) monitored [here](https://github.com/Parsely/wp-parsely/issues/483). I'll bring your comment there for context, and we can resume conversation from there.

This PR doesn't replace that proposal. Rather, it's a relatively fast way to stop users from doing something they shouldn't do and the system shouldn't allow them to do. Regardless of the UI we'll finally implement, we should also be needing server-side validation in case anything else goes wrong.

pauarge:
> we should also be needing server-side validation in case anything else goes wrong.

This is especially important, I think this PR is definitely positive.

acicovic:
I've just converted this PR to draft as it still needs some work and I've uncovered a bug.

acicovic:
Since we will be migrating to https://github.com/Parsely/wp-parsely/issues/483, we've agreed with @pauarge that an end-to-end test is not needed for this since it will change anyway.

So I'm flagging this for review.

src/UI/class-settings-page.php
```diff
@@ -638,12 +638,19 @@ public function validate_options( array $input ): array {
 		$input['track_page_types'] = self::validate_option_array( $input['track_page_types'] );
 
 		// Detect and prevent duplicate tracking.
-		$duplicate_tracking_found = ! empty( array_intersect( $input['track_post_types'], $input['track_page_types'] ) );
-		if ( $duplicate_tracking_found ) {
+		$duplicate_items = array_intersect( $input['track_post_types'], $input['track_page_types'] );
+		if ( ! empty( $duplicate_items ) ) {
```
GaryJones: Instead of the generic `empty()`, can we use a more specific `0 === count( $duplicate_items )`?

https://beberlei.de/2021/02/19/when_to_use_empty_in_php_i_say_never.html

src/UI/class-settings-page.php
```diff
@@ -638,12 +638,19 @@ public function validate_options( array $input ): array {
...
 				Parsely::OPTIONS_KEY,
 				'track_page_types',
-				__( 'An element is selected twice in post and page types to track. Each element can be selected only once.', 'wp-parsely' )
+				sprintf(
+					/* translators: %s: Item(s) being tracked both as posts and pages. */
+					__(
+						'%s cannot be tracked as both posts and pages. Please select every element only once.',
+						'wp-parsely'
+					),
+					implode( ', ', $duplicate_items )
```
GaryJones: Can this be further refined to swap the last `, ` for an ` and `?

Right now it would say: "Foo, Bar cannot be tracked..." or "Foo, Bar, Baz" cannot be tracked...". Ideally, we could reframe it as "Foo and Bar cannot be tracked..." and "Foo, Bar and Baz cannot be tracked...".

Note also that some languages may handle this differently (or may not use `,` as comma punctuation at all), so correct i18n support may need a bit more thinking.

GaryJones (dismissed):

src/UI/class-settings-page.php
acicovic: I'm certainly not against it.

Looking at the current codebase of the plugin, we use `empty()` in a variety of places, which brings up the question of whether a refactor should be done at some point. In general I would say that `empty()` is more readable for many people as well as shorter, but we could go with the alternatives if we want to aim on better quality code.

For the time being, I'll post a commit addressing this.

src/UI/class-settings-page.php
acicovic: We would indeed need more tinkering for real/correct i18n support. Ideally, I would add post types in quotes which I think could make the message even more clear:

> "Foo", "Bar" and "Baz" cannot be tracked...

However, from the 3 languages I know, each one handles quotes differently. So for the time being I stepped back from this. For the time being I'll post a commit to swap the `','` with `and` and we can open a discussion about correct i18n support with the others.

acicovic:
@GaryJones I've marked the discussions as resolved but feel free to reopen anything if needed.

acicovic:
That `0 === count( $duplicate_items )` broke the tests so I'll look into it.

src/UI/class-settings-page.php
GaryJones: In terms of i18n support, I think swapping all commas for an "and" is going to be worse. I'll have a think about this.

tests/Integration/UI/SettingsPageTest.php
```diff
@@ -0,0 +1,91 @@
...
+declare(strict_types=1);
+
+namespace Parsely\Tests\Integration\UI;
+
+use Parsely\Parsely;
+use Parsely\Tests\Integration\TestCase;
+use Parsely\UI\Settings_Page;
+
+/**
+ * Row actions tests.
```
pauarge: I think this might be a leftover from copy pasting, but this tests are not testing row actions.

tests/Integration/UI/SettingsPageTest.php
```diff
@@ -0,0 +1,91 @@
...
+	 */
+	public function test_validate_duplicate_tracking_values_returns_errors(): void {
+		// Initializations.
+		$expected = self::$parsely->get_options();
+		$options  = self::$parsely->get_options();
+
+		// Duplicate selection in Post Types.
+		$options['track_post_types'] = array( 'post', 'page' );
+		$options['track_page_types'] = array( 'page' );
+		self::assertSame( self::$settings_page->validate_options( $options ), $expected );
```
pauarge: When asserting something, I think it's more readable to put first the `$expected` variable and then the computed one. Even better, we could separate the call to `validate_options` in a different line within a variable.

This comment applies for the other assertions as well.

tests/Integration/UI/SettingsPageTest.php
```diff
@@ -0,0 +1,91 @@
...
+
+		// Duplicate selection in Page Types.
+		$options['track_post_types'] = array( 'post' );
+		$options['track_page_types'] = array( 'post', 'page' );
+		self::assertSame( self::$settings_page->validate_options( $options ), $expected );
```
pauarge: I would separate the three assertions on three separate tests. In the event of a test failing, we would get better granularity.

pauarge (changes requested):
Great work! I've left some minor improvements in the test.

tests/Integration/UI/SettingsPageTest.php
GaryJones: Agreed on `$expectation` coming first: https://phpunit.readthedocs.io/en/9.5/assertions.html#assertsame

tests/Integration/UI/SettingsPageTest.php
acicovic: This needs to be there (we use `self::$settings_page->validate_options()`).

tests/Integration/UI/SettingsPageTest.php
acicovic: This should be OK now. I've implemented both suggestions.

tests/Integration/UI/SettingsPageTest.php
pauarge: I mean the `Row actions tests.` comment. It should be `Settings page tests.`

acicovic: You are completely right. My eye didn't catch this.

pauarge (approved):
LGTM, great work!

PR #547, opened 2021-12-16
Fix metadata on password protected posts

pauarge:
## Description

Password-protected posts in WordPress would get metadata tags from Parse.ly, regardless of the user having input the password or not. This PR fixes that by adding a check on the post's password.

## Motivation and Context

Hide data for password protected posts.

## How Has This Been Tested?

Create a password protected post and try to access it. With this changes, metadata should not be rendered unless the user inputs the password.


src/class-parsely.php
```diff
@@ -287,8 +287,9 @@ public static function post_has_trackable_status( $post ): bool {
 		 * @param string[]    $trackable_statuses The list of post statuses that are allowed to be tracked.
 		 * @param int|WP_Post $post               Which post object or ID is being checked.
 		 */
-		$statuses          = apply_filters( 'wp_parsely_trackable_statuses', array( 'publish' ), $post );
-		$cache[ $post_id ] = in_array( get_post_status( $post ), $statuses, true );
+		$statuses = apply_filters( 'wp_parsely_trackable_statuses', array( 'publish' ), $post );
+
+		$cache[ $post_id ] = in_array( get_post_status( $post ), $statuses, true ) && ! post_password_required( $post );
```
jblz: I'd probably move the `post_password_required` to the top of the function since there's no need to fire the `wp_parsely_trackable_statuses` filters if they've no chance of passing the check on protected posts.

jblz:
This seems to do the trick for me.  Thanks for fixing so quickly.

tests/Integration/OtherTest.php
```diff
@@ -211,4 +211,47 @@ public function test_corrupted_options(): void {
...
+	 */
+	public function test_post_has_trackable_status_password_protected_with_filter(): void {
+		add_filter(
+			'wp_parsely_skip_post_password_check',
+			function( bool $skip, $post ): bool {
+				return true;
+			},
+			10,
+			2
+		);
```
jblz: This can be simplified, if desired:

```suggestion
		add_filter( 'wp_parsely_skip_post_password_check', '__return_true' );
```

jblz (dismissed):
I left a take-it-or-leave-it suggestion on the `add_filter` callback, but otherwise, this looks good to me.  Thanks again!

jblz (approved):

PR #554, opened 2021-12-21
Try: Block Editor integration for pre-publish experience features

jblz:
## Description

This is a work in progress.

https://user-images.githubusercontent.com/1587282/146972085-40039d47-017c-49ac-9480-607c84364315.mov


acicovic:
Is this still relevant with work being done on #642?

pauarge:
> Is this still relevant with work being done on https://github.com/Parsely/wp-parsely/pull/642?

Yes, this is a separate piece of work regarding the side panel of the editor. We'll want to pick in the future.

pauarge:
This PR has some conflicts already and we won't be merging it as-is. Closing it, but we can use its code as a reference for future work.

PR #555, opened 2021-12-21
Remove phpunit-integration.xml.dist from exported plugin files

jblz:
## Description

Add a wildcard to the phpunit line in the .gitattributes file

## Motivation and Context

This is a development file and does not belong in a published plugin archive (be it zip or dir). We already ignore the file called `phpunit.xml.dist` -- this change adds a wildcard such that it will ignore any similarly named files in the future.

## How Has This Been Tested?

* Check out this branch
* Run ./bin/export-plugin.sh ~/Downloads/wp-parsely-dev DIR
* `ls ~/Downloads/wp-parsely-dev` -- should be a properly formed dir of "exported" plugin files without either of the `phpunit*.xml.dist` files


jblz:
Hrm...it appears that this breaks the github integration test action. Looks like it pulls the archive behind the scenes:

https://github.com/actions/checkout/blob/eb8a193c1dbf4bbb2053320cef52bacc1a485839/src/github-api-helper.ts#L28-L32

The fact that the "unit test" action isn't affected by this was surprising. I suspect it's just using default settings.

jblz:
Closing for now... 😕

PR #556, opened 2021-12-22
name hooked functions & init in after_setup_theme instead of plugins_loaded

jblz:
## Description

- Instead of using anonymous functions in our various initialization hooks, name the functions.
- Kick off instantiation and initialization in the `after_setup_theme` action (instead of `plugins_loaded`).

## Motivation and Context

#411 moved this functionality from top-level in the plugin entry file into a function hooked into `plugins_loaded`:
https://github.com/Parsely/wp-parsely/pull/411/files#diff-41f0c604e0848ecb9ac1b6e07d5d2fab90b2ef3f715792c15243c5eed180894cL52-R43

When loaded this early, it doesn't give theme code the ability to affect loading. While rare (& possibly suboptimal), this is something that does happen from time to time.

Moving to `after_setup_theme` allows for those cases.

Additionally, naming the functions allows for selective `remove_action` calls for whatever reason.

## How Has This Been Tested?

TBD

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -45,46 +45,58 @@ const PARSELY_FILE    = __FILE__;
 require __DIR__ . '/src/class-parsely.php';
 require __DIR__ . '/src/class-rest.php';
 require __DIR__ . '/src/class-scripts.php';
-add_action(
-	'plugins_loaded',
-	function(): void {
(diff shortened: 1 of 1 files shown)


jblz:
Closing in favor of #595.

Will revisit the hook in a separate PR.

Issue #338, opened 2021-06-17
Tests for JS loading are flawed


GaryJones:
## Describe the bug
The following tests seem flawed - I was able to remove an `echo` (to fix a CS issue), and the test didn't fail.

- `All_Test::test_load_js_tracker()`
- `All_Test::test_load_js_api_no_secret()`
- `All_Test::test_load_js_api_with_secret()`
- `All_Test::test_user_logged_in()`
- `All_Test::test_user_logged_in_multisite()` x2
- `All_Test::test_load_js_tracker_filter()`
- `All_Test::test_deprecated_insert_javascript_filter()`

They use `ob_get_contents()` around a call to a function that enqueues scripts - but under a test condition, these scripts may not be echo'd into output. At the very least, we probably want a call to `wp_print_scripts()` call in the output buffer, and maybe `wp_footer()`, and then check the output for the presence of the script.

There is a smarter way, however: seeing if a script was enqueued or not, since that decision-making logic to enqueue or not is ours; the logic of printing to screen is WP's, so it's not ours to test.

https://wordpress.stackexchange.com/questions/11041/check-if-a-script-style-was-enqueued-registered/11584 outlines how `wp_script_is()` can be used to check the status of a script. This can be tested immediately after the function-under-test is called, rather than having to set up an output buffer and recreate printing to screen.

## Additional context
The above tests will need to be rewritten.

PR #557, opened 2021-12-22
Try to improve script integration tests

acicovic:
# Goal

This PR tries to address the issues reported in #338 and opens a discussion so we can focus on finding the most correct approach to this.

**I will need your feedback!**

# Points made and code examination results

If I correctly understand the points made by @GaryJones, it is said that:  
1. `wp_script_is()` could be enough for our testing, as well as that testing the output is "out of our scope".
2. If we stick to testing output, we should also make sure that any footer-injected scripts are taken into account.

## Point 1
Looking at the code, I see that tests use a combination of output checking **and** `wp_script_is()`. `wp_script_is()` should definitely be checked, but the only way to really know if something actually made it to output is output examination. So I think that using both together can be considered as a valid approach.

Having said that, I also do agree that this is in WP scope and that we're testing something that's "out of our control and business". It would be easy to just remove output checking and it would mitigate point 2 altogether, but I'd rather have feedback from contributors first as I wouldn't like to decide by myself on this.

## Point 2

I made some tryouts using `wp_print_scripts()`, `wp_print_footer_scripts()` and `wp_footer()` functions. `wp_footer` gave me errors and I gave up on it quite quickly on the basis that `wp_print_footer_scripts()` is much more fitting for our use case and that `wp_footer` could result in unpredictable output.

I used `wp_print_scripts()` and `wp_print_footer_scripts()` in combination (in both orders) as well as isolated, with scripts being included in the footer and header. My findings were that they behaved exactly the same, and that when calling both, the second call wouldn't give any output. I was a bit surprised by that.

So from my findings, using `wp_print_scripts()` should be enough for our needs, but we could add both functions for good measure as it's easy and doesn't introduce any side effects (at least from my own testing).

# Final notes

- I find that test messages like "Failed to confirm nothing was printed by register_scripts()" can be confusing for me. I think that going for something like "register_scripts() unexpected output" or "register_scripts() should not produce any output" is more easily understandable and I've started modifying test messages towards that direction. Let me know if I shouldn't.
- So far I've made changes only to the `test_parsely_register_scripts()` and `test_enqueue_js_tracker()` functions just to give you an idea where I'm going with this.
- Let me know of any comments or feedback so we can decide on the final direction this will take.

Thanks!

### DIFF
--- a/tests/Integration/ScriptsTest.php
+++ b/tests/Integration/ScriptsTest.php
@@ -139,6 +139,37 @@ final class ScriptsTest extends TestCase {
 	}
 
 	/**
+	 * Test the wp_parsely_enqueue_js_tracker filter
+	 * When it returns false, the tracking script should not be enqueued.
+	 *
+	 * @covers \Parsely\Scripts::enqueue_js_tracker
+	 * @uses \Parsely\Parsely::api_key_is_missing
+	 * @uses \Parsely\Parsely::api_key_is_set
+	 * @uses \Parsely\Parsely::get_options
+	 * @uses \Parsely\Parsely::post_has_trackable_status
+	 * @uses \Parsely\Parsely::update_metadata_endpoint
+	 */
+	public function test_enqueue_js_tracker_filter(): void {
+		add_filter( 'wp_parsely_load_js_tracker', '__return_false' );
+
+		$post_array = $this->create_test_post_array();
+		$post       = $this->factory->post->create( $post_array );
+		$this->go_to( '/?p=' . $post );
+		self::$scripts->register_scripts();
+		self::$scripts->enqueue_js_tracker();
+
+		// Confirm that tracker script is registered but not enqueued.
+		self::assertTrue(
+			wp_script_is( 'wp-parsely-tracker', 'registered' ),
+			'Script wp-parsely-tracker was not registered'
+		);
+		self::assertFalse(
+			wp_script_is( 'wp-parsely-tracker', 'enqueued' ),
+			'Script wp-parsely-tracker should not be enqueued'
+		);
+	}
+
+	/**
 	 * Test the API init script enqueue.
 	 *
 	 * @covers \Parsely\Scripts::enqueue_js_api
@@ -290,37 +321,6 @@ final class ScriptsTest extends TestCase {
 		self::assertTrue(
 			wp_script_is( 'wp-parsely-tracker', 'enqueued' ),
 			'Failed to confirm tracker script was enqueued'
-		);
-	}
-
-	/**
-	 * Test the wp_parsely_enqueue_js_tracker filter
-	 * When it returns false, the tracking script should not be enqueued.
-	 *
-	 * @covers \Parsely\Scripts::enqueue_js_tracker
-	 * @uses \Parsely\Parsely::api_key_is_missing
-	 * @uses \Parsely\Parsely::api_key_is_set
-	 * @uses \Parsely\Parsely::get_options
-	 * @uses \Parsely\Parsely::post_has_trackable_status
-	 * @uses \Parsely\Parsely::update_metadata_endpoint
-	 */
-	public function test_enqueue_js_tracker_filter(): void {
-		add_filter( 'wp_parsely_load_js_tracker', '__return_false' );
-
-		$post_array = $this->create_test_post_array();
-		$post       = $this->factory->post->create( $post_array );
-		$this->go_to( '/?p=' . $post );
-		self::$scripts->register_scripts();
-		self::$scripts->enqueue_js_tracker();
-
-		// Confirm that tracker script is registered but not enqueued.
-		self::assertTrue(
-			wp_script_is( 'wp-parsely-tracker', 'registered' ),
-			'Script wp-parsely-tracker was not registered'
-		);
-		self::assertFalse(
-			wp_script_is( 'wp-parsely-tracker', 'enqueued' ),
-			'Script wp-parsely-tracker should not be enqueued'
 		);
 	}
 }
--- a/tests/Integration/ScriptsTest.php
+++ b/tests/Integration/ScriptsTest.php
@@ -248,12 +248,18 @@ final class ScriptsTest extends TestCase {
 	 * @group settings
 	 */
 	public function test_do_not_track_logged_in_users(): void {
-		TestCase::set_options( array( 'track_authenticated_users' => false ) );
+		TestCase::set_options(
+			array(
+				'api_secret'                => 'hunter2',
+				'track_authenticated_users' => false,
+			)
+		);
 		$new_user = $this->create_test_user( 'bill_brasky' );
 		wp_set_current_user( $new_user );
 
 		self::$scripts->register_scripts();
 		self::$scripts->enqueue_js_tracker();
+		self::$scripts->enqueue_js_api();
 
 		// As track_authenticated_users options is false, enqueue should fail.
 		// Confirm that tracker script is registered but not enqueued.
@@ -261,6 +267,13 @@ final class ScriptsTest extends TestCase {
 			'wp-parsely-tracker',
 			array( 'registered' ),
 			array( 'enqueued' )
+		);
+
+		// API should be unaffected by track_authenticated_users setting.
+		// Confirm that API script is registered and enqueued.
+		$this->assert_script_statuses(
+			'wp-parsely-api',
+			array( 'registered', 'enqueued' )
 		);
 	}
 
--- a/tests/Integration/ScriptsTest.php
+++ b/tests/Integration/ScriptsTest.php
@@ -88,10 +88,40 @@ final class ScriptsTest extends TestCase {
 		self::$scripts->register_scripts();
 		self::$scripts->enqueue_js_tracker();
 
-		// Confirm that JS tracker script is registered and enqueued.
+		// Confirm that tracker script is registered and enqueued.
 		$this->assert_script_statuses(
 			'wp-parsely-tracker',
 			array( 'registered', 'enqueued' )
+		);
+	}
+
+	/**
+	 * Test the tracker script enqueue with the disable_javascript option set
+	 * to true.
+	 *
+	 * @covers \Parsely\Scripts::enqueue_js_tracker
+	 * @uses \Parsely\Parsely::get_asset_cache_buster
+	 * @uses \Parsely\Parsely::api_key_is_missing
+	 * @uses \Parsely\Parsely::api_key_is_set
+	 * @uses \Parsely\Parsely::get_options
+	 * @uses \Parsely\Parsely::post_has_trackable_status
+	 * @uses \Parsely\Parsely::update_metadata_endpoint
+	 * @uses \Parsely\Scripts::register_scripts
+	 * @uses \Parsely\Scripts::script_loader_tag
+	 * @group enqueue-js
+	 */
+	public function test_enqueue_js_tracker_with_javascript_option_disabled(): void {
+		TestCase::set_options( array( 'disable_javascript' => true ) );
+
+		$this->go_to_new_post();
+		self::$scripts->register_scripts();
+		self::$scripts->enqueue_js_tracker();
+
+		// Confirm that tracker script is registered but not enqueued.
+		$this->assert_script_statuses(
+			'wp-parsely-tracker',
+			array( 'registered' ),
+			array( 'enqueued' )
 		);
 	}
 
@@ -282,7 +312,7 @@ final class ScriptsTest extends TestCase {
 		self::assertEquals( get_current_blog_id(), $first_blog );
 		self::assertTrue( is_user_member_of_blog( $first_blog_admin, $first_blog ) );
 
-		// Enqueue JS tracker.
+		// Enqueue tracker script.
 		self::$scripts->register_scripts();
 		self::$scripts->enqueue_js_tracker();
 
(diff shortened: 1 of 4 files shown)


jblz:
The changes already in this PR make sense to me.

As to the output checking, since we have e2e tests in place, I think it probably makes good sense to cover the situational presence and lack there of for the script tags with puppeteer selectors on the front end.

GaryJones:
> the only way to really know if something actually made it to output is output examination

But that logic not our code to test. Our plugin tells WP core what the enqueue, and we leave it to WordPress core to build the markup and output it, in whatever shape or method it desires.

Let's consider it another way - if we just had a completely unconditional `wp_enqueue_script()` call like below, would we set up a test for it?

```php
add_action( 'init', function() { wp_enqueue_script(...) }
```

I would say we wouldn't - because if we can't trust WordPress to get that right in our tests, we have bigger problems.

As such, the only reason we want to have tests, is to ensure that the conditions we wrap around whether to enqueue various scripts, is correct or not, and doesn't accidentally break if some other part or condition (maybe elsewhere in the plugin) starts interfering. That means, we only care if the script is enqueued, by the time the output would have started (which is different to "just after the enqueueing method has been called).

acicovic:
Thank you for you input, points well taken!

I'll be stripping out the output tests and improving commenting as I go forward. Don't hesitate to comment at any point.

Thanks!

tests/Integration/ScriptsTest.php
```diff
@@ -219,110 +222,52 @@ public function test_enqueue_js_api_no_secret(): void {
...
-
-		self::assertFalse(
-			wp_script_is( 'wp-parsely-api', 'registered' ),
-			'Failed to confirm API script was not registered'
-		);
-
-		self::assertFalse(
-			wp_script_is( 'wp-parsely-api', 'enqueued' ),
-			'Failed to confirm API script was not enqueued'
-		);
```
GaryJones, line 302: This is no longer being tested - is that your intent?

GaryJones (dismissed):
One query, but otherwise looking great.

tests/Integration/ScriptsTest.php
acicovic: Yes, I left this out because the test is aimed to test what happens with the tracker when the user is logged in and the `track_authenticated_users` option is disabled. So I considered testing the API a bit irrelevant and removed it. Let me know if you'd prefer to have it though.

GaryJones: The previous test was checking what happened to logged in users, compared to non-logged in users, so the presence of checking the status of the API before made sense.

If I asked you to prove (through tests) that the API script is _not_ affected for logged in users, we can no longer do that.

acicovic: I've included an assertion to verify that that API is unaffected for logged-in users. Feel free to resolve the conversation or comment if you think anything additional would be needed.

tests/Integration/ScriptsTest.php
```diff
@@ -334,131 +292,94 @@ public function test_user_logged_in(): void {
...
 	 *
-	 * @covers \Parsely\Scripts::enqueue_js_tracker
-	 * @uses \Parsely\Parsely::api_key_is_missing
-	 * @uses \Parsely\Parsely::api_key_is_set
-	 * @uses \Parsely\Parsely::get_options
-	 * @uses \Parsely\Parsely::post_has_trackable_status
-	 * @uses \Parsely\Parsely::update_metadata_endpoint
+	 * @param string $handle Script handle to test.
+	 * @param array  $assert_true Statuses that should assert to true.
+	 * @param array  $assert_false Statuses that should assert to false.
```
GaryJones:
```suggestion
	 * @param string $handle       Script handle to test.
	 * @param array  $assert_true  Optional. Statuses that should assert to true. Accepts 'enqueued', 'registered', 'queue', 'to_do', and 'done'. Default is an empty array.
	 * @param array  $assert_false Optional. Statuses that should assert to false. Accepts 'enqueued', 'registered', 'queue', 'to_do', and 'done'. Default is an empty array.
```
Do we need to make this method slightly more protected, so that someone doesn't try to take advantage of the array args being optional and simply call `assert__script_statuses( 'foo' )` and get a pass? 

(This may already be handled via no assertions being called, therefore it is marked as [Risky](https://phpunit.readthedocs.io/en/9.5/risky-tests.html#useless-tests) automatically.)

tests/Integration/ScriptsTest.php
```diff
@@ -177,34 +196,18 @@ public function test_enqueue_js_tracker_with_cloudflare(): void {
...
 	 * @uses \Parsely\Parsely::update_metadata_endpoint
 	 * @uses \Parsely\Scripts::register_scripts
-	 * @group insert-js
+	 * @group enqueue-js
 	 */
 	public function test_enqueue_js_api_no_secret(): void {
-		ob_start();
-		$post_array = $this->create_test_post_array();
-		$post       = $this->factory->post->create( $post_array );
-		$this->go_to( '/?p=' . $post );
```
GaryJones, line 186: Does this need a `$this->go_to_new_post();` replacement or not?

tests/Integration/ScriptsTest.php
```diff
@@ -219,110 +222,65 @@ public function test_enqueue_js_api_no_secret(): void {
...
 	 * @uses \Parsely\Scripts::register_scripts
 	 * @uses \Parsely\Scripts::script_loader_tag
-	 * @group insert-js
+	 * @group enqueue-js
 	 */
 	public function test_enqueue_js_api_with_secret(): void {
-		ob_start();
-		$post_array = $this->create_test_post_array();
-		$post       = $this->factory->post->create( $post_array );
-		$this->go_to( '/?p=' . $post );
```
GaryJones, line 228: Does this need a `$this->go_to_new_post();` replacement?

tests/Integration/ScriptsTest.php
```diff
@@ -144,15 +137,14 @@ public function test_enqueue_js_tracker(): void {
 	 * @uses \Parsely\Parsely::update_metadata_endpoint
 	 * @uses \Parsely\Scripts::register_scripts
 	 * @uses \Parsely\Scripts::script_loader_tag
+	 * @group enqueue-js
 	 * @group insert-js
 	 */
 	public function test_enqueue_js_tracker_with_cloudflare(): void {
```
GaryJones: This test method isn't checking the enqueueing, but the correct building of the output markup based on a filter value. As such, it seems like this test (the only one left with output buffering) could be moved out of the middle of the other tests that check for correct enqueueing, and the method renamed to indicate it's more about the markup. This test is checking two lines of https://github.com/Parsely/wp-parsely/blob/b9cf935255c63e707ff5b86dc0ecb9d142ab1e46/src/class-scripts.php#L172.

tests/Integration/ScriptsTest.php
```diff
@@ -51,40 +49,23 @@ public function set_up(): void {
...
 	 * @uses \Parsely\Parsely::update_metadata_endpoint
-	 * @group insert-js
+	 * @group enqueue-js
 	 */
 	public function test_parsely_register_scripts(): void {
-		ob_start();
-		$post_array = $this->create_test_post_array();
-		$post       = $this->factory->post->create( $post_array );
-		$this->go_to( '/?p=' . $post );
 		self::$scripts->register_scripts();
```
GaryJones, line 70: Can you add a couple of `assert_script_statuses()` to show that both scripts are not even `registered`, before calling this `register_scripts()`?

tests/Integration/ScriptsTest.php
```diff
@@ -51,40 +49,23 @@ public function set_up(): void {
 	 * @uses \Parsely\Parsely::api_key_is_set
 	 * @uses \Parsely\Parsely::get_options
 	 * @uses \Parsely\Parsely::update_metadata_endpoint
-	 * @group insert-js
+	 * @group enqueue-js
```
GaryJones: Since this script handling is now moved to its own file and own test file, we can probably simplify these `@group` throughout this file down to `scripts`.

GaryJones (changes requested):
A few minor suggestions and questions, but I really like where this is going.

tests/Integration/ScriptsTest.php
acicovic: I left it out because the only thing the API seems to need is an API secret key. Adding or removing a `$this->go_to_new_post();` doesn't seem to have any influence on the test and the functionality of `enqueue_js_api()`.

tests/Integration/ScriptsTest.php
acicovic: I left it out because the only thing the API seems to need is an API secret key. Adding or removing a `$this->go_to_new_post();` doesn't seem to have any influence on the test and the functionality of `enqueue_js_api()`.

tests/Integration/ScriptsTest.php
acicovic: Correct, I will do this. We could also move it to another file which would be dedicated to any output-specific tests.

tests/Integration/ScriptsTest.php
acicovic: Sure!

tests/Integration/ScriptsTest.php
acicovic: Agreed.

tests/Integration/ScriptsTest.php
acicovic: We could do this. Would you suggest a particular approach to this?

GaryJones: If the test is run without the args, does it get flagged as Risky? If so, no change is needed, since I think we have that configured as a fail (if not, we should).

acicovic: Passing no arguments does flag a test as risky. However, this is not the case if we have multiple calls to `assert_script_statuses` and at least one of them has arguments, as the test does assert something in that case.

tests/Integration/ScriptsTest.php
```diff
@@ -382,4 +348,39 @@ public function assert_script_statuses( string $handle, array $assert_true = arr
...
+	 * @uses \Parsely\Parsely::api_key_is_set
+	 * @uses \Parsely\Parsely::get_options
+	 * @uses \Parsely\Parsely::post_has_trackable_status
+	 * @uses \Parsely\Parsely::update_metadata_endpoint
+	 * @uses \Parsely\Scripts::register_scripts
+	 * @uses \Parsely\Scripts::script_loader_tag
+	 * @group enqueue-js
+	 * @group insert-js
+	 */
+	public function test_tracker_html_markup_when_cfasync_filter_is_called(): void {
```
GaryJones:
```suggestion
	public function test_tracker_markup_has_attribute_when_cfasync_filter_is_used(): void {
```

This change will read slightly better with Testdox format.

tests/Integration/ScriptsTest.php
GaryJones: Good point - in which case, having a check inside `assert_script_statuses()` itself and mark test as risky / incomplete seems sensible then.

tests/Integration/ScriptsTest.php
acicovic: This has not been marked as resolved yet, so I'm not sure if there's anything additional to do. I have updated the `@group` designation and have also moved the function towards the bottom.

tests/Integration/ScriptsTest.php
acicovic: Done

tests/Integration/ScriptsTest.php
```diff
@@ -334,131 +272,134 @@ public function test_user_logged_in(): void {
...
+	/**
+	 * Assert multiple enqueueing statuses for a script.
+	 *
+	 * @param string $handle       Script handle to test.
+	 * @param array  $assert_true  Optional. Statuses that should assert to true. Accepts 'enqueued', 'registered', 'queue', 'to_do', and 'done'. Default is an empty array.
+	 * @param array  $assert_false Optional. Statuses that should assert to false. Accepts 'enqueued', 'registered', 'queue', 'to_do', and 'done'. Default is an empty array.
+	 *
+	 * @throws RiskyTestError If no assertions ($assert_true, $assert_false) get passed to the function.
+	 */
+	public function assert_script_statuses( string $handle, array $assert_true = array(), array $assert_false = array() ): void {
```
GaryJones, line 386: I think we can change the type declarations from `array` to the wider `iterable` (PHP 7.1). We're only looping through the collections; we're not using access the values through array syntax. This potentially gives other tests the ability to use anything that implements Traversable

```suggestion
	public function assert_script_statuses( string $handle, iterable $assert_true = array(), iterable $assert_false = array() ): void {
```

If it still works, then the docs need updating as well.

I'm rusty on whether `count()` will work on a Traversable / Iterator though, or if there would be another way to detect that the object was empty of values - if that makes it messy, ignore this suggestion and stick with `array`.

GaryJones (approved):
Approved with one suggestion.

tests/Integration/ScriptsTest.php
acicovic: I checked on this and would prefer to stick with `array`. From my research, `Traversable` doesn't automatically imply `Countable` (only some `Traversable` objects implement `Countable`). So adding `iterable` would imply a broader scope of possibilities and introduce more complexity. Using `iterator_to_array()` within the function probably defeats the purpose, and empty() has its own issues and would probably need array conversion as well. So I wouldn't go forward with this unless we were to find a really good reason for it.

pauarge (approved):
🚢 🚢 🚢

PR #560, opened 2021-12-24
Fix incorrect last modified date in json+ld

### LINKED ISSUE #558: Last Modified date is wrong in the json+ld
## Describe the bug
The last modified date in the json+ld output by the parsely plugin is wrong. It always outputs the created date due to the if statement.

On: 
https://github.com/Parsely/wp-parsely/blob/7c372ca9e26bb3c0af14d2b81af355644ed2cea3/src/class-parsely.php#L452
https://github.com/Parsely/wp-parsely/blob/7c372ca9e26bb3c0af14d2b81af355644ed2cea3/src/class-parsely.php#L453

The function code is using: get_the_modified_date( 'U', true )

The documentation: https://developer.wordpress.org/reference/functions/get_the_modified_date/ says that the second argument should be post object or post ID. Using true returns a bad value. Changing true to $post should fix the issue although will probably not be in GMT time.

Given the code, I'm assuming you probably meant to use: get_post_modified_time  
https://developer.wordpress.org/reference/functions/get_post_modified_time/

In which case swapping out the functions would work fine with the original code (although still probably should explicitly pass in the $post object for the third argument).

## To Reproduce
Go to any page with parsely plugin enabled and the view the json+ld data.

## Expected behaviour
Returns the last modified date of the post.

### LINKED ISSUE #562: Reports of fatal error coming from plugin
Multiple Parse.ly users have reported the following error coming from the plugin:

```
Fatal error: Uncaught Error: gmdate() expects parameter 2 to be int, bool given
in /var/www/wp-content/plugins/wp-parsely/src/class-parsely.php on line 450
```

One shared data from its log files. I've obscured the customer name:

```
2021-12-27 14:08:23.821171 [NOTICE] [9118] [T0] [204.237.142.109:33881#APVH_xxxxxxx:443 (http://204.237.142.109:33881/#APVH_xxxxxxx:443) ] [STDERR] PHP Fatal error: Uncaught TypeError: gmdate() expects parameter 2 to be int, bool given in /var/www/vhosts/xxxxxxx/httpdocs/wp-content/plugins/wp-parsely/src/class-parsely.php:450

Stack trace:

#0 /var/www/vhosts/xxxxxxx/httpdocs/wp-content/plugins/wp-parsely/src/class-parsely.php(450): gmdate()

#1 /var/www/vhosts/xxxxxxx/httpdocs/wp-content/plugins/wp-parsely/src/class-parsely.php(189): Parsely\Parsely->construct_parsely_metadata()

#2 /var/www/vhosts/xxxxxxx/httpdocs/wp-includes/class-wp-hook.php(292): Parsely\Parsely->insert_page_header_metadata()

#3 /var/www/vhosts/xxxxxxx/httpdocs/wp-includes/class-wp-hook.php(316): WP_Hook->apply_filters()

#4 /var/www/vhosts/xxxxxxx/httpdocs/wp-includes/plugin.php(484): WP_Hook->do_action()

#5 /var/www/vhosts/xxxxxxx/httpdocs/wp-includes/general-template.php(3009): do_action()

#6 /var/www/vhosts/xxxxxxx/httpdocs/wp-content/themes/tpd-theme/header.php(18): wp_head()

#7 /var/www/vhosts/xxxxxxx/httpdocs/wp-includes/template.php(730): require_once('/var/www in /var/www/vhosts/xxxxxxx/httpdocs/wp-content/plugins/wp-parsely/src/class-parsely.php on line 450

When we de-activated the plugin, the site performance returned to normal levels.
```

acicovic:
## Description
This PR attempts to resolve an issue where the last modified date in the json+ld output could be wrong.  Additionally, this fixes potential fatal errors caused by passing `false` into the second param of the `gmdate` function now that we're enforcing strict types for function parameters.

Fixes #558
Fixes #562

## Motivation and Context
More details about the issue can be found in #558. This is something that should be fixed and we should aim to add a test for it.

## How Has This Been Tested?
So far only by examining browser output and by examining variables using step debugging, but will also look into writing a test for this.

### DIFF
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -449,8 +449,8 @@ class Parsely {
 			);
 			$parsely_page['dateCreated']      = gmdate( 'Y-m-d\TH:i:s\Z', get_post_time( 'U', true, $post ) );
 			$parsely_page['datePublished']    = gmdate( 'Y-m-d\TH:i:s\Z', get_post_time( 'U', true, $post ) );
-			if ( get_the_modified_date( 'U', true ) >= get_post_time( 'U', true, $post ) ) {
-				$parsely_page['dateModified'] = gmdate( 'Y-m-d\TH:i:s\Z', get_the_modified_date( 'U', true ) );
+			if ( get_post_modified_time( 'U', true, $post ) >= get_post_time( 'U', true, $post ) ) {
+				$parsely_page['dateModified'] = gmdate( 'Y-m-d\TH:i:s\Z', get_post_modified_time( 'U', true, $post ) );
 			} else {
 				// Use the post time as the earliest possible modification date.
 				$parsely_page['dateModified'] = gmdate( 'Y-m-d\TH:i:s\Z', get_post_time( 'U', true, $post ) );
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -502,19 +502,15 @@ class Parsely {
 			return;
 		}
 
-		$post_time_gmt = gmdate( $date_format, $post_time );
+		$post_time_created_gmt = gmdate( $date_format, $post_time );
+		$post_time_modified    = get_post_modified_time( 'U', true, $post );
 
-		// Set post created and published time.
-		$metadata['dateCreated']   = $post_time_gmt;
-		$metadata['datePublished'] = $post_time_gmt;
+		$metadata['dateCreated']   = $post_time_created_gmt;
+		$metadata['datePublished'] = $post_time_created_gmt;
+		$metadata['dateModified']  = $post_time_created_gmt;
 
-		// Set post modified time.
-		$metadata['dateModified'] = $post_time_gmt;
-
-		$post_modified_time = get_post_modified_time( 'U', true, $post );
-
-		if ( false !== $post_modified_time && $post_modified_time > $post_time ) {
-			$metadata['dateModified'] = gmdate( $date_format, $post_modified_time );
+		if ( false !== $post_time_modified && $post_time_modified > $post_time ) {
+			$metadata['dateModified'] = gmdate( $date_format, $post_time_modified );
 		}
 	}
 
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -491,7 +491,7 @@ class Parsely {
 	 * Sets all metadata values related to post time.
 	 *
 	 * @param array   $metadata Array containing all metadata. It will be potentially mutated to add keys: dateCreated, dateModified, & datePublished.
-	 * @param WP_Post $post Post object from which to extract time data.
+	 * @param WP_Post $post     Post object from which to extract time data.
 	 * @return void
 	 */
 	private function set_metadata_post_times( array &$metadata, WP_Post $post ): void {
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -447,14 +447,9 @@ class Parsely {
 				'@type' => 'ImageObject',
 				'url'   => $image_url,
 			);
-			$parsely_page['dateCreated']      = gmdate( 'Y-m-d\TH:i:s\Z', get_post_time( 'U', true, $post ) );
-			$parsely_page['datePublished']    = gmdate( 'Y-m-d\TH:i:s\Z', get_post_time( 'U', true, $post ) );
-			if ( get_post_modified_time( 'U', true, $post ) >= get_post_time( 'U', true, $post ) ) {
-				$parsely_page['dateModified'] = gmdate( 'Y-m-d\TH:i:s\Z', get_post_modified_time( 'U', true, $post ) );
-			} else {
-				// Use the post time as the earliest possible modification date.
-				$parsely_page['dateModified'] = gmdate( 'Y-m-d\TH:i:s\Z', get_post_time( 'U', true, $post ) );
-			}
+
+			$this->set_metadata_post_times( $parsely_page, $post );
+
 			$parsely_page['articleSection'] = $category;
 			$author_objects                 = array();
 			foreach ( $authors as $author ) {
@@ -490,6 +485,31 @@ class Parsely {
 		 * @param array   $parsely_options The Parsely options.
 		 */
 		return apply_filters( 'wp_parsely_metadata', $parsely_page, $post, $parsely_options );
+	}
+
+	/**
+	 * Sets all metadata values related to post time.
+	 *
+	 * @param array   $metadata Array containing all metadata.
+	 * @param WP_Post $post Post object from which to extract time data.
+	 * @return void
+	 */
+	private function set_metadata_post_times( array &$metadata, WP_Post $post ): void {
+		// Initializations.
+		$date_format   = 'Y-m-d\TH:i:s\Z';
+		$post_time     = get_post_time( 'U', true, $post );
+		$post_time_gmt = gmdate( $date_format, $post_time );
+
+		// Set post created and published time.
+		$metadata['dateCreated']   = $post_time_gmt;
+		$metadata['datePublished'] = $post_time_gmt;
+
+		// Set post modified time.
+		$metadata['dateModified'] = $post_time_gmt;
+		$post_modified_time       = get_post_modified_time( 'U', true, $post );
+		if ( false !== $post_modified_time && $post_modified_time > $post_time ) {
+			$metadata['dateModified'] = $post_modified_time;
+		}
 	}
 
 	/**
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -495,22 +495,21 @@ class Parsely {
 	 * @return void
 	 */
 	private function set_metadata_post_times( array &$metadata, WP_Post $post ): void {
-		$date_format = 'Y-m-d\TH:i:s\Z';
-		$post_time   = get_post_time( 'U', true, $post );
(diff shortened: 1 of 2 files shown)


GaryJones:
Can we avoid calling the same function with the same args twice, and also defensively check it didn't return false before passing it to a function that requires an integer?

acicovic:
@GaryJones I've updated this based on your feedback and took the liberty to extract it as a function. Let me know should you have any feedback.

src/class-parsely.php
```diff
@@ -492,6 +487,31 @@ public function construct_parsely_metadata( array $parsely_options, WP_Post $pos
...
+	 * Sets all metadata values related to post time.
+	 *
+	 * @param array   $metadata Array containing all metadata.
+	 * @param WP_Post $post Post object from which to extract time data.
+	 * @return void
+	 */
+	private function set_metadata_post_times( array &$metadata, WP_Post $post ): void {
+		// Initializations.
+		$date_format   = 'Y-m-d\TH:i:s\Z';
+		$post_time     = get_post_time( 'U', true, $post );
```
GaryJones: As per https://developer.wordpress.org/reference/functions/get_post_time/ - this function can return `false`, so a guard needs adding here.

src/class-parsely.php
```diff
@@ -492,6 +487,31 @@ public function construct_parsely_metadata( array $parsely_options, WP_Post $pos
 		return apply_filters( 'wp_parsely_metadata', $parsely_page, $post, $parsely_options );
 	}
 
+	/**
+	 * Sets all metadata values related to post time.
+	 *
+	 * @param array   $metadata Array containing all metadata.
+	 * @param WP_Post $post Post object from which to extract time data.
```
```suggestion
	 * @param array   $metadata Array containing all metadata.
	 * @param WP_Post $post     Post object from which to extract time data.
```

GaryJones (dismissed):

tests/Integration/StructuredData/SinglePostTest.php
```diff
@@ -430,6 +430,191 @@ public function test_http_canonicals_for_single_post(): void {
...
+		self::assertSame( $date_created, $metadata['dateModified'] );
+
+		// Update the post and reload metadata.
+		wp_update_post( array( 'ID' => $post_id ) );
+		$post_updated     = get_post( $post_id );
+		$metadata_updated = $parsely->construct_parsely_metadata( $parsely_options, $post_updated );
+
+		// In the metadata, check that the last modified date has been updated.
+		self::assertSame( $date_created, $metadata_updated['dateCreated'] );
+		self::assertNotSame( $date_created, $metadata_updated['dateModified'] );
```
jblz, line 482: This assertion missed that `dateModified` was not formatted to a date string -- it should check for a particular value (or at least a meaningful relationship to the `$date_created`) instead of calling `assertNotSame`.

jblz (changes requested):
I've added some fixes & more specific tests.

I think we can probably pull out `test_metadata_post_modified_date` altogether as it's covered by the other tests.

I avoided juggling `wp_update_post`, etc. and just manually annotated the `WP_Post` object which I think will suffice for these tests, but I'm happy to hear why it won't if there's reason.

tests/Integration/StructuredData/SinglePostTest.php
acicovic: I have no objections about `test_metadata_post_modified_date()` being removed if reviewers think it was rendered redundant.

My opinion (and rationale behind this implementation) is that in all integration tests we should use WP functions as much as possible, since in a regular workflow the data is manipulated by WP functions, and as such we should test with the data received by them. So I don't find "modified object" tests bad, but testing with WP built-in functions is a "reality check" that should be done IMO.

This doesn't rule out modifying objects with extreme or unexpected values to see what happens. However, the premise is that WP functions are upstream and should have their own tests to verify their quality and results. If their tests are lacking, they should be added there.

So regarding the missed assertion:

- I do expect `wp_update_post()` to add a correct date (in format and value) to the database.
- As I could not hardcode a time using `wp_update_post()`, the time could (at least theoretically) change at the time of comparison. But something that could be done to improve this is compare that modified time is in a 1-2 second window from the current time. But even this could fail under certain circumstances (for example performance problems).
- Given the previous point, I decided just to test for `NotSame` and that this would be good enough.

Having said that, team feedback is more important than my own opinion, so if any adjustments must be made or the test must be removed, that's fine by me.

jblz: > in all integration tests we should use WP functions as much as possible, since in a regular workflow the data is manipulated by WP functions, and as such we should test with the data received by them

I agree in general. If `set_metadata_post_times` (& by proxy `construct_parsely_metadata`) were operating against the "current" post or some global, I'd probably be in favor of going the extra mile to call `wp_update_post`. Since they just take a post object, I'm less pressed about it, personally.

Maybe we leave both and move the tests I added to `tests/Unit`...?  Your call.

GaryJones: > Maybe we leave both and move the tests I added to `tests/Unit`...? 

By definition of the tests including `$this->factory->post->create();`, they can't be unit tests. You'd have to instantiate `WP_Post` directly, and I suspect that might be painful to populate without a DB entry to pull from.

src/class-parsely.php
```diff
@@ -492,6 +487,37 @@ public function construct_parsely_metadata( array $parsely_options, WP_Post $pos
...
 
+	/**
+	 * Sets all metadata values related to post time.
+	 *
+	 * @param array   $metadata Array containing all metadata. It will be potentially mutated to add keys: dateCreated, dateModified, & datePublished.
+	 * @param WP_Post $post     Post object from which to extract time data.
+	 * @return void
+	 */
+	private function set_metadata_post_times( array &$metadata, WP_Post $post ): void {
+		$date_format = 'Y-m-d\TH:i:s\Z';
```
pauarge: Could we extract this to a constant within the class?

src/class-parsely.php
```diff
@@ -492,6 +487,37 @@ public function construct_parsely_metadata( array $parsely_options, WP_Post $pos
...
+		$post_time   = get_post_time( 'U', true, $post );
+
+		if ( false === $post_time ) {
+			return;
+		}
+
+		$post_time_gmt = gmdate( $date_format, $post_time );
+
+		// Set post created and published time.
```
pauarge: I would get rid of this comment since it is quite straight forward what the code is doing.

src/class-parsely.php
```diff
@@ -492,6 +487,37 @@ public function construct_parsely_metadata( array $parsely_options, WP_Post $pos
...
+		$metadata['dateCreated']   = $post_time_gmt;
+		$metadata['datePublished'] = $post_time_gmt;
+
+		// Set post modified time.
+		$metadata['dateModified'] = $post_time_gmt;
```
pauarge: Could we set this a inside (for instance) and `else` in the later conditional? We are setting this variable twice, when ideally we should check for the condition and then set it accordingly.

pauarge (changes requested):
Thanks for working on this. I left some possible refinements in the code.

pauarge:
Also, we should first merge @jblz 's PR into this one and then this one into `develop`.

src/class-parsely.php
GaryJones: What benefit does that serve?

src/class-parsely.php
GaryJones: Since it's going to be set anyway, it's better practice to avoid `else` clauses - [set it once, and overwrite it if necessary](https://phpmd.org/rules/cleancode.html#elseexpression). It reduces the [cyclomatic complexity](https://www.axelerant.com/blog/reducing-cyclomatic-complexity-and-npath-complexity-steps-for-refactoring).

pauarge: I see your point on cyclomatic complexity and I agree with it. My comment comes from trying to avoid doing useless overwrites, which do waste resources (granted, not many of them).

src/class-parsely.php
pauarge: I see this is only used inside this function. Dismiss my comment.

src/class-parsely.php
acicovic: @pauarge can you recheck the function and see if it is OK with you?
I also in general tend to avoid if/else when possible.

tests/Integration/StructuredData/SinglePostTest.php
acicovic: I'm not very clear on what needs to be done here. I suggest leaving all tests as they are (no move to unit tests) unless something needs to be removed. If we start adding lots of tests for metadata perhaps it would be good to split the file into many parts (?) - though splitting the file is not in the scope of this PR.

pauarge (approved):

acicovic:
Credit: The commit https://github.com/Parsely/wp-parsely/pull/560/commits/11167e95d519acf39fbe63b50f20dc0cfa9eeb75 is based on the work in https://github.com/Parsely/wp-parsely/pull/565 by @jblz.

tests/Integration/StructuredData/SinglePostTest.php
jblz: > I suggest leaving all tests as they are

Works for me

jblz (approved):

PR #565, opened 2021-12-31
Metadata: Format dates & directly compare instead of making extra gmdate calls

jblz:
## Description

Since our date format is comparable (ISO 8601), we can avoid the intermediate step of pulling and comparing timestamps and just pass the format string to `get_post_time` & `get_post_modified_time` directly. 

NOTE: This is targeting the branch in #560. 

## Motivation and Context

Fewer LOC, fewer variables to grok, etc.

## How Has This Been Tested?

Automated tests still pass. Fatal errors still not present when we coerce the post creation and modification date values to `false` via filter

### DIFF
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -495,14 +495,12 @@ class Parsely {
 	 * @return void
 	 */
 	private function set_metadata_post_times( array &$metadata, WP_Post $post ): void {
-		$date_format = 'Y-m-d\TH:i:s\Z';
-		$post_time   = get_post_time( 'U', true, $post );
+		$date_format   = 'Y-m-d\TH:i:s\Z';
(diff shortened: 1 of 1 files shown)


GaryJones:
Worth checking against https://make.wordpress.org/core/2019/09/23/date-time-improvements-wp-5-3/ and see if we can improve anything further around this.

pauarge (approved):
Nice improvement 👍

acicovic:
Thanks @jblz, great suggestion

I've implemented this as part of #558 in https://github.com/Parsely/wp-parsely/pull/560/commits/11167e95d519acf39fbe63b50f20dc0cfa9eeb75

jblz:
incorporated in upstream branch

PR #569, opened 2022-01-03
Adding Parse.ly stats button on admin bar

pauarge:
## Description

This PR introduces a _Parsely Stats_ button visible on the top WordPress admin bar whenever the user is logged in and in a post or page that they can edit. The button redirects to the Parse.ly Dash, similarly to what's happening in the row actions link on the posts list.

This PR also creates a utilities class to place reusable methods there.

## Motivation and Context

Improving the pre-publish experience by giving editors a quick way to see stats on a post.

## How Has This Been Tested?

Go to a post that the current user can edit while being logged in. Click on the _Parsely Stats_ button and expect to be redirected to the Parse.ly Dash.

_Notice the Parse.ly Stats button on the top bar._

Screen Shot 2022-01-04 at 15 28 05


pauarge:
Opening for review, but I'm still missing some tests.

src/class-utils.php
```diff
@@ -0,0 +1,43 @@
...
+namespace Parsely;
+
+use WP_Post;
+
+/**
+ * Utility methods to be reused across the codebase.
+ *
+ * @since 3.2.0
+ */
+class Utils {
```
jblz: > This PR also creates a utilities class to place reusable methods there.

Personally, I vote we avoid having a class called something generic like `Utils` to avoid it becoming catch-all file. If it's important enough to become a named function, we should probably be able to come up with a grouping for it.

e.g.

```suggestion
class Parsely_Dashboard_Links {
```

...or similar.

src/class-utils.php
```diff
@@ -0,0 +1,43 @@
...
+	 * Generate the URL for the link.
+	 *
+	 * @since 2.6.0
+	 * @since 3.2.0 Moved to class-utils.php. Renamed from `generate_url`.
+	 *
+	 * @param WP_Post $post   Which post object or ID to check.
+	 * @param string  $apikey API key or empty string.
+	 * @return string
+	 */
+	public static function generate_parsely_post_url( WP_Post $post, string $apikey ): string {
```
jblz: If the containing class is more descriptive, this function name can be a bit more terse:

```suggestion
	public static function post_stats_url( WP_Post $post, string $apikey ): string {
```

...making the call signature something like: `Parsely_Dashboard_Links::post_stats_url( ...`

src/class-utils.php
```diff
@@ -0,0 +1,43 @@
...
+		$query_args = array(
+			'url'          => rawurlencode( get_permalink( $post ) ),
+			'utm_campaign' => 'wp-admin-posts-list',
+			'utm_medium'   => 'wp-parsely',
+			'utm_source'   => 'wp-admin',
```
jblz: We should consider making `utm_source` a function param.

If the new interaction is happening via the logged-in toolbar on the frontend, it's definitely not from `wp-admin` (at least as far as the browser and user are concerned).  I'd probably pass in `admin-bar` or similar in the `admin_bar_parsely_stats_button` function.

src/UI/class-admin-bar.php
```diff
@@ -0,0 +1,78 @@
...
+	 * @return void
+	 */
+	public function admin_bar_parsely_stats_button( WP_Admin_Bar $admin_bar ): void {
+		$current_object = $GLOBALS['wp_the_query']->get_queried_object();
+
+		if ( empty( $current_object ) ) {
+			return;
+		}
+
+		if ( ! empty( $current_object->post_type ) ) {
```
jblz: Minor, but I'd combine these two cases and avoid the indentation level:

```suggestion
		if ( empty( $current_object ) || empty( $current_object->post_type ) ) {
			return;
		}
```

jblz: Should we also..?

* Consult the `track_post_types` option and only show the link if it's a "trackable" type?
* Consult `Parsely::post_has_trackable_status` and only show the link if it's published, etc?

I can see arguments for and against both.

jblz (changes requested):
Nice!  This is working pretty well.

I left some feedback.  I'd say the only real blocker is distinguishing the `utm_source` from the other usage.

src/UI/class-admin-bar.php
```diff
@@ -0,0 +1,78 @@
...
+			$post_type_object = get_post_type_object( $current_object->post_type );
+			if ( $post_type_object && current_user_can( 'edit_post', $current_object->ID ) && $post_type_object->show_in_admin_bar ) {
```
jblz: `edit_post` is more restrictive than our "Post row actions" links:

https://github.com/Parsely/wp-parsely/blob/b667f6f707e215b267485658289f972274e2d54a/src/UI/class-row-actions.php#L54-L55

Should we also limit the post row actions to those with `edit_post` capabilities? As is, we show the links to "Contributors" (even for posts they haven't authored).

...or should we loosen this to allow the link to show when the current user is the post author?  I think we should make the behavior consistent regardless.

src/class-utils.php
pauarge: I don't fully agree with this. I'm not sure we'll have more functions related to generate dashboard links in the near future, so a full-blown class for this kind of functions seems overkill.

The other option I considered was to simply put it in `class-parsely.php`, which has effectively become a catch-all file. When we tackle splitting that file, we might want to reconsider this.

src/class-utils.php
pauarge: This goes very tight with the comment above.

src/class-utils.php
pauarge: I like that.

pauarge: I'd even do a similar argument for `utm_campaign`. I will add both as arguments.

src/UI/class-admin-bar.php
pauarge: I agree we should be consistent, and I would go down the road of being less restrictive (as we've been doing up until now). I've removed the `can_edit` post check and added the `has_trackable_status` that you suggest in the other comment. 

With this, we should have the same conditions on both admin bar and posts list.

src/UI/class-admin-bar.php
pauarge: I think the second option is straight forward to implement and by calling it through `cannot_show_parsely_link`, is in line with the row action.

src/class-dashboard-links.php
```diff
@@ -0,0 +1,59 @@
...
+	 * Determine whether Parse.ly dashboard link should be shown or not.
+	 *
+	 * @since 2.6.0
+	 * @since 3.2.0 Moved to class-utils.php. Renamed from `cannot_show_parsely_link`.
+	 *
+	 * @param WP_Post $post    Which post object or ID to check.
+	 * @param Parsely $parsely Parsely object.
+	 * @return bool True if the link cannot be shown, false if the link can be shown.
+	 */
+	public static function cannot_show_link( WP_Post $post, Parsely $parsely ): bool {
```
GaryJones: Can we make this a positive check `can_show_link()`? Current usage has a case of `! cannot_show_link()` (double-negative).

src/class-dashboard-links.php
```diff
@@ -0,0 +1,59 @@
...
+namespace Parsely;
+
+use WP_Post;
+
+/**
+ * Utility methods to build and generate dashboard links.
+ *
+ * @since 3.2.0
+ */
+class Dashboard_Links {
```
GaryJones: Right now, the methods are only dealing with the creation/handling of single links at a time (rather than a collection of dashboard links), so I think class name (and file) should be referenced in the singular (`Dashboard_Link`).

tests/Integration/DashboardLinksTest.php
```diff
@@ -0,0 +1,150 @@
...
+	 *
+	 * @covers \Parsely\UI\Row_Actions::cannot_show_parsely_link
+	 * @uses \Parsely\Parsely::api_key_is_set
+	 * @uses \Parsely\Parsely::api_key_is_missing
+	 * @uses \Parsely\Parsely::get_options
+	 * @uses \Parsely\Parsely::post_has_trackable_status
+	 * @uses \Parsely\Parsely::update_metadata_endpoint
+	 * @group ui
+	 */
+	public function test_can_correctly_determine_if_Parsely_link_can_be_shown_when_actions_are_an_array_or_not(): void {
```
GaryJones: > array_or_not

Which is it? What's being tested for here?

(`or` in a test name is a code smell IMO)

tests/Integration/DashboardLinksTest.php
```diff
@@ -0,0 +1,150 @@
...
+	 */
+	public function test_can_correctly_determine_if_Parsely_link_can_be_shown_when_post_is_viewable_or_not(): void {
+		$non_publicly_queryable_post = self::factory()->post->create_and_get( array( 'post_type' => 'parsely_tests_pt' ) );
+		$published_post              = self::factory()->post->create_and_get();
+
+		self::set_options( array( 'apikey' => 'somekey' ) );
+
+		// Test if post is not viewable status.
+		self::assertTrue( Dashboard_Links::cannot_show_link( $non_publicly_queryable_post, self::$parsely ) );
+		self::assertFalse( Dashboard_Links::cannot_show_link( $published_post, self::$parsely ) );
```
GaryJones: This assertion is testing exactly the same thing as line 97 and 72. i.e. a viewable and trackable (not sure how a post can be trackable if it's not viewable...) post with an array of row actions.

GaryJones (changes requested):

src/class-dashboard-links.php
pauarge: Agreed. I was trying to be conservative in terms of the function, but avoiding the double negative makes it more readable.

tests/Integration/UI/AdminBarTest.php
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * @package Parsely
+ */
+
+declare(strict_types=1);
+
+namespace Parsely\Tests\Integration\UI;
+
+// Explicit require as the `Rest` class could not be found in some environments.
+require_once __DIR__ . '/../../../src/UI/class-admin-bar.php';
```
jblz: Is there any more background on why we have to explicitly require UI classes such as this in an issue / PR somewhere? Just curious.

jblz (dismissed):
LGTM!

tests/Integration/UI/AdminBarTest.php
pauarge: I have removed this line as it works fine in CI. This should be tackled in #586

GaryJones (dismissed):
✅

pauarge:
Please, do not merge this PR before 3.1 has been released.

GaryJones:
> Please, do not merge this PR before 3.1 has been released.

Use a release branch for the 3.1 work, and then you don't hold up continued development on `develop` ;-)

acicovic (approved):

PR #570, opened 2022-01-03
Allow user to choose logo using media library in plugin settings

### LINKED ISSUE #227: Logo setting - allow to choose from Media library
Screenshot 2021-04-16 at 00 19 25

The Logo setting is currently a text input field (and a narrow one at that) - we can make it more useful by letting users select an image directly from the media library.

acicovic:
## Description
- In the current settings page, the only way to specify a logo is to type or copy/paste its path into the respective textbox.
- This PR attempts to provide the ability to select the logo from within WP's media library.

## Motivation and Context
- The motivation is to render the experience of selecting a logo more user-friendly.
- Fixes #227.

## How Has This Been Tested?
- Manual local testing.
- Using the included end-to-end test.

## Screenshots
Before:
image

After:
image


acicovic:
Getting this out of draft status, but still needs work to implement an end to end test.

src/UI/class-settings-page.php
```diff
@@ -598,6 +612,30 @@ public function print_multiple_checkboxes( array $args ): void {
...
 
+	/**
+	 * Print out a "single-image browse control" which includes a text
+	 * input to store image path and a button to browse for images.
+	 *
+	 * @param array $args The arguments for the control.
+	 * @return void
+	 */
+	public function print_media_single_image( array $args ) {
+		$key         = esc_attr( $args['option_key'] );
```
GaryJones: The output escaping is already included in the markup below - it's not needed here as well.

src/UI/class-settings-page.php
```diff
@@ -598,6 +612,30 @@ public function print_multiple_checkboxes( array $args ): void {
...
+		$input_value = $this->parsely->get_options()[ $key ];
+		$input_name  = Parsely::OPTIONS_KEY . "[$key]";
+		$button_text = __( 'Browse', 'wp-parsely' );
+		?>
+
+		<fieldset class="media-single-image" id="media-single-image-<?php echo esc_attr( $key ); ?>">
+			<legend class="screen-reader-text"><span><?php echo esc_html( $args['title'] ); ?></span></legend>
+			<input class="large-text file-path" type="text" name="<?php echo esc_attr( $input_name ); ?>" value="<?php echo esc_attr( $input_value ); ?>" />
+			<button data-option="<?php echo esc_attr( $key ); ?>" class="browse button" style="margin-top: 0.4rem" type="button"><?php echo esc_html( $button_text ); ?></button>
```
GaryJones: Does the custom style still work zoomed into or out of the page (CMD-+)?

src/js/admin/index.js
```diff
@@ -0,0 +1,25 @@
+document.querySelector( '.media-single-image button.browse' ).addEventListener( 'click', selectImage );
+
+function selectImage() {
+	const optionName = this.dataset.option;
+	let imageFrame = null;
+
+	if ( imageFrame ) {
```
GaryJones: In what usecase would imageFrame be true-ish here?

GaryJones (dismissed):

src/UI/class-settings-page.php
acicovic: Here's what I get in Firefox and Chromium. I don't see inconsistencies with the other elements as I zoom in/out. So my response to this question would be yes.

image

image

image

src/UI/class-settings-page.php
acicovic: We currently do this consistently in many other places in the code (for example see `$id` variable in `print_select_tag()`, `print_binary_radio_tag()`, `print_text_tag()` etc.

I assume that removing this doesn't have any security risk since we escape when outputting it, but if we are to do it then it would be better to do it everywhere to keep consistency. Else it just looks like an omission.

Would you still go for removing this?

src/js/admin/index.js
acicovic: None 😄

Removed this.

src/UI/class-settings-page.php
pauarge: Not a huge fan of having inlined custom styles. Do we really need them or can we use some wp-admin default class? Could we not have it at all? If both answers are no, could we move this into a CSS file?

acicovic: Not a fan of inline styles either but I made this exception because this is really a single statement and I didn't think this would warrant a file of its own. Having said that:

- I can move it to wp-parsely.css.
- I can put this in a wp-admin CSS file that would be useful if we're thinking of introducing more styling in the admin in the future.

Another approach (that WP employs but IMO find worse from all the above) would be to put the button in a `<p>` tag. That's what WP does for the "Save Changes" button in settings. But it doesn't require external files or inline styles. 😄

Edit: forgot to say that I didn't find any WP built-in classes for this.

pauarge: I like the `<p>` idea, actually.

src/UI/class-settings-page.php
GaryJones: Yes - escaping twice is unnecessary.

Escaping should be done as late as possible (i.e. when pushing into echoed markup). Until we know the _context_ with which a value be displayed (HTML vs attribute vs XML, etc.) then we can't confidently escape it, so this escaping on line 623 is incorrect.

src/UI/class-settings-page.php
acicovic: Custom style is not relevant anymore so I'm marking this as resolved.

acicovic: Done! I've also updated the PR's screenshot since the spacing is a bit less.

src/UI/class-settings-page.php
acicovic: Pretty good point. Removed.

pauarge:
Could we make the logo box the same length as the boxes above? 
Screen Shot 2022-01-20 at 3 15 42 PM

src/js/admin/index.js
```diff
@@ -0,0 +1,20 @@
...
+
+	const imageFrame = wp.media( {
+		multiple: false,
+		library: {
+			type: 'image',
+		},
+	} );
+
+	imageFrame.on( 'select', function() {
+		const url = imageFrame.state().get( 'selection' ).first().toJSON().url;
```
pauarge, line 14: I would be a bit more defensive in here. Every call here is nullable, so could we have some kind of checks to ensure that nothing fails? If something fails, I would display an `alert`.

src/js/admin/index.js
```diff
@@ -0,0 +1,20 @@
...
+		const inputSelector = '#media-single-image-' + optionName + ' input.file-path';
+		document.querySelector( inputSelector ).value = url;
```
pauarge, line 16: Something similar here. What if we can't select the DOM element?

tests/e2e/specs/browse-logo-button.spec.js
```diff
@@ -0,0 +1,118 @@
+/**
+ * External dependencies
+ */
+const resolve = require( 'path' ).resolve;
```
pauarge:
```suggestion
import * as path from 'path';
```

tests/e2e/specs/browse-logo-button.spec.js
```diff
@@ -0,0 +1,118 @@
...
+	visitAdminPage,
+} from '@wordpress/e2e-test-utils';
+
+/**
+ * Internal dependencies
+ */
+import { waitForWpAdmin } from '../utils';
+
+// General initializations.
+const imageLocalPath = resolve( __dirname, '../../../.wordpress-org/icon-256x256.png' );
```
pauarge:
```suggestion
const imageLocalPath = path.resolve( __dirname, '../../../.wordpress-org/icon-256x256.png' );
```

tests/e2e/specs/browse-logo-button.spec.js
```diff
@@ -0,0 +1,118 @@
...
+// Edit attachment modal selectors.
+const modalEditAttachment = 'div.media-modal div.edit-attachment-frame';
+const modalDeleteAttachmentLink = `${ modalEditAttachment } button.delete-attachment`;
+
+/**
+ * Browse button tests
+ */
+describe( 'Browse for logo button', () => {
+	/**
+	 * Before all tests, login and activate the Parse.ly plugin.
```
pauarge:
```suggestion
	 * Login and activate the Parse.ly plugin.
```

tests/e2e/specs/browse-logo-button.spec.js
```diff
@@ -0,0 +1,118 @@
...
+	 */
+	beforeAll( async () => {
+		await loginUser();
+		await activatePlugin( 'wp-parsely' );
+	} );
+
+	/**
+	 * After all tests, remove the uploaded image.
```
pauarge:
```suggestion
	 * Remove the uploaded image.
```

tests/e2e/specs/browse-logo-button.spec.js
```diff
@@ -0,0 +1,118 @@
...
+		// Confirm image deletion.
+		// Note: Dialog handling must be placed before the click event.
+		await page.on( 'dialog', async ( dialog ) => {
+			await dialog.accept();
+		} );
+		await page.click( modalDeleteAttachmentLink );
+	} );
+
+	/**
+	 * Before each test, go to settings page and click the browse button.
```
pauarge:
```suggestion
	 * Go to settings page and click the browse button.
```

tests/e2e/specs/browse-logo-button.spec.js
```diff
@@ -0,0 +1,118 @@
...
+	 */
+	beforeEach( async () => {
+	// Go to settings page (resets page selections).
```
pauarge: This comments are redundant given the function description.

acicovic:
Regarding textbox size, I did this on purpose (refer to the PR's screenshots to see why). I Personally prefer a large textbox to show the URL instead of having to click in it and move left/right with the arrow keys. But let me know if you think I should resize it.

pauarge:
> Regarding textbox size, I did this on purpose (refer to the PR's screenshots to see why). I Personally prefer a large textbox to show the URL instead of having to click in it and move left/right with the arrow keys. But let me know if you think I should resize it.

The site logo is not a parameter that will be modified often. Also, in most of the cases you won't even have to touch the URL since it comes from the media uploader. To me, it is more important to prioritize visual cohesion (especially on large monitors) than addressing an edge case.

acicovic:
> The site logo is not a parameter that will be modified often. Also, in most of the cases you won't even have to touch the URL since it comes from the media uploader. To me, it is more important to prioritize visual cohesion (especially on large monitors) than addressing an edge case.

I do think that current textboxes are too narrow. Look for example the default textbox size in `/wp-admin/options-general.php` and `/wp-admin/options-writing.php` pages. Also notice that in the second page there is a full-width textarea for URLs which could be much shorter than what we'll have in our own textbox. Let me know if you still want this.

pauarge:
I don't necessarily agree that they are too narrow. I don't feel too strongly about that, though. What is relevant for me is that they all look the same so they maintain consistency. I think changing all of them to be wide it's outside the scope of this PR, so I'd stick to having them all narrow and then let's make them wider in a separate piece of work if needed.

src/js/admin/index.js
acicovic: - Displaying alerts with messages would imply JS localization. Do you want us to go that route?
- I also think the above aren't dealbreakers. If the browser can't use the media library correctly as well as detect the input, the user already has much bigger problems. Are you confirming we should get into that level of checking?

src/js/admin/index.js
```diff
@@ -0,0 +1,20 @@
+document.querySelector( '.media-single-image button.browse' ).addEventListener( 'click', selectImage );
```
jblz, line 1: I think we should reinstate the built admin page js file that was removed in #518 to keep consistency with the rest of the scripts and provide the expectation of a common runtime when working in JS.

That would make it fairly straightforward to use the [MediaUpload](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-media-utils/#mediaupload) component ([usage](https://github.com/WordPress/gutenberg/blob/9675b99c6e7c88af16749acc3dd0dcdfbe35db27/packages/block-editor/src/components/media-upload/README.md)).

I'm ok if we land this in develop and see what that approach would look like in a follow-up branch, though.

src/js/admin/index.js
pauarge: You have a point. The properties come from a fairly stable object, so let's go with your solution. Resolving this conversation for now.

src/js/admin/index.js
pauarge: I like the idea, but maybe this is a bit out of scope of this PR. Maybe we could create an issue and we tackle it later?

pauarge (approved):
🚢 🚢 🚢

src/js/admin/index.js
acicovic: Related discussion should be tracked in #617 from now on, so marking this as resolved.

pauarge:
Great work Alex!

PR #582, opened 2022-01-10
Adding network support 

### LINKED ISSUE #575: Support Network Installation
https://www.parse.ly/help/integration/wordpress mentions that:

> Installation of the Parse.ly WordPress plugin must be done individually per site. A network install will not work because each site needs to specify the site ID.

We should make handling that task as simple as possible, so that it can be network activated successfully.

@jblz says that a simple network enable / site ID setting UI would be super straightforward.

We might also consider providing a solution or documentation on how someone can use a pre-defined set of non-default settings across all subsites (other than than the known-to-be-unique apikey and API secret).

We could do nothing and not support network installation, but that hinders adoption - considers customer applications with hundreds of subsites, and the pain there would be in activating them and configuring them individually (even with bash scripting with WP-CLI).

## Additional context
We only have one integration test that requires multisite.

pauarge:
## Description

Closes #575

### DIFF
--- a/src/UI/class-parsely-sites-table.php
+++ b/src/UI/class-parsely-sites-table.php
@@ -37,6 +37,7 @@ final class Parsely_Sites_Table extends WP_List_Table {
 	public function get_columns(): array {
 		return array(
 			'blog_id'     => __( 'Blog ID', 'wp-parsely' ),
+			'blog_domain' => __('Blog Domain', 'wp-parsely'),
 			'blog_name'   => __( 'Blog Name', 'wp-parsely' ),
 			'api_key_set' => __( 'Status', 'wp-parsely' ),
 			'settings'    => __( 'Settings', 'wp-admin' ),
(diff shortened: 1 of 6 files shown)


wp-parsely.php
```diff
@@ -86,6 +87,16 @@ function(): void {
 	}
 );
 
+require __DIR__ . '/src/UI/class-settings-page-network.php';
+require __DIR__ . '/src/UI/class-parsely-sites-table.php';
+add_action(
+	'_network_admin_menu',
```
jblz: Is there a good reason to hook into `_network_admin_menu` instead of `network_admin_menu`?

src/UI/class-parsely-sites-table.php
```diff
@@ -0,0 +1,66 @@
...
+		);
+	}
+
+	public function prepare_items(): void {
+		$this->_column_headers = $this->get_column_info();
+		$this->items = $this->fetch_table_data();
+	}
+
+	public function column_default( $item, $column_name ): string {
+		return strval($item[$column_name]);
```
jblz:
```suggestion
		return strval( $item[$column_name] );
```

jblz (commented):
Great work so far!

src/UI/class-parsely-sites-table.php
```diff
@@ -0,0 +1,85 @@
...
+
+		return strval( $item[ $column_name ] );
+	}
+
+	/**
+	 * @return array
+	 */
+	private function fetch_table_data(): array {
+		$parsely_network_sites = array();
+		foreach ( get_sites() as $site ) {
```
jblz: I think `get_sites` will cap at 100 results per:
https://github.com/WordPress/wordpress-develop/blob/5.8.1/src/wp-includes/class-wp-site-query.php#L176

So, we should probably either paginate or add an indication somewhere that only the first 100 results will be shown.  I think the latter is fine.  If someone is running that many subsites, they're likely going to be reaching for a scripting approach to management.

wp-parsely.php
pauarge: It would collide with the action in `class-settings-page-network.php`. I've added a change that gets rid of it and plays with priorities, but I'm not sure about it.

src/UI/class-parsely-sites-table.php
pauarge: I see _number_ is a parameter, so we could hypothetically bump that number up. I would expect a performance drop if that number is very big, though.

I like your messaging proposal. We can implement pagination later if we see a lot of demand for it.

wp-parsely.php
jblz: Ah. I see. `_network_admin_menu` works for me, then!

pauarge:
Closing in favor of #583

PR #583, opened 2022-01-10
Show settings link & API Key in Network Admin sites list

jblz:
## Description

In the Multisite Site Listing Table for each subsite:
* Include a link to the Parsely Settings page in the "row actions"
* Include info about the API key to see at-a-glance which subsites need to be configured

https://user-images.githubusercontent.com/1587282/149400665-9e76d93b-111c-4407-98b9-147ad884873b.mov

## Motivation and Context

This seeks to provide a good multisite management experience (#575)

## How Has This Been Tested?

* Configure a new multisite (can be either subdirectory or subdomain)
* Configure some subsites
* Browse to `/wp-admin/network/sites.php`
  * Note that all sites (including the main site) inform you that the `Parse.ly API key is missing`
* Hover over a site and click `Parse.ly Settings`
* Populate the API Key field with some test string
* Return to `/wp-admin/network/sites.php`
  * Confirm that the API key is reflected in the appropriate column in the table
* Rinse & repeat -- confirm the interactions are otherwise ideal


wp-parsely.php
```diff
@@ -86,6 +86,33 @@ function(): void {
 	}
 );
 
+add_action(
```
jblz: I'm not 100% sure this will get merged, so I'm punting on putting functionality in a proper file for now.

jblz:
SonarCloud says:

> Rename class "Network_Admin_Sites_List" to match the regular expression ^[A-Z][a-zA-Z0-9]*$.

I don't think that that regex matches our sniffs.

src/UI/class-network-admin-sites-list.php
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	/**
+	 * Attach network admin page functionality to the appropriate action and filter hooks.
+	 *
+	 * @since 3.2.0
+	 *
+	 * @return void
+	 */
+	public function run() {
```
pauarge:
```suggestion
	public function run(): void {
```

src/UI/class-network-admin-sites-list.php
```diff
@@ -0,0 +1,110 @@
...
+	 * @return array The list of actions including ours.
+	 */
+	public static function add_action_link( array $actions, int $_blog_id ): array {
+		if ( ! current_user_can( Parsely::CAPABILITY ) ) {
+			return $actions;
+		}
+
+		if ( ! is_numeric( $_blog_id ) ) {
+			return $actions;
+		}
```
pauarge: What do you think about merging this two conditionals into one?

src/UI/class-network-admin-sites-list.php
```diff
@@ -0,0 +1,110 @@
...
+
+		$settings_url                = Parsely::get_settings_url( (int) $_blog_id );
```
pauarge: Blog id should always be an integer if we're using strict types, so I think we don't need the casting here.

src/UI/class-network-admin-sites-list.php
```diff
@@ -0,0 +1,110 @@
...
+		$actions['parsely-settings'] = '<a href="' . esc_url( $settings_url ) . '">' .
```
pauarge: Question -- should we add `target='_blank'` on the link?

src/UI/class-network-admin-sites-list.php
```diff
@@ -0,0 +1,110 @@
...
+	 *
+	 * @param string $column_name The column name for the current context.
+	 * @param int    $_blog_id The blog ID for the current context.
+	 *
+	 * @since 3.2.0
+	 *
+	 * @return void
+	 */
+	public function populate_api_key_column( string $column_name, int $_blog_id ): void {
+		if ( 'parsely-api-key' !== $column_name ) {
```
pauarge: We could move the `parsely-api-key` as a constant of this class to signify that it should be same string in the different functions.

pauarge (changes requested):
Changes looking great! Left some small comments, but if we fix those and add some tests this is ready to be shipped 🚢

src/UI/class-network-admin-sites-list.php
GaryJones: I'd go with keeping them separate. The conditions are different contexts - one is about the state of the environment (whether logged in user has certain capabilities), while the other is about the data type of a specific variable.

src/UI/class-network-admin-sites-list.php
GaryJones: I'd strongly prefer not to - that creates an accessibility issue. Folks know how to open a link in a new tab if that's what they want - let's not interfere with the default behaviour of their browser.

src/UI/class-network-admin-sites-list.php
```diff
@@ -0,0 +1,110 @@
...
+		add_action( 'manage_sites_custom_column', array( $this, 'populate_api_key_column' ), 10, 2 );
+	}
+
+	/**
+	 * Use the manage_sites_action_links filter to append a link to the settings page in the "row actions."
+	 *
+	 * @param array $actions The list of actions meant to be displayed for the current site's context in the row actions.
+	 * @param int   $_blog_id The blog ID for the current context.
+	 *
+	 * @since 3.2.0
```
GaryJones, line 48: `@since` should go after the summary and any description.

`@param` should be right before (no clear linespace) the `@returm`.

src/UI/class-network-admin-sites-list.php
```diff
@@ -0,0 +1,110 @@
...
+
+	/**
+	 * Use the manage_sites_custom_column action to output each site's API Key (if configured).
+	 *
+	 * @param string $column_name The column name for the current context.
+	 * @param int    $_blog_id The blog ID for the current context.
+	 *
+	 * @since 3.2.0
+	 *
+	 * @return void
```
GaryJones: As above (order of tags).

src/UI/class-network-admin-sites-list.php
```diff
@@ -0,0 +1,110 @@
...
+
+		// phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.switch_to_blog_switch_to_blog
+		switch_to_blog( $_blog_id );
+		$apikey = $this->parsely->get_api_key();
+		restore_current_blog();
+
+		if ( strlen( $apikey ) > 0 ) {
+			echo esc_html( $apikey );
+		} else {
+			echo '<em>' . esc_html( 'Parse.ly API key is missing' ) . '</em>';
```
GaryJones: This string isn't internationalised.

src/UI/class-network-admin-sites-list.php
```diff
@@ -0,0 +1,110 @@
...
+	 */
+	public function populate_api_key_column( string $column_name, int $_blog_id ): void {
+		if ( 'parsely-api-key' !== $column_name ) {
+			return;
+		}
+
+		// phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.switch_to_blog_switch_to_blog
+		switch_to_blog( $_blog_id );
+		$apikey = $this->parsely->get_api_key();
+		restore_current_blog();
```
GaryJones, line 117: If a site has 100 or 4000 sites, how painful will this bit of code be? Does the database retrieval across all of the options tables for the subsites need to be cached in the object cache, or stored in an option in the root site, if say, the number of subsites is above a certain threshold?

src/class-parsely.php
```diff
@@ -937,10 +937,12 @@ public function get_clean_parsely_page_value( string $val ): string {
...
 	 * Get the URL of the plugin settings page.
 	 *
+	 * @param int $_blog_id The Blog ID for the multisite subsite to use for context (Default null for current).
+	 *
 	 * @return string
 	 */
-	public static function get_settings_url(): string {
-		return admin_url( 'options-general.php?page=' . self::MENU_SLUG );
+	public static function get_settings_url( int $_blog_id = null ): string {
+		return get_admin_url( $_blog_id, 'options-general.php?page=' . self::MENU_SLUG );
```
GaryJones, line 982: Have we got any multisite tests for this that could be updated?

GaryJones (dismissed):

src/UI/class-network-admin-sites-list.php
jblz: +1 on leaving off the link target. Maybe if we were sending users into an entirely separate context (e.g. the Parsely dashboard), but this is still an intra-(multi)site link. There's navigation in the admin bar to return to this screen or we can just use the back button to return.

src/UI/class-network-admin-sites-list.php
jblz: I could go either way, so I'll default to leaving it as is :)

src/UI/class-network-admin-sites-list.php
jblz: Ah, cool.  I didn't realize that adding the `int` type declaration still coerces a numeric string to an `int` with strict types declared. :TIL:

I had encountered a TypeError prior to adding the type declarations because the function was passed a string for the blog id.

That means I can take the `is_numeric` check out above.  Neat.

src/UI/class-network-admin-sites-list.php
jblz: > If a site has 100 or 4000 sites, how painful will this bit of code be? 

Short answer is:
It's probably ok.

Here's some context:

1. The list table is [paginated](https://github.com/WordPress/WordPress/blob/270f2011f8ec7265c3f4ddce39c77ef5b496ed1c/wp-admin/network/sites.php#L17-L18), so I think it will be capped at 100 sites at a time (the [default](https://github.com/WordPress/WordPress/blob/fa62df07749a3bfaa44e944c02512215ab70a5db/wp-includes/class-wp-site-query.php#L127) for a `WP_Site_Query`).
1. `Parsely::get_api_key` is a pretty lightweight wrapper around `get_option`, which is [backed by an object cache](https://github.com/WordPress/WordPress/blob/2a3ad7c4304143c4761726c7c49beda995d6f2f0/wp-includes/option.php#L171).
1. `get_option` [calls](https://github.com/WordPress/WordPress/blob/2a3ad7c4304143c4761726c7c49beda995d6f2f0/wp-includes/option.php#L166) `wp_load_alloptions` prior to the lookup. That result is also [backed](https://github.com/WordPress/WordPress/blob/2a3ad7c4304143c4761726c7c49beda995d6f2f0/wp-includes/option.php#L279-L280) by object cache. Our option is set to autoload by default, so, this is the codepath that will provide the data in most cases when the option is set for a subsite.
1. For subsites without the option set, a [query](https://github.com/WordPress/WordPress/blob/2a3ad7c4304143c4761726c7c49beda995d6f2f0/wp-includes/option.php#L174) will be executed the first time the option is attempted to be read for each subsite.  When the key is not found, `get_option` stores that fact in [another cache](https://github.com/WordPress/WordPress/blob/2a3ad7c4304143c4761726c7c49beda995d6f2f0/wp-includes/option.php#L146-L186), so this query will not happen on subsequent calls to `get_option`.
1. The option query linked above is super fast thanks to a `UNIQUE KEY` on `` `wp_#_options`.`option_name` ``, so 100 times shouldn't be a huge deal. The bottleneck is likely the IO between the web server & db as opposed to any processing load on the db.
1. The option will be populated and cache hot the very first time a subsite is accessed in any manner owing to [the way `Parsely::run` works](https://github.com/Parsely/wp-parsely/blob/f536ee6e61a738a4843e2b7aa532c0321b440211/src/class-parsely.php#L98-L109)

All that to say, I don't think this will have a meaningful impact on site performance -- particularly on a production site with persistent caching.

I'd say the largest impact we could expect on a large multisite would be while network-activating the plugin. The first time a subsite is accessed in any manner afterward, the option will be written and cache populated.

pauarge (dismissed):
LGTM

.github/workflows/integration-tests.yml
```diff
@@ -92,6 +92,10 @@ jobs:
         if: ${{ matrix.php != 8.0 }}
         run: composer testwp --no-interaction
 
+      - name: Run integration tests (multi site)
+        if: ${{ matrix.php != 8.0 }}
+        run: composer testwp-ms --no-interaction
+
```
GaryJones: Why is it not being run for PHP 8.0?

jblz: > Why is it not being run for PHP 8.0?

I wasn't entirely sure about that -- I copied the block above. I'm not 100% sure how the coverage portion works, but it is running the Integration tests with multisite enabled.  Happy to remove the `if` line here, though.

src/class-parsely.php
jblz: we do now.

.github/workflows/integration-tests.yml
pauarge: +1 on removing the line if it works.

wp-parsely.php
```diff
@@ -97,6 +98,15 @@ function parsely_admin_menu_register(): void {
...
 
+require __DIR__ . '/src/UI/class-network-admin-sites-list.php';
+add_action(
+	'admin_init',
+	function(): void {
+		$network_admin_sites_list = new Network_Admin_Sites_List( $GLOBALS['parsely'] );
+		$network_admin_sites_list->run();
+	}
+);
+
```
pauarge: As per the changes in 3.0.4, this function should be a named function.

pauarge (changes requested):
Looking super promising. It just needs a small change.

src/UI/class-network-admin-sites-list.php
```diff
@@ -0,0 +1,104 @@
...
+	 */
+	public static function add_action_link( array $actions, int $_blog_id ): array {
+		if ( ! current_user_can( Parsely::CAPABILITY ) ) {
+			return $actions;
+		}
+
+		$actions['parsely-settings'] =
+			'<a href="' . esc_url( Parsely::get_settings_url( $_blog_id ) ) . '">' .
+				__( 'Parse.ly Settings', 'wp-parsely' ) .
+			'</a>';
```
GaryJones: This will generate links with the exact same computed anchor text, but pointing to different URLs, which is an accessibility issue. It needs some screen reader text to differentiate them.

See https://github.com/Parsely/wp-parsely/blob/e54b25bfd10fea86cc1751b8419dc710a024d249/src/UI/class-row-actions.php#L106-L113

tests/Integration/UI/NetworkAdminSitesListTest.php
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * @package Parsely
+ */
+
+declare(strict_types=1);
+
+namespace Parsely\Tests\Integration\UI;
+
+// Explicit require as the `Network_Admin_Sites_List` class could not be found in test environment.
+require_once __DIR__ . '/../../../src/UI/class-network-admin-sites-list.php';
```
GaryJones: This shouldn't be needed. Probably needs a `composer dumpautoload` to make it be found.

tests/Integration/UI/NetworkAdminSitesListTest.php
```diff
@@ -0,0 +1,116 @@
...
+
+use Parsely\Parsely;
+use Parsely\Tests\Integration\TestCase;
+use Parsely\UI\Network_Admin_Sites_List;
+use \WP_Site;
```
GaryJones: Why does this one have a leading `\` when the others don't?

tests/Integration/UI/NetworkAdminSitesListTest.php
```diff
@@ -0,0 +1,116 @@
...
+	 * @covers \Parsely\UI\Network_Admin_Sites_List::run
+	 * @uses \Parsely\UI\Network_Admin_Sites_List::__construct
+	 * @return void
+	 */
+	public function test_api_key_column_is_present(): void {
+		self::$sites_list->run();
+		$columns = $this->table->get_columns();
+
+		self::assertArrayHasKey( 'parsely-api-key', $columns );
+		self::assertSame( 'Parse.ly API Key', $columns['parsely-api-key'] );
```
GaryJones: Can you test that the array didn't have the key before you called `->run()`?

( I know it's highly unlikely that core will ever add a column key by that name, but if we're going to prove that a function did something, let's also show that it wasn't already like that anyway.)

GaryJones (dismissed):
Some minor bits.

GaryJones:
Wondering if this could be registered to Screen Options, so that folks have the opportunity to hide the column if they don't need the data? They may already have a bunch of other plugin-added columns, and seeing the state of Parse.ly is not their focus.

pauarge (dismissed):
The PR looks good to me, great work! I agree with the minor suggestions that @GaryJones is saying, but other than that we're good to merge.

jblz:
> Wondering if this could be registered to Screen Options, so that folks have the opportunity to hide the column if they don't need the data?

 Cool!  Apparently, the column is already automatically toggle-able via Screen Options:

https://user-images.githubusercontent.com/1587282/151018133-c0769bef-6c42-4b1e-85c4-ce23da6754bf.mov

pauarge (approved):
LGTM

PR #588, opened 2022-01-11
Fix fatal TypeError in get_categories function

### LINKED ISSUE #587: Fatal TypeError reported in get_categories function
## Describe the bug

We have received a report from a customer. When they attempt to upgrade to version 3.0.2, they are met with an error:

`explode() expects parameter 2 to be string, bool given in ... src/class-parsely.php:675`

Per logs, the function was called thusly:

`explode( '/', false )`

Here's the pertinent line:

https://github.com/Parsely/wp-parsely/blob/77d032037d00624600e823f516880f10b16ea7a2/src/class-parsely.php#L675

The customer also added context:

> The posts (namely, links to posts) stopped working when navigating from the Home page. The archive pages worked correctly, and the posts opened correctly when accessing from the archive pages, but not from the Home page.

## To Reproduce

No fatals.
Links present, correct, & functioning.

## Additional context

140584-zd-wordpressvip
p1641908716117600-slack-C011HP930Q7

pauarge:
## Description

When a site had the plugin configured with the categories as tags option, the plugin would fail if that post wouldn't have categories. This was because we were calling `end()` of an empty array.

This PR adds a check to prevent that from failing plus two integration tests to prevent this regression from happening again. It is also addressing some other type issues in which a `WP_Error` might be returned and passed into a function that would not accept it.

## Motivation and Context

Closes #587

## How Has This Been Tested?

Integration tests.

### DIFF
--- a/tests/Integration/OtherTest.php
+++ b/tests/Integration/OtherTest.php
@@ -248,7 +248,6 @@ final class OtherTest extends TestCase {
 		self::assertTrue( $result );
 	}
 
-
 	/**
 	 * Tests if keywords in Parse.ly metadata are generated correctly when categories are tags.
 	 *
@@ -258,14 +257,14 @@ final class OtherTest extends TestCase {
 	 * @uses \Parsely\Parsely::get_categories
 	 */
 	public function test_post_with_categories_as_tags(): void {
-		self::set_options( array('cats_as_tags' => true) );
+		self::set_options( array( 'cats_as_tags' => true ) );
 		$options = self::$parsely->get_options();
 
 		// Create a single post.
 		$post_id = $this->factory->post->create();
 		$post    = get_post( $post_id );
 
-		$expected = array( "uncategorized", );
+		$expected = array( 'uncategorized' );
 		$metadata = self::$parsely->construct_parsely_metadata( $options, $post );
 
 		self::assertSame( $expected, $metadata['keywords'] );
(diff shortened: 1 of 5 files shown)


acicovic (approved):

GaryJones:
Why is this being pulled against `trunk`? If this is a hotfix, then create a hotfix branch and handle fixes, changelog and version bumps there. Only when that is fully ready should it be merged back into `trunk` and tagged.

tests/Integration/OtherTest.php
```diff
@@ -247,4 +247,50 @@ public function test_post_has_trackable_status_password_protected_with_filter():
...
+
+	/**
+	 * Tests if keywords in Parse.ly metadata are generated correctly when categories are tags.
+	 *
+	 * @since 3.0.3
+	 *
+	 * @covers \Parsely\Parsely::construct_parsely_metadata
+	 * @uses \Parsely\Parsely::get_categories
+	 */
+	public function test_post_with_categories_as_tags(): void {
```
GaryJones, line 259: Since the structured data output is being tested here, then putting these tests in SinglePostTest.php would seem a better fit, and we already have some tests there that may cover things: https://github.com/Parsely/wp-parsely/blob/develop/tests/Integration/StructuredData/SinglePostTest.php#L172-L196

GaryJones (changes requested):
Rejecting on the principle that fixing a fatal error shouldn't also be bundled with version changes and changelog updates. If these 6 commits get squashed, then the ability to revert becomes trickier (and commits that amend changelogs/version numbers are harder to find in a list of commits). If the commits don't get squashed on merge, then they should first by tidied up / rebased into atomic commits, at which point they can be separated out into two PRs - one for the fixes (or one for each fix), one of the administrative tasks for doing a release.

pauarge:
> Why is this being pulled against trunk? If this is a hotfix, then create a hotfix branch and handle fixes, changelog and version bumps there. Only when that is fully ready should it be merged back into trunk and tagged.

This is a hotfix. We have customers that found that error on their sites and we want to fix it as quick as possible. Hence, the hotfix branch is this one, which includes the actual code changes and the maintenance tasks. If this branch gets merged to `trunk`, then we can tag it and do the release.

PR #590, opened 2022-01-12
Adding PHPStan to perform static analysis

pauarge:
## Description

We are enforcing strict type checks since version 3.0 of the plugin. That gives us an opportunity to run a static code analyzer to improve the quality and stability of our code.

This PR adds PHPStan (plus a WordPress compatibility PHPStan extension) for that purpose. We are adding the analyzer as a composer command, and running it to its highest level (9) so it can detect all issues. We have plenty of errors right now, so the command is purely informative so we can bring that number down gradually. 

## Motivation and Context

Improve stability of the plugin and prevent execution errors.

## How Has This Been Tested?

Run the following commands locally:

```
composer install
composer static-analysis
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
 		"squizlabs/php_codesniffer": "^3.5",
 		"wp-coding-standards/wpcs": "^2.3.0",
 		"yoast/wp-test-utils": "^1",
-		"phpstan/phpstan": "^1.3",
+		"phpstan/phpstan": "^1.4",
 		"phpstan/extension-installer": "^1.1",
 		"phpstan/phpstan-strict-rules": "^1.1",
 		"szepeviktor/phpstan-wordpress": "^1.0"
--- a/composer.json
+++ b/composer.json
@@ -80,7 +80,7 @@
 			"@php ./vendor/bin/phpunit --no-coverage --order-by=random"
 		],
 	   	"static-analysis": [
-		  "@php ./vendor/bin/phpstan analyze src views wp-parsely.php uninstall.php -l 9 --memory-limit 536870912"
+		  "@php ./vendor/bin/phpstan analyze --memory-limit 536870912"
 		],
 		"test-ms": [
 			"@putenv WP_MULTISITE=1",
--- a/composer.json
+++ b/composer.json
@@ -25,6 +25,7 @@
 		"yoast/wp-test-utils": "^1",
 		"phpstan/phpstan": "^1.3",
 		"phpstan/extension-installer": "^1.1",
+		"phpstan/phpstan-strict-rules": "^1.1",
 		"szepeviktor/phpstan-wordpress": "^1.0"
 	},
 	"config": {
--- a/composer.json
+++ b/composer.json
@@ -22,12 +22,16 @@
 		"phpcompatibility/phpcompatibility-wp": "^2.1",
 		"squizlabs/php_codesniffer": "^3.5",
 		"wp-coding-standards/wpcs": "^2.3.0",
-		"yoast/wp-test-utils": "^1"
+		"yoast/wp-test-utils": "^1",
+		"phpstan/phpstan": "^1.3",
+		"phpstan/extension-installer": "^1.1",
+		"szepeviktor/phpstan-wordpress": "^1.0"
 	},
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -75,6 +79,9 @@
 		"test": [
 			"@php ./vendor/bin/phpunit --no-coverage --order-by=random"
 		],
+	   	"static-analysis": [
+		  "@php ./vendor/bin/phpstan analyze src views wp-parsely.php uninstall.php -l 9 --memory-limit 536870912"
```
GaryJones: May as well do `tests` as well.

composer.json
```diff
@@ -22,12 +22,16 @@
 		"phpcompatibility/phpcompatibility-wp": "^2.1",
 		"squizlabs/php_codesniffer": "^3.5",
 		"wp-coding-standards/wpcs": "^2.3.0",
-		"yoast/wp-test-utils": "^1"
+		"yoast/wp-test-utils": "^1",
+		"phpstan/phpstan": "^1.3",
+		"phpstan/extension-installer": "^1.1",
```
GaryJones, line 27: https://github.com/ergebnis/phpstan-rules might be another good extension to look into - a good selection of extra rules.

composer.json
pauarge: I excluded `tests` on purpose. There are some errors derived from PHPStan not finding the Yoast test helpers, which I want to address in the future. We can ignore those errors, but I think at this point they are just a distraction.

composer.json
pauarge: I'm not sure on this one. Rules are way more strict, and when running it, we get plenty of errors for using things like `isset` or `empty`. I can see the point of not using them, but I'm not sure if those are real issues or noise.

composer.json
GaryJones: Let's move the configuration to a `.neon` file ([example](https://github.com/google/web-stories-wp/pull/9816/files#diff-6f19df6a6307a48db0940e6897591fb08776d2db8a6134737aa708defdb5c92d)),

composer.json
GaryJones: We don't have to use them straight away - but we can work towards some of them over time.

https://github.com/phpstan/phpstan-strict-rules would be another set to consider to improve the overall codebase.

composer.json
pauarge: I've moved the configuration to a `.neon` file and added tests, but I'm still trying to figure out how to make it work with Yoast.

GaryJones (dismissed):
✅  (Other rules can be added in new PRs)

composer.json
pauarge: I've actually tried this last proposal and I like it. It will give us more work to clean all the errors up, but we can do that over time.

GaryJones (dismissed):
✅

composer.json
GaryJones: https://phpstan.org/user-guide/discovering-symbols

phpstan.neon
```diff
@@ -0,0 +1,7 @@
+parameters:
+  level: 9
+  paths:
+    - src/
+    - views/
+    - wp-parsely.php
+    - uninstall.php
```
GaryJones, line 7:
```suggestion
    - uninstall.php
  inferPrivatePropertyTypeFromConstructor: true
  ignoreErrors:
    # Uses func_get_args()
    - '#^Function apply_filters(_ref_array)? invoked with [34567] parameters, 2 required\.$#'
```
The new property is documented at https://phpstan.org/config-reference#inferprivatepropertytypefromconstructor

This ignore removes 11 Level 1 issues that we wouldn't be able to otherwise address.

GaryJones (dismissed):
✅

GaryJones:
While `^1.3` would catch it anyway, noting that [1.4.0 was released yesterday](https://github.com/phpstan/phpstan/releases/tag/1.4.0).

GaryJones:
Noting that I started fixing some of the level 0 strict items (not using `empty()`), but these are going to need some careful thinking to replace accurately, so it may be better to write some tests that specifically check that the condition results doesn't change for each instance.

acicovic (approved):

PR #591, opened 2022-01-12
Fix fatal TypeError in get_categories function

### LINKED ISSUE #587: Fatal TypeError reported in get_categories function

pauarge:
## Description

When a site had the plugin configured with the categories as tags option, the plugin would fail if that post wouldn't have categories. This was because we were calling `end()` of an empty array.

This PR adds a check to prevent that from failing plus two integration tests to prevent this regression from happening again. It is also addressing some other type issues in which a `WP_Error` might be returned and passed into a function that would not accept it.

## Motivation and Context

Closes #587

## How Has This Been Tested?

Integration tests.

### DIFF
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -646,11 +646,13 @@ class Parsely {
 	 * @return array The tags of the post represented by the post id.
 	 */
 	private function get_tags( int $post_id ): array {
-		$tags = array();
-		foreach ( wp_get_post_tags( $post_id ) as $wp_tag ) {
-			array_push( $tags, $wp_tag->name );
+		$tags      = array();
+		$post_tags = wp_get_post_tags( $post_id );
+		if ( ! is_wp_error( $post_tags ) ) {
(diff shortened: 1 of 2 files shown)


GaryJones (approved):
✅

PR #595, opened 2022-01-14
Moving plugin loading to named functions

pauarge:
## Description

This PR moves the functions that instantiate and load some different classes of the plugin to named functions.

## Motivation and Context

We want to be able to reference the functions that load some parts of the code from other parts of WordPress outside of this plugin.

## How Has This Been Tested?

The plugin should load and operate normally. This has been tested on VIP's environment as well.

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -59,32 +59,39 @@ add_action(
 	}
 );
 
-// Until auto-loading happens, we need to include this file for tests as well.
 require __DIR__ . '/src/UI/class-admin-warning.php';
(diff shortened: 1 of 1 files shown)


wp-parsely.php
```diff
@@ -59,32 +59,39 @@ function(): void {
 	}
```
jblz: I'd go ahead and move `plugins_loaded` to a named function as well.

acicovic: I agree

acicovic (approved):

jblz (approved):
lgtm

PR #602, opened 2022-01-17
Adding support for Google Web Stories

### LINKED ISSUE #600: Add web stories support

We'd like to add native support to [Google's Web Stories](https://github.com/google/web-stories-wp). Stories using Google's WordPress plugin are not tracked on Parse.ly unless custom code is inserted by the customer. We want to support those stories as if they were regular posts.

We'd like to implement a solution very similar to this gist: https://gist.github.com/swissspidy/bdd58b2de495d5a691d015eb67736775

Not implementing the support. However, given that the required changes on our end are very small, we can solve this pain for a handful of customers.

## Additional context

See: https://github.com/google/web-stories-wp/issues/9936

pauarge:
## Description

This PR adds native tracking support for Google Web Stories. Instead of inserting the usual tracking code, we rely on a Parse.ly integration with AMP to track the different stories. 

See https://www.parse.ly/help/integration/google-amp

## Motivation and Context

Closes #600.

## How Has This Been Tested?

Install the Google Web Stories plugin and create a story. Check that the tracking code (the site's API key) gets inserted on that story.


GaryJones (changes requested):
Since this is an integratation, I'd rather see it moved to an its own `Parsely\Integrations\...` class, so it is properly identified, rather than pushed into the always-enabled Scripts class.

src/Integrations/class-google-web-stories.php
```diff
@@ -0,0 +1,71 @@
...
+	/**
+	 * Apply the hooks that integrate the plugin or theme with the Parse.ly plugin.
+	 *
+	 * @since 3.2.0
+	 *
+	 * @return void
+	 */
+	public function integrate(): void {
+		if ( defined( 'WEBSTORIES_PLUGIN_FILE' ) ) {
+			add_action( 'web_stories_print_analytics', array( $this, 'render_amp_analytics_tracker' ) );
```
GaryJones, line 46: Why is the constant needing to be checked? If the plugin isn't active, the hook won't be defined and won't run.

src/Integrations/class-google-web-stories.php
```diff
@@ -0,0 +1,71 @@
...
+		}
+	}
+
+	/**
+	 * Load additional Javascript for Google's Web Stories WordPress plugin. This relies on the `amp-analytics` tag.
```
GaryJones:
```suggestion
	 * Load additional JavaScript for Google's Web Stories WordPress plugin. This relies on the `amp-analytics` element.
```

tests/Integration/ScriptsTest.php
```diff
@@ -402,4 +402,50 @@ public function assert_script_statuses( string $handle, array $assert_true = arr
...
+	/**
+	 * Test if the web stories analytics got enqueued.
+	 *
+	 * @since 3.2.0
+	 *
+	 * @covers \Parsely\Scripts::run()
+	 * @group scripts
+	 */
+	public function test_web_stories_script_is_enqueued(): void {
+		self::$scripts->run();
```
GaryJones: These tests should be moved to their own file - the code being called is not part of the Scripts class.

tests/Integration/ScriptsTest.php
```diff
@@ -402,4 +402,50 @@ public function assert_script_statuses( string $handle, array $assert_true = arr
...
+	/**
+	 * Test if the AMP tracker render outputs the correct script.
+	 *
+	 * @since 3.2.0
+	 *
+	 * @covers \Parsely\Scripts::render_amp_analytics_tracker
+	 * @group scripts
+	 */
+	public function test_render_amp_analytics_tracker(): void {
+		ob_start();
```
GaryJones: Don't use output buffering in the test - the TestCase already has this built-in - see https://phpunit.readthedocs.io/en/9.5/writing-tests-for-phpunit.html#testing-output and use `expectOutputString()` instead.

GaryJones (changes requested):

src/Integrations/class-google-web-stories.php
pauarge: I agree. However, I expect that most of our users won't have the plugin, so we'll always be enqueuing an action that it is never executed. I think it has a smaller performance toll to do a `defined` check than an unconditional `add_action`.  We are also doing something similar in the FBIA class.

tests/Integration/ScriptsTest.php
pauarge: Yes, the PR was a WIP :)

tests/Integration/Integrations/GoogleWebStoriesTest.php
```diff
@@ -0,0 +1,95 @@
...
+	/**
+	 * Test if the web stories analytics got enqueued.
+	 *
+	 * @since 3.2.0
+	 *
+	 * @covers \Parsely\Scripts::run()
+	 * @group scripts
+	 */
+	public function test_web_stories_script_is_enqueued(): void {
+		self::$google->integrate();
```
GaryJones, line 59: Let's test that it wasn't already added before this point.

tests/Integration/Integrations/GoogleWebStoriesTest.php
```diff
@@ -0,0 +1,95 @@
...
+	 *
+	 * @since 3.2.0
+	 *
+	 * @covers \Parsely\Scripts::render_amp_analytics_tracker
+	 * @group scripts
+	 */
+	public function test_render_amp_analytics_tracker(): void {
+		$this::set_options( array( 'apikey' => 'blog.parsely.com' ) );
+
+		ob_start();
```
GaryJones: Any reason not to use the `setExpectedOutput()` method?

tests/Integration/Integrations/IntegrationsTest.php
```diff
@@ -20,6 +21,15 @@
...
  */
 final class IntegrationsTest extends TestCase {
+	/**
+	 * The setUpBeforeClass run before all tests
+	 */
+	public static function setUpBeforeClass(): void {
+		parent::setUpBeforeClass();
+
+		$GLOBALS['parsely'] = new Parsely();
+	}
```
GaryJones: Why is this change needed?

GaryJones (changes requested):

tests/Integration/Integrations/GoogleWebStoriesTest.php
pauarge: I missed your initial suggestion, adding it now.

tests/Integration/Integrations/IntegrationsTest.php
pauarge: As per the changes in `wp-parsely.php`, the Web Stories class needs a Parse.ly object to be initialized. We take that object from `$GLOBALS` as we do with all the other classes we init there.

GaryJones: The `Google_Web_Stories::__construct()` accepts an argument of a Parsely instance, and that is passed in from `wp-parsely.php`. Nothing in the class itself references the global that I can see.

The `GoogleWebStoriesTest` class instantiates a new instance of `Google_Web_Stories` with a new instance of `Parsely` - all good, and `GoogleWebStoriesTest` doesn't reference the global either.

I can't see anywhere in this `IntegrationsTest` where the global is used either.

Using globals in tests is a code smell, as is creating globals in tests.

What breaks if you remove this global creation?

pauarge: The issue is in line 52 of `IntegrationsTest`. That calls a function on `wp-parsely.php`, `parsely_integrations`, which instantiates GWS class on line 137. That constructor is called with the global variable. That global variable is always present on normal execution because it's created in line 60 of the same file, but that hook doesn't get executed on tests.

If I remove the global declaration on the test, tests fail. I agree though that using globals on tests is not the nicest.

GaryJones: I understand now - thanks for the explanation.

Can you please add a comment to this line 29 then, to explain the reasoning? Something like:

```php
// Add a global instance, so that calls to `parsely_integrations()` in tests can correctly instantiate and register Google_Web_Stories as an integration.
```

GaryJones (approved):
✅  Not tested, but code looks good.

PR #606, opened 2022-01-19
CI: Harden e2e workflows

GaryJones:
## Description
Makes some changes to workflow config files to harden them.

## Motivation and Context

pcURaU-6Z-p2

### Pin actions to a full length commit SHA
https://app.stepsecurity.io/ was a handy tool for this.

Instead of using, for example, `actions/checkout@v2`, then we use `actions/checkout@629c2de402a417ea7690ca6ce3f33229e27606a5` instead. While this stops the workfow from benefitting from incremental fixes to the action, it also stops us from encountering potentially bad or malicious commits. If the action is doing what we want on the current commit, then we can have maybe quarterly reviews to see if new versions are available for these actions, and treat it like other development dependencies.

I’ve not been able to fully test the `deploy.yml` workflow as that requires a new tag, once the PR has been merged.

### Restrict external endpoints that GitHub Actions can call

By adding the `step-security/harden-runner action`, it captures requests the workflow does to external endpoints. It also creates a link in the workflow output. At the bottom of that output there's a recommended list of endpoints which we can mark as an allowlist. For the CS & Lint one, the `allowed-endpoints` was:

```yaml
- uses: step-security/harden-runner@v1
  with:
    allowed-endpoints:
      artifactcache.actions.githubusercontent.com:443
      github.com:443
      kv4gacprodeus2file3.blob.core.windows.net:443
      objects.githubusercontent.com:443
      repo.packagist.org:443
```

I did notice that `packagist.org:443` was being rejected for three workflows where Composer was used, so to avoid noise I added that to the allowlist as well.

By using an allowlist of endpoints, we can be informed when our actions are hitting new, perhaps malicious, endpoints, and adjust as necessary.

At the moment, then `deploy.yml` workflow is still in `egress-policy: audit`, which means the audit link is created, but there are no endpoints listed or blocked. This can be updated once the workflow has been used after creating a tag.

See https://github.com/step-security/supply-chain-goat/blob/main/RestrictOutboundTraffic.md and https://github.com/step-security/supply-chain-goat/blob/main/DNSExfiltration.md for background info.

## How Has This Been Tested?
Ran the CIs as part of this PR, and checked the results.


.github/workflows/codeql-analysis.yml
```diff
@@ -37,18 +37,23 @@ jobs:
...
-        # By default, queries listed here will override any specified in a config file.
-        # Prefix the list here with "+" to use these queries and those in the config file.
-        # queries: ./path/to/local/query, your-org/your-repo/queries@main
-
-    - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@v1
+      - name: Harden Runner
+        uses: step-security/harden-runner@14dc64f30986eaa2ad2dddcec073f5aab18e5a24 # v1
+        with:
+          egress-policy: audit
```
pauarge, line 43: As discussed internally, could we get rid of this external dependency? I acknowledge we'd be less hardened in terms of the calls the action does, but we wouldn't be introducing an additional external dependency.

pauarge (changes requested):
Looks good but I would not have the Harden Runner.

.github/workflows/codeql-analysis.yml
GaryJones: As the result of the action already show, we already have several dependencies, and actions that reach out to third-party sites - one more (that gives us insight into those we already use) isn't going to be detrimental.

I don't feel strongly though, so if you do feel stronger, it can be removed.

pauarge: Sorry for the late response, this completely slipped through my radar. 

I would still remove this external dependency. Other than that the PR looks great.

varunsh-coder: Hi @pauarge, can you please help me understand the concern? You don't have to use it, but understanding the concern might help me improve it. In an end-to-end test workflow, adding `harden-runner` is not going to compromise the build output or anything. Best case you might find a compromised dependency making outbound calls (which they typically do), but what is the worst case of using `harden-runner` in an end-to-end test workflow?

pauarge: Hi @varunsh-coder, thanks for reaching out. I am not particularly keen on a third-party dependency running on our CI that's actively keeping a log of our builds: https://app.stepsecurity.io/github/Parsely/wp-parsely/actions/runs/1704105637

varunsh-coder: > Hi @varunsh-coder, thanks for reaching out. I am not particularly keen on a third-party dependency running on our CI that's actively keeping a log of our builds: https://app.stepsecurity.io/github/Parsely/wp-parsely/actions/runs/1704105637

Thanks a lot @pauarge for the feedback! Would it help if: 
1. The logs were not publicly visible, but were private only to those who have access to the repo? This is in `preview` for [private repositories](https://github.com/step-security/harden-runner#support-for-private-repositories), but I could also enable it for Public repositories. AND/OR
2. The logs expired and were purged in a few days (e.g. in 1 or 2 days) or whoever has access to the repo had the ability to delete the logs,  AND/OR
3. In `block` mode, you could configure it to not send out telemetry. This was a request from another developer using `harden-runner`, and is being tracked [here](https://github.com/step-security/harden-runner/issues/89). 

Please let me know which of these options will help with your concern or if you have other ideas. The goal is really to help developers find compromised dependencies/ build tools, and if anything in the method is causing concern with that goal, I can try to address that. Thanks again for the feedback!

pauarge:
Closing due to conflicts and being stale for a while.

PR #597, opened 2022-01-14
WIP Recommended Content Block

jblz:
## Description

Pick up where #318 left off.

## Motivation and Context

See: https://github.com/Parsely/web/issues/12585

## How Has This Been Tested?

TBD

## Screenshots (if appropriate)

TODO

PR #611, opened 2022-01-21
Add an endpoint to proxy Recommendations requests to the Parsely API

jblz:
## Description

Pulled out of #597 and tidied up a bit, this endpoint will allow us to serve first-party responses from the [`/related` endpoint](https://www.parse.ly/help/api/recommendations#get-related).

For now, this is behind a feature flag. If `wp_parsely_enable_recommendations_endpoint` isn't set to return true, this code will do nothing.

## Motivation and Context

Cross-origin requests are increasingly unsupported by modern browsers. This endpoint gives us a way to serve front-end features without reasonable concerns they'll be blocked for the preponderance of visitors.

## How Has This Been Tested?

* WP REST API Testing in this PR
* Manually browse to `/wp-json/wp-parsely/v1/recommendations` (if you don't have a valid api key on your site, override it [here](https://github.com/Parsely/wp-parsely/blob/e6ad76194517acd59fb95c3a1462b0402557e2f9/src/class-recommended-content.php#L39) to test)


tests/Integration/Endpoints/RecommendationsProxyEndpointTest.php
```diff
@@ -0,0 +1,171 @@
...
+use Parsely\Endpoints\Recommendations_API_Proxy;
+use WP_Error;
+use WP_REST_Request;
+use WP_REST_Server;
+use WP_Test_REST_Controller_Testcase;
+
+/**
+ * Parsely REST API tests.
+ */
+final class RecommendationsProxyEndpointTest extends WP_Test_REST_Controller_Testcase {
```
jblz: I'm attempting to extend our `TestCase` so we don't have so many skipped tests d/t having to implement all the abstract methods of `WP_Test_REST_Controller_Testcase`, but it's not immediately obvious why it's not working.

wp-parsely.php
```diff
@@ -103,6 +104,20 @@ function parsely_admin_menu_register(): void {
...
+/**
+ * Register the Recommendatiosn API Proxy WP-API REST Endpoint
+ *
+ * @return void
+ */
+function rest_api_init_recommendations_proxy() {
+	$api = new Recommendations_API_Proxy();
+	$api->run();
+}
+add_action( 'rest_api_init', __NAMESPACE__ . '\\rest_api_init_recommendations_proxy' );
```
jblz: Should this be feature flagged?

src/Endpoints/class-recommendations-api-proxy.php
```diff
@@ -0,0 +1,174 @@
...
+		$params  = $request->get_params();
+
+		if ( empty( $apikey ) ) {
+			return (object) array(
+				'data'  => array(),
+				'error' => new WP_Error( 400, __( 'A Parsely API Key must be set in site options to use this Endpoint', 'wp-parsely' ) ),
+			);
+		}
+
+		$cache_key = 'api_recos-' . md5( wp_json_encode( compact( 'apikey', 'params' ) ) );
```
jblz: sonarcloud is complaining that md5 is a security risk...  Should I replace this with a more robust hashing algorithm?
```suggestion
		$cache_key = 'api_recos-' . hash( 'sha512', wp_json_encode( compact( 'apikey', 'params' ) ) );
```

GaryJones: `wp_hash()`?

src/class-recommended-content.php
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Recommended_Content class file.
+ *
+ * @package Parsely
+ * @since 3.2.0
+ */
+
+declare(strict_types=1);
+
+namespace Parsely;
```
GaryJones: How about putting this into a `Parsely\Api` namespace and corresponding file path?

src/class-recommended-content.php
```diff
@@ -0,0 +1,54 @@
...
+ */
+class Recommended_Content {
+	const RELATED_API_ENDPOINT = 'https://api.parsely.com/v2/related';
+
+	/**
+	 * Get the URL for the Recommendation API (GET /related).
+	 *
+	 * @see https://www.parse.ly/help/api/recommendations#get-related
+	 *
+	 * @since 2.6.0
```
GaryJones: Add another `@since` here to show it's been moved.

src/Endpoints/class-recommendations-api-proxy.php
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	/**
+	 * Cached "proxy" to the Parsely `/related` endpoint
+	 *
+	 * @param WP_Rest_Request $request The request object.
+	 * @return array
+	 */
+	public static function get_items( WP_Rest_Request $request ) {
+		global $parsely;
```
pauarge: Could we avoid using the the global object? I see that this function is static, but how problematic would it be to hold a $parsely variable in the class?

src/Endpoints/class-recommendations-api-proxy.php
pauarge: +1 on using `wp_hash` and a modern hashing algorithm.

src/Endpoints/class-recommendations-api-proxy.php
```diff
@@ -0,0 +1,174 @@
...
+					'image_url' => $link->image_url,
+					'title'     => $link->title,
+					'url'       => $link->url,
+				);
+			},
+			$links
+		);
+
+		$response = (object) array( 'data' => $data );
+		wp_cache_set( $cache_key, $data, 'wp-parsely', 5 * MINUTE_IN_SECONDS );
```
pauarge: Could we extract the expiration time as a constant?

src/UI/class-recommended-widget.php
```diff
@@ -14,6 +14,7 @@
 namespace Parsely\UI;
 
 use Parsely\Parsely;
+use Parsely\Recommended_Content;
```
pauarge: The changes on the recommended widget look good, but could we introduce them in a separate PR after this one gets merged?

wp-parsely.php
pauarge: Yes, I would put a filter inside the `run` method.

pauarge (changes requested):
The PR looks very promising. 

Open question, I'd like to see what folks think about it. Should we move the `Rest` class together with the `Enpoints` one? I know that one is extending an existing endpoint vs the other is adding a new one, but it seems to me that they could be together.

src/class-recommended-content.php
jblz: I think if we are to put it into a separate namespace, it makes sense to disambiguate the API to mean the Parse.ly service API vs the REST API on the site itself. What about `Parsely\RemoteApi` or similar...?

jblz:
> Should we move the Rest class together with the Endpoints one?

I think so, yes.  Let's do it in a separate PR, though, IMO.

src/Endpoints/class-recommendations-api-proxy.php
```diff
@@ -0,0 +1,181 @@
+<?php
+/**
+ * Parsely Recommendations REST API Endpoint
+ *
+ * @package Parsely
+ */
+
+namespace Parsely\Endpoints;
```
pauarge, line 10: We're missing the strict types declaration on this file.

wp-parsely.php
```diff
@@ -105,6 +106,20 @@ function parsely_admin_menu_register(): void {
 	$settings_page->run();
 }
 
+require __DIR__ . '/src/class-recommended-content.php';
+
```
pauarge, line 112: To keep consistency with the rest of the file, could we remove the line break on 110?

wp-parsely.php
```diff
@@ -105,6 +106,20 @@ function parsely_admin_menu_register(): void {
...
+require __DIR__ . '/src/Endpoints/class-recommendations-api-proxy.php';
+/**
+ * Register the Recommendatiosn API Proxy WP-API REST Endpoint
+ *
+ * @return void
+ */
+function rest_api_init_recommendations_proxy() {
```
pauarge:
```suggestion
function rest_api_init_recommendations_proxy(): void {
```

pauarge (changes requested):
We're almost there. I left some super minor tweaks, but overall looks good.

src/Endpoints/class-recommendations-api-proxy.php
```diff
@@ -0,0 +1,181 @@
...
+	 */
+	public function get_items( WP_Rest_Request $request ) {
+		$options = $this->parsely->get_options();
+		$apikey  = $options['apikey'];
+		$params  = $request->get_params();
+
+		if ( empty( $apikey ) ) {
+			return (object) array(
+				'data'  => array(),
+				'error' => new WP_Error( 400, __( 'A Parsely API Key must be set in site options to use this Endpoint', 'wp-parsely' ) ),
```
GaryJones:
```suggestion
				'error' => new WP_Error( 400, __( 'A Parse.ly API Key must be set in site options to use this endpoint', 'wp-parsely' ) ),
```

tests/Integration/Endpoints/RecommendationsProxyEndpointTest.php
```diff
@@ -0,0 +1,126 @@
...
+ * Parsely REST API tests.
+ */
+final class RecommendationsProxyEndpointTest extends TestCase {
+	/**
+	 * Set up globals & initialize the Endpoint.
+	 */
+	public function setUp(): void {
+		parent::setUp();
+		add_filter( 'wp_parsely_enable_recommendations_endpoint', '__return_true' );
+		update_option( 'parsely', array( 'apikey' => 'example.com' ) );
```
GaryJones: This will wipe out all of the other Parsely settings, which isn't reflective of real-world happenings.

Please use [`setOptions()` method](https://github.com/Parsely/wp-parsely/blob/ebd84b8d3c124b178b7ca8a8f3136eb9d7b24e73/tests/Integration/TestCase.php#L61) so that the APIkey value is merged in with the rest.

tests/Integration/Endpoints/RecommendationsProxyEndpointTest.php
```diff
@@ -0,0 +1,126 @@
...
+
+		$this->wp_rest_server_global_backup        = $GLOBALS['wp_rest_server'] ?? null;
+		$this->rest_api_init_recommendations_proxy = function () {
```
GaryJones, line 48: These dynamic properties aren't defined, and that will be an issue in PHP 8.1.

tests/Integration/Endpoints/RecommendationsProxyEndpointTest.php
```diff
@@ -0,0 +1,126 @@
...
+		add_action( 'rest_api_init', $this->rest_api_init_recommendations_proxy );
+	}
+
+	/**
+	 * Reset globals.
+	 */
+	public function tearDown(): void {
+		parent::tearDown();
+		remove_action( 'rest_api_init', $this->rest_api_init_recommendations_proxy );
+		$GLOBALS['parsely'] = $this->parsely_global_backup;
```
GaryJones: There doesn't seem to be any setting of this `parsely_global_backup` property (and it's not defined).

tests/Integration/Endpoints/RecommendationsProxyEndpointTest.php
```diff
@@ -0,0 +1,126 @@
...
+		// phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedVariableFound
+		$GLOBALS['wp_rest_server'] = $this->wp_rest_server_global_backup;
+		delete_option( 'parsely' );
```
GaryJones: Is it realistic to have all the options deleted? Or would resetting it to the defaults be better, if that is indeed needed?

tests/Integration/Endpoints/RecommendationsProxyEndpointTest.php
```diff
@@ -0,0 +1,126 @@
...
+
+	/**
+	 * Confirm the the route is registered.
+	 *
+	 * @covers Recommendations_API_Proxy::register_rest_route
+	 */
+	public function test_register_routes() {
+		$routes = rest_get_server()->get_routes();
+		$this->assertArrayHasKey( '/wp-parsely/v1/recommendations', $routes );
+		$this->assertSame( 1, count( $routes['/wp-parsely/v1/recommendations'] ) );
```
GaryJones: Use `assertCount()` instead.

```suggestion
		$this->assertCount( 1, $routes['/wp-parsely/v1/recommendations'] );
```

tests/Integration/Endpoints/RecommendationsProxyEndpointTest.php
```diff
@@ -0,0 +1,126 @@
...
+		$this->assertEquals( array( 'GET' => true ), $routes['/wp-parsely/v1/recommendations'][0]['methods'] );
```
GaryJones: Can `assertSame()` be used here?

tests/Integration/Endpoints/RecommendationsProxyEndpointTest.php
```diff
@@ -0,0 +1,126 @@
...
+	/**
+	 * Confirm that calls to `GET /wp-parsely/v1/recommendations` gets and error and makes no remote call when the apikey is not populated in site options.
+	 *
+	 * @covers Recommendations_API_Proxy::get_items
+	 */
+	public function test_get_items_fails_without_apikey_set() {
+		delete_option( 'parsely' );
+		$request    = new WP_REST_Request( 'GET', '/wp-parsely/v1/recommendations' );
+		$response   = null;
+		$dispatched = TestCase::mock_remote_network_request(
```
GaryJones: This uses `TestCase`, but the same call above uses `self`. Any reason for the inconsistency?

tests/Integration/Endpoints/RecommendationsProxyEndpointTest.php
```diff
@@ -0,0 +1,126 @@
...
+						'url'       => 'https://example.com/2',
+					),
+				),
+			),
+			$response->get_data()
+		);
+	}
+
+	/**
+	 * Confirm that calls to `GET /wp-parsely/v1/recommendations` gets and error and makes no remote call when the apikey is not populated in site options.
```
GaryJones:
```suggestion
	 * Confirm that calls to `GET /wp-parsely/v1/recommendations` gets an error and makes no remote call when the apikey is not populated in site options.
```

tests/Integration/Endpoints/RecommendationsProxyEndpointTest.php
```diff
@@ -0,0 +1,126 @@
...
+	 *
+	 * @covers Recommendations_API_Proxy::get_items
+	 */
+	public function test_get_items_fails_without_apikey_set() {
+		delete_option( 'parsely' );
```
GaryJones: As before, use `setOptions()` to set `apikey` as an empty string, rather than deleting the whole `parsely` option.

tests/Integration/Endpoints/RecommendationsProxyEndpointTest.php
```diff
@@ -0,0 +1,126 @@
...
+		$this->assertSame( 200, $response->get_status() );
+		$data = $response->get_data();
+
+		$this->assertSame( 0, $dispatched );
+		$this->assertObjectHasAttribute( 'data', $data );
+		$this->assertEmpty( $data->data );
+
+		$this->assertObjectHasAttribute( 'error', $data );
+		$this->assertEquals(
+			new WP_Error( 400, 'A Parsely API Key must be set in site options to use this Endpoint' ),
```
GaryJones:
```suggestion
			new WP_Error( 400, 'A Parse.ly API Key must be set in site options to use this endpoint' ),
```

tests/Integration/RecommendedApiTest.php
```diff
@@ -10,6 +10,9 @@
 namespace Parsely\Tests\Integration;
 
 use Parsely\UI\Recommended_Widget;
+use Parsely\Recommended_Content;
+
+require_once __DIR__ . '/../../src/class-recommended-content.php';
```
GaryJones: This require shouldn't be needed - all of the file-loading for running tests should either already be handled, or it should be handled in a bootstrap file.

tests/Integration/TestCase.php
```diff
@@ -176,4 +176,46 @@ public function go_to_new_post(): int {
 
 		return $post_id;
 	}
+
+	/**
+	 * Simulate a single outbound HTTP request and response.
```
GaryJones: How does WP core tests handle mocking remote network requests?

tests/Integration/TestCase.php
```diff
@@ -176,4 +176,46 @@ public function go_to_new_post(): int {
...
+	 *
```
GaryJones: I think including docs for an example call in here would be useful to better understand how to use it.

tests/Integration/TestCase.php
```diff
@@ -176,4 +176,46 @@ public function go_to_new_post(): int {
...
+	 * @param callable $callback The function that will be called while the network response is mocked.
+	 * @param string   $response_body The data that will be used in the HTTP response body. Default ''.
+	 * @param integer  $response_code The numerical code that will be used in the HTTP response. Default 200.
```
GaryJones:
```suggestion
	 * @param int      $response_code The numerical code that will be used in the HTTP response. Default 200.
```

tests/Integration/TestCase.php
```diff
@@ -176,4 +176,46 @@ public function go_to_new_post(): int {
...
+	 * @param string   $response_message The data that will be used in the HTTP response body. Default 'OK'.
+	 * @param array    $response_headers The data that will be used in the HTTP response headers. Default empty array.
+	 * @param array    $response_cookies The data that will be used in the HTTP response cookies. Default empty array.
+	 * @return int     The number of times the remote request was dispatched.
```
GaryJones:
```suggestion
	 * @return int The number of times the remote request was dispatched.
```

`@return` doesn't need to line up.

tests/Integration/TestCase.php
```diff
@@ -176,4 +176,46 @@ public function go_to_new_post(): int {
...
+	 */
+	public static function mock_remote_network_request(
+		callable $callback,
+		string $response_body = '',
+		int $response_code = 200,
+		string $response_message = 'OK',
+		array $response_headers = array(),
+		array $response_cookies = array()
+	) {
```
GaryJones: Return type can be added.

```suggestion
	): int {
```

tests/Integration/TestCase.php
```diff
@@ -176,4 +176,46 @@ public function go_to_new_post(): int {
...
+				'cookies'  => $response_cookies,
+				'filename' => null,
+			);
+
+			return $response;
+		};
+
+		add_filter( 'pre_http_request', $pre_http_request_filter, 10, 3 );
+		$callback();
+		remove_filter( 'pre_http_request', $pre_http_request_filter, 10 );
```
GaryJones:
```suggestion
		remove_filter( 'pre_http_request', $pre_http_request_filter );
```

10 is the default priority arg value.

GaryJones (dismissed):

tests/Integration/TestCase.php
jblz: > How does WP core tests handle mocking remote network requests?

From this search:
https://github.com/WordPress/wordpress-develop/search?q=pre_http_request

...it looks like it's just manually hooking into `pre_http_request` and not unhooking (I guess it relies on some sort of isolation of the test setup).  That could work here as well, I suppose.

wp-parsely.php
```diff
@@ -105,6 +106,19 @@ function parsely_admin_menu_register(): void {
 	$settings_page->run();
 }
 
+require __DIR__ . '/src/class-recommended-content.php';
+require __DIR__ . '/src/Endpoints/class-recommendations-api-proxy.php';
+/**
+ * Register the Recommendatiosn API Proxy WP-API REST Endpoint
```
jblz:
```suggestion
 * Register the Recommendations API Proxy WP-API REST Endpoint
```

pauarge (dismissed):
🚢 🚢 🚢

pauarge (dismissed):

jblz:
I went ahead and just used filters directly in the test to mock the request as per core (after verifying that they don't persist to different tests).

src/class-recommended-content.php
jblz: :+1:  I went ahead and moved it to `Parsely\RemoteAPI` in namespace and file path.

src/RemoteAPI/class-recommended-content.php
```diff
@@ -0,0 +1,54 @@
...
+ * @package Parsely
+ * @since 3.2.0
+ */
+
+declare(strict_types=1);
+
+namespace Parsely\RemoteAPI;
+
+/**
+ * Recommended_Content
```
acicovic, line 14: Consider a more descriptive comment if possible. What does this class do or what is its use?

acicovic:
Do you think it would make sense to write a test for when the feature flags are not set?

acicovic (approved):

pauarge (approved):

GaryJones (commented):
I'd love to see this PR get documentation before it is merged.

Why is the endpoint needed?
How is it used within the wp-parsely plugin?
How could other plugin developers or site developers use it themselves?
How should it be used (calls, arguments, API reference)?

acicovic:
> I'd love to see this PR get documentation before it is merged.

Here are my 2 cents:

- PR #642 is based upon this, so it would be nice to be able to merge this as soon as possible, before putting more work on 642.
- As far as I know, the endpoint is mostly currently aimed for our own use, and that we may need to modify code further if we want to officially support it for public use. I know that @jblz has thought about this but I don't exactly know where we're at with that.

So, If docs need time or we need time consuming code adjustments to officially support public use, I would probably prefer to merge this as soon as possible as an intermediary release, rather than having to wait for whatever else needs to be done.

Now of course, if any code adjustment would mean breaking things, we should finish with code adjustments here before doing anything else.

pauarge:
I agree with @acicovic's point. For now, this endpoint is for internal use (although we're not actively blocking it). I am happy to properly document this plus the block feature right before we release 3.2.

Issue #586, opened 2022-01-11
Reorganize wp-parsely.php and tackle autoloading issues


pauarge:
## Is your feature request related to a problem?

During the past few months, we have put a significant effort into splitting the plugin's logic into smaller files and classes. That has lead to the plugin's base file, `wp-parsely.php`, growing quite a bit in size. The file has multiple repeated hooks and a mixture of lambda and named functions. It also handles all the `require` of the plugin (except for tests).

## Describe the solution you'd like

- We'd like a file that is leaner, cleaner, and easier to maintain. 
- We would like to reduce the number of hooks that we use and have a unified decision on whether all hooks are lambdas or named functions. 
- We'd like to solve the autoloading issues, especially with tests.

## Describe alternatives you've considered

@jblz did some exploration with this topic https://github.com/Parsely/wp-parsely/pull/556

## Additional context

We'd like to prevent things like this from happening: https://github.com/Parsely/wp-parsely/pull/569#discussion_r781180713

GaryJones:
I could get behind the structure described at https://github.com/szepeviktor/small-project/blob/master/MAIN-FILE-PARTS.md

Among other things, it suggests using a config class (like [this](https://github.com/szepeviktor/small-project/blob/master/src/Config.php)) to replace global (namespaced) constants - big +1 for that.

PR #615, opened 2022-01-24
Remove Plugin Actions global variable on load

pauarge:
## Description

This PR gets rid of a class that was being put available globally on plugin load. This is not being used elsewhere and it shouldn't be a way to access the class. 

Given that's a simple class and that it can be extended/changed anyway by using other filters, we are removing this global variable.

## Motivation and Context

Tighten up improper uses of the plugin.

See https://github.com/Parsely/wp-parsely/issues/586

## How Has This Been Tested?

Plugin runs and loads as normal.

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -83,8 +83,8 @@ function parsely_admin_init_register(): void {
 	$admin_warning = new Admin_Warning( $GLOBALS['parsely'] );
 	$admin_warning->run();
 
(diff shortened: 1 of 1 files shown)


sjinks (approved):

jblz (approved):
This has a chance to break integrations since 3.0.0 if anyone were relying on the global, but I'd not expect it to have any real-world impact, so 👍

acicovic (approved):

PR #619, opened 2022-01-26
Add a WINDOWS.md file for Windows developers

acicovic:
## Description
This PR introduces a WINDOWS.md file containing some troubleshooting/setup instructions for developers who work on Windows, so they can set up a fully working development environment. It also puts a couple of links in CONTRIBUTING.md towards WINDOWS.md in the related sections.

Thanks to @jblz for helping in finding the solution to the composer issue.

## Motivation and Context
Although we don't support Windows development officially, possible contributors might be working on Windows. Unfortunately, certain things don't seem to be working as expected out of the box in Windows, and this could be a deterrent factor to developers trying to contribute. This PR provides some help regarding that and is also a note to ourselves should we ever need to develop on Windows.

## How Has This Been Tested?
1. The problems described have been experienced on a Windows 10 installation.
2. After researching and doing trial and error operations, working solutions have been found.
3. The above working solutions were documented.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -106,7 +106,9 @@ npm run dev:stop
 
 We're leveraging [husky](https://typicode.github.io/husky) to automate some tasks around things like code and commit message quality. You can browse the configured hooks in [this directory](.husky/).
 
-For example, the [coding standards](#coding-standards) and [lint rules](#linting) are applied [prior to commit](.husky/pre-commit). If violations are encountered, the commit is rejected.
+For example, the [coding standards](#coding-standards) and [lint rules](#linting) are applied [prior to commit](.husky/pre-commit). If violations are encountered, the commit is rejected. Please note that this quality assurance process introduces a delay before every commit.
+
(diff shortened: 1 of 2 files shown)


GaryJones (approved):
✅

pauarge (approved):
Great work!

This is out of the scope of this PR and I'm happy to merge, but I wonder if we should create in the future a `docs` folder where we hold documents like this one.

acicovic:
> This is out of the scope of this PR and I'm happy to merge, but I wonder if we should create in the future a `docs` folder where we hold documents like this one.

I wondered about the same thing. We should probably see if anything would need to stay out of that directory (for example README.md).

pauarge:
> I wondered about the same thing. We should probably see if anything would need to stay out of that directory (for example README.md).

Let's create an issue for it to keep track of the problem, but happy to merge this PR.

acicovic:
> Let's create an issue for it to keep track of the problem, but happy to merge this PR.

Tracked in #621.

PR #620, opened 2022-01-26
Un-hard code references to Uncategorized

### LINKED ISSUE #593: Un-hard code references to Uncategorized
We have [some instances](https://github.com/Parsely/wp-parsely/search?q=Uncategorized) of "Uncategorized" / `uncategorized` in the codebase. What we really probably mean is "the default category". This then allows for non-US-English installs which may use a different name or slug by default (even en-GB uses an `s` not a `z`), and supports when someone has changed/renamed the default category (i.e. to "General").

This is stored in wp_options as `default_category`, so we should retrieve and use that instead.

acicovic:
## Description
In #593 it was reported that we have some instances of hardcoded "Uncategorized" strings instead of referring to the default category name. This can cause issues in non-US English WP installations or if the user changes the default category title. This PR fixes this issue.

## Motivation and Context
Closes #593.

## How Has This Been Tested?
- Manually browsed through a post page
- Integration tests that call `get_default_category()` directly or indirectly pass


acicovic:
Opening this for discussion.

1. I replaced 'Uncategorized' references with a function that gets the default category name (`get_default_post_category_name()`). It is placed within the `Parsely` class (which I don't like but don't see currently any other place for it - I think having a utils class or other solutions have been mentioned).

2. I'm unsure whether I should have declared the function as `public` (no other class except tests should probably call it) but this is what I've done for now.

3. It seems that the `test_post_with_categories_as_tags_without_categories()` test was resulting in an empty category array in `get_default_post_category_name()`. I still don't understand how this is possible, but I added a check to return `''` if the category array is empty. Not sure this is the right thing to do, but tests pass.

4. The function `get_default_post_category_name()` is currently being indirectly tested, but I can add a direct test for it if needed

Any feedback is greatly appreciated!

tests/Integration/StructuredData/SinglePostTest.php
```diff
@@ -652,7 +652,8 @@ public function test_post_with_categories_as_tags_without_categories(): void {
 		$parsely_options['cats_as_tags'] = true;
 		update_option( 'parsely', $parsely_options );
 
-		wp_remove_object_terms( $post_id, 'uncategorized', 'category' );
+		$default_category_name = $parsely->get_default_post_category_name();
+		wp_remove_object_terms( $post_id, $default_category_name, 'category' );
```
GaryJones: The second argument here is using the slug (but it could also use the ID), but you're now passing in the category name, not the category slug.

src/class-parsely.php
```diff
@@ -1105,4 +1105,22 @@ public function get_api_key(): string {
...
+	public function get_default_post_category_name(): string {
+		$category_id = get_option( 'default_category' );
+		$category    = get_categories( array( 'ID' => $category_id ) );
+
+		if ( 0 === count( $category ) ) {
+			return '';
+		}
+
+		return $category[0]->name;
+	}
```
GaryJones: I think I would change this to `get_default_category()` and either return an array of:
 - slug
 - name
 - ID
 
 Or just return the object. That way, we know everything we need to know about the default category in one call.

```php
get_category( get_option( 'default_category', 1 )->slug;
```

If it's this simple, you may find that you don't even need a separate method to handle this, unless we want to mock it for unit tests.

GaryJones (changes requested):

tests/Integration/StructuredData/SinglePostTest.php
acicovic: You are correct. I've fixed this.

src/class-parsely.php
acicovic: I've renamed the function according to your suggestion and made it to return the object or null.

src/class-parsely.php
```diff
@@ -694,8 +695,14 @@ private function get_categories( int $post_id, string $delimiter = '/' ): array
...
 			$tags = explode( '/', $last_tag );
 		}
-		// remove uncategorized value from tags.
-		return array_diff( $tags, array( 'Uncategorized' ) );
+
+		// Remove default category name from tags if needed.
+		$default_category_name = $this->get_default_category()->name ?? null;
+		if ( null !== $default_category_name ) {
+			return array_diff( $tags, array( $default_category_name ) );
+		}
```
GaryJones: Let's move the check into `get_default_category()`, and have it always return a WP_Term. That way, it should always have a `name` set as well, or at least be an empty string - in which case, doing the array_diff doesn't need the conditional.

src/class-parsely.php
```diff
@@ -727,8 +734,8 @@ private function get_category_name( WP_Post $post_obj, array $parsely_options ):
 		$taxonomy_dropdown_choice = get_the_terms( $post_obj->ID, $parsely_options['custom_taxonomy_section'] );
 		// Get top-level taxonomy name for chosen taxonomy and assign to $parent_name; it will be used
 		// as the category value if 'use_top_level_cats' option is checked.
-		// Assign as "Uncategorized" if no value is checked for the chosen taxonomy.
-		$category = 'Uncategorized';
+		// Assign as the default category name if no value is checked for the chosen taxonomy.
+		$category_name = $this->get_default_category()->name ?? '';
```
GaryJones: ...no `??` check would be needed here then either.

src/class-parsely.php
```diff
@@ -1107,4 +1114,23 @@ public function get_api_key(): string {
...
+	 * Get the default category for posts.
+	 *
+	 * @since 3.2.0
+	 *
+	 * @return WP_Term|null The category, or null if a category couldn't be retrieved.
+	 */
+	public function get_default_category(): ?WP_Term {
+		$result      = null;
+		$category_id = get_option( 'default_category' );
+		$categories  = get_categories( array( 'ID' => $category_id ) );
```
GaryJones: Why using `get_categories()` with a single ID, and then the first from the resulting array? Why not `get_category()` (singular) with the ID?

src/class-parsely.php
```diff
@@ -1107,4 +1114,23 @@ public function get_api_key(): string {
...
+	/**
+	 * Get the default category for posts.
+	 *
+	 * @since 3.2.0
+	 *
+	 * @return WP_Term|null The category, or null if a category couldn't be retrieved.
+	 */
+	public function get_default_category(): ?WP_Term {
+		$result      = null;
+		$category_id = get_option( 'default_category' );
```
GaryJones: The [usage](https://github.com/wordpress/wordpress/search?q=default_category) of `get_option( 'default_category' )` in WP core doesn't have any fallback option, so if that is understood to always be an integer value, then:

```php
public function get_default_category(): WP_Term {
	return get_term( get_option( 'default_category' ), 'category' );
}
```

should be all you need for this function, which as noted before, could just as easily be used as a one-liner in the places where it's actually needed, rather than define and test a new method.

GaryJones (changes requested):

src/class-parsely.php
```diff
@@ -694,8 +695,14 @@ private function get_categories( int $post_id, string $delimiter = '/' ): array
 		if ( $last_tag ) {
 			$tags = explode( '/', $last_tag );
 		}
-		// remove uncategorized value from tags.
-		return array_diff( $tags, array( 'Uncategorized' ) );
```
GaryJones, line 698:
```diff
-		return array_diff( $tags, array( 'Uncategorized' ) );
+ 		return array_diff( $tags, get_cat_name( get_option( 'default_category' ) ) );
```

would seem sufficient here. It's what WP core uses.

src/class-parsely.php
```diff
@@ -727,8 +734,8 @@ private function get_category_name( WP_Post $post_obj, array $parsely_options ):
 		$taxonomy_dropdown_choice = get_the_terms( $post_obj->ID, $parsely_options['custom_taxonomy_section'] );
 		// Get top-level taxonomy name for chosen taxonomy and assign to $parent_name; it will be used
 		// as the category value if 'use_top_level_cats' option is checked.
-		// Assign as "Uncategorized" if no value is checked for the chosen taxonomy.
-		$category = 'Uncategorized';
```
GaryJones, line 731:
```diff
- 		$category = 'Uncategorized';
+		$category = get_cat_name( get_option( 'default_category' ) );
```

tests/Integration/StructuredData/SinglePostTest.php
```diff
@@ -652,7 +652,8 @@ public function test_post_with_categories_as_tags_without_categories(): void {
 		$parsely_options['cats_as_tags'] = true;
 		update_option( 'parsely', $parsely_options );
 
-		wp_remove_object_terms( $post_id, 'uncategorized', 'category' );
```
GaryJones, line 655:
```diff
-		wp_remove_object_terms( $post_id, 'uncategorized', 'category' );
+		wp_remove_object_terms( $post_id, get_category( get_option( 'default_category' ) )->slug, 'category' );

**Comment by GaryJones** [comment]
Needs testing, but those last three comments with diffs are probably all that is really needed to solve this.

**Comment by acicovic** [comment]
Hi @GaryJones, great suggestions, thank you very much. I've mostly done what you've proposed, with the difference that I extracted the slug into an external variable for easier readability.

I'm marking all your comments as resolved, but feel free to repost if you think that something could be better.

File: src/class-parsely.php [hunk 10]
```diff
@@ -694,8 +694,14 @@ private function get_categories( int $post_id, string $delimiter = '/' ): array
 		if ( $last_tag ) {
 			$tags = explode( '/', $last_tag );
 		}
-		// remove uncategorized value from tags.
-		return array_diff( $tags, array( 'Uncategorized' ) );
+
+		// Remove default category name from tags if needed.
+		$default_category_name = get_cat_name( get_option( 'default_category' ) );
+		if ( null !== $default_category_name ) {
```
pauarge: We should check for the type of the option returned in `get_option`. `get_cat_name` only accepts integers. Then, we don't need to check if that's null, because it will never be. `get_cat_name` always returns a string. If it can't find the category, it will return an empty string.

src/class-parsely.php
```diff
@@ -727,8 +733,8 @@ private function get_category_name( WP_Post $post_obj, array $parsely_options ):
 		$taxonomy_dropdown_choice = get_the_terms( $post_obj->ID, $parsely_options['custom_taxonomy_section'] );
 		// Get top-level taxonomy name for chosen taxonomy and assign to $parent_name; it will be used
 		// as the category value if 'use_top_level_cats' option is checked.
-		// Assign as "Uncategorized" if no value is checked for the chosen taxonomy.
-		$category = 'Uncategorized';
+		// Assign as the default category name if no value is checked for the chosen taxonomy.
+		$category_name = get_cat_name( get_option( 'default_category' ) );
```
pauarge, line 733: Same here. It's unlikely that `get_option('default_category')` fails, but I'd introduce an additional check just in case.

pauarge (changes requested):
Left a couple of comments regarding type safety.

src/class-parsely.php
acicovic: How about something like:
```php
$default_category_name = get_cat_name( (int) get_option( 'default_category' ) );
```
or:
```php
$default_category_name = get_cat_name( intval( get_option( 'default_category' ) ) );
```
I prefer the first one.

GaryJones: WordPress core generally doesn't add fallback values or casting for when `get_option( 'default_category' );` returns anything but an integer - it's not something that can be unset, short of running WP-CLI command or editing the database directly.

There is just [one place](https://github.com/WordPress/WordPress/blob/56d10309d0a20b877d8f50b48b5e02988e3954aa/wp-includes/taxonomy.php#L1974) where it is typecast, and that is for protecting against it being deleted.

While I'm in favour of adding useful protections and guards, it doesn't seem relevant here. If the default_option needs deleting, then we want it to shout and scream and be a fatal if necessary, as that's a bigger issue that needs to be fixed; covering it up doesn't help.

src/class-parsely.php
GaryJones: As above - if it's going to fail, then we need it to fail so the site admin can fix it.

src/class-parsely.php
GaryJones: `get_cat_name()` will never be `null`. If it fails, it will return an empty string, and that's fine to pass to `array_diff()`.

Please, just use the diff I suggested.

acicovic: I've removed the problematic `if` check.

pauarge (approved):
LGTM

GaryJones (approved):
✅

PR #622, opened 2022-01-27
Unifying AMP and Google Web Stories implementation

pauarge:
## Description

Google Web Stories is an extension over AMP. Hence, we can reuse some code of our existing AMP implementation. This PR introduces some changes in the AMP to make it modular so its pieces can be used. Then, those pieces are used in the GWS class.

## Motivation and Context

DRY 😄 

## How Has This Been Tested?

Install the AMP and GWS plugins and check that the tracking codes get inserted.


src/Integrations/class-google-web-stories.php
```diff
@@ -56,16 +38,19 @@ public function integrate(): void {
...
-			<script type="application/json">
-				{
-					"vars": {
-						"apikey": "<?php echo esc_js( $this->parsely->get_api_key() ); ?>"
-					}
-				}
-			</script>
-		</amp-analytics>
-		<?php
+		$json = Amp::construct_amp_json();
```
GaryJones, line 41: I wonder if this `$json` should be a method arg, so that other JSON could be passed in?

Maybe fine for now, but this static call is like a global, couples this class to `Amp` class, and it wouldn't changeable for tests, etc.

pauarge: This method is the one that gets called by the GWS plugin in the `web_stories_print_analytics` action. We could make a second function with that as an argument, but I think it's complicating the code for a fairly simple function (and strict in terms of what data is expected).

GaryJones (approved):
👍🏻

PR #623, opened 2022-01-27
Move documentation to docs directory

### LINKED ISSUE #621: Move some .md files into a docs directory
Our project has now multiple .md files like:
- README.md
- CHANGELOG.md
- CONTRIBUTING.md
- WINDOWS.md
- README.md (in `tests/e2e`)

We should perhaps consider grouping some of these files into a docs (or similarly-named) directory. Are there any opinions on how this should be done?

acicovic:
## Description
This PR creates a new `docs` folder and moves documentation `.md` files there.

## Motivation and Context
- Grouping documentation in one place is better than having it scattered in the project's root directory.
- Closes #621


acicovic:
A question here would be whether we would want to move any additional `.md` files to the `docs` directory.

pauarge (changes requested):
Let's keep in mind to change the links to these files on README and others.

acicovic:
> Let's keep in mind to change the links to these files on README and others.

Well, that was a glaring oversight. Will check this out.

.gitattributes
```diff
@@ -23,19 +23,20 @@
 /.eslintrc                       export-ignore
 /.gitattributes                  export-ignore
 /.gitignore                      export-ignore
+/.nvmrc                          export-ignore
 /.phpcs.xml.dist                 export-ignore
 /.prettierrc                     export-ignore
-/.nvmrc                          export-ignore
```
pauarge: Could we revert the changes that are not strictly necessary? `.nvmrc`, `package.json` and others do not need to be shuffled around.

acicovic: In my opinion, in small PRs like these such simple changes are not an issue. Having said that, I've reverted the alphabetic reordering.

GaryJones:
Is it really worth creating a directory for the two files? If there was a `docs/` directory, I'd expect a lot more documentation in there. I'm not sure what problem moving two files actually does.

acicovic:
> Is it really worth creating a directory for the two files? If there was a `docs/` directory, I'd expect a lot more documentation in there. I'm not sure what problem moving two files actually does.

I think it doesn't hurt to have it, and that it's ready for the future if we want to add more files.

pauarge:
I've been thinking about it and I think I lean towards what @GaryJones's saying. Maybe it's a bit overkill to have a dedicated folder for just two files. If there's a time in the future where we have a few more, by all means, let's do this change then.

Issue #609, opened 2022-01-20
Improve plugin page on wordpress.org


acicovic:
## Is your feature request related to a problem?
Our [wordpress.org plugin page](https://wordpress.org/plugins/wp-parsely/) could use some love. For example, as a beginning we could add more tags and fix the page's current display problems. Also, it seems that the plugin page doesn't completely obey to our README.md (for example, we do have extra tags in our README.md). Our plugin page is our interface with the world, so we should be able to manipulate it as best as possible.

## Describe the solution you'd like
Create a readme.txt file to be able to control the plugin's page contents. My understanding is that this will also allow us to have a different readme for the plugin's page and another one for this repository (which is be default the README.md).

Related material:
- [readme.txt documentation](https://developer.wordpress.org/plugins/wordpress-org/how-your-readme-txt-works/).
- The [readme.txt from Yoast](https://github.com/Yoast/wordpress-seo/blob/trunk/readme.txt) can be used as a reference as it contains everything (screenshots, FAQs, code blocks) and I think it powers everything in [their plugin page](https://wordpress.org/plugins/wordpress-seo/).

pauarge:
Noting that the broken syntax will be fixed with the release of 3.1.0.

acicovic:
OK, so we may need to see if it does make sense to go towards a readme.txt file. No need to do it if we the README.md is enough.

acicovic:
I've been informed by @GaryJones that having a single README.md file is intentional in order to avoid documentation duplication.

This should be OK, since we now have many different documentation files and we can move things from the main readme to other places if needed (which is already the case for the "Local Development" section for example).

Having said that, our [wp.org page](https://wordpress.org/plugins/wp-parsely/) still seems broken after the 3.1 release.

acicovic:
Here are some thing we can work on (will keep updating as I spot issues):
- [x] Fix broken code blocks.
- [x] Make all tags show.
> **Update:** Only some tags are shown, and this is a default behavior.
- [x] The "parsely" contributor has a person picture. I think a logo picture would make more sense for that.
> **Update:** @hbbtstar has been informed and is taking care of this.
- [x] Check what needs to be done to show that the plugin provides a block.
> **Update:** We don't provide a block yet, but we will soon. Using `block.json` would be the best way to go.
- [x] New screenshots of the plugin settings page are needed, as some of the latest PRs introduce new interface elements there.

GaryJones:
> Make all tags show.

[...]

Issue #564, opened 2021-12-29
Proposal: Extract classes from main Parsely class


acicovic:
## Description
In an ideal world, each class should have a specific responsibility, encapsulate it and expose members so other classes can use the functionality. The `Parsely` class still seems to have several responsibilities that could be extracted as external classes. This would improve code quality and simplify the main class.

Here are the ones I can think off the top of my head:

- A currently obvious one is `Metadata` as the `construct_parsely_metadata()` function is very long and (IMO) should be broken into sub-functions. I would propose the creation of a `Metadata` (or similarly named) class that would include this function and all its sub-functions, as well as any other metadata-related functionality of course.
- Another one that could be discussed is `Options`. This is a more complicated one as we're getting options in many places in different ways (haven't looked at this in detail).

## Next steps

1. Discuss and decide if we want to go forward with this idea
2. Propose/determine the extractions that would be useful
3. Determine the cost/implications of each possible extraction, as to decide if we should go forward with implementation (for example, is it simple? Will it require a lot of refactoring in the `Parsely` or other classes? Is the effort worth it?). Something to always keep in mind that the plugin is still small and that it's better to do expensive operations now, instead of possibly regretting it in the future.
4. Implement the good ones
5. Profit!

## Request for feedback from contributors

So here it is. Contributors, let me know of your comments, objections or feedback about this. I'll be happy to work on this, provided we decide what to extract. Please make your comments and propositions so we can determine the most useful extractions.

Thanks!

acicovic:
Related: #323

GaryJones:
> A currently obvious one is Metadata as the construct_parsely_metadata() function is very long and (IMO) should be broken into sub-functions. I would propose the creation of a Metadata (or similarly named) class that would include this function and all its sub-functions, as well as any other metadata-related functionality of course.

I would take this further - one class that handles the logic of getting the right values for the context, and two other polymorphic classes for the different outputs (JSON, repeated metas). Use a [polymorphism refactor](https://refactoring.guru/replace-conditional-with-polymorphism) to replace [this conditional](https://github.com/Parsely/wp-parsely/blob/501a7e5f7410206d62b9126883ff0622b76939b3/src/class-parsely.php#L198-L226).

[...]

PR #633, opened 2022-01-28
Settings page: Improve UI for post/page types to track

### LINKED ISSUE #483: Settings page: Improve UI for post/page types to track

### LINKED ISSUE #643: Consider building production CSS files
We generate production builds for our JS files but not for our CSS files. I'm not sure if this is by design or just something we haven't done yet.

We should consider if we want to be building our CSS files. On that same path, we could also think of including a CSS pre-processor in our pipeline which could probably do this job itself.

acicovic:
## Description
In the settings page, the UI for selecting the post/page types to track needs to be improved as to make it more straightforward and user friendly.

## Motivation and Context
- A more thorough discussion about this can be found in #483
- Closes #483
- Closes #643

## How Has This Been Tested?
- Manual testing of data saving and loading, including wiping the plugin's settings in the database to see how the change reacts to default/unsaved settings
- Testing that built CSS works by removing and re-enabling its wp_enqueue_style() function
- Checking on a browser that the table is responsive and works well in big and small widths
- Integration tests pass

### Before:
image

### After:
image


acicovic:
Hello there,

I've just committed the UI prototype for the Track Post Types functionality, the aim being to experience the UI as well as examine the code from an accessibility perspective.

Things to know:
- I've introduced a new CSS file to hold admin CSS, but I haven't created a built file for it (yet?).
- This is a UI-only commit and I haven't tested what happens when the form saves. A saving/loading commit will come later. For the time being, you're saving at your own risk!
- Fun fact: while trying to commit this, I stumbled upon #639

Regarding accessibility
- I've replaced the usual `<legend>` that we have in most fieldsets in the settings page with a `<caption>` within the table.
- Tabbing works as expected.
- In this case, accessibility can be tricky and there are many different screen readers. I decided to use `aria-labelledby`.

Please let me know of any findings, suggestions, etc.

acicovic:
Hi, I'm opening this for review.

SonarCloud is complaining because `class-settings-page.php` now has 21 methods from the "20 max allowed". It also suggests adding a `<legend>` in the `<fieldset>`, but I'm not sure this is needed in our case because we do have the `<caption>` within the `<table>`. If anyone has a better understanding of this, please let me know.

Thanks!

src/UI/class-settings-page.php
```diff
@@ -285,41 +287,19 @@ function (): void {
...
-			'option_key'     => 'track_page_types',
-			'help_text'      => $h,
-			'select_options' => get_post_types( array( 'public' => true ) ),
-			'label_for'      => Parsely::OPTIONS_KEY . "[$field_id]",
-		);
-		add_settings_field(
-			'track_page_types',
-			__( 'Page Types to Track', 'wp-parsely' ),
-			array( $this, 'print_multiple_checkboxes' ),
+			__( 'Track Post Types as...', 'wp-parsely' ),
```
pauarge: I would remove the `...` from the label, so leave it as `Track Post Types as`

acicovic:
Noting here some related improvements for completeness:
- ~~Build the CSS file~~ - Created as a new issue for all our CSS files in #643 
- ~~Add new UI screenshot to our WP.org page~~ - Added to #609 
 - [x] Make adjustments so the whole table is visible in very small screens (for example320px) when the post type name is really long

src/UI/class-settings-page.php
acicovic: Not pushed yet, but done.

src/UI/class-settings-page.php
```diff
@@ -76,23 +76,34 @@ public function run(): void {
...
 	 */
-	public function enqueue_settings_scripts( string $hook_suffix ): void {
+	public function enqueue_settings_assets( string $hook_suffix ): void {
 		if ( 'settings_page_parsely' === $hook_suffix ) {
 			wp_enqueue_media();
 
 			$admin_settings_asset = require plugin_dir_path( PARSELY_FILE ) . 'build/admin-settings.asset.php';
+			$built_assets_url     = plugin_dir_url( PARSELY_FILE ) . '/build/';
+
+			// Admin settings JS.
```
pauarge: I think this comment and the one in line 100 are redundant.

src/UI/class-settings-page.php
```diff
@@ -442,41 +453,19 @@ function (): void {
 		);
 
 		// Allow use of custom taxonomy to populate articleSection in parselyPage; defaults to category.
-		$h          = __( 'By default, Parse.ly only tracks the default post type as a post page. If you want to track custom post types, select them here!', 'wp-parsely' );
```
pauarge, line 445: Could we keep this help text (or something that's similar enough but fits with the current UI)?

src/UI/class-settings-page.php
```diff
@@ -442,41 +453,19 @@ function (): void {
...
-		$field_id   = 'track_post_types';
-		$field_args = array(
-			'title'          => __( 'Post Types to Track', 'wp-parsely' ),
-			'option_key'     => $field_id,
-			'help_text'      => $h,
-			'select_options' => get_post_types( array( 'public' => true ) ),
-			'label_for'      => Parsely::OPTIONS_KEY . "[$field_id]",
-		);
+		$field_id   = 'track_post_types_as';
+		$field_help = __( 'Select how you want to track every Post Type.', 'wp-parsely' );
```
pauarge: No need to capitalize `Post Type`

src/UI/class-settings-page.php
```diff
@@ -653,7 +642,7 @@ public function print_description_text( $args ) {
 	}
 
 	/**
-	 * Print out the radio buttons.
+	 * Print out an input text tag.
```
pauarge, line 643: 👍

src/UI/class-settings-page.php
```diff
@@ -821,6 +810,89 @@ public function print_media_single_image( array $args ) {
 		$this->print_description_text( $args );
 	}
 
+	/**
+	 * Print out the post tracking options table.
+	 *
```
pauarge, line 813: We're missing a `@since` annotation here.

src/UI/class-settings-page.php
```diff
@@ -821,6 +810,89 @@ public function print_media_single_image( array $args ) {
...
+	 * @param array $args The arguments used in the output HTML elements.
```
pauarge: Would it be possible to make the `array` type more concrete here, something similar to what we do in: https://github.com/Parsely/wp-parsely/blob/89c057b5783f99b90d9d7a37659b53221eee8503/src/Integrations/class-amp.php#L158?

(if the object type is too complex to represent, feel free to dismiss my comment)

src/UI/class-settings-page.php
```diff
@@ -821,6 +810,89 @@ public function print_media_single_image( array $args ) {
...
+
+	/**
+	 * Return the custom post type tracking values in a format that is easily
+	 * consumable by the print_track_post_types_table() function.
+	 *
+	 * @since 3.2.0
+	 *
+	 * @return array Key-value pairs with post type and their 'track as' value.
+	 */
+	public function get_tracking_values_for_display() {
```
pauarge: Missing return type

src/UI/class-settings-page.php
```diff
@@ -1054,6 +1089,63 @@ public function validate_options( array $input ): array {
...
+	 *
+	 * @since 3.2.0
+	 *
+	 * @param array $input Array passed to validate_options() function.
+	 * @return void
+	 */
+	private function validate_options_post_type_tracking( array &$input ): void {
+		$options         = $this->parsely->get_options();
+		$posts           = 'track_post_types';
+		$pages           = 'track_page_types';
```
pauarge, line 1104: Maybe we could pull these two out of the function as constants?

src/UI/class-settings-page.php
```diff
@@ -1094,7 +1186,7 @@ private static function get_logo_default(): string {
 	 * @param array $array Array of options to be sanitized.
 	 * @return array
 	 */
-	private static function validate_option_array( array $array ): array {
+	private static function sanitize_option_array( array $array ): array {
```
pauarge, line 1185: Why the name change?

pauarge (changes requested):
Amazing work! Just some small tweaks left.

src/UI/class-settings-page.php
acicovic: This was done for clarity, as this function doesn't validate, but rather _**sanitizes**_ the array passed to it. As such, `sanitize` is much more descriptive of what the function does in reality, and makes it stand out from the functions starting with `validate`.

As this is a private function, changing it wasn't risky, so I went ahead with it.

src/UI/class-settings-page.php
acicovic: So far we're only passing string key-value pairs so I've changed it to `array<string, string>`. By the way, syntax highlighting in VSCode for this isn't super great.

src/UI/class-settings-page.php
acicovic: Removed both.

src/UI/class-settings-page.php
acicovic: I thought of this myself, but in the bigger picture I would consider this as part of #564 as an `options` class or similar which would expose constants for every option key. However, I can do this as an intermediary step if you wish.

src/UI/class-settings-page.php
pauarge: Thanks for the change. This really helps PHPStan to infer types, in my setup (PHPStorm) works pretty well.

src/UI/class-settings-page.php
pauarge: Fair, we can leave it as is.

src/UI/class-settings-page.php
```diff
@@ -452,7 +452,7 @@ function (): void {
 
 		// Allow use of custom taxonomy to populate articleSection in parselyPage; defaults to category.
 		$field_id   = 'track_post_types_as';
-		$field_help = __( 'Select how you want to track every post type.', 'wp-parsely' );
+		$field_help = __( 'By default, Parse.ly only tracks posts and pages. If you want to track other post types, select how you want to track them here!', 'wp-parsely' );
```
pauarge: Can we remove the final exclamation mark? The text looks good otherwise.

tests/e2e/utils.js
```diff
@@ -45,3 +45,43 @@ export const checkH2DoesNotExist = async ( text ) => {
...
+	}
+
+	const [ input ] = await page.$x( '//p[contains(@class, \'submit\')]//input[contains(@name, \'screen-options-apply\')]' );
+	await input.click();
+};
+
+/**
+ * Save settings in the settings page and force a hard refresh.
+ */
+export const saveSettingsAndHardRefresh = async () => {
```
pauarge, line 82: Could we have similar docstring as the method above in this one?

pauarge (changes requested):
Super small change, nice work on the testing!

src/UI/class-settings-page.php
```diff
@@ -76,23 +76,32 @@ public function run(): void {
...
 		if ( 'settings_page_parsely' === $hook_suffix ) {
 			wp_enqueue_media();
 
 			$admin_settings_asset = require plugin_dir_path( PARSELY_FILE ) . 'build/admin-settings.asset.php';
+			$built_assets_url     = plugin_dir_url( PARSELY_FILE ) . '/build/';
+
 			wp_enqueue_script(
-				'parsely_admin_settings',
-				plugin_dir_url( PARSELY_FILE ) . '/build/admin-settings.js',
+				'parsely-admin-settings-js',
```
pauarge: Could we remove the `js` suffix to this script and the `css` to the CSS one? They'd have the same name but be different by the fact that one is a `script` enqueued and the other is a `style`.

pauarge (dismissed):
LGTM

pauarge (approved):

PR #634, opened 2022-01-28
Use WordPress scripts to export plugin

pauarge:
## Description

A recent update on the `@wordpress/scripts` package introduced the `wp-scripts plugin-zip` command, which creates a zip file containing the plugin. We do have a custom script that does the same.

This PR removes that custom script and adds a new command on `package.json`, `plugin-zip`, which relies on the one included on wp-scripts.

## Motivation and Context

Don't maintain a custom script with very similar functionality to the one provided by a package we already use.

## How Has This Been Tested?

Execute `npm run plugin-zip` on the root folder of the plugin. `package.json` linting passes.


acicovic (dismissed):

GaryJones (commented):
I was only able to get this working by doing a `npm update` first.

The list of includes files in the root is different. For the original script:

Screenshot 2022-01-31 at 09 38 33

For the new `plugin-zip` script:

Screenshot 2022-01-31 at 09 38 36

I'm fine with dropping the WINDOWS file, but does the export really need the `package.json`? Is there a way to exclude that?

Also, the `plugin-zip` seems to drop the `views/` directory as well, which is going to break things...

GaryJones (changes requested):
`views/` dir is missing.

pauarge:
> I'm fine with dropping the WINDOWS file, but does the export really need the package.json? Is there a way to exclude that?

I'm not sure we can easily exclude `package.json`, given that the tool is based in https://www.npmjs.com/package/npm-packlist.

GaryJones (approved):
✅

jblz (approved):
This adds an extra step to deploy to a remote filesystem, but no big deal. Worth it for less code to maintain.  👍🏼

PR #636, opened 2022-01-28
Use built version strings & deprecate Parsely::get_asset_cache_buster

jblz:
## Description

Instead of using `Parsely::get_asset_cache_buster` for the version of our scripts:
* Use the already present `version` in the backing asset  PHP file for each built script file
* Use the plugin version for the tracker (what gets used in production installs anyhow).

Deprecate `Parsely::get_asset_cache_buster` as it's a public method and no longer in use internally.

## Motivation and Context

Instead of tying this query argument to the plugin version, tie it to the version of the built script.  This allows us to continue busting the cache during development and simplifies the code (once the deprecated function is removed).

## How Has This Been Tested?

* Access the features powered by each script with a clean checkout of this branch
* Access the features powered by each script with a while running `npm run start` and making test changes. Ensure changes are reflected when you reload the page.


src/class-parsely.php
```diff
@@ -653,7 +655,9 @@ public static function get_asset_cache_buster(): string {
 		 *
 		 * @param string $cache_buster Plugin version, unless WP_DEBUG is defined and truthy, and tests are not running.
 		 */
-		return apply_filters( 'wp_parsely_cache_buster', (string) $cache_buster );
+		$cache_buster = apply_filters( 'wp_parsely_cache_buster', (string) $cache_buster );
+
+		return $cache_buster;
```
jblz: This is a fix for a bug that would result in inconsistent behavior if anyone has this `wp_parsely_cache_buster ` filter set.

Namely, the static variable would be set above, the result filtered once and returned, & the unfiltered version used on subsequent calls.  

Since this change deprecates the function altogether, this filter should also be deprecated, but I wasn't sure if we should fix the bug or leave it as is.  Thoughts?

GaryJones: Use `apply_filters_deprecated()`.

src/class-parsely.php
```diff
@@ -653,7 +655,9 @@ public static function get_asset_cache_buster(): string {
 		 *
 		 * @param string $cache_buster Plugin version, unless WP_DEBUG is defined and truthy, and tests are not running.
 		 */
-		return apply_filters( 'wp_parsely_cache_buster', (string) $cache_buster );
+		$cache_buster = apply_filters_deprecated( 'wp_parsely_cache_buster', array( (string) $cache_buster ), '3.2.0' );
+
+		return $cache_buster;
```
pauarge, line 660: Is there any reason to introduce a variable instead of returning the value directly?

acicovic: Although I'll let @jblz respond to this, I do use variables to store results of long functions or complicated expressions, as it makes the code easier to follow.

`return $cache_buster` is easier to understand much more quickly than `return apply_filters_deprecated( 'wp_parsely_cache_buster', array( (string) $cache_buster ), '3.2.0' )`

pauarge: Given that this is the return statement, we can assume that whatever that line returns is described in the function header documentation. In this case, I see that we're applying a filter that does something that "Returns a random number string or plugin version string". We also have a description of what the filter does on top of it.

IMO it's better to avoid declaring variables unless we really need them.

acicovic: No strong preference for me in the specific case.

GaryJones: I agree with Pau here - no point in assigning the value to a variable only to return the variable - the value could be returned instead. 

If the issue is with the value not being self-documented, then we need to look at the naming of the function.

jblz: This assignment was actually intended to serve a purpose. `$cache_buster` is static across calls to the function serving as an in-memory cache of the calculated result of the result of any filtering functions.  

As [noted in an earlier comment](https://github.com/Parsely/wp-parsely/pull/636#discussion_r794624378), this is a bug fix, but I'm not sure if we should change it or leave it as is given the function is being deprecated. Thoughts?

pauarge: That is a fair point. In this case, happy to leave it as-is.

pauarge (approved):
🚢

PR #642, opened 2022-02-02
Add Recommendations Block

jblz:
## Description
Introduce a `Recommendations Block` that displays links to relevant pages, powered by Parse.ly's Recommendations (`/related`) API.

## Motivation and Context
See: https://github.com/Parsely/web/issues/12585

## How to perform manual testing
1. Add the following line to an mu-plugin to enable the flags for the API Endpoint and Block:
    ```
    add_filter( 'wp_parsely_recommendations_block_enabled', '__return_true' );
    ```
2. Specify a valid API key on the plugin's options page, in order to test against a "real" site's recommendations.
3. Drop the block in various places (a post, page, full-site editor).
4. Tinker with different Block settings to see the effects they have.

## Screenshots
Screenshot from 2022-02-02 17-28-02


acicovic:
Note: See what minimum WP version we should target.

pauarge:
On bumping the WordPress version: I think we should stick to 5.0 and disable the block for lower than 5.2 with a version check. I see us bumping the minimum requirement to 5.2 in the next major version.

GaryJones:
> disable the block for lower than 5.2

Why?

jblz:
RE: bumping the minimum WordPress version:

Our current usage of `registerBlockType` with `{ apiVersion: 2 }` actually [requires a minimum of 5.6](https://make.wordpress.org/core/2020/11/18/block-api-version-2/).

I think the `5.2` is likely a typo on my part.  There are still a [couple of vestiges](https://github.com/Parsely/wp-parsely/blob/2018c98f1ee0f7e4e8572b837836d87bd7849582/src/blocks/recommendations/class-recommendations-block.php#L17-L25) in this branch of the [old one](https://github.com/Parsely/wp-parsely/pull/318).

It would likely be possible to refactor to use the old API, but 5.6 has been [out for 14 months](https://wordpress.org/news/2020/12/simone/) at this point, so I don't think it'd be a huge lift to bump the minimum.

I vote we change the minimum tag whichever release this lands outside of the filter / feature flag gate.

acicovic:
Other things to see here for our first version are:
- Is there anything that should be added?
- Is there anything that should be removed?
- Is there anything that should be changed?

src/blocks/recommendations/README.md
```diff
@@ -0,0 +1,66 @@
+# Parse.ly Recommendations Block
+
+This block is designed to showcase links to content on your site as provided by the [Parse.ly `/related` API endpoint](https://www.parse.ly/help/api/recommendations#get-related).
+
+You can use the Parse.ly Recommendations Block wherever you use the WordPress Block Editor ("Gutenberg") to build your site. That means that not only can leverage this feature in single `Post`'s content (or `Page` or nearly any custom post type), you can also leverage it in the Full Site Editor's template editor [slated to debut with WordPress version 5.8](https://make.wordpress.org/core/2021/06/16/introducing-the-template-editor-in-wordpress-5-8/) to add it to various layout elements (e.g. a Header, Footer, or Sidebar).
```
acicovic: I found the wording a bit complicated for me, trying to simplify a bit:

```suggestion
You can use the Parse.ly Recommendations Block wherever you use the WordPress Block Editor ("Gutenberg") to build your site. That means that you can leverage this feature in `Posts`, `Pages`, or nearly any other custom post type. You can also leverage it in the Full Site Editor's template editor [slated to debut with WordPress version 5.8](https://make.wordpress.org/core/2021/06/16/introducing-the-template-editor-in-wordpress-5-8/) to add it to various layout elements (e.g. a Header, Footer, or Sidebar).
```

src/Endpoints/README.md
```diff
@@ -0,0 +1 @@
+# wp-parsely Plugin WordPress REST API Endpoints
```
jblz: We should pull out `src/Endpoints/README.md` to separate PR & flesh it out with information about how we build our endpoints.

src/blocks/recommendations/README.md
```diff
@@ -0,0 +1,66 @@
+# Parse.ly Recommendations Block
+
+This block is designed to showcase links to content on your site as provided by the [Parse.ly `/related` API endpoint](https://www.parse.ly/help/api/recommendations#get-related).
```
jblz: We should link to documentation for the related proxy endpoint here as well as the backend API.

src/blocks/recommendations/README.md
```diff
@@ -0,0 +1,66 @@
+# Parse.ly Recommendations Block
```
jblz, line 1:
```suggestion
# Parse.ly Related Links Block
```

Should we consider changing the name of the block (& all underlying nomenclature) to match the endpoint? I can see arguments both ways. For example, we may end up composing the results of multiple endpoints to offer recommendations.

src/blocks/recommendations/components/parsely-recommendations-block-controls.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { BlockControls } from '@wordpress/block-editor';
+import { Dashicon, ToolbarGroup } from '@wordpress/components';
+
+const ParselyRecommendationsBlockControls = ( { attributes: { layoutstyle }, setAttributes } ) => {
```
jblz: Let's consider removing the "Layout Style" toggle here and in the "Inspector Controls" (if it's pulled out, we should remove support for the `layoutstyle` attribute throughout).

src/blocks/recommendations/components/parsely-recommendations-inspector-controls.js
```diff
@@ -0,0 +1,200 @@
...
+								{ label: __( 'Thumbnail from Parse.ly', 'wp-parsely' ), value: 'thumbnail' },
+							] }
+							onChange={ ( newval ) =>
+								setAttributes( {
+									imagestyle: newval === 'original' ? 'original' : 'thumbnail',
+								} )
+							}
+						/>
+					</PanelRow>
+				) }
```
jblz, line 67: The images that we pull for this are of varying quality. Let's consider how to ensure that when images are enabled, we only show recommendations with a reasonably large image file.

src/blocks/recommendations/README.md
acicovic: I personally prefer "Related Links" over "Recommendations" as it is more descriptive/precise of what the block actually does. What do other contributors think about this?

src/blocks/recommendations/components/parsely-recommendations-block-controls.js
acicovic: @jblz could you remind us of the reasoning for this, and what would be the desired result/advantage of doing so?

src/Endpoints/README.md
acicovic: Now tracked in #702.

src/blocks/recommendations/README.md
acicovic: Now tracked in #702.

src/blocks/recommendations/components/parsely-recommendations-inspector-controls.js
acicovic: Now tracked in #703.

src/blocks/recommendations/components/parsely-recommendations-block-controls.js
jblz: @acicovic Just to simplify the MVP. The least complex rendering would be as simple an unordered-list as possible (minimal CSS and intermediate container nesting added by our block). This would push responsibility to the theme / site implementer to ensure the design elements flow with the rest of the site and would not lock us into supporting a more complex layout going forward until we're confident it's a real winner.

acicovic:
We also have 2 feature flags for this to work. Should this remain as is, or should we put this under a single flag?

acicovic:
Also, we have 2 feature flags in order to enable this. Should this be changed to a single feature flag?

src/blocks/recommendations/README.md
pauarge: Let's go with Recommendations Block. The related name is tied to a specific endpoint, but Parse.ly refers to this feature as Recommendations: https://www.parse.ly/help/api/recommendations

src/blocks/recommendations/components/parsely-recommendations-block-controls.js
pauarge: Let's remove it. We should also create an issue to track it and implement it later, perhaps we can paste some code in that issue to pick it up easily.

pauarge:
> Also, we have 2 feature flags in order to enable this. Should this be changed to a single feature flag?

It does make sense to have just a single feature flag. I would leave `wp_parsely_recommendations_block_enabled` on.

pauarge:
As per the minimum WordPress version, let's stick to 5.0 and only enable the block if we're above that.

pauarge:
What's left here is:

- [ ] Adding unit/integration tests for the JavaScript bits.
- [ ] Adding unit/integration tests for the PHP bits.
- [ ] Adding E2E tests.

acicovic:
> As per the minimum WordPress version, let's stick to 5.0 and only enable the block if we're above that.

Per https://github.com/Parsely/wp-parsely/pull/642#issuecomment-1032574708:
> Our current usage of registerBlockType with { apiVersion: 2 } actually [requires a minimum of 5.6](https://make.wordpress.org/core/2020/11/18/block-api-version-2/).

This means that our WP minimum is 5.0, but the minimum to use the **Block** will be 5.6 as this is what is required to use `apiVersion 2`. WordPress 5.6 “Simone” was released to the public on December 8, 2020.

src/blocks/recommendations/components/parsely-recommendations-block-controls.js
acicovic: This has been removed in ffefdce and is being tracked in #708, so marking this as resolved.

src/blocks/recommendations/components/parsely-recommendations-fetcher.js
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Internal dependencies
+ */
+import { setError, setRecommendations } from '../actions';
+import { useRecommendationsStore } from '../recommendations-store';
+
+const ParselyRecommendationsFetcher = ( { boost, limit, sort } ) => {
+	const {
+		state: {},
```
acicovic: SonarCloud complains about this. `state = {}` could possibly fix the complaint, but then my Intelisense complains about the variable not being used. If anyone knows of a graceful way to handle this, let me know!

acicovic:
I'm opening this for review and will start writing some tests.

src/RemoteAPI/class-base-proxy.php
```diff
@@ -43,7 +43,7 @@ public function __construct( Parsely $parsely ) {
 	}
 
 	/**
-	 * Get URL for Recommendation API (GET /related).
+	 * Get URL for Recommendations API (GET /related).
```
GaryJones:
```suggestion
	 * Get URL for a particular Parse.ly API endpoint.
```

This base proxy class doesn't know anything about Recommendations or other endpoints.

src/blocks/recommendations/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+This block is designed to showcase links to content on your site as provided by the [Parse.ly `/related` API endpoint](https://www.parse.ly/help/api/recommendations#get-related).
+
+You can use the Parse.ly Recommendations Block wherever you use the WordPress Block Editor ("Gutenberg") to build your site. That means that you can leverage this feature in `Posts`, `Pages`, or nearly any other custom post type. You can also leverage it in the Full Site Editor's template editor [slated to debut with WordPress version 5.8](https://make.wordpress.org/core/2021/06/16/introducing-the-template-editor-in-wordpress-5-8/) to add it to various layout elements (e.g. a Header, Footer, or Sidebar).
```
GaryJones:
```suggestion
You can use the Parse.ly Recommendations Block wherever you use the WordPress Block Editor to build your site. That means that you can leverage this feature in `Posts`, `Pages`, or nearly any other custom post type. You can also leverage it in the Full Site Editor's template editor [slated to debut with WordPress version 5.8](https://make.wordpress.org/core/2021/06/16/introducing-the-template-editor-in-wordpress-5-8/) to add it to various layout elements (e.g. a Header, Footer, or Sidebar).
```

While there are still references to the code feature code name (and plugin name) Gutenberg, the feature now in core is called the block editor, and we should be comfortable just using that terminology.

Not changed in my suggestion above, but we should be able to replace that incorrect "coming soon in 5.8 reference" with something that better describes what landed in 5.9.

src/blocks/recommendations/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Disabled by Default
+
+Note that this block is currently behind a "feature flag" and must be explicitly enabled for the code to load.
+
+To enable the block on your site, add the following line anywhere in your theme code (or a custom plugin) that is sourced [prior to the `init` WordPress action](https://codex.wordpress.org/Plugin_API/Action_Reference) (top-level in your theme or plugin is fine):
+
+`add_filter( 'wp_parsely_recommendations_block_enabled', '__return_true' );`
+
+This restriction will be lifted and the block code loading will become opt-out in a future release.
```
GaryJones, line 15: Why are we putting this feature behind a flag in the first place? Are we not confident about our own code?

src/blocks/recommendations/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+## How to Use
+
+- Inside the Block Editor, add the `Parse.ly Recommendations` block via the [standard block controls](https://wordpress.org/support/article/adding-a-new-block/).
+- Use the Block and Inspector Controls to configure the [block attributes](#block-attributes)
```
GaryJones:
```suggestion
- Use the Block and Inspector Controls to configure the [block attributes](#block-attributes).
```

src/blocks/recommendations/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Block Attributes
```
GaryJones, line 22: Feels like some of these entries need more description (and not just "it gets passed to the API"), and maybe a list of available options. We can include links to more detailed documentation elsewhere to keep it canonical.

src/blocks/recommendations/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+Note that this block is currently behind a "feature flag" and must be explicitly enabled for the code to load.
+
+To enable the block on your site, add the following line anywhere in your theme code (or a custom plugin) that is sourced [prior to the `init` WordPress action](https://codex.wordpress.org/Plugin_API/Action_Reference) (top-level in your theme or plugin is fine):
+
+`add_filter( 'wp_parsely_recommendations_block_enabled', '__return_true' );`
+
+This restriction will be lifted and the block code loading will become opt-out in a future release.
+
+## How to Use
```
GaryJones, line 17: This README isn't going to be surfaced to anyone except developers looking through source code, so customer-facing information from here should go elsewhere and leave this file to be aimed at a developer audience.

src/blocks/recommendations/constants.js
```diff
@@ -0,0 +1,3 @@
+export const RECO_BLOCK_ERROR = 'RECO_BLOCK_ERROR';
+export const RECO_BLOCK_LOADED = 'RECO_BLOCK_LOADED';
+export const RECO_BLOCK_RECOMMENDATIONS = 'RECO_BLOCK_RECOMMENDATIONS';
```
GaryJones: Do we have a limit on the length of constant names? Why are these un-necessarily abbreviated?

src/blocks/recommendations/class-recommendations-block.php
```diff
@@ -0,0 +1,145 @@
...
+
+declare(strict_types=1);
+
+namespace Parsely;
+
+/**
+ * Parse.ly Recommendations Gutenberg Block.
+ *
+ * Gutenberg Block that lists Parse.ly recommendations using the /related
+ * Parse.ly API, which is made possible by the Related_API_Proxy endpoint.
```
GaryJones:
```suggestion
 * Parse.ly Recommendations Block.
```

I don't think there's any value in describing the implementation in the class DocBlock.

src/blocks/recommendations/class-recommendations-block.php
```diff
@@ -0,0 +1,145 @@
...
+ */
+class Recommendations_Block {
+	const MINIMUM_WORDPRESS_VERSION = '5.6';
+
+	/**
+	 * Determine whether the block and its assets should be registered.
+	 *
```
GaryJones, line 23: All of the methods in this class should have a `@since 3.2.0` tag.

src/blocks/recommendations/class-recommendations-block.php
```diff
@@ -0,0 +1,145 @@
...
+		if ( ! isset( $wp_version ) || version_compare( $wp_version, self::MINIMUM_WORDPRESS_VERSION ) < 0 ) {
+			// WordPress is not recent enough to run this block.
+			return;
+		}
+
+		self::register_block_and_assets();
+	}
+
+	/**
+	 * Registers all block assets so that they can be enqueued through Gutenberg in
```
GaryJones:
```suggestion
	 * Registers all block assets so that they can be enqueued through the block editor in
```

src/blocks/recommendations/class-recommendations-block.php
```diff
@@ -0,0 +1,145 @@
...
+			'wp-parsely/recommendations',
+			array(
+				'editor_script'   => 'wp-parsely-recommendations-block-editor',
+				'render_callback' => __CLASS__ . '::render_callback',
+				'script'          => 'wp-parsely-recommendations-block',
+				'style'           => 'wp-parsely-recommendations-block',
+				'supports'        => array(
+					'html' => false,
+				),
+				'attributes'      => array(
```
GaryJones: Let's return this array from its own method.

That reduces the level of indentation here (and the whole args array starting on L79 could be pulled into its own variable to keep the `register_block_type()` to a single line for code readability.

With the attributes array in its own method, it can be re-used in `render_callback()`, loop through the array keys and create a `data-{$attribute_name}="<?php echo esc_attr( $attributes[ $attribute_name ] ); ?>"`, instead of duplicating lines 135-140.
Looks like `tag` is an exception here, so either `break` if the `$attribute_name` matches that, or store the attributes array in the new method without it, and add `tag` item to the array in this method instead. The latter is probably cleaner.

src/blocks/recommendations/class-recommendations-block.php
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	/**
+	 * The Server-side render_callback for the wp-parsely/recommendations block.
+	 *
+	 * @uses wp_validate_redirect If the stored results aren't considered "safe" by this function, they're skipped.
+	 * @param array $attributes The user-controlled settings for this block.
+	 * @return string
+	 */
+	public static function render_callback( $attributes ) {
```
GaryJones: Can type declarations for the parameter and return be added here?

src/blocks/recommendations/class-recommendations-block.php
```diff
@@ -0,0 +1,145 @@
...
+					),
+				),
+			)
+		);
+	}
+
+	/**
+	 * The Server-side render_callback for the wp-parsely/recommendations block.
+	 *
+	 * @uses wp_validate_redirect If the stored results aren't considered "safe" by this function, they're skipped.
```
GaryJones: I don't see `wp_validate_redirect()` being used in this method.

src/blocks/recommendations/components/parsely-recommendations-fetcher.js
```diff
@@ -0,0 +1,75 @@
...
+
+		const data = response?.data || [];
+		dispatch( setRecommendations( { recommendations: data } ) );
+	}
+
+	const apiMemoProps = [ ...Object.values( query ) ];
+
+	const updateRecosWhenPropsChange = useCallback( fetchRecos, apiMemoProps );
+
+	const debouncedUpdate = useDebounce( updateRecosWhenPropsChange, 300 );
```
GaryJones: Can this magic number be extracted into a variable, or be commented to say what it means? (Presumably 300 seconds?)

src/blocks/recommendations/components/parsely-recommendations-inspector-controls.js
```diff
@@ -0,0 +1,183 @@
...
+				<PanelRow>
+					<TextControl
+						label={ __( 'Title', 'wp-parsely' ) }
+						value={ title }
+						onChange={ ( newval ) => setAttributes( { title: newval } ) }
+					/>
+				</PanelRow>
+				<PanelRow>
+					<RangeControl
+						label={ __( 'Maximum Results', 'wp-parsely' ) }
```
GaryJones, line 32:
```suggestion
						label={ __( 'Maximum Number Of Recommendations', 'wp-parsely' ) }
```

Might be a bit too wordy, but trying to make it more contextual for the user.

src/blocks/recommendations/components/parsely-recommendations-inspector-controls.js
```diff
@@ -0,0 +1,183 @@
...
+						min="1"
+						max="99"
```
GaryJones: 99 seems redundantly excessive. How many does the Parse.ly API return by default?

src/blocks/recommendations/components/parsely-recommendations-inspector-controls.js
```diff
@@ -0,0 +1,183 @@
...
+						value={ tag }
+						onChange={ ( newval ) => setAttributes( { tag: newval } ) }
+					/>
+				</PanelRow>
+				<PanelRow>
+					<SelectControl
+						label={ __( 'Sort Recommendations', 'wp-parsely' ) }
+						value={ sort }
+						options={ [
+							{ label: 'Score', value: 'score' },
```
GaryJones: The `Score` string isn't internationalised.

src/blocks/recommendations/components/parsely-recommendations-inspector-controls.js
```diff
@@ -0,0 +1,183 @@
...
+							{
+								label: __( 'Desktop Views', 'wp-parsely' ),
+								value: 'desktop_views',
+							},
+							{
+								label: __( 'Visitors', 'wp-parsely' ),
+								value: 'visitors',
+							},
+							{
+								label: __( 'Visitors New', 'wp-parsely' ),
```
GaryJones: Can we use the Description columns from https://www.parse.ly/help/api/available-metrics to populate the labels in these options? (Or at least get the labels reading better in English.)

```suggestion
								label: __( 'New Visitors', 'wp-parsely' ),
```

src/blocks/recommendations/components/parsely-recommendations-list-item.js
```diff
@@ -0,0 +1,45 @@
...
+	const imageForLink = showimages && getImageForLink( { imagestyle, imageUrl, thumbUrlMedium } );
+
+	return (
+		<li>
+			<a href={ linkUrl } className="parsely-recommendations-link">
+				<Card className="parsely-recommendations-card" size="custom">
+					{ imageForLink && (
+						<CardMedia className="parsely-recommendations-cardmedia">
+							<img
+								className="parsely-recommendations-list-img"
```
GaryJones: Why does the classname here include `-list` when the other classes (also in a list), don't?

Also, the element is already called `img`, so should the class name end with `-image` instead?

src/blocks/recommendations/components/parsely-recommendations-list.js
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Internal dependencies
+ */
+import ParselyRecommendationsListItem from './parsely-recommendations-list-item';
+
+const ParselyRecommendationsList = ( { imagestyle, recommendations, showimages } ) => (
+	<ul
+		className={ classNames(
+			'parsely-recommendations-ul'
```
GaryJones:
```suggestion
			'parsely-recommendations-list'
```

The person writing the CSS shouldn't really care if this is a `ul` or `ol`. Chances are, the returned articles **_are_** ordered in some way, so maybe `ol` is correct for the element anyway?

src/blocks/recommendations/components/parsely-recommendations-title.js
```diff
@@ -0,0 +1,5 @@
+const ParselyRecommendationsTitle = ( { title } ) => (
+	<>{ title && <p className="parsely-recommendations-list-title">{ title }</p> }</>
```
GaryJones: What do the opening and closing elements on this line do?

src/blocks/recommendations/edit.js
```diff
@@ -0,0 +1,53 @@
...
+import { createBlock, registerBlockType } from '@wordpress/blocks';
+import { useBlockProps } from '@wordpress/block-editor';
+
+/**
+ * Internal dependencies
+ */
+import ParselyRecommendations from './components/parsely-recommendations';
+import ParselyRecommendationsInspectorControls from './components/parsely-recommendations-inspector-controls';
+import RecommendationsStore from './recommendations-store';
+import { ReactComponent as LeafIcon } from './Parsely-Logo-sRGB-Leaf-Green.svg';
```
GaryJones: Can the file that this relates to be renamed to use all lowercase characters in the file name?

src/blocks/recommendations/style.scss
```diff
@@ -0,0 +1,18 @@
...
+
+.parsely-recommendations-cardbody {
+	overflow: hidden;
+	padding: 0.8em;
+	text-overflow: ellipsis;
+	white-space: nowrap;
+}
+
+.parsely-recommendations-cardmedia {
+	padding: 0.8em 0.8em 0 0.8em;
```
GaryJones:
```suggestion
	padding: 0.8em 0.8em 0;
```

If the last positional item is removed, then it uses the value from the 2nd positional item.

wp-parsely.php
```diff
@@ -139,6 +139,18 @@ function parsely_rest_api_init(): void {
...
+/**
+ * Register the Recommendations Block.
+ *
+ * @return void
+ */
+function init_recommendations_block() {
+	$recommendations_block = new Recommendations_Block();
+	$recommendations_block->run();
+}
+add_action( 'init', __NAMESPACE__ . '\\init_recommendations_block' );
```
GaryJones: Just shuffles the order around so the add_action() call comes right before the DocBlock of the callback. This order makes it easier to understand what parameters might be going into the callback.

```suggestion
require __DIR__ . '/src/blocks/recommendations/class-recommendations-block.php';

add_action( 'init', __NAMESPACE__ . '\\init_recommendations_block' );
/**
 * Register the Recommendations Block.
 *
 * @return void
 */
function init_recommendations_block(): void {
	$recommendations_block = new Recommendations_Block();
	$recommendations_block->run();
}
```

Adds `: void` as well.

GaryJones (dismissed):
I love the work that's gone into it. Had some feedback and questions.

src/RemoteAPI/class-base-proxy.php
acicovic: Good catch. Will also fix another comment in the class and push it in a single commit.

src/blocks/recommendations/README.md
acicovic: I went ahead and replaced "Gutenberg" with "WordPress Block Editor" in the Recommendations Block's code.

Concerning the bit about FSE, this will have to be revised anyway, as the Block is currently not stable in FSE mode.

src/blocks/recommendations/README.md
acicovic: The block is in beta and currently has known issues (#711, #703). We are scheduled for release on 21 March, and probably the time is not enough to stabilize it and remove the feature flag.

src/blocks/recommendations/README.md
acicovic: Thanks for the input. Docs need to be re-read and revised. This is now tracked in #714 and I will be adding all related info which is here in that issue. I'll resolve the other document-related conversations here, so we can focus on code in this space.

acicovic:
FYI: All documentation issues are tracked in #714

src/blocks/recommendations/constants.js
acicovic: I agree, names changed.

src/blocks/recommendations/style.scss
acicovic: TIL - though we could leave this as is for readability/simplicity?

src/blocks/recommendations/components/parsely-recommendations-fetcher.js
acicovic: This seems to be the minimum milliseconds of wait before re-invoking the function (if needed), to ensure that calls are not executed too frequently. So in this case, even if props were to change every 50ms, we would be calling the respective function every 300ms.

I've updated the code but left this open for visibility, feel free to comment on this or mark it as resolved. As a bonus, I've changed "Recos" to "Recommendations".

src/blocks/recommendations/components/parsely-recommendations-inspector-controls.js
acicovic: I get the intention. IMO "results" is clear enough and widely used. But I'll keep this open just in case.

src/blocks/recommendations/components/parsely-recommendations-inspector-controls.js
acicovic: I sent an API request without the `limit` parameter and I'm getting 9 results, but I guess it could be more had I passed another site ID. Parse.ly docs list a maximum **pagination** limit of 2000 (!!!) records, but there's no mention of the default number of results.

For the Block, I think that a sensible number by default could be 5, and we could probably lower the maximum to something like 40 or 50? (still much but maybe there's a use case?) Open to suggestions...

src/blocks/recommendations/style.scss
GaryJones: For those who would be reading this, they should know what this means, and if not, it's a good excuse to spend 2 minutes and learn about how positional values work in CSS.

Code that doesn't exist, is the easiest to read in this case.

acicovic:
Re-copy/pasting this here, as GitHub hides a lot from the conversations:

What's left here is:

- [x] Rename to Recommendations Block
- [x] Only require one feature flag
- [x] Remove support for the `layoutstyle` attribute and track it in a separate issue
- [x] Finalize Block CSS
- [x] Perform some manual testing and fix/address/file possible issues
- [ ] Add unit/integration tests for the JavaScript bits
- [x] Add unit/integration tests for the PHP bits
- [ ] Add E2E tests
- [x] Integrating feedback from code reviews
- [x] Write/update minimum documentation

src/blocks/recommendations/components/parsely-recommendations-list.js
acicovic: I renamed the class with the `list` suffix (as part of a bigger commit including the previous suggestion), but haven't updated the`ul` to an `ol`. There is some logic from the Parse.ly API that obviously sorts the items in the specific order, but I'm unsure I would go for this with a `ol`. Any additional opinions/feedback welcome.

src/blocks/recommendations/components/parsely-recommendations-title.js
acicovic: They are required, else we get parsing errors.

pauarge: This is the short syntax for a React Fragment: https://reactjs.org/docs/fragments.html

They are technically required for all components, but sometimes if the component is simple enough, they can be omitted. This is not the case so they are needed. If we don't want them, we should rewrite this as:

``` ({title}) => title && <p>...</p>```

File: src/blocks/recommendations/class-recommendations-block.php (hunk 18 above)
Comment (acicovic) [reply src/blocks/recommendations/class-recommendations-block.php]: I've extracted the method.

Concerning `register_block_type()`, even with the `foreach` approach, the tag attribute was not output. I could have left as is, but for good measure I decided to unset the tag attribute from the array before processing (and we can do similarly for attributes that we would like to keep off the output in the future if this occurs). Previously, I had created an `$ignored_attributes` array, but this needed a call to `in_array` on each iteration and the code was more wordy. So I opted for unsetting.

Leaving this open for possible feedback.

File: src/blocks/recommendations/README.md [hunk 32]
```diff
@@ -1,10 +1,10 @@
...
 
-You can use the Parse.ly Recommendations Block wherever you use the WordPress Block Editor to build your site. That means that you can leverage this feature in `Posts`, `Pages`, or nearly any other custom post type. You can also leverage it in the Full Site Editor's template editor [slated to debut with WordPress version 5.8](https://make.wordpress.org/core/2021/06/16/introducing-the-template-editor-in-wordpress-5-8/) to add it to various layout elements (e.g. a Header, Footer, or Sidebar).
+You can add it in `Posts`, `Pages`, or nearly any other custom post type. From WordPress 5.8 and up, the block can also be used as a [Block-based Widget](https://wordpress.org/support/article/block-based-widgets-editor/). Note however, that the Recommendations Block is not available in the Full Site Editor's template editor. This restriction will be lifted in the future.
 
-### Disabled by Default
+## Disabled by Default
 
 Note that this block is currently behind a "feature flag" and must be explicitly enabled for the code to load.
```
pauarge:
```suggestion
Note that this block is currently behind a feature flag and must be explicitly enabled for the code to load.
```

src/blocks/recommendations/class-recommendations-block.php
```diff
@@ -0,0 +1,168 @@
...
+
+		register_block_type( 'wp-parsely/recommendations', self::get_block_registration_args() );
+	}
+
+	/**
+	 * Return the Block's registration arguments.
+	 *
+	 * @since 3.2.0
+	 *
+	 * @return array
```
pauarge:
```suggestion
	 * @return array<string, mixed>
```

README.md
```diff
@@ -85,6 +85,10 @@ A sample `JSON-LD` meta tag and structured data for an article or post:
 </script>
 ~~~
 
+## The Parse.ly Recommendations Block (beta)
+
+The plugin includes a (currently in beta and disabled by default) Recommendations Block that uses the [Parse.ly API](https://www.parse.ly/help/api/recommendations#get-related) to showcase links to content on your site. More information about enabling the Recommendations Block can be found in its [documentation](https://github.com/Parsely/wp-parsely/blob/trunk/src/blocks/recommendations/README.md).
```
pauarge:
```suggestion
The plugin includes a Recommendations Block that uses the [Parse.ly API](https://www.parse.ly/help/api/recommendations#get-related) to showcase links to content on your site. More information about enabling the Recommendations Block can be found in its [documentation](https://github.com/Parsely/wp-parsely/blob/trunk/src/blocks/recommendations/README.md). This feature is currently in beta and disabled by default.
```

src/blocks/recommendations/components/parsely-recommendations-list.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * External dependencies
+ */
+import classNames from 'classnames';
```
pauarge: I don't think we need the `classNames` package at all. Notice that its raison d'être is to join different classes, but we're not really joining anything.

We could remove it from `package.json` and then have to rerun `npm install` to update `package-lock.json`.

src/blocks/recommendations/components/parsely-recommendations-list.js
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Internal dependencies
+ */
+import ParselyRecommendationsListItem from './parsely-recommendations-list-item';
+
+const ParselyRecommendationsList = ( { imagestyle, recommendations, showimages } ) => (
+	<ul
+		className={ classNames(
+			'parsely-recommendations-list'
+		) }
```
pauarge:
```suggestion
		className=recommendations-list'
```

package.json
```diff
@@ -37,16 +37,27 @@
...
 		"js-cookie": "^3.0.1"
 	},
 	"devDependencies": {
+		"@wordpress/api-fetch": "^6.1.1",
 		"@wordpress/babel-preset-default": "^6.6.1",
+		"@wordpress/block-editor": "^8.3.1",
+		"@wordpress/blocks": "^11.3.1",
+		"@wordpress/components": "^19.6.1",
+		"@wordpress/compose": "^5.2.1",
+		"@wordpress/dom-ready": "^3.4.1",
```
pauarge: This could be removed as a dev-dependency since it's a dependency already.

package.json
```diff
@@ -37,16 +37,27 @@
...
 		"@wordpress/e2e-test-utils": "^7.0.1",
+		"@wordpress/editor": "^12.3.1",
```
pauarge: I'm not sure we're using this dependency. We could remove it.

src/Endpoints/README.md
pauarge: I see that this documentation will be handled in a separate ticket. If so, I'd wouldn't include this README file (it's essentially empty) and add it later when it has content.

pauarge (changes requested):
Left some changes related with the added `npm` included packages.

src/blocks/recommendations/recommendations-store.js
```diff
@@ -0,0 +1,55 @@
+/**
+ * External dependencies
+ */
+import { createContext, useContext, useReducer } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import { RECOMMENDATIONS_BLOCK_ERROR, RECOMMENDATIONS_BLOCK_LOADED, RECOMMENDATIONS_BLOCK_RECOMMENDATIONS } from './constants';
+// TODO: consider using createReduxStore https://github.com/WordPress/gutenberg/tree/a7e2895829c16ecd77a5ba22d84f1dee1cfb0977/packages/data#registering-a-store
```
pauarge: I don't thing using Redux is a good call. I'd remove this comment.

src/blocks/recommendations/components/parsely-recommendations-list-item.js
```diff
@@ -0,0 +1,45 @@
...
+ * External dependencies
+ */
+import { Card, CardBody, CardMedia } from '@wordpress/components';
+
+const getImageForLink = ( { imagestyle, imageUrl, thumbUrlMedium } ) => {
+	if ( imagestyle === 'original' ) {
+		return imageUrl;
+	}
+	return thumbUrlMedium;
+};
```
pauarge:
```suggestion
const getImageForLink = ( { imagestyle, imageUrl, thumbUrlMedium } ) => imagestyle === 'original' ? imageUrl : thumbUrlMedium;
```

src/blocks/recommendations/class-recommendations-block.php
```diff
@@ -0,0 +1,168 @@
...
+		// Temporary workaround - remove block when in FSE due to issues.
+		global $pagenow;
+		if ( 'site-editor.php' === $pagenow ) {
+			add_action(
+				'admin_init',
+				function() {
+					unregister_block_type( 'wp-parsely/recommendations' );
+				}
+			);
+		}
```
pauarge: I wonder if we could do this check inside `register_block_and_assets` and instead of regitering and then unregistering, we could not register at all in the first place.

src/blocks/recommendations/class-recommendations-block.php
```diff
@@ -0,0 +1,168 @@
...
+	 *
+	 * @return void
+	 */
+	public static function register_block_and_assets(): void {
+		$plugin_url = plugin_dir_url( PARSELY_FILE );
+
+		$editor_asset_file = require plugin_dir_path( PARSELY_FILE ) . 'build/recommendations-edit.asset.php';
+		wp_register_script(
+			'wp-parsely-recommendations-block-editor',
+			$plugin_url . 'build/recommendations-edit.js',
```
pauarge, line 64: What do you think about renaming the file from `recommendations-edit.js` to `recommendations-editor.js`. I think it better represents the intent of the file.

package.json
acicovic: I've removed it and tests pass and I've also done a couple of manual tests. It seems to be working.

`depcheck` also lists those are unused:
```
@wordpress/babel-preset-default
concurrently
```

However, I see `concurrently` in `package.json`'s scripts and mention of `babel-prset-default`in `babel.config.js`, so assume `depcheck` is incorrect on those?

pauarge: We definitely need those packages. They are not used in the code itself but in our tooling, so I guess that's why they are being flagged as not used.

acicovic: 👍 Marking this as resolved. If anything arises we'll revert.

src/blocks/recommendations/class-recommendations-block.php
acicovic: I think I had the same idea, but I don't remember why I didn't do it. Will recheck this.

src/blocks/recommendations/components/parsely-recommendations-inspector-controls.js
pauarge: +1 on @acicovic's point.

src/blocks/recommendations/components/parsely-recommendations-inspector-controls.js
pauarge: I agree that 99 is excessive. I'd even say 50 is too much. Let's limit it to 25 for now?

src/blocks/recommendations/components/parsely-recommendations-title.js
pauarge: I've pushed a simpler and easier to understand ternary.

src/blocks/recommendations/components/parsely-recommendations-inspector-controls.js
acicovic: OK, marking this as resolved but feel free to come back if needed.

src/blocks/recommendations/components/parsely-recommendations-title.js
acicovic: Marking this as resolved.

src/blocks/recommendations/components/parsely-recommendations-fetcher.js
acicovic: Resolved in https://github.com/Parsely/wp-parsely/pull/642/commits/cf9227c69c3e372cae8b1ed9bc3c2dd5cac21cdd by @pauarge.

src/blocks/recommendations/components/parsely-recommendations-inspector-controls.js
acicovic: I've updated the max to 25.

src/blocks/recommendations/components/parsely-recommendations-inspector-controls.js
```diff
@@ -31,7 +31,7 @@ const ParselyRecommendationsInspectorControls = ( {
 					<RangeControl
 						label={ __( 'Maximum Results', 'wp-parsely' ) }
 						min="1"
-						max="99"
+						max="25"
 						onChange={ ( newval ) => setAttributes( { limit: newval } ) }
```
pauarge, line 35:
```suggestion
						onChange={ ( limit ) => setAttributes( { limit } ) }
```

src/blocks/recommendations/components/parsely-recommendations-fetcher.js
```diff
@@ -0,0 +1,71 @@
...
+		if ( errorMessage ) {
+			dispatch( setError( { error: response.error } ) );
+			return;
+		}
+
+		const data = response?.data || [];
+		dispatch( setRecommendations( { recommendations: data } ) );
+	}
+
+	const reinvocationDelay = 300; // Reinvoke function at most every this amount of milliseconds.
```
pauarge: Could we move this out of the function, at the top of the file? We don't need to redeclare each time we run the function.

src/blocks/recommendations/components/parsely-recommendations-fetcher.js
```diff
@@ -0,0 +1,71 @@
...
+	/**
+	 * Fetch recommendations:
+	 * - On component mount
+	 * - When an attribute changes that affects the API call.
+	 *   (This happens in the Editor context when someone changes a setting.)
+	 */
+	useEffect( debouncedUpdate, apiMemoProps );
+
+	// This is a data-only component and does not render
+	return null;
```
pauarge, line 69: Is it necessary to return null? Could we omit these lines all together?

acicovic: I've tested this and it is needed. Else, the block errors out.

src/blocks/recommendations/class-recommendations-block.php
acicovic: I guess the reason I did it was for it to be grouped with the other code that would not register the Block (no filter, old WP version). But what you actually suggest is cleaner, and I'm pushing a new commit with this solution right now.

src/blocks/recommendations/class-recommendations-block.php
acicovic: For Blocks, creating an `edit.js` and `save.js` file (if needed - we don't have a save file in our case) is a convention, and it's also the way files are generated by the [create-block](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-create-block/) package.

I don't have a strong opinion on renaming the rebuilt file to recommendations-editor, although recommendations-edit is more representative of where the built file comes from. I don't think it matters that much because it's a built file anyway.

Let me know if you want me to go ahead with this.

pauarge: If it's the convention, then let's leave it as-is.

src/blocks/recommendations/class-recommendations-block.php
```diff
@@ -0,0 +1,164 @@
...
+	/**
+	 * Registers all block assets so that they can be enqueued through the
+	 * WordPress Block Editor in the corresponding context.
+	 *
+	 * @since 3.2.0
+	 *
+	 * @return void
+	 */
+	public static function register_block_and_assets(): void {
+
```
pauarge:
```suggestion
```

pauarge: Let's remove this additional line break

pauarge (approved):
Amazing work @acicovic @jblz 🚢

acicovic:
@jblz and @GaryJones, the plan is to release this early next week. If you want to (re)review this within this week, feel free to do so.

Thanks! 😃

PR #646, opened 2022-02-04
Correct "since" annotations to 3.1.0

jblz:
## Description

These annotations were released in version 3.1.0, but erroneously cite version 3.2.0 (which is not yet released).

To make this PR, I did:

* `git checkout 3.1.0`
* `git checkout -b fix/since-annotations-3.1.0`
* search replace for `@since 3.2.0` in my IDE
* added and committed the changes
* `git rebase origin/develop`
* `git push origin fix/since-annotations-3.1.0`

## Motivation and Context

Keeping our inline documentation as correct as possbile.

### DIFF
--- a/src/UI/class-admin-bar.php
+++ b/src/UI/class-admin-bar.php
@@ -3,7 +3,7 @@
  * Parsely tweaks to WordPress admin bar
  *
  * @package Parsely
- * @since 3.2.0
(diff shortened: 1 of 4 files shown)


pauarge (approved):
Nice catch

GaryJones (approved):
✅

PR #647, opened 2022-02-07
Improve reliability of E2E tests

pauarge:
## Description

The screen options test might be left on a broken state, hence always failing to pass the tests. This was caused by the fact that the screen options was always clicked, instead of checking the value that the checkbox contains. Hence, if the script clicks and the value was true, the result would be false (which seems counterintuitive).

This PR makes the helper function a bit smarter and only clicks the input in case it's needed.

## Motivation and Context

Improve reliability of the E2E tests.

## How Has This Been Tested?

Tests pass consistently locally and on GitHub.

### DIFF
--- a/tests/e2e/specs/front-end-code.spec.js
+++ b/tests/e2e/specs/front-end-code.spec.js
@@ -6,11 +6,19 @@ import {
 	createURL,
 	loginUser,
 } from '@wordpress/e2e-test-utils';
+import * as fs from 'fs';
 
 /**
  * Internal dependencies
  */
 import { changeKeysState, PLUGIN_VERSION } from '../utils';
+
+const getAssetVersion = () => {
+	const data = fs.readFileSync( 'build/init-api.asset.php', { encoding: 'utf8', flag: 'r' } );
+	const re = new RegExp( "\'version\' => \'(.*)\'" );
+	const r = data.match( re );
+	return r[ 1 ];
+};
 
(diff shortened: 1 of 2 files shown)


acicovic:
> Hence, if the script clicks and the value was true, the result would be false

Maybe write this somewhere as a comment in the code?

acicovic (dismissed):

acicovic (dismissed):

tests/e2e/specs/front-end-code.spec.js
```diff
@@ -41,5 +50,6 @@ describe( 'Front end code insertion', () => {
 		expect( content ).toContain( `<script data-parsely-site="e2etest.example.com" src="https://cdn.parsely.com/keys/e2etest.example.com/p.js?ver=${ PLUGIN_VERSION }" id="parsely-cfg"></script>` );
 		expect( content ).toContain( '<script id="wp-parsely-api-js-extra">' );
 		expect( content ).toContain( 'var wpParsely = {"apikey":"e2etest.example.com"};' );
+		expect( content ).toContain( `<script src="http://localhost:8889/wp-content/plugins/wp-parsely/build/init-api.js?ver=${ getAssetVersion() }" id="wp-parsely-api-js"></script>` );
```
jblz, line 54: Let's also assert that the asset version is not empty.

jblz (approved):
:ship:

acicovic (approved):

PR #649, opened 2022-02-07
Completely delegate handling of query args to the backend API

jblz:
## Description

Instead of maintaining logic for manipulating arguments that will be sent along to the Parse.ly API, accept a single parameter called `query` and forward that along.

Additionally, this moves away from the `Recommended` or `Recommendations` nomenclature and settles on `related` instead.

## Motivation and Context

This is a move to simplify the API endpoint powering the block. The ultimate source of truth of what is acceptable is determined by the Parse.ly API service. If that is ever to change, this change will prevent us from also having to update the logic in this plugin to match.

As to the nomenclature, this aims for consistency between our symbols and the path of the remote API call.

## How Has This Been Tested?

* Automated testing included in this PR
* Manual testing with #642 adapted to work against this new data shape format
* Manual testing in the browser, e.g. browse to: http://localhost:8888/wp-json/wp-parsely/v1/related?query[limit]=2&query[sort]=score

Note: for manual testing, you'll need to add the following in a separate plugin:
`add_filter( 'wp_parsely_enable_related_endpoint', '__return_true' );`


GaryJones:
Right now, the class handles both the API call, and the caching of its own results.

I'd love to see this refactored to use a _Decorator pattern_ to provide the caching behaviour, to really separate out responsibilities.

src/Endpoints/index.php
```diff
@@ -0,0 +1,12 @@
...
+ * Loader for REST API endpoints served by this site
+ *
+ * @package Parsely
+ */
+
+declare(strict_types=1);
+
+namespace Parsely;
+
+require_once __DIR__ . '/class-related-api-proxy.php';
```
pauarge, line 12: Interesting approach for adding a loader pattern. I like it, but I'd rather do this for all files in the project rather than just this one. 

Given that's only one file for now, could we add it as usual (in `wp-parsely.php`) and then add this structure in a further PR?

src/Endpoints/index.php
```diff
@@ -0,0 +1,12 @@
+<?php
+/**
+ * Loader for REST API endpoints served by this site
+ *
+ * @package Parsely
```
pauarge: Could we add a `@since` annotation here?

src/RemoteAPI/index.php
```diff
@@ -0,0 +1,20 @@
+<?php
+/**
+ * Loader for Parsely Remote API functionality.
```
pauarge: Could we add a `@since` annotation here?

src/Endpoints/class-related-api-proxy.php
```diff
@@ -0,0 +1,121 @@
...
+use WP_Error;
+use WP_Object_Cache;
+use WP_REST_Server;
+use WP_REST_Request;
+
+/**
+ * Configure a REST API endpoint for use e.g. by the Related Block.
+ */
+final class Related_API_Proxy {
+	const OBJECT_CACHE_TTL = 5 * MINUTE_IN_SECONDS;
```
GaryJones: Now that there's a decorator, this constant seems to be unused.

src/Endpoints/class-related-api-proxy.php
```diff
@@ -0,0 +1,121 @@
...
+					'args'                => array(
+						'query' => array(
+							'default'           => array(),
+							'sanitize_callback' => function ( $query ) {
+								// question: how should we sanitize these?
+								return (array) $query;
+							},
+						),
+					),
+				),
```
GaryJones: That's a lot of indentation. Suggest extract into variable, and use the variable in this call.

src/Endpoints/class-related-api-proxy.php
```diff
@@ -0,0 +1,121 @@
...
+			)
+		);
+	}
+
+	/**
+	 * Determine if the endpoint can be called.
+	 *
+	 * @return boolean
```
GaryJones:
```suggestion
	 * @return bool
```

src/Endpoints/class-related-api-proxy.php
```diff
@@ -0,0 +1,121 @@
...
+	public function permission_callback() {
+		// Unauthenticated.
+		return true;
+	}
+
+	/**
+	 * Cached "proxy" to the Parsely `/related` endpoint
+	 *
+	 * @param WP_Rest_Request $request The request object.
+	 * @return array
```
GaryJones: This method returns objects, not arrays.

src/RemoteAPI/related/class-related-caching-decorator.php
```diff
@@ -0,0 +1,62 @@
...
+	 */
+	private $cache_key;
+
+	/**
+	 * Constructor.
+	 *
+	 * @param Proxy           $proxy The Proxy object to cache.
+	 * @param WP_Object_Cache $cache The WordPress object cache instance.
+	 */
+	public function __construct( Proxy $proxy, WP_Object_Cache $cache ) {
```
GaryJones: The presence of the class and not an interface as the type declaration ties this directly to the `WP_Object_Cache` class, and any child classes of it. Another cache object couldn't be used, even though this class shouldn't care _where_ the data is cached.

The way around this is to create an adaptor interface and an adaptor which consumes an `WP_Object_Cache` object and exposes methods for the `->get()` and `->set()` methods which are used in this class.

(This is more about good OOP practices, and it can be added later if necessary, so not a blocker to this PR.)

Ref:
- https://devdocs.magento.com/guides/v2.4/extension-dev-guide/adapters.html
- https://designpatternsphp.readthedocs.io/en/latest/Structural/Adapter/README.html

src/RemoteAPI/related/class-related-caching-decorator.php
```diff
@@ -0,0 +1,62 @@
...
+		$this->decorated_proxy = $proxy;
+		$this->cache           = $cache;
+		$this->cache_key       = 'api_recos-' . wp_hash( wp_json_encode( $this->decorated_proxy->get_query() ) );
+	}
+
+	/**
+	 * Implements caching for the proxy interface.
+	 */
+	public function get_items() {
+		$items = $this->cache->get( $this->cache_key, 'wp-parsely' );
```
GaryJones: Can we make `wp-parsely` a class constant?

src/RemoteAPI/related/class-related-caching-decorator.php
```diff
@@ -0,0 +1,62 @@
...
+		if ( ! is_object( $items ) ) {
+			$items = $this->decorated_proxy->get_items();
+			$this->cache->set( $this->cache_key, $items, 'wp-parsely', 5 * MINUTE_IN_SECONDS );
```
GaryJones: Can this duration argument be made a class constant as well, to better surface this value?

src/RemoteAPI/related/class-related-proxy.php
```diff
@@ -0,0 +1,109 @@
...
+	 *
+	 * @var array
+	 */
+	private $query;
+
+	/**
+	 * Constructor.
+	 *
+	 * @param Parsely $parsely Parsely instance.
+	 * @param array   $query The query arguments to send to the remote API.
```
GaryJones: Can we at least define the shape of the array, even if we don't want to specify the keys? `array<string, mixed>` or something like that. (Likewise for the class property as well)

src/RemoteAPI/related/class-related-proxy.php
```diff
@@ -0,0 +1,109 @@
...
+		$this->parsely = $parsely;
+		$this->query   = $query;
+	}
+
+	/**
+	 * Accessor for private query var.
+	 *
+	 * @since 3.2.0
+	 */
+	public function get_query() {
```
GaryJones: Missing a `@return` doc, and return type declaration.

src/RemoteAPI/related/class-related-proxy.php
```diff
@@ -0,0 +1,109 @@
...
+
+		$query = apply_filters( 'wp_parsely_related_endpoint_args', $query );
+		return add_query_arg( $query, self::RELATED_API_ENDPOINT );
+	}
+
+	/**
+	 * Implements the fetcher for the proxy interface.
+	 *
+	 * @since 3.2.0
+	 */
```
GaryJones: Missing a `@return` tag.

tests/Integration/RemoteAPI/RelatedRemoteAPITest.php
```diff
@@ -0,0 +1,176 @@
...
+		self::$parsely = new Parsely();
+	}
+
+	/**
+	 * Data provider for test_related_api_url().
+	 *
+	 * @return array[]
+	 */
+	public function data_related_api_url(): array {
+		return array(
```
GaryJones: Consider using `yield` here instead of a giant array. See https://github.com/Parsely/wp-parsely/pull/383 which made the exact same switch.

src/RemoteAPI/related/class-related-caching-decorator.php
```diff
@@ -0,0 +1,62 @@
...
+	/**
+	 * Constructor.
+	 *
+	 * @param Proxy           $proxy The Proxy object to cache.
+	 * @param WP_Object_Cache $cache The WordPress object cache instance.
+	 */
+	public function __construct( Proxy $proxy, WP_Object_Cache $cache ) {
+		$this->decorated_proxy = $proxy;
+		$this->cache           = $cache;
+		$this->cache_key       = 'api_recos-' . wp_hash( wp_json_encode( $this->decorated_proxy->get_query() ) );
```
GaryJones: Does this cache key prefix need updating to reflect we're now using the nomenclature `related`?

src/RemoteAPI/related/class-related-proxy.php
GaryJones: I can't think of a better way to generate a unique cache key, so I think you're right that this needs to be exposed. As such, should it be added to the interface as well?

src/RemoteAPI/related/class-related-proxy.php
```diff
@@ -0,0 +1,109 @@
...
+		$full_api_url = $this->get_api_url();
+
+		$result = wp_safe_remote_get( $full_api_url, array() );
+
+		if ( is_wp_error( $result ) ) {
+			return $result;
+		}
+
+		$body    = wp_remote_retrieve_body( $result );
+		$decoded = json_decode( $body );
```
GaryJones: All of the decoding, and checks, are technically a new task from the actual getting a response, so it could be handled with a `JSON_Decoder` class (which implements a `Decoder` interface, to allow for future expansion of retrieving XML, CSV, plain text, etc.). This would then centralise "the ability to decode some JSON and return an error if it fails" into one place, allowing multiple proxy classes to make use of it.

GaryJones (changes requested):
I love the work to move to the caching decorator!

A few small and larger suggestions.

src/RemoteAPI/related/class-related-caching-decorator.php
```diff
@@ -42,10 +41,10 @@ class Related_Caching_Decorator implements Proxy {
 	/**
 	 * Constructor.
 	 *
-	 * @param Proxy           $proxy The Proxy object to cache.
-	 * @param WP_Object_Cache $cache The WordPress object cache instance.
+	 * @param Proxy         $proxy The Proxy object to cache.
+	 * @param Cache_Adapter $cache An object cache instance.
 	 */
-	public function __construct( Proxy $proxy, WP_Object_Cache $cache ) {
+	public function __construct( Proxy $proxy, Cache_Adapter $cache ) {
```
GaryJones:
```suggestion
	public function __construct( Proxy $proxy, Cache $cache ) {
```

Cache_Adapter is the class, but the interface is what should be type declared here.

GaryJones: The doc on this constructor, and `$cache` class property doc will also need updating.

src/RemoteAPI/related/class-related-caching-decorator.php
```diff
@@ -42,10 +41,10 @@ class Related_Caching_Decorator implements Proxy {
...
 		$this->decorated_proxy = $proxy;
```
GaryJones: Since it's _this_ class that is doing the decorating, then this property could just be named as `$proxy`, rather than `$decorated_proxy`.

GaryJones (dismissed):
Rather than `Related_Caching_Decorator` as the name of the class that provides the caching behaviour for a proxy, I think `Cached_Proxy` would be a better nomenclature. That way, we're not leaking our choice of design pattern into the naming of the class (a _Decorator_ makes sense to us as engineers, but it has no standing as part of our business modelling here) and with a more general approach, then it could be easily used with other proxies (after all, we just want to add caching behaviour, regardless of the proxy being used):

```php
$cache = new Cache_Adaptor( $GLOBALS['wp_object_cache'] );
$related_proxy = new Cached_Proxy( new Related_Proxy( $this->parsely, $params['query'] ), $cache );
$another_proxy = new Cached_Proxy( new Another_Proxy( $this->parsely, $params['query'] ), $cache );
```

The only thing then would be to find a way to generate the unique keys. We could make the proxy interface have a `get_key()` method, and add that to the Proxy concrete classes. But this would mean the Proxy classes needing to know that it might be cached (it should have no knowledge of any decorations that might be applied to it).

So alternatively, have the Cached_Proxy do a hash of the object given to it, maybe extracting the prefix from the class name, so that it can generate a unique key that way, regardless of  whatever proxy was passed to it.

```php
class Cached_Proxy() {
...
	public function generate_key() {
		$prefix = strtolower( $this->proxy::class ); // Maybe limit to the first or last 10 characters?
		$unique = hash( $this->proxy );

		return $prefix . $unique;
	}
}
```

GaryJones:
As of [95a8aa2](https://github.com/Parsely/wp-parsely/pull/649/commits/95a8aa220f84faf723a7c7a207b43ffc6a8edb82), this is working for me locally, with various changes I've suggested now implemented.

Default endpoint: `http://localhost:10024/wp-json/wp-parsely/v1/related`
Screenshot 2022-02-12 at 16 12 21

Endpoint with a query passed through: `http://localhost:10024/wp-json/wp-parsely/v1/related?query[limit]=2`
Screenshot 2022-02-12 at 16 18 58

I've not looked into the JSON_Decoder suggestion yet, but wanted to share the progress so far.

GaryJones:
With [b79d11c](https://github.com/Parsely/wp-parsely/pull/649/commits/b79d11c39163ede462c72a488b25dd698dfa75bb), we now have the `/related` API being proxied with this class:

```php
class Related_Proxy extends Base_Proxy {
	protected const ENDPOINT     = 'https://api.parsely.com/v2/related';
	protected const QUERY_FILTER = 'wp_parsely_related_endpoint_args';
}
```

Other Parse.ly API endpoints that we want to proxy in the future, and pass all of the query args to, only need a new child class with two protected constants defined!

src/Endpoints/class-related-api-proxy.php
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	/**
+	 * Determine if the endpoint can be called.
+	 *
+	 * @return bool
+	 */
+	public function permission_callback(): bool {
+		// Unauthenticated.
+		return true;
```
jblz, line 88: Should we impose a nonce check or any rate limiting?  Any changes can be done in a follow up, but I thought I'd raise it here at least.

jblz (commented):
Thanks so much for tidying this up @GaryJones. This really sets us on solid ground for adding additional endpoints.

src/Endpoints/class-related-api-proxy.php
GaryJones: Probably needs more thought, so would suggest opening an issue to follow-up.

pauarge (approved):

src/Endpoints/class-related-api-proxy.php
GaryJones: https://github.com/Parsely/wp-parsely/issues/667

src/RemoteAPI/related/class-related-proxy.php
GaryJones: Since we now have a single proxy point, then this line will only appear in one place - this could be switched out later on, but it's not a blocker to this PR.

src/RemoteAPI/related/class-related-proxy.php
GaryJones: I changed the approach, so this comment is no longer relevant.

PR #650, opened 2022-02-08
Support JS hooks for front-end actions

### LINKED ISSUE #640: Expose Parse.ly tracker onload event

The initial building block for advanced integrations is the `window.PARSELY` global object loaded whenever the `p.js` tracking script is inserted on the page. The majority of instructions on our docs page refer to hooking into the `onload` method that the Parse.ly library offers.

With a catch, users can absolutely do that hook right now on wp-parsely. The onload definition must be done before the script is loaded for the snippets to work. This is non-trivial, and a developer would have to reverse-engineer our plugin to insert their code in a specific place on their plugin code.

We must provide a hook within wp-parsely to let users run their JS code to address this issue. This way, we can ensure that it will run before the tracking script is inserted.

We must create a new WordPress action in our `Scripts` class that lets a user define their own `onload` function. If that action is defined, we should insert the developer's code *before* the Parse.ly tracking script is loaded. We should provide the boilerplate for onload and the user would only have to provide the actual function.

Schematically, this is what we should output to the HTML front-end.

```
<script>
    window.PARSELY = window.PARSELY || {
        onload: // code of the JS function extracted from the filter goes here.
    };
</script>

<!-- START Parse.ly Include -->
<!-- ...insert the parsely tracker code here... -->
<!-- END Parse.ly Include -->
```

Not implementing it and leaving it to the developer properly hooking their calls.

## Additional context

https://www.parsely.com/help/integration/category/tracking

pauarge:
## Description

This PR provides a novel interface to interact with the `onload` method of the Parse.ly JavaScript tracker. We make use of the [WordPress JavaScript hooks package](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-hooks/) to provide an interface similar to what we achieve in the rest of the plugin through PHP filters.

The idea is that we have a new file, `loader.js`, that constructs a function with all functions that have been hooked to the `wpParselyOnLoad` action. That constructed function then gets combined with `window.PARSELY`'s onload.

We used to have an `init-api.js` file which used a similar way of combining a custom method for an API call plus whatever `onload` had. We have renamed this file to `uuid-profile-call.js` and make use of the hooking mechanism to to add it in the same queue.

## Motivation and Context

Closes #640 

## How Has This Been Tested?

Make sure you're in this branch and have an API key set for your Parse.ly site. Then, add some JS code:

```
$script = '
window.wpParselyHooks.addAction("wpParselyOnLoad", "wpParsely", testFunc, 10);
function testFunc() {
	console.log("This is a hook");
}';
wp_add_inline_script( 'wp-parsely-loader', $script );
```

With this function, you should get a `console.log` executed when the Parse.ly tracker loads.


src/class-scripts.php
```diff
@@ -127,30 +117,15 @@ public function enqueue_js_tracker(): void {
...
-	public function enqueue_js_api(): void {
-		$parsely_options = $this->parsely->get_options();
-
-		// If we don't have an API secret, there's no need to proceed.
-		if ( empty( $parsely_options['api_secret'] ) ) {
-			return;
+		// If we don't have an API secret, there's no need to set the API key.
+		// Setting the API key triggers the UUID Profile Call function.
+		if ( isset( $parsely_options['api_secret'] ) && is_string( $parsely_options['api_secret'] ) && '' !== $parsely_options['api_secret'] ) {
+			$js_api_key = "window.wpParselyApiKey = '" . $this->parsely->get_api_key() . "';";
```
jblz:
```suggestion
			$js_api_key = "window.wpParselyApiKey = '" . esc_js( $this->parsely->get_api_key() ) . "';";
```

jblz (commented):
I'm digging the approach :heart_eyes:

src/class-scripts.php
pauarge: Oooops

src/class-scripts.php
```diff
@@ -69,22 +67,14 @@ public function register_scripts(): void {
...
+		$custom_loader_asset = require plugin_dir_path( PARSELY_FILE ) . 'build/loader.asset.php';
 		wp_register_script(
-			'wp-parsely-api',
-			plugin_dir_url( PARSELY_FILE ) . 'build/init-api.js',
-			$api_script_asset['dependencies'],
-			$api_script_asset['version'],
+			'wp-parsely-loader',
+			plugin_dir_url( PARSELY_FILE ) . 'build/loader.js',
+			$custom_loader_asset['dependencies'],
+			$custom_loader_asset['version'],
```
acicovic: As the `custom_loader` name was replaced with `loader`, consider changing this?

acicovic:
I wouldn't call those suggestions, they are more of a thinking out loud thing:
- Is the name 'loader' too generic, and can we find a more describing name for this?
- Would it be useful, and how easy would it be to extend this system to hold other events?

pauarge:
> Is the name 'loader' too generic, and can we find a more describing name for this? 

I thought about this. This is all namespaced within the plugin and it's a script, so it can be inferred that is something that loads scripts. Do you have a proposal for what you'd name it?

> Would it be useful, and how easy would it be to extend this system to hold other events?

Absolutely. That is the plan actually. The good thing is that this hook system works pretty similarly to what the WordPress PHP hooks work, so we can extend it as much as we want. I'd keep it out of the scope of this PR, though.

acicovic:
> > Is the name 'loader' too generic, and can we find a more describing name for this?
> 
> I thought about this. This is all namespaced within the plugin and it's a script, so it can be inferred that is something that loads scripts. Do you have a proposal for what you'd name it?

Currently unsure as I'm not very imaginative today 🙂

acicovic (dismissed):

acicovic (dismissed):

README.md
```diff
@@ -119,6 +119,19 @@ You can use the `wp_parsely_metadata` filter, which sends three arguments: the a
...
+For some advanced integrations, you need to hook custom JavaScript code to some events on the `window.PARSELY` object. To do so, our plugin supports WordPress JavaScript hooks. For example:
+
+```
+$script = '
+window.wpParselyHooks.addAction("wpParselyOnLoad", "wpParsely", testFunc, 10);
+function testFunc() {
+	console.log("This is a hook");
+}';
+wp_add_inline_script( 'wp-parsely-loader', $script );
+```
```
GaryJones:
```suggestion
~~~php
$script = '
window.wpParselyHooks.addAction("wpParselyOnLoad", "wpParsely", testFunc, 10);
function testFunc() {
	console.log("This is a hook");
}';
wp_add_inline_script( 'wp-parsely-loader', $script );
~~~
```

README markdown should use tilda instead of backtick fences for code blocks so that it renders correctly on WordPress.org.

acicovic (approved):

PR #655, opened 2022-02-10
Replace PHP_EOL with "\n" (LF)

### LINKED ISSUE #630: Some integration tests may fail on Windows due to line feed differences
## Describe the bug
Some integration tests may fail on Windows due to line feed differences.

## To Reproduce
1. Run integration tests on Windows
2. See tests output

## Expected behaviour
Integration tests should pass.

## Screenshots
image

acicovic:
## Description
`PHP_EOL` outputs `"\n"` (LF) feeds under Linux/Mac and `"\r\n"` (CRLF) feeds under Windows, which makes it the right choice in a wide variety of scenarios.

However, In our case, we were using a mix of `"\n"` (LF) and `PHP_EOL` calls for adding line feeds. In Windows, this resulted in having mixed line feeds, causing integration tests to fail. This PR fixes that by bringing LF line feeds everywhere.

## Motivation and Context
- Being able to make tests pass on windows (fixes #630), making this a better experience
- Always serving LF line feeds (which is not a problem for Windows anymore), as mixed line feeds are inconsistent and messy

## How Has This Been Tested?
 - Tests pass on a Windows machine


GaryJones (approved):
✅

pauarge (changes requested):
Speaking out of ignorance here, so we might end up ignoring my comment, but it feels like we're going from something that's platform agnostic to not. 

Maybe the tests fail on Windows because they run on Docker (Linux-based), hence PHP_EOL is \n?

acicovic:
>Speaking out of ignorance here, so we might end up ignoring my comment, but it feels like we're going from something that's platform agnostic to not.

I know where you're coming from, but this is not an issue. Windows recognizes `"\n"` perfectly and this is demonstrated by the tests that now pass. Nothing on Mac and Linux is affected, since those systems already work with `"\n"`. All modern software in Windows recognizes `"\n"` line feeds, so we're not jumping to something that would cause problems on Windows. In general, `"\n"` would be much more platform-agnostic than `"\r\n"` which is a Windows-only thing (and something that should have never existed IMO).

> Maybe the tests fail on Windows because they run on Docker (Linux-based), hence PHP_EOL is \n?

`PHP_EOL` resulted in `"\r\n"` even though it was running under Docker. This is what resulted in the differences, combined with a hardcoded `"\n"` in at least one place in our existing code. So we were trying to test for something that couldn't work realistically on a Windows environment.

Trying to put `PHP_EOL` everywhere would be much more tedious, because in many cases we would need to terminate every string line with it, not also to mention that it would give us no real benefit. All this to say, that I'm pretty confident that what we have here is the best solution with the less possible code changes and no breakages.

pauarge:
> Trying to put PHP_EOL everywhere would be much more tedious

Why? I think it could be relatively quick with search-and-replace within the IDE. Would this work on Windows?

acicovic:
It would be doable with tinkering (I don't think this will work with just search and replace). We would most probably need to append `PHP_EOL` to the expected strings in the tests. But I can't see what would be the advantage of this approach.

acicovic:
I'll look into posting an alternative PR using `PHP_EOL`.

acicovic:
Closed in favor of #670

acicovic:
Reopening this after discussion in #670.

PR #656, opened 2022-02-10
Moving recommended widget CSS to build system

pauarge:
## Description

The plugin has a `wp-parsely.css` global styles file. This file contained more things historically, but now it only contains styles for the recommended widget. Instead of having it as a global file, this PR moves the styles into the webpack build system and only enqueues the CSS file if the recommended widget is used.

## Motivation and Context

Get rid of generic CSS files that fall outside our build process.

## How Has This Been Tested?

The recommended widget looks and behaves as expected.


src/UI/class-recommended-widget.php
```diff
@@ -136,7 +133,15 @@ public function widget( $args, $instance ): void {
 			true
 		);
 
+		wp_register_style(
+			'wp-parsely-recommended-widget',
+			plugin_dir_url( PARSELY_FILE ) . 'build/recommended-widget.css',
+			array(),
```
acicovic, line 139: Maybe this makes sense?

```suggestion
			$recommended_widget_script_asset['dependencies'],
```

pauarge: That is very much intentional. If we put the dependencies in, the styles are added too late and don't get rendered correctly.

acicovic (approved):

PR #657, opened 2022-02-10
Fixing some type inconsistencies suggested by PHPStan

pauarge:
## Description

This PR adds type checks to ensure that we don't get errors derived from strict typing. Those checks were suggested by running PHPStan.

## Motivation and Context

Better type safety.

## How Has This Been Tested?

Test run normally and PHPStan doesn't complain about those errors anymore.


src/class-rest.php
```diff
@@ -90,13 +90,19 @@ public function register_meta(): void {
 	 * response and the `meta` object containing the actual metadata.
 	 */
 	public function get_callback( array $object ): array {
-		$post_id = $object['ID'] ?? $object['id'];
+		$post_id = $object['ID'] ?? $object['id'] ?? '';
```
GaryJones: `WP_Post::get_instance()` is expecting an integer, so passing an empty string doesn't seem right. The value gets cast to an integer anyway (so would be an integer `0`), so may as well just use `0` instead of an empty string here.

```suggestion
		$post_id = $object['ID'] ?? $object['id'] ?? 0;
```

src/class-rest.php
```diff
@@ -121,6 +127,12 @@ public function get_callback( array $object ): array {
...
 		ob_start();
 		$this->parsely->insert_page_header_metadata();
-		return ob_get_clean();
+		$meta = ob_get_clean();
+
+		if ( false === $meta ) {
+			return '';
+		}
+
+		return $meta;
```
GaryJones: This change seems a little redundant, as output buffering is turned on two lines above.

GaryJones (changes requested):

GaryJones (approved):
✅

GaryJones:
What was the PHPStan counts before and after this change?

PR #665, opened 2022-02-11
Not installing wp-env globally on GitHub Actions

pauarge:
## Description

This PR removes a redundant install of wp-env on the e2e tests pipeline.

## Motivation and Context

Fix failing E2E tests.


jblz (approved):
Desirable change even if it doesn't fix the e2e tests :)

PR #666, opened 2022-02-11
CI: Update e2e invocations

jblz:
## Description

* Vary the jest cache for the puppeteer run on a hash of the package-lock.json file
* Use our npm scripts to kick off the e2e runs

## Motivation and Context

We've been seeing intermittent e2e test failures. The theory behind this change is that concurrent runs were polluting the jest cache leading to an inconsistent state.  This seeks to avoid such and make the tests more reliable by using a separate cache for separate javascript package environments.


jblz:
so far no fix

PR #668, opened 2022-02-14
Allow settings to be saved when "Requires recrawl" section is hidden

acicovic:
## Description
@GaryJones reported a bug where saving the plugin's settings while the "Requires recrawl" section was hidden, would result in an error message and settings not being saved. The desired behavior would be saving to work correctly. Here's also a screenshot provided by @GaryJones:

image

## Motivation and Context
An error message should not be appearing and settings should be saved.

## How Has This Been Tested?
Integration tests pass.


src/UI/class-settings-page.php
```diff
@@ -1125,20 +1125,12 @@ private function validate_options_post_type_tracking( array &$input ): void {
...
-			add_settings_error(
-				Parsely::OPTIONS_KEY,
-				$track_as,
-				sprintf(
-					__(
-						'Settings could not be saved because post tracking data is invalid.',
-						'wp-parsely'
-					)
-				)
-			);
```
pauarge, line 1141: Is this error message no longer relevant?

src/UI/class-settings-page.php
```diff
@@ -1125,20 +1125,12 @@ private function validate_options_post_type_tracking( array &$input ): void {
...
+		}
+
+		if ( isset( $input[ $track_as ] ) ) {
```
pauarge, line 1132: Would it make sense to check if the requires recrawl section is indeed hidden?

src/UI/class-settings-page.php
acicovic: Correct. We don't display an error message anymore. If the given data is empty (which is the case when the recrawl settings section is invisible) or tampered/invalid in any way, the data in the database remains as it is, preserving data integrity.

src/UI/class-settings-page.php
acicovic: This needs to be stripped off from the `$input` array in all cases, regardless of the "Requires Recrawl" section's visibility status. This array item is a temporary store for values that we convert to something that all versions of the plugin understand, and we don't want to save it in the database.

pauarge (approved):

PR #670, opened 2022-02-15
Use PHP_EOL instead of "\n"

### LINKED ISSUE #630: Some integration tests may fail on Windows due to line feed differences

acicovic:
## Description
Some integration tests were failing on Windows environments due to line feed differences (`\n` vs `\r\n)`. This PR fixes those issues.

## Motivation and Context
- Being able to make tests pass on windows (fixes #630)

## How Has This Been Tested?
- Integration tests pass on a Windows machine


acicovic:
Personally I still prefer #655 because it is less verbose, requires less changes and it doesn't have the danger of introducing files with mixed line feeds.

pauarge (dismissed):
Great work

tests/Integration/Integrations/GoogleWebStoriesTest.php
```diff
@@ -71,16 +71,21 @@ public function test_web_stories_script_is_enqueued(): void {
...
+				</amp-analytics>'
+			)
+		);
 
 		self::expectOutputString( $expected );
 
 		$this::set_options( array( 'apikey' => 'blog.parsely.com' ) );
+		ob_start();
 		$this::$google->render_amp_analytics_tracker();
+		echo esc_html( $this::normalize_string( ob_get_clean() ) );
```
GaryJones: Why does this need to be echo'd? If it's to meet the expectOutputString() above, that that suggests that the `render_amp_analytics_tracker()` call can be removed, which means this isn't actually testing the thing that it's supposed to be testing.

tests/Integration/bootstrap.php
```diff
@@ -41,7 +41,7 @@
 	WPIntegration\bootstrap_it();
 
 	if ( ! defined( 'WP_PLUGIN_DIR' ) || file_exists( WP_PLUGIN_DIR . '/wp-parsely/wp-parsely.php' ) === false ) {
-		echo PHP_EOL, 'ERROR: Please check whether the WP_PLUGIN_DIR environment variable is set and set to the correct value. The unit test suite won\'t be able to run without it.', PHP_EOL;
+		echo PHP_EOL . "ERROR: Please check whether the WP_PLUGIN_DIR environment variable is set and set to the correct value. The unit test suite won't be able to run without it." . PHP_EOL;
```
GaryJones: It's more performant to use commas here (echo three strings), than it is to concat three strings into a single one, and the concat that.

tests/Integration/TestCase.php
```diff
@@ -192,4 +192,14 @@ public function go_to_new_post(): int {
...
+
+	/**
+	 * Normalize a string, removing line feeds and tab characters.
+	 *
+	 * @param string $string The string to be normalized.
+	 * @return string The normalized string.
+	 */
+	public static function normalize_string( string $string ): string {
+		return str_replace( array( "\r", "\n", "\t" ), '', $string );
+	}
```
GaryJones: While a clever approach, I'm not a fan. There may be times when we want to check for line-breaks and tabs - for instance, after being normalized `<meta name="parsely-type" content="post	" />` (that's a tab at the end of the content attribute value) would match `<meta name="parsely-type" content="post" />`, and yet the values are technically different, and may be interpreted by the Parse.ly parser as such. Blanket normalizing just in the tests isn't representative of what the actual output from a function or method will be.

GaryJones (changes requested):
Thanks @acicovic for testing this out.

All things considered, this seems like a lot of changes and hoop-jumping, compared to #655, which was much simpler and cleaner, by being consistent and letting Windows catch up with the rest of the world. The number of web servers and local setups using Windows would only be a small fraction of use cases anyway.

I'd strongly go for #655 instead of this PR.

tests/Integration/bootstrap.php
acicovic: Thank you for bringing this into my attention, it makes sense. I've updated this.

tests/Integration/Integrations/GoogleWebStoriesTest.php
acicovic: Not sure what is happening there, will look into this if we decide to go with this PR.

tests/Integration/TestCase.php
acicovic: You're right of course, and this isn't by any means a perfect solution. Having said that, it should work for most of our testing purposes. I moved the replacements as a parameter so we can have at least a choice of what to remove.

acicovic:
Concerning which PR to merge, I've expressed my personal preference of #655 here and in that PR. However, I'm probably not the one who has the final say on this, and I don't know if there's a specific process when there are different opinions.

tests/Integration/Integrations/GoogleWebStoriesTest.php
acicovic: Good catch. This was erroneous thinking from my side. I have removed the bad code and adjusted the expected string.

acicovic:
OK, having said all that, both PRs should be in a ready-to-merge state. I've reopened #655 as well to make our final choice.

pauarge:
We'll use `\n` as WP Core does.

PR #673, opened 2022-02-16
Improve plugin page on wordpress.org 

### LINKED ISSUE #609: Improve plugin page on wordpress.org

acicovic:
## Description
This PR solves some of the issues in our [plugin page](https://wordpress.org/plugins/wp-parsely/) that are described [here](https://github.com/Parsely/wp-parsely/issues/609#issuecomment-1019801977).

## Motivation and Context
Fixes #609

## How Has This Been Tested?
Making changes and previewing them.


GaryJones:
The same Basic Settings are now shown three times in the first three updated screenshots - this seems excessive and doesn't help to focus on how this plugin can help the user.

I wouldn't bother highlighting the presence of Help or Screen Options - they are nice-to-have features, but they don't need to be the focus of screenshots.

acicovic:
> The same Basic Settings are now shown three times in the first three updated screenshots - this seems excessive and doesn't help to focus on how this plugin can help the user.
> 
> I wouldn't bother highlighting the presence of Help or Screen Options - they are nice-to-have features, but they don't need to be the focus of screenshots.

Fair enough. What would you suggest?

GaryJones:
What was wrong with [fresh] screenshots of the existing sections?

acicovic:
So far we had a separate screenshot for each section (I guess not to have really long screenshots). This is not representative of how the plugin looks now, though whether this is a problem or not is personal opinion. I could go that route again (screenshot for every section) or just post a single screenshot with everything enabled in the options page (but this would be long).

GaryJones:
> screenshot for every section

This looked fine and made most sense to me.

pauarge (dismissed):
🚢 🚢 🚢

acicovic:
I'm working on updating the second screenshot to include the new tracking table UI element.

acicovic:
I think we should be good to go now.

GaryJones (approved):
✅

PR #676, opened 2022-02-17
Update some outdated references in function comments

acicovic:
## Description
While skimming code, I found function comments that were referring to a non-existent file and some functions that had been since renamed. This PR fixes this by pointing everything to their current names.

## Motivation and Context
Pointing to non-existent functions or files in comments can become confusing.


pauarge (approved):
LGTM

PR #677, opened 2022-02-17
Removing redundant EOL

pauarge:
## Description

This PR removes some line breaks that we have in our code that do not add any value and are redundant.

## Motivation and Context

We've been exploring how to have EOLs in our code, but the best way is to not have them :) 

## How Has This Been Tested?

Code still renders with sufficient line breaks on the front end

## Screenshots

** Old**
Screen Shot 2022-02-17 at 10 44 26 AM

** New **

Screen Shot 2022-02-17 at 10 44 01 AM


GaryJones (approved):
✅

acicovic (approved):

PR #678, opened 2022-02-17
Updating composer dependencies

pauarge:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
 	],
 	"require": {
 		"php": ">=7.1",
-		"composer/installers": "^1.12.0"
+		"composer/installers": "^1||^2"
 	},
 	"require-dev": {
 		"automattic/vipwpcs": "^2.3.3",
--- a/composer.json
+++ b/composer.json
@@ -75,7 +75,7 @@
 			"@php ./vendor/php-parallel-lint/php-parallel-lint/parallel-lint . -e php --exclude vendor --exclude .git --checkstyle"
 		],
 		"prepare-ci": [
-			"bash bin/install-wp-tests.sh wordpress_test root '' localhost trunk"
+			"bash bin/install-wp-tests.sh wordpress_test root root localhost trunk"
 		],
 		"test": [
 			"@php ./vendor/bin/phpunit --no-coverage --order-by=random"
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -13,20 +13,20 @@
 	],
 	"require": {
 		"php": ">=7.1",
-		"composer/installers": "^1"
+		"composer/installers": "^1.12.0"
```
GaryJones: The WordPress plugin installers are available in `1.0.0`, so why is `1.12.0` a minimum requirement?

I'd rather see us test and add support for `^2`:

```suggestion
		"composer/installers": "^1||^2"
```

composer.json
```diff
@@ -13,20 +13,20 @@
...
+		"phpcompatibility/phpcompatibility-wp": "^2.1.3",
+		"squizlabs/php_codesniffer": "^3.6.2",
 		"wp-coding-standards/wpcs": "^2.3.0",
 		"yoast/wp-test-utils": "^1",
-		"phpstan/phpstan": "^1.4",
+		"phpstan/phpstan": "^1.4.6",
 		"phpstan/extension-installer": "^1.1",
 		"phpstan/phpstan-strict-rules": "^1.1",
-		"szepeviktor/phpstan-wordpress": "^1.0"
+		"szepeviktor/phpstan-wordpress": "^1.0.2"
```
GaryJones, line 29: Are there particular bugs that you're trying to avoid with these new minimums?

The default `composer install`  / `composer update` would already install the latest versions with the existing constraints anyway.

composer.json
pauarge: Good point, let me try this out.

composer.json
pauarge: I prefer having tighter requirements, so everyone that's using/developing the plugin has a similar environment. I would even pitch the idea of getting rid of the carets on version numbers, but that would require us to keep an active eye on updates.

acicovic: Personally I update my deps very often. If there are features we want in the newer versions, we should probably go for them IMO.

GaryJones: Whether you have a tight constraint or not, using the latest versions requires `composer update` often - so there's not really any value in using tight constraints, when the current constraints give exactly the same (latest) versions anyway. 

The issue you're trying to solve for, is a people-problem, not a technology problem.

pauarge: Fair point

pauarge:
Closing exploration after feedback.

PR #624, opened 2022-01-27
WIP: Exploring various interactions & flows around network-wide installs

jblz:

PR #679, opened 2022-02-17
Improving network support

### LINKED ISSUE #575: Support Network Installation

pauarge:
## Description

This PR improves how the plugin supports WordPress networks by adding the following tweaks:

- Warning when editing the plugin's settings on the multisite's root site.
- Link to the sites list on multisites from the WordPress plugin list.
- Parse.ly documentation link on the WordPress plugin list.
- Tweak admin warning CSS classes.

## Motivation and Context

Borrowing from @jblz's work in #624

Closes #575.

## How Has This Been Tested?

Check out this code into a multisite network. See the new links and alerts in wp-admin.


src/UI/class-plugins-actions.php
```diff
@@ -39,13 +39,25 @@ public function run(): void {
 	 * @return array
 	 */
 	public function add_plugin_meta_links( array $actions ): array {
-		$settings_link = sprintf(
+		if ( is_multisite() && is_plugin_active_for_network( plugin_basename( PARSELY_FILE ) ) ) {
+			$actions['siteslist'] = sprintf(
+				'<a href="%s" target="_blank">%s</a>',
```
GaryJones: Do existing links have a target attribute? It has accessibility implications, and I'd rather we let folks choose to open it in a new tab if they wish, instead of forcing it.

src/UI/class-plugins-actions.php
```diff
@@ -39,13 +39,25 @@ public function run(): void {
...
+
+		$actions['settings'] = sprintf(
 			'<a href="%s">%s</a>',
 			esc_url( Parsely::get_settings_url() ),
 			esc_html__( 'Settings', 'wp-parsely' )
 		);
 
-		$actions['settings'] = $settings_link;
+		$actions['documentation'] = sprintf(
+			'<a href="%s" target="_blank">%s</a>',
```
GaryJones: See previous comment.

src/UI/class-plugins-actions.php
```diff
@@ -39,13 +39,25 @@ public function run(): void {
 	 * @return array
 	 */
 	public function add_plugin_meta_links( array $actions ): array {
-		$settings_link = sprintf(
+		if ( is_multisite() && is_plugin_active_for_network( plugin_basename( PARSELY_FILE ) ) ) {
+			$actions['siteslist'] = sprintf(
+				'<a href="%s" target="_blank">%s</a>',
+				network_admin_url( 'sites.php' ),
```
GaryJones: Let's escape this, as it has a filter at the end of `network_admin_url()`.

src/UI/class-plugins-actions.php
```diff
@@ -39,13 +39,25 @@ public function run(): void {
...
+		$actions['settings'] = sprintf(
 			'<a href="%s">%s</a>',
 			esc_url( Parsely::get_settings_url() ),
 			esc_html__( 'Settings', 'wp-parsely' )
 		);
 
-		$actions['settings'] = $settings_link;
+		$actions['documentation'] = sprintf(
+			'<a href="%s" target="_blank">%s</a>',
+			esc_url( 'https://www.parse.ly/help/integration/wordpress' ),
```
GaryJones: Don't technically need to escape a static string.

src/UI/class-plugins-actions.php
```diff
@@ -39,13 +39,25 @@ public function run(): void {
 	 * @return array
 	 */
 	public function add_plugin_meta_links( array $actions ): array {
-		$settings_link = sprintf(
+		if ( is_multisite() && is_plugin_active_for_network( plugin_basename( PARSELY_FILE ) ) ) {
```
GaryJones, line 42: Not a blocker, but would be nice (in another PR) to see the `PARSELY_FILE` global injected via a constructor arg, so we reduce the class's dependency on a global constant.

GaryJones (approved):
✅

PR #688, opened 2022-02-22
Don't return metadata on REST if no API key

pauarge:
## Description

This PR makes the REST API fields consistently behave when no API key is set. We used to return some data if no API key was set on the `meta` field, but with this change we only do that after the API key check.

## Motivation and Context

Consistency on the REST API behavior.

## How Has This Been Tested?

An additional integration test has been added.

### DIFF
--- a/src/class-rest.php
+++ b/src/class-rest.php
@@ -86,18 +86,18 @@ class Rest {
 	 * the `parsely` object in the REST API.
 	 *
 	 * @param array $object The WordPress object to extract to render the metadata for, usually a post or a page.
-	 * @return array The `parsely` object to be rendered in the REST API. Contains a version number describing the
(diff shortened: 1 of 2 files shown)


acicovic (approved):

GaryJones:
The API key is for talking to parse.ly - not for getting data out of WordPress.

Why should an API key be present for the REST API to work?

pauarge:
> Why should an API key be present for the REST API to work?

I would argue historical reasons. On a classic WordPress install, we need the API key to insert the tracking script. That is not true for the metadata, which could be generated anyway. I understand the reasoning behind not showing the metadata if no API key is set. Although it is possible to generate it, there's no point on inserting something that won't be tracked/crawled if you don't have an API key, anyway. 

The point of this PR is to give a consistent behavior. That is, if we don't return metadata on WP when there's no API key, the user expectation is that we don't return it either on REST.

PR #706, opened 2022-03-04
Initial experiments for GraphQL

pauarge:

### DIFF
--- a/src/Endpoints/class-graphql-metadata.php
+++ b/src/Endpoints/class-graphql-metadata.php
@@ -43,7 +43,7 @@ class GraphQL_Metadata extends Metadata_Endpoint {
 			 *
 			 * @param bool $enabled True if enabled, false if not.
 			 */
-			if ( apply_filters( 'wp_parsely_enable_graphql_support', true ) ) {
+			if ( apply_filters( 'wp_parsely_enable_graphql_support', true ) && $this->parsely->api_key_is_set() ) {
 				add_action( 'graphql_register_types', array( $this, 'register_meta' ) );
 			}
 		}
--- a/src/Endpoints/class-graphql-metadata.php
+++ b/src/Endpoints/class-graphql-metadata.php
@@ -40,19 +40,6 @@ class GraphQL_Metadata extends Metadata_Endpoint {
 		}
 	}
 
-	public function example_extend_wpgraphql_schema() {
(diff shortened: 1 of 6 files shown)


src/Endpoints/class-graphql-metadata.php
```diff
@@ -0,0 +1,262 @@
...
+
+	/**
+	 * Register fields in WPGraphQL plugin
+	 *
+	 * @since 3.2.0
+	 *
+	 * @return void
+	 */
+	public function run(): void {
+		if ( class_exists( 'WPGraphQL' ) ) {
```
GaryJones, line 38: I'd like to see this check moved outside of this class. Wrap it around the call to `run()` instead.

Since it's protecting a call to an action that only appears in WPGraphQL, then it may not even be needed anyway, since the action would never run if the WPGraphQL isn't active anyway.

pauarge: > I'd like to see this check moved outside of this class. Wrap it around the call to run() instead.

Would this mean putting moving the check in `wp-parsely.php`? I think it's good that we have as least logic as possible in that file, so I'm not sure where else to put it.

> Since it's protecting a call to an action that only appears in WPGraphQL, then it may not even be needed anyway, since the action would never run if the WPGraphQL isn't active anyway.

Does it have an impact on performance/memory usage to enqueue too many things? I see it as a waste of resources to enqueue a function for all `wp-parsely` users, most of which won't be using GraphQL. Doing just one conditional is more performant than an `add_action`.

pauarge:
Superseded by #710

src/Endpoints/class-graphql-metadata.php
GaryJones: > Would this mean putting moving the check in wp-parsely.php?

Yes, but that's a good thing.

>  I think it's good that we have as least logic as possible in that file, so I'm not sure where else to put it.

That's flawed thinking IMO. The main plugin file is the perfect place to be the [_composition root_](https://freecontent.manning.com/dependency-injection-in-net-2nd-edition-understanding-the-composition-root/) where classes and integrations are bootstrapped and started (or not).

> Does it have an impact on performance/memory usage to enqueue too many things? I see it as a waste of resources to enqueue a function for all wp-parsely users, most of which won't be using GraphQL. Doing just one conditional is more performant than an add_action.

That's fair 👍🏻

Issue #190, opened 2021-04-05
Support GraphQL output


GaryJones:
A decoupled app powered by WordPress may want to get Parse.ly constructed metadata for a post/page/archive/home/etc. via GraphQL (WPGraphQL plugin), so that it can be inserted into the decoupled application.

We should find a way to provide that. WPGraphQL has a [bunch of extensions](https://www.wpgraphql.com/extensions) which are used to make existing community plugins output to GraphQL.

pauarge:
Related with #487

PR #707, opened 2022-03-04
Moving Rest endpoint to Rest Metadata

pauarge:
## Description

This PR moves the `Rest` class into the `Endpoints` package. Given its generic name, the class has been renamed `Rest_Metadata` to better reflect what's doing, exposing metadata on the REST endpoint. We are also adding an abstract class for implementing metadata endpoints. We're doing this in preparation for the GraphQL work, which is going to expose the same metadata on GraphQL.

## Motivation and Context

See #190 

## How Has This Been Tested?

Go to any of the usual REST API endpoints and check that data is returned the same way. For instance, `http://localhost:8888/?rest_route=/wp/v2/posts/1`

### DIFF
--- a/src/Endpoints/class-metadata-endpoint.php
+++ b/src/Endpoints/class-metadata-endpoint.php
@@ -3,7 +3,7 @@
  * Metadata endpoint abstract class
  *
  * @package Parsely\Endpoints
- * @since 3.3.0
+ * @since 3.2.0
  */
 
 declare(strict_types=1);
@@ -15,7 +15,7 @@ use Parsely\Parsely;
 /**
  * Metadata endpoint classes are expected to implement the remaining functions of the class.
  *
- * @since 3.3.0
+ * @since 3.2.0
  */
 abstract class Metadata_Endpoint {
 	protected const VERSION    = '1.0.0';
@@ -40,7 +40,7 @@ abstract class Metadata_Endpoint {
 	/**
 	 * Function to start up the class and enqueue necessary actions.
 	 *
-	 * @since 3.3.0
+	 * @since 3.2.0
 	 *
 	 * @return void
 	 */
@@ -58,7 +58,7 @@ abstract class Metadata_Endpoint {
 	/**
 	 * Get the metadata in string format.
 	 *
-	 * @since 3.3.0
+	 * @since 3.2.0
 	 *
 	 * @return string String containing the metadata as HTML code that can be directly inserted in into the page.
 	 */
--- a/src/Endpoints/class-metadata-endpoint.php
+++ b/src/Endpoints/class-metadata-endpoint.php
@@ -1,0 +1,61 @@
+<?php
+/**
+ * Metadata endpoint abstract class
+ *
+ * @package Parsely\Endpoints
+ * @since 3.3.0
+ */
+
+declare(strict_types=1);
+
+namespace Parsely\Endpoints;
+
+use Parsely\Parsely;
+
+/**
+ * Metadata endpoint classes are expected to implement the remaining functions of the class.
+ *
+ * @since 3.3.0
+ */
+abstract class Metadata_Endpoint {
+	protected const VERSION    = '1.0.0';
+	protected const FIELD_NAME = 'parsely';
+
+	/**
+	 * Instance of Parsely class.
+	 *
+	 * @var Parsely
+	 */
+	protected $parsely;
+
+	/**
+	 * Constructor.
+	 *
+	 * @param Parsely $parsely Instance of Parsely class.
+	 */
+	public function __construct( Parsely $parsely ) {
+		$this->parsely = $parsely;
+	}
+
+	/**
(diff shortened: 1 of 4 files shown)


src/Endpoints/class-rest-metadata.php
```diff
@@ -8,35 +8,17 @@
...
-namespace Parsely;
+namespace Parsely\Endpoints;
 
 use WP_Post;
 
 /**
  * Injects Parse.ly Metadata to WordPress REST API
  *
  * @since 3.1.0
+ * @since 3.3.0 Renamed to from `Rest` to `Rest_Metadata`
```
GaryJones:
```suggestion
 * @since 3.3.0 Renamed FQCN from `Parsely\Rest` to `Parsely\Endpoints\Rest_Metadata`.
```

src/Endpoints/class-metadata-endpoint.php
```diff
@@ -0,0 +1,70 @@
...
+	 * @return void
+	 */
+	abstract public function register_meta(): void;
+
+	/**
+	 * Get the metadata in string format.
+	 *
+	 * @since 3.2.0
+	 *
+	 * @return string String containing the metadata as HTML code that can be directly inserted in into the page.
```
GaryJones:
```suggestion
	 * @return string Contains the metadata as HTML code that can be directly inserted into the page.
```

src/Endpoints/class-metadata-endpoint.php
```diff
@@ -0,0 +1,70 @@
...
+	 * @since 3.2.0
+	 *
+	 * @return void
+	 */
+	abstract public function run(): void;
+
+	/**
+	 * Registers the metadata fields on the appropriate resource types.
+	 *
+	 * @since 3.1.0
```
GaryJones:
```suggestion
	 * @since 3.2.0
```

src/Endpoints/class-metadata-endpoint.php
```diff
@@ -0,0 +1,70 @@
...
+	/**
+	 * Get the metadata in string format.
+	 *
+	 * @since 3.2.0
+	 *
+	 * @return string String containing the metadata as HTML code that can be directly inserted in into the page.
+	 */
+	public function get_rendered_meta(): string {
+		ob_start();
+		$this->parsely->insert_page_header_metadata();
```
GaryJones, line 66: Is this still suitable generic for other endpoints? What if there's data in a REST API endpoint that is something we don't make available in the page header data, like Settings values etc.

I think this `get_rendered_meta()` method should probably stay in the Rest_Metadata class for now.

src/Endpoints/class-metadata-endpoint.php
```diff
@@ -0,0 +1,70 @@
...
+
+use Parsely\Parsely;
+
+/**
+ * Metadata endpoint classes are expected to implement the remaining functions of the class.
+ *
+ * @since 3.2.0
+ */
+abstract class Metadata_Endpoint {
+	protected const VERSION    = '1.0.0';
```
GaryJones: By putting that here, then consumers will need to check their logic for the endpoints every time we make a breaking change to _any_ endpoint output, even if the endpoint they are calling with the parsely data has not changed.

I would suggest leaving this in the Rest_Metadata for now, since that's really the output we're making claims against - it doesn't have any bearing on changes for GraphQL or other endpoints.

src/Endpoints/class-rest-metadata.php
```diff
@@ -101,7 +83,7 @@ public function get_callback( array $object ): array {
 		}
 
 		$response = array(
-			'version' => self::REST_VERSION,
+			'version' => self::VERSION,
```
GaryJones: `VERSION` is too generic and needs context. `REST_VERSION` is more descriptive.

src/Endpoints/class-rest-metadata.php
acicovic: Should this comment be 3.3.0 or 3.2.0?

src/Endpoints/class-rest-metadata.php
pauarge: I agree with what you're saying. However, I intend to use this for versioning GraphQL too and GraphQL is not REST. The alternative that I can see is using a separate version for each, but I think it would introduce too many variables and end up being confusing.

src/Endpoints/class-metadata-endpoint.php
pauarge: Thinking about the future `GraphQL_Metadata` class, this is for sure a method we'll reuse. I have some experiments going on in https://github.com/Parsely/wp-parsely/pull/706 for some additional context.

src/Endpoints/class-metadata-endpoint.php
pauarge: Fair point.

acicovic (approved):

PR #709, opened 2022-03-08
Checking for API key existence in REST API earlier

pauarge:
## Description

If the API key is not set, we don't want to return Parse.ly data on the REST API. Instead of having to check for this condition each time we render a field (which can happen multiple times in a single request), this PR moves the check at initialization time of the class.

Some tests are added to check for this condition.

## Motivation and Context

More efficient checks.

## How Has This Been Tested?

Remove the API key from your test site and try to fetch metadata from the REST API. It shouldn't appear. Then, add the key back and it should appear.

### DIFF
--- a/tests/Integration/Endpoints/RestMetadataTest.php
+++ b/tests/Integration/Endpoints/RestMetadataTest.php
@@ -43,14 +43,49 @@ final class RestMetadataTest extends TestCase {
 	}
 
 	/**
-	 * Test whether the logic has been enqueued in the `rest_api_init` hook with a filter that disables it.
+	 * Test whether the logic has been enqueued in the when the `run` method is called.
+	 *
+	 * @covers \Parsely\Endpoints\Rest_Metadata::run
+	 * @uses \Parsely\Endpoints\Rest_Metadata::register_meta
+	 */
+	public function test_register_enqueued_rest_init(): void {
+		global $wp_rest_additional_fields;
+
+		self::set_options( array( 'apikey' => 'testkey' ) );
+		self::$rest->run();
+
+		$this->assertParselyRestFieldIsConstructedCorrectly( 'page', $wp_rest_additional_fields );
+
+		// Cleaning up the registered fields.
+		// phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedVariableFound
+		$wp_rest_additional_fields = array();
+	}
+
+	/**
+	 * Test whether the logic has been enqueued in the when the `run` method is called with a filter that disables it.
 	 *
 	 * @covers \Parsely\Endpoints\Rest_Metadata::run
 	 */
 	public function test_register_enqueued_rest_init_filter(): void {
+		global $wp_rest_additional_fields;
+
 		add_filter( 'wp_parsely_enable_rest_api_support', '__return_false' );
 		self::$rest->run();
-		self::assertFalse( has_action( 'rest_api_init', array( self::$rest, 'register_meta' ) ) );
+
+		self::assertEmpty( $wp_rest_additional_fields );
+	}
+
+	/**
+	 * Test whether the logic has been enqueued in the when the `run` method is called with no API key.
+	 *
+	 * @covers \Parsely\Endpoints\Rest_Metadata::run
+	 */
+	public function test_register_enqueued_rest_init_no_api_key(): void {
+		global $wp_rest_additional_fields;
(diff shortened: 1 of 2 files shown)


tests/Integration/Endpoints/RestMetadataTest.php
```diff
@@ -43,14 +43,49 @@ public function set_up(): void {
 	}
 
 	/**
-	 * Test whether the logic has been enqueued in the `rest_api_init` hook with a filter that disables it.
+	 * Test whether the logic has been enqueued in the when the `run` method is called.
```
acicovic:
```suggestion
	 * Test whether the logic has been enqueued when the `run` method is called.
```

tests/Integration/Endpoints/RestMetadataTest.php
```diff
@@ -43,14 +43,49 @@ public function set_up(): void {
...
+
+		$this->assertParselyRestFieldIsConstructedCorrectly( 'page', $wp_rest_additional_fields );
+
+		// Cleaning up the registered fields.
+		// phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedVariableFound
+		$wp_rest_additional_fields = array();
+	}
+
+	/**
+	 * Test whether the logic has been enqueued in the when the `run` method is called with a filter that disables it.
```
acicovic:
```suggestion
	 * Test whether the logic has been enqueued when the `run` method is called with a filter that disables it.
```

tests/Integration/Endpoints/RestMetadataTest.php
```diff
@@ -43,14 +43,49 @@ public function set_up(): void {
...
+
 		add_filter( 'wp_parsely_enable_rest_api_support', '__return_false' );
 		self::$rest->run();
-		self::assertFalse( has_action( 'rest_api_init', array( self::$rest, 'register_meta' ) ) );
+
+		self::assertEmpty( $wp_rest_additional_fields );
+	}
+
+	/**
+	 * Test whether the logic has been enqueued in the when the `run` method is called with no API key.
```
acicovic:
```suggestion
	 * Test whether the logic has been enqueued when the `run` method is called with no API key.
```

tests/Integration/Endpoints/RestMetadataTest.php
```diff
@@ -43,14 +43,49 @@ public function set_up(): void {
...
+	}
+
+	/**
+	 * Test whether the logic has been enqueued when the `run` method is called with a filter that disables it.
 	 *
 	 * @covers \Parsely\Endpoints\Rest_Metadata::run
 	 */
 	public function test_register_enqueued_rest_init_filter(): void {
+		global $wp_rest_additional_fields;
+
```
acicovic, line 71: Maybe we should be setting an API key here just to prove the point that this works?

tests/Integration/Endpoints/RestMetadataTest.php
```diff
@@ -43,14 +43,49 @@ public function set_up(): void {
...
 		add_filter( 'wp_parsely_enable_rest_api_support', '__return_false' );
 		self::$rest->run();
-		self::assertFalse( has_action( 'rest_api_init', array( self::$rest, 'register_meta' ) ) );
+
+		self::assertEmpty( $wp_rest_additional_fields );
+	}
+
+	/**
+	 * Test whether the logic has been enqueued when the `run` method is called with no API key.
```
acicovic:
```suggestion
	 * Verify that the logic has not been enqueued when the `run` method is called with no API key.
```

tests/Integration/Endpoints/RestMetadataTest.php
```diff
@@ -43,14 +43,49 @@ public function set_up(): void {
...
+
+		$this->assertParselyRestFieldIsConstructedCorrectly( 'page', $wp_rest_additional_fields );
+
+		// Cleaning up the registered fields.
+		// phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedVariableFound
+		$wp_rest_additional_fields = array();
+	}
+
+	/**
+	 * Test whether the logic has been enqueued when the `run` method is called with a filter that disables it.
```
acicovic:
```suggestion
	 * Verify that the logic has not been enqueued when the `run` method is called with a filter that disables it.
```

acicovic (approved):

PR #710, opened 2022-03-09
Adding GraphQL support

### LINKED ISSUE #190: Support GraphQL output

pauarge:
## Description

This PR adds GraphQL support to output metadata. We are doing so by extending the WPGraphQL plugin (thus, it is a requirement that it is installed). If it's not installed, wp-parsely will continue to work normally but without GraphQL support.

Since GraphQL is strongly typed, we have had to define the types of our metadata. Therefore, in this change we're registering a number of custom types that support our metadata (see screenshots). Then, we are registering the `parsely` field in all fields that should be tracked (added by the user).

In this implementation, we've also added a few filters so user can build on top of our default support.

## Motivation and Context

Closes #190 

## How Has This Been Tested?

1. Install WPGraphQL plugin.
2. Make sure wp-parsely is enabled and with an API key.
3. Perform queries on GraphiQL in wp-admin (see screenshots for examples).

Screen Shot 2022-03-09 at 12 15 20 PM

Screen Shot 2022-03-09 at 12 18 59 PM

Screen Shot 2022-03-09 at 12 18 19 PM

### DIFF
--- a/src/Endpoints/class-graphql-metadata.php
+++ b/src/Endpoints/class-graphql-metadata.php
@@ -222,7 +222,6 @@ class GraphQL_Metadata extends Metadata_Endpoint {
 	 * @since 3.2.0
 	 *
 	 * @param array<string, mixed> $meta Page metadata array.
-	 *
 	 * @return array<string, mixed>
 	 */
 	private function process_meta_for_graphql( array $meta ): array {
--- a/src/Endpoints/class-graphql-metadata.php
+++ b/src/Endpoints/class-graphql-metadata.php
@@ -180,7 +180,7 @@ class GraphQL_Metadata extends Metadata_Endpoint {
 				self::FIELD_NAME,
 				array(
 					'type'        => self::GRAPHQL_CONTAINER_TYPE,
-					'description' => 'Parse.ly metadata fields, to be rendered in the front-end so they can be parsed by the crawler. See https://www.parse.ly/help/integration/crawler.',
+					'description' => __( 'Parse.ly metadata fields, to be rendered in the front-end so they can be parsed by the crawler. See https://www.parse.ly/help/integration/crawler.', 'wp-parsely' ),
 					'resolve'     => function ( $graphql_post ) {
 						$post_id = $graphql_post->__get( 'ID' );
 						$post    = WP_Post::get_instance( $post_id );
--- a/src/Endpoints/class-graphql-metadata.php
+++ b/src/Endpoints/class-graphql-metadata.php
@@ -87,6 +87,13 @@ class GraphQL_Metadata extends Metadata_Endpoint {
 						'wp-parsely'
 					),
 				),
+				'isTracked' => array(
+					'type'        => 'Boolean',
+					'description' => __(
+						'Boolean indicating whether the current object\'s page type should be tracked according to user\'s settings.',
+						'wp-parsely'
+					),
+				),
 			),
 		);
 		register_graphql_object_type( self::GRAPHQL_CONTAINER_TYPE, $container_type );
@@ -100,18 +107,6 @@ class GraphQL_Metadata extends Metadata_Endpoint {
 	 * @return void
 	 */
 	private function register_fields(): void {
-		$options      = $this->parsely->get_options();
-		$object_types = array_unique( array_merge( $options['track_post_types'], $options['track_page_types'] ) );
-
-		/**
-		 * Filters the list of post object types that the Parse.ly GraphQL API is hooked into.
-		 *
-		 * @since 3.2.0
-		 *
-		 * @param string[] $object_types Array of strings containing the object types, i.e. `page`, `post`, `term`.
-		 */
-		$object_types = apply_filters( 'wp_parsely_graphql_object_types', $object_types );
-
 		$resolve = function ( \WPGraphQL\Model\Post $graphql_post ) {
 			$post_id = $graphql_post->ID;
 			$post    = WP_Post::get_instance( $post_id );
@@ -120,28 +115,27 @@ class GraphQL_Metadata extends Metadata_Endpoint {
 				return array();
 			}
 
-			$meta = $this->parsely->construct_parsely_metadata( $this->parsely->get_options(), $post );
-			$meta = $this->process_meta_for_graphql( $meta );
+			$options             = $this->parsely->get_options();
+			$object_types        = array_unique( array_merge( $options['track_post_types'], $options['track_page_types'] ) );
+			$current_object_type = get_post_type( $post );
 
 			return array(
 				'version'   => self::GRAPHQL_VERSION,
 				'scriptUrl' => $this->parsely->get_tracker_url(),
 				'metaTags'  => self::get_rendered_meta( 'meta_tags' ),
 				'jsonLd'    => self::get_rendered_meta( 'json_ld' ),
+				'isTracked' => is_string( $current_object_type ) && array_key_exists( $current_object_type, $object_types ),
 			);
 		};
 
-		foreach ( $object_types as $object_type ) {
-			$post_type_object = get_post_type_object( $object_type );
-			$config           = array(
-				'type'        => self::GRAPHQL_CONTAINER_TYPE,
-				'description' => __(
-					'Parse.ly metadata fields, to be rendered in the front-end so they can be parsed by the crawler. See https://www.parse.ly/help/integration/crawler.',
-					'wp-parsely'
-				),
-				'resolve'     => $resolve,
-			);
-			register_graphql_field( $post_type_object->graphql_single_name, self::FIELD_NAME, $config );
-		}
+		$config = array(
+			'type'        => self::GRAPHQL_CONTAINER_TYPE,
+			'description' => __(
+				'Parse.ly metadata fields, to be rendered in the front-end so they can be parsed by the crawler. See https://www.parse.ly/help/integration/crawler.',
+				'wp-parsely'
+			),
+			'resolve'     => $resolve,
+		);
+		register_graphql_field( 'ContentNode', self::FIELD_NAME, $config );
 	}
 }
--- a/src/Endpoints/class-graphql-metadata.php
+++ b/src/Endpoints/class-graphql-metadata.php
@@ -170,7 +170,7 @@ class GraphQL_Metadata extends Metadata_Endpoint {
 		$object_types = apply_filters( 'wp_parsely_graphql_object_types', $object_types );
 
 		$resolve = function ( \WPGraphQL\Model\Post $graphql_post ) {
-			$post_id = $graphql_post->__get( 'ID' );
+			$post_id = $graphql_post->ID;
 			$post    = WP_Post::get_instance( $post_id );
 
 			if ( false === $post ) {
--- a/src/Endpoints/class-graphql-metadata.php
+++ b/src/Endpoints/class-graphql-metadata.php
@@ -19,13 +19,8 @@ use WP_Post;
  * @since 3.2.0
  */
 class GraphQL_Metadata extends Metadata_Endpoint {
-	private const GRAPHQL_VERSION          = '1.0.0';
-	private const GRAPHQL_CONTAINER_TYPE   = 'ParselyMetaContainer';
-	private const GRAPHQL_AUTHOR_TYPE      = 'ParselyAuthor';
-	private const GRAPHQL_MAIN_ENTITY_TYPE = 'ParselyMainEntityOfPage';
-	private const GRAPHQL_IMAGE_TYPE       = 'ParselyImage';
-	private const GRAPHQL_PUBLISHER_TYPE   = 'ParselyPublisher';
-	private const GRAPHQL_META_TYPE        = 'ParselyMeta';
+	private const GRAPHQL_VERSION        = '1.0.0';
+	private const GRAPHQL_CONTAINER_TYPE = 'ParselyMetaContainer';
 
 	/**
 	 * Register fields in WPGraphQL plugin
@@ -67,65 +62,6 @@ class GraphQL_Metadata extends Metadata_Endpoint {
 	 * @return void
 	 */
 	private function register_object_types(): void {
-		$author_type = array(
-			'description' => __( 'Parse.ly container type for author meta object.', 'wp-parsely' ),
-			'fields'      => array(
-				'name' => array( 'type' => 'String' ),
-				'type' => array( 'type' => 'String' ),
-			),
-		);
-		register_graphql_object_type( self::GRAPHQL_AUTHOR_TYPE, $author_type );
-
-		$main_entity_type = array(
-			'description' => __( 'Parse.ly container type for main entity meta object.', 'wp-parsely' ),
-			'fields'      => array(
-				'id'   => array( 'type' => 'String' ),
-				'type' => array( 'type' => 'String' ),
-			),
-		);
-		register_graphql_object_type( self::GRAPHQL_MAIN_ENTITY_TYPE, $main_entity_type );
-
-		$image_type = array(
-			'description' => __( 'Parse.ly container type for image meta object.', 'wp-parsely' ),
-			'fields'      => array(
-				'type' => array( 'type' => 'String' ),
-				'url'  => array( 'type' => 'String' ),
-			),
-		);
-		register_graphql_object_type( self::GRAPHQL_IMAGE_TYPE, $image_type );
-
-		$publisher_type = array(
-			'description' => __( 'Parse.ly container type for publisher meta object.', 'wp-parsely' ),
-			'fields'      => array(
-				'logo' => array( 'type' => 'String' ),
-				'name' => array( 'type' => 'String' ),
-				'type' => array( 'type' => 'String' ),
-			),
-		);
-		register_graphql_object_type( self::GRAPHQL_PUBLISHER_TYPE, $publisher_type );
-
-		$meta_type = array(
-			'description' => __(
-				'Metadata fields to be rendered in the front-end. They follow Parse.ly\'s metadata structure. See https://www.parse.ly/help/integration/category/metadata',
-				'wp-parsely'
-			),
-			'fields'      => array(
-				'articleSection'   => array( 'type' => 'String' ),
-				'author'           => array( 'type' => array( 'list_of' => self::GRAPHQL_AUTHOR_TYPE ) ),
-				'context'          => array( 'type' => 'String' ),
-				'creator'          => array( 'type' => 'String' ),
-				'headline'         => array( 'type' => 'String' ),
-				'image'            => array( 'type' => self::GRAPHQL_IMAGE_TYPE ),
-				'keywords'         => array( 'type' => array( 'list_of' => 'String' ) ),
-				'mainEntityOfPage' => array( 'type' => self::GRAPHQL_MAIN_ENTITY_TYPE ),
-				'publisher'        => array( 'type' => self::GRAPHQL_PUBLISHER_TYPE ),
-				'thumbnailUrl'     => array( 'type' => 'String' ),
-				'type'             => array( 'type' => 'String' ),
-				'url'              => array( 'type' => 'String' ),
-			),
-		);
-		register_graphql_object_type( self::GRAPHQL_META_TYPE, $meta_type );
-
 		$container_type = array(
 			'description' => __( 'Parse.ly Metadata root type.', 'wp-parsely' ),
 			'fields'      => array(
@@ -133,14 +69,17 @@ class GraphQL_Metadata extends Metadata_Endpoint {
 					'type'        => 'String',
 					'description' => __( 'Revision of the metadata format.', 'wp-parsely' ),
 				),
-				'meta'     => array(
-					'type'        => 'ParselyMeta',
-					'description' => __( 'Structured and filterable metadata.', 'wp-parsely' ),
-				),
-				'rendered' => array(
+				'metaTags' => array(
 					'type'        => 'String',
 					'description' => __(
-						'HTML string containing the metadata. Intended to be rendered in the front-end as is.',
+						'HTML string containing the metadata in JSON-LD. Intended to be rendered in the front-end as is.',
+						'wp-parsely'
+					),
+				),
+				'jsonLd'   => array(
+					'type'        => 'String',
+					'description' => __(
+						'HTML string containing the metadata in JSON-LD. Intended to be rendered in the front-end as is.',
 						'wp-parsely'
 					),
 				),
@@ -182,8 +121,8 @@ class GraphQL_Metadata extends Metadata_Endpoint {
 
 			return array(
 				'version'  => self::GRAPHQL_VERSION,
-				'meta'     => $meta,
-				'rendered' => self::get_rendered_meta(),
+				'metaTags' => self::get_rendered_meta( 'meta_tags' ),
+				'jsonLd'   => self::get_rendered_meta( 'json_ld' ),
 			);
 		};
 
@@ -199,52 +138,5 @@ class GraphQL_Metadata extends Metadata_Endpoint {
 			);
 			register_graphql_field( $post_type_object->graphql_single_name, self::FIELD_NAME, $config );
 		}
-	}
-
-	/**
-	 * Adapts an array of metadata to GraphQL. Namely, all keys with a leading `@` symbol are duplicated by the same
-	 * key without the symbol.
-	 *
-	 * @since 3.2.0
-	 *
-	 * @param array<string, mixed> $meta Page metadata array.
-	 * @return array<string, mixed>
-	 */
-	private function process_meta_for_graphql( array $meta ): array {
-		$meta = $this->add_type_key( $meta );
-
-		$meta['context'] = $meta['@context'];
-
-		if ( array_key_exists( 'author', $meta ) ) {
-			$meta['author'] = array_map( array( $this, 'add_type_key' ), $meta['author'] );
-		}
-
-		if ( array_key_exists( 'mainEntityofPage', $meta ) ) {
-			$meta['mainEntityOfPage'] = array(
-				'type' => $meta['mainEntityOfPage']['@type'],
-				'id'   => $meta['mainEntityOfPage']['@id'],
-			);
-		}
-
-		if ( array_key_exists( 'publisher', $meta ) ) {
-			$meta['publisher'] = $this->add_type_key( $meta['publisher'] );
-		}
-
-		return $meta;
-	}
-
-	/**
-	 * Adds the `type` key in an array, sourcing its value from `@type` if that key exists.
-	 *
-	 * @since 3.2.0
-	 *
-	 * @param array<string, mixed> $data Some array that may or may not contain the `@type` key.
-	 * @return array<string, mixed>
-	 */
-	private function add_type_key( array $data ): array {
-		if ( array_key_exists( '@type', $data ) ) {
-			$data['type'] = $data['@type'];
-		}
-		return $data;
 	}
 }
--- a/src/Endpoints/class-graphql-metadata.php
+++ b/src/Endpoints/class-graphql-metadata.php
@@ -180,15 +180,6 @@ class GraphQL_Metadata extends Metadata_Endpoint {
 			$meta = $this->parsely->construct_parsely_metadata( $this->parsely->get_options(), $post );
 			$meta = $this->process_meta_for_graphql( $meta );
 
-			/**
-			 * Filters the array with the actual metadata that is exposed through GraphQL.
-			 *
-			 * @see https://wpgraphqldocs.gatsbyjs.io/functions/register_graphql_field/
-			 * @since 3.2.0
-			 *
-			 * @param array $meta Array with the fields of the new type.
-			 */
-			$meta = apply_filters( 'wp_parsely_graphql_meta_object', $meta );
 
 			return array(
 				'version'  => self::GRAPHQL_VERSION,
--- a/src/Endpoints/class-graphql-metadata.php
+++ b/src/Endpoints/class-graphql-metadata.php
@@ -124,7 +124,7 @@ class GraphQL_Metadata extends Metadata_Endpoint {
 				'scriptUrl'     => $this->parsely->get_tracker_url(),
 				'repeatedMetas' => self::get_rendered_meta( 'repeated_metas' ),
 				'jsonLd'        => self::get_rendered_meta( 'json_ld' ),
-				'isTracked'     => is_string( $current_object_type ) && array_key_exists( $current_object_type, $object_types ),
+				'isTracked'     => in_array( $current_object_type, $object_types, true ),
 			);
 		};
 
--- a/src/Endpoints/class-graphql-metadata.php
+++ b/src/Endpoints/class-graphql-metadata.php
@@ -172,6 +172,34 @@ class GraphQL_Metadata extends Metadata_Endpoint {
 		 */
 		$object_types = apply_filters( 'wp_parsely_graphql_object_types', $object_types );
 
+		$resolve = function ( \WPGraphQL\Model\Post $graphql_post ) {
+			$post_id = $graphql_post->__get( 'ID' );
+			$post    = WP_Post::get_instance( $post_id );
+
+			if ( false === $post ) {
+				return array();
+			}
+
+			$meta = $this->parsely->construct_parsely_metadata( $this->parsely->get_options(), $post );
+			$meta = $this->process_meta_for_graphql( $meta );
+
+			/**
+			 * Filters the array with the actual metadata that is exposed through GraphQL.
+			 *
+			 * @see https://wpgraphqldocs.gatsbyjs.io/functions/register_graphql_field/
+			 * @since 3.2.0
+			 *
+			 * @param array $meta Array with the fields of the new type.
+			 */
+			$meta = apply_filters( 'wp_parsely_graphql_meta_object', $meta );
+
+			return array(
+				'version'  => self::GRAPHQL_VERSION,
+				'meta'     => $meta,
+				'rendered' => self::get_rendered_meta(),
+			);
+		};
+
 		foreach ( $object_types as $object_type ) {
 			$post_type_object = get_post_type_object( $object_type );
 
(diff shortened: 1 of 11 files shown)


src/Endpoints/class-graphql-metadata.php
```diff
@@ -0,0 +1,266 @@
...
+
+	/**
+	 * Register fields in WPGraphQL plugin
+	 *
+	 * @since 3.2.0
+	 *
+	 * @return void
+	 */
+	public function run(): void {
+		if ( class_exists( 'WPGraphQL' ) ) {
```
GaryJones: See https://github.com/Parsely/wp-parsely/pull/706#discussion_r822547658

src/Endpoints/class-graphql-metadata.php
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	/**
+	 * Registers the new custom types for Parse.ly Metadata into the GraphQL instance.
+	 *
+	 * @since 3.2.0
+	 *
+	 * @return void
+	 */
+	private function register_object_types(): void {
```
GaryJones, line 64: I would suggest breaking this method down into smaller methods, one for each type of object graph, and then have this method call those new methods. It's more granular, can be better documented (now, and with any future changes that only affect one object type), avoids an 84-line function call with 6 different sets of argument arrays, and with reflection can be integration tested for each individual registration.

src/Endpoints/class-graphql-metadata.php
```diff
@@ -0,0 +1,266 @@
...
+
+		foreach ( $object_types as $object_type ) {
+			$post_type_object = get_post_type_object( $object_type );
+
+			register_graphql_field(
+				$post_type_object->graphql_single_name,
+				self::FIELD_NAME,
+				array(
+					'type'        => self::GRAPHQL_CONTAINER_TYPE,
+					'description' => 'Parse.ly metadata fields, to be rendered in the front-end so they can be parsed by the crawler. See https://www.parse.ly/help/integration/crawler.',
```
GaryJones: Should this be translatable?

src/Endpoints/class-graphql-metadata.php
```diff
@@ -0,0 +1,266 @@
...
+		/**
+		 * Filters the list of post object types that the Parse.ly GraphQL API is hooked into.
+		 *
+		 * @since 3.2.0
+		 *
+		 * @param string[] $object_types Array of strings containing the object types, i.e. `page`, `post`, `term`.
+		 */
+		$object_types = apply_filters( 'wp_parsely_graphql_object_types', $object_types );
+
+		foreach ( $object_types as $object_type ) {
```
GaryJones: Consider extracting the body of this `foreach` into a new `register_field()` method. That way, this method collects the object types, and another method handles the registration of one. Separation of concerns, and it will help to avoid so many levels of indentation.

src/Endpoints/class-graphql-metadata.php
```diff
@@ -0,0 +1,266 @@
...
+			$post_type_object = get_post_type_object( $object_type );
+
+			register_graphql_field(
+				$post_type_object->graphql_single_name,
+				self::FIELD_NAME,
+				array(
```
GaryJones: Rather than such a large array making the end of the `register_graphql_field()` call difficult to find, extract this array into a variable, and make the `register_graphql_field()` a single line.

This will also help with reducing the significant levels of indentation.

src/Endpoints/class-graphql-metadata.php
```diff
@@ -0,0 +1,266 @@
...
+					'type'        => self::GRAPHQL_CONTAINER_TYPE,
+					'description' => 'Parse.ly metadata fields, to be rendered in the front-end so they can be parsed by the crawler. See https://www.parse.ly/help/integration/crawler.',
+					'resolve'     => function ( $graphql_post ) {
```
GaryJones: Assigning this anonymous function to a variable and then using that variable as the value of this `resolve` key will also help with the indentation levels.

```
$resolve = function ( $graphql_post ) {
	...
};
```
```
	'resolve' => $resolve,
```

src/Endpoints/class-graphql-metadata.php
```diff
@@ -0,0 +1,266 @@
...
+							'rendered' => self::get_rendered_meta(),
+						);
+					},
+				)
+			);
+		}
+	}
+
+	/**
+	 * Adapts an array of metadata to GraphQL. Namely, all keys with a leading `@` symbol are duplicated by the same
```
GaryJones:
```suggestion
	 * Adapts an array of metadata to GraphQL. Namely, all keys with a leading `@` symbol are duplicated to the same
```

Also, this needs (in phpDoc Description, not Summary (first line)) an explanation as to why this is needed. Maybe GraphQL doesn't support `@` in keys? Maybe something else?

src/Endpoints/class-graphql-metadata.php
```diff
@@ -0,0 +1,266 @@
...
+	 * key without the symbol.
+	 *
+	 * @since 3.2.0
+	 *
+	 * @param array<string, mixed> $meta Page metadata array.
+	 *
```
GaryJones:
```suggestion
```

src/Endpoints/class-graphql-metadata.php
```diff
@@ -0,0 +1,266 @@
...
+		return $meta;
+	}
+
+	/**
+	 * Adds the `type` key in an array, sourcing its value from `@type` if that key exists.
+	 *
+	 * @since 3.2.0
+	 *
+	 * @param array<string, mixed> $data Some array that may or may not contain the `@type` key.
+	 *
```
GaryJones:
```suggestion
```

GaryJones (changes requested):

src/Endpoints/class-graphql-metadata.php
```diff
@@ -0,0 +1,269 @@
...
+				'rendered' => self::get_rendered_meta(),
+			);
+		};
+
+		foreach ( $object_types as $object_type ) {
+			$post_type_object = get_post_type_object( $object_type );
+			$config           = array(
+				'type'        => self::GRAPHQL_CONTAINER_TYPE,
+				'description' => __(
+					'Parse.ly metadata fields, to be rendered in the front-end so they can be parsed by the crawler. See https://www.parse.ly/help/integration/crawler.',
```
chriszarate, line 214: Would it be more helpful to link to a description of JSON-LD, since that will help explain the purpose of these metadata fields? Information on the crawler feels like a non-sequitur.
```suggestion
					'Parse.ly metadata fields, to be rendered in the front-end so they can be parsed by the crawler. See https://www.parse.ly/help/integration/jsonld.',
```

chriszarate: Would it be more helpful to link to a description of JSON-LD, since that will help explain the purpose of these metadata fields? Information on the crawler feels like a non-sequitur.
```suggestion
					'Parse.ly metadata fields, to be rendered in the front-end so they can be parsed by the crawler. See https://www.parse.ly/help/integration/jsonld.',
```

src/Endpoints/class-graphql-metadata.php
```diff
@@ -0,0 +1,269 @@
...
+		$meta_type = array(
+			'description' => __(
+				'Metadata fields to be rendered in the front-end. They follow Parse.ly\'s metadata structure. See https://www.parse.ly/help/integration/category/metadata',
+				'wp-parsely'
+			),
+			'fields'      => array(
+				'articleSection'   => array( 'type' => 'String' ),
+				'author'           => array( 'type' => array( 'list_of' => self::GRAPHQL_AUTHOR_TYPE ) ),
+				'context'          => array( 'type' => 'String' ),
+				'creator'          => array( 'type' => 'String' ),
```
chriszarate: This leads to a 500 in my local testing, since `creator` is supplied as an array and not a string. However, isn't `creator` duplicative? Can it be removed, since:

> For author, creator, and contributor properties, Parse.ly will combine all the unique values into a single list.

```suggestion
```

src/Endpoints/class-graphql-metadata.php
chriszarate: > Maybe GraphQL doesn't support @ in keys

Yes, because `@` is reserved for [directives](http://graphql.org/learn/queries/#directives).

src/Endpoints/class-graphql-metadata.php
```diff
@@ -0,0 +1,269 @@
...
+	/**
+	 * Adapts an array of metadata to GraphQL. Namely, all keys with a leading `@` symbol are duplicated by the same
+	 * key without the symbol.
+	 *
+	 * @since 3.2.0
+	 *
+	 * @param array<string, mixed> $meta Page metadata array.
+	 * @return array<string, mixed>
+	 */
+	private function process_meta_for_graphql( array $meta ): array {
```
chriszarate: This function does not process the meta for `image`, so `image.type` is always `null`. It does make me wonder if it should use an `array_walk_recursive` or similar to DRY this up.

src/Endpoints/class-graphql-metadata.php
```diff
@@ -0,0 +1,269 @@
...
+					'description' => __( 'Revision of the metadata format.', 'wp-parsely' ),
+				),
+				'meta'     => array(
+					'type'        => 'ParselyMeta',
+					'description' => __( 'Structured and filterable metadata.', 'wp-parsely' ),
+				),
+				'rendered' => array(
+					'type'        => 'String',
+					'description' => __(
+						'HTML string containing the metadata. Intended to be rendered in the front-end as is.',
```
chriszarate: Rendering a script tag "as-is" can be very difficult in React when the script tag itself is included in the string, as discussed in this earlier thread:

pcURaU-3T-p2

I can recommend offering a GraphQL directive that allows the consumer to choose whether the wrapping script tag is included. We take that approach here in our blocks extension:

https://github.com/Automattic/vip-decoupled-bundle/blob/2fe6cdb3e52a37ec4a320961148df92dac41df7b/blocks/blocks.php

src/Endpoints/class-graphql-metadata.php
```diff
@@ -0,0 +1,269 @@
...
+		 *
+		 * @param array $meta_type Array with the fields of the new type.
+		 */
+		$meta_type = apply_filters( 'wp_parsely_graphql_meta_type', $meta_type );
+		register_graphql_object_type( self::GRAPHQL_META_TYPE, $meta_type );
+
+		$container_type = array(
+			'description' => __( 'Parse.ly Metadata root type.', 'wp-parsely' ),
+			'fields'      => array(
+				'version'  => array(
```
chriszarate: I love that we have "edge" fields here as a place to provide information about the Parse.ly plugin itself! Would we consider adding a `scriptUrl` field or similar that provides a URL to the Parse.ly script?

src/Endpoints/class-graphql-metadata.php
chriszarate: I like this suggestion for readability, but since the types are heavily interdependent I think every test will benefit from having all them registered.

src/Endpoints/class-graphql-metadata.php
```diff
@@ -0,0 +1,269 @@
...
+		);
+		/**
+		 * Filters the array for the custom type to represent metadata on GraphQL.
+		 *
+		 * @see https://wpgraphqldocs.gatsbyjs.io/functions/register_graphql_object_type/
+		 * @since 3.2.0
+		 *
+		 * @param array $meta_type Array with the fields of the new type.
+		 */
+		$meta_type = apply_filters( 'wp_parsely_graphql_meta_type', $meta_type );
```
chriszarate: This filter is redundant since WPGraphQL already provides multiple filters to hook into type registration (search for `apply_filters` in this file):

https://github.com/wp-graphql/wp-graphql/blob/a40f46b935e0e4d2cd46b92603c900cbb017a29d/src/Type/WPObjectType.php#L68

While not well-documented, they are extremely stable. They also provide additional arguments that are very useful when mutating types. I'd recommend pointing to those filters as a standard, central integration point.

```suggestion

```

src/Endpoints/class-graphql-metadata.php
```diff
@@ -0,0 +1,269 @@
...
+
+			/**
+			 * Filters the array with the actual metadata that is exposed through GraphQL.
+			 *
+			 * @see https://wpgraphqldocs.gatsbyjs.io/functions/register_graphql_field/
+			 * @since 3.2.0
+			 *
+			 * @param array $meta Array with the fields of the new type.
+			 */
+			$meta = apply_filters( 'wp_parsely_graphql_meta_object', $meta );
```
chriszarate: This filter is also redundant. WPGraphQL provides multiple filters for field resolution with additional helpful arguments:

https://www.wpgraphql.com/docs/graphql-resolvers/#overriding-existing-resolvers

I'd also point out that it's _very_ easy to introduce critical bugs in field resolvers for large object types like this one, since we don't know, at this location, which fields are being requested. It would be better to encourage plugin users to use the official filters to filter the actual fields that are being resolved (e.g., `authors`) so that they are operating more narrowly and with context.

```suggestion
```

src/Endpoints/class-graphql-metadata.php
```diff
@@ -0,0 +1,269 @@
...
+	/**
+	 * Register the custom metadata fields so they can be queried in GraphQL.
+	 *
+	 * @since 3.2.0
+	 *
+	 * @return void
+	 */
+	private function register_fields(): void {
+		$options      = $this->parsely->get_options();
+		$object_types = array_unique( array_merge( $options['track_post_types'], $options['track_page_types'] ) );
```
chriszarate: This approach works great, but I have an alternative to propose.

First, tying field registration to this setting introduces the possibility of painful breaking schema changes via the UI. Because this option is controlled via the plugin settings, non-technical users probably have no idea that checking or unchecking a box there has side effects in GraphQL. Unchecking the box to track a custom post type would introduce a breaking schema change and could immediately make all queries to GraphQL return 500.

Second, I think this data can be useful even if the post type is not being tracked by Parse.ly. It benefits SEO and makes rendering consistent. Put another way: Turning off "tracking" for a post type shouldn't change how it looks to Google.

I would make two suggestions:

1. Register this field on an interface instead of on individual post types. The field would be consistently available across all post types that implement the interface (a schema-level decision) and be immune to changes in Parse.ly plugin settings (generally, a UI-level decision).

2. Provide an additional boolean field named `isTracked` that informs whether the option to track this post type is enabled in the UI.

Here are the interfaces that `Post` implements:

Screen Shot 2022-03-09 at 6 58 09 PM

While this meta conceptually spans a few of these interfaces, I think `ContentNode` is most appropriate since it is the least common denominator for content that can be described by schema.org data.

src/Endpoints/class-graphql-metadata.php
```diff
@@ -0,0 +1,269 @@
...
+		 * Filters the list of post object types that the Parse.ly GraphQL API is hooked into.
+		 *
+		 * @since 3.2.0
+		 *
+		 * @param string[] $object_types Array of strings containing the object types, i.e. `page`, `post`, `term`.
+		 */
+		$object_types = apply_filters( 'wp_parsely_graphql_object_types', $object_types );
+
+		$resolve = function ( \WPGraphQL\Model\Post $graphql_post ) {
+			$post_id = $graphql_post->__get( 'ID' );
```
chriszarate: Curious: This is a magic method, is it an antipattern to just use it?
```suggestion
			$post_id = $graphql_post->ID;
```

chriszarate (changes requested):
Thanks for this important work! I made some suggestions around hooks and type registration.

I have to admit I am a bit conflicted here. I went back and read [our conversation from a few months back](https://vipanalyticsintegration.wordpress.com/2021/11/09/supporting-decoupled-wordpress-architectures-on-wp-parsely/), because I wanted to look again at this imagined code snippet that would be consuming the data that this PR would provide:

```js
import React from 'react';
import { Helmet } from 'react-helmet';
 
export default function SchemaTag( props ) {
  const schema = /* load Parse.ly schema from API */;
 
  // Add paywall status to the schema
  Object.assign( schema, { isAccessibleForFree: ! props.hasPaywall } );
 
  return (
    <Helmet>
      <script
        dangerouslySetInnerHTML={{ __html: JSON.stringify( schema ) }}
        type="application/ld+json"
      />
    </Helmet>
  );
}
```

This kind of simple integration is no longer possible, because the customer now has a lot of mapping to do, mostly to restore the `@` to the field names that had to be removed when creating full-fledged GraphQL object types. It kind of echoes something you said at the time in support of offering two different data types (which was the right call):

> The code snippet above is precisely what I wanted to avoid. Notice that the user is responsible for the JSON-LD => metas conversion, which might lead to an error. We cannot prevent them from doing so, but if they incorrectly associate a field, that might break their analytics.

I'm really not sure if providing this as structured data is providing much value—or at least not enough to overcome the friction we are adding. Did you have something in mind to help with the front-end integration? Would it be better to supply this as stringified JSON?

```js
"jsonLd": ""{\"@context\":\"http://schema.org\",\"@type\":\"NewsArticle\",\"headline\":\"Zipf's Law of the Internet: Explaining Online Behavior\",\"url\":\"https://blog.parse.ly/post/57821746552\",\"thumbnailUrl\":\"https://blog.parse.ly/inline_mra670hTvL1qz4rgp.png\",\"datePublished\":\"2013-08-15T13:00:00Z\",\"articleSection\":\"Programming\",\"creator\":[\"Alan Alexander Milne\"],\"keywords\":[\"statistics\",\"zipf\",\"internet\",\"behavior\"]}""
```

Or perhaps as a third format? 😬 As it stands in this PR, I think the front-end implementation will be extremely error-prone. What do you think?

src/Endpoints/class-graphql-metadata.php
pauarge: I wasn't aware that those filters existed. It is clear that our filters are redundant.

src/Endpoints/class-graphql-metadata.php
pauarge: For some reason (probably having too much spaghetti code at that stage) it was not working. It is now, so applying your suggestion.

pauarge:
@chriszarate Thanks for your feedback! I see the other comments, but I think we should address how do we want to represent data.

I think we all agree that we need the `rendered` field. Keep in mind that we're using the JSON-ld examples here, but the user can also select from the UI to return repeated meta tags. I like the approach of refining the field with what you're proposing here:

> I can recommend offering a GraphQL directive that allows the consumer to choose whether the wrapping script tag is included. We take that approach here in our blocks extension

For the `meta` field, I agree that the way we have it right now is super error prone. The reason I decided to go with it anyway was because a stringified array wasn't that different from the `rendered` field. However, I see three main points against the current implementation: 

- As you pointed out, it's very error prone and we're trusting that the user will do the right job.
- Having to explicitly define all types introduces a lot of code that we also have to maintain (imagine we need to add a new meta field).
- It doesn't make sense to render the meta tags partially. If they are available, they should all be rendered.

For those reasons, I think it would make sense to return the rendered string (JSON-LD or meta tags) and a stringified version of the meta array that we produce internally in the plugin. This would be very similar to what we have in the REST API, with the convenience of having it available on GraphQL.

What do you think?

chriszarate:
Hi @pauarge! 

> I think we all agree that we need the rendered field. Keep in mind that we're using the JSON-ld examples here, but the user can also select from the UI to return repeated meta tags.

> For those reasons, I think it would make sense to return the rendered string (JSON-LD or meta tags) and a stringified version of the meta array that we produce internally in the plugin.

I think this generally makes sense as a strategy. However, we should try to prevent the `rendered` field from becoming overloaded. If the format of the `rendered` field changes depending on a UI setting, then a UI change can still break front-end code—or require the front-end to conduct a brittle inspection of the content to determine what format it is. 

In GraphQL, if there is choice about the format of data, we should try to place control in the schema itself. This allows the queryer to have confidence about what data they will receive.

We can use separate fields or directives to accomplish this, but separate fields feels better to me because the field names become descriptive. This is useful because field names generally become variable names in front-end code. Directives also introduce a "default" which is impossible to change in the future without disrupting consumers.

Here is what I would suggest. I am just making up field names, feel free to use your own.

```gql
parsely {
  jsonLd(@removeWrappingTag: true)
  meta
  version
}
```

Since we are resolving `jsonLd` to stringified JSON, it probably makes sense to provide a WP-Parsely-specific filter to allow plugin users to filter an associative array before it is converted to JSON.

src/Endpoints/class-graphql-metadata.php
GaryJones: TIL - I suspected that would be the case, but referencing the avoidance of clashing with directives would be good rationale to include in the docs here IMO.

src/Endpoints/class-graphql-metadata.php
pauarge: We can absolutely do that.

src/Endpoints/class-graphql-metadata.php
pauarge: I think `crawler` is more generic, given that we're returning repeated metas and json-ld.

src/Endpoints/class-graphql-metadata.php
pauarge: Given that we've pulled out all types but one, I think it's fine to leave the function as is.

src/Endpoints/class-graphql-metadata.php
```diff
@@ -0,0 +1,147 @@
...
+			$post_type_object = get_post_type_object( $object_type );
+			$config           = array(
+				'type'        => self::GRAPHQL_CONTAINER_TYPE,
+				'description' => __(
+					'Parse.ly metadata fields, to be rendered in the front-end so they can be parsed by the crawler. See https://www.parse.ly/help/integration/crawler.',
+					'wp-parsely'
+				),
+				'resolve'     => $resolve,
+			);
+			register_graphql_field( $post_type_object->graphql_single_name, self::FIELD_NAME, $config );
```
chriszarate: @pauarge FYI since I don't think the docs give an example of registering on an interface, all that's necessary is to pass a string reference to the interface name:

```suggestion
			register_graphql_field( 'ContentNode', self::FIELD_NAME, $config );
```

Then the surrounding code that loops through post types is no longer needed.

chriszarate (commented):
This is looking great! There is one piece of feedback from my earlier review that didn't get addressed or discussed:

> Rendering a script tag "as-is" can be very difficult in React when the script tag itself is included in the string, as discussed in this earlier thread:
> 
> pcURaU-3T-p2
> 
> I can recommend offering a GraphQL directive that allows the consumer to choose whether the wrapping script tag is included. We take that approach here in our blocks extension:
> 
> https://github.com/Automattic/vip-decoupled-bundle/blob/2fe6cdb3e52a37ec4a320961148df92dac41df7b/blocks/blocks.php

I think this is fairly important to improve the ergonomics of using `jsonLd` data. Because frameworks like React must create the wrapping tag with JSX or equivalent, the implementor will be forced to manually strip out the wrapping tag from this plugin's response with RegEx. An example:

```
import React from 'react';
import { Helmet } from 'react-helmet';
 
export default function SchemaTag( props ) {
  const jsonLd = /* load Parse.ly jsonLd from API */;
 
  // Remove wrapping tag.
  const jsonLdWithoutWrappingTag = jsonLd
    .trim()
    .replace( /^<script\s[^>]+>/ )
    .replace( /<\/script>$/ );
 
  return (
    <Helmet>
      <script
        dangerouslySetInnerHTML={{ __html: JSON.stringify( jsonLdWithoutWrappingTag ) }}
        type="application/ld+json"
      />
    </Helmet>
  );
}
```

Providing a directive to give the API consumer the choice of whether to receive the wrapping tag would be much easier to work with.

chriszarate (dismissed):
GraphQL code looks great! I'll defer to @GaryJones for the tests.

GaryJones (dismissed):
I'm not a fan of the approach to mocking used in the tests but I haven't got time right now to test and demonstrate the approach I would use, so adding a review block until I get chance.

(You probably want to read https://brain-wp.github.io/BrainMonkey/docs/functions-when.html or https://brain-wp.github.io/BrainMonkey/docs/functions-expect.html to monkey-patch the externally-expected functions)

GaryJones:
As per the commit message of db5a510:

> Instead of stubbing out WPGraphQL functions under a `Parsely\Endpoints` namespace, and having it set public properties in the test class, let's use BrainMonkey to simply patch the missing external functions instead.
>
> This commit adds a new unit test that includes a test for whether these functions defined in the WPGraphQL plugin are called. We use `Brain\Monkey\Functions\expect()` to name the functions we're expecting to be called, and then also state how many times we expect them to be called, and also what arguments (values or type) they should be called with. If in the course of running the test, anything differs, the test fails. Otherwise, it passes.
>
> The `Brain\Monkey\Functions\when()` call is there to mock the WordPress-defined `__()` function, since in a unit test, WordPress isn't loaded. In this case, we don't care what it was called with or how many times, but we patch it to simply return the first argument that was passed to it (the string to be translated).

After pulling down my commit, feel free to change the `once()` to `twice()` (or to other [Mockery call count expectations](https://docs.mockery.io/en/latest/reference/expectations.html#declaring-call-count-expectations)), or add a typo / make a change in the expected arguments, and see the test fail to convince yourself that it works! 😄 

Docs:
- https://brain-wp.github.io/BrainMonkey/docs/functions-expect.html
- https://brain-wp.github.io/BrainMonkey/docs/functions-when.html

pauarge:
Thanks for the updated tests @GaryJones. I did try BrainMonkey but I hit a wall, so I opted for the manual approach. You code looks way better, so let's stick with it.

src/Endpoints/class-graphql-metadata.php
```diff
@@ -0,0 +1,159 @@
...
+		 *
+		 * @param bool $enabled True if enabled, false if not.
+		 */
+		if ( apply_filters( 'wp_parsely_enable_graphql_support', true ) && $this->parsely->api_key_is_set() ) {
+			add_action( 'graphql_register_types', array( $this, 'register_meta' ) );
+		}
+	}
+
+	/**
+	 * Registers the meta field on the appropriate resource types in the REST API.
```
acicovic:
```suggestion
	 * Register the meta field on the appropriate resource types in the REST API.
```

src/Endpoints/class-graphql-metadata.php
```diff
@@ -0,0 +1,159 @@
...
+	 *
+	 * @return void
+	 */
+	public function register_meta(): void {
+		$this->register_object_types();
+		$this->register_fields();
+	}
+
+	/**
+	 * Registers the new custom types for Parse.ly Metadata into the GraphQL instance.
```
acicovic:
```suggestion
	 * Register the new custom types for Parse.ly Metadata into the GraphQL instance.
```

src/Endpoints/class-metadata-endpoint.php
```diff
@@ -59,11 +59,18 @@ abstract public function register_meta(): void;
 	 *
 	 * @since 3.2.0
 	 *
+	 * @param string $meta_type `json_ld` or `repeated_metas`.
 	 * @return string Contains the metadata as HTML code that can be directly inserted into the page.
```
acicovic:
```suggestion
	 * @return string The metadata as HTML code.
```

src/class-parsely.php
```diff
@@ -129,14 +129,30 @@ public function wpparsely_add_cron_interval( array $schedules ): array {
...
+	public function get_tracker_url(): string {
+		if ( $this->api_key_is_set() ) {
+			$tracker_url = 'https://cdn.parsely.com/keys/' . $this->get_api_key() . '/p.js';
+			return esc_url( $tracker_url );
+		}
+		return '';
+	}
+
 	/**
 	 * Actually inserts the code for the <meta name='parsely-page'> parameter within the <head></head> tag.
```
```suggestion
	 * Insert the code for the <meta name='parsely-page'> parameter within the <head></head> tag.
```

src/class-parsely.php
```diff
@@ -129,14 +129,30 @@ public function wpparsely_add_cron_interval( array $schedules ): array {
...
-	 * @since 3.0.0
+	 * @since 3.2.0
 	 *
+	 * @param string $meta_type `json_ld` or `repeated_metas`.
 	 * @return void
 	 */
-	public function insert_page_header_metadata(): void {
+	public function render_metadata( string $meta_type ): void {
 		/**
 		 * Filters whether the Parse.ly meta tags should be inserted in the page.
```
```suggestion
		 * Filter whether the Parse.ly meta tags should be inserted in the page.
```

src/class-parsely.php
```diff
@@ -224,6 +236,18 @@ public function insert_page_header_metadata(): void {
 		}
 	}
 
+	/**
+	 * Actually inserts the code for the <meta name='parsely-page'> parameter within the <head></head> tag.
```
```suggestion
	 * Insert the code for the <meta name='parsely-page'> parameter within the <head></head> tag.
```

src/class-parsely.php
```diff
@@ -252,11 +276,13 @@ public function insert_parsely_page(): array {
 	/**
 	 * Function to be used in `array_filter` to clean up repeated metas
```
```suggestion
	 * Function to be used in `array_filter` to clean up repeated metas.
```

src/class-parsely.php
```diff
@@ -252,11 +276,13 @@ public function insert_parsely_page(): array {
...
 	 *
+	 * @since 2.6.0
+	 *
 	 * @param mixed $var Value to filter from the array.
 	 * @return bool Returns true if the variable is not empty, and it's a string
```
```suggestion
	 * @return bool True if the variable is not empty, and it's a string.
```

tests/Unit/Endpoints/GraphQLMetadataTest.php
```diff
@@ -0,0 +1,68 @@
...
+	 */
+	public function set_up(): void {
+		parent::set_up();
+
+		self::$parsely = new Parsely();
+		self::$graphql = new GraphQL_Metadata( self::$parsely );
+	}
+
+	/**
+	 * Tests if the functions to register items into GraphQL have been called.
```
acicovic:
```suggestion
	 * Test if the functions to register items into GraphQL have been called.
```

tests/Unit/Endpoints/GraphQLMetadataTest.php
```diff
@@ -0,0 +1,68 @@
...
+
+	/**
+	 * Internal Parsely variable
+	 *
+	 * @var Parsely $parsely Holds the Parsely object
+	 */
+	private static $parsely;
+
+	/**
+	 * The setUp run before each test
```
acicovic:
```suggestion
	 * The setup run before each test.
```

tests/Integration/OtherTest.php
```diff
@@ -245,4 +245,29 @@ public function test_post_has_trackable_status_password_protected_with_filter():
...
+	 *
+	 * @covers \Parsely\Parsely::get_tracker_url
+	 */
+	public function test_get_tracker_url(): void {
+		$expected = 'https://cdn.parsely.com/keys/blog.parsely.com/p.js';
+		self::assertEquals( $expected, self::$parsely->get_tracker_url() );
+	}
+
+	/**
+	 * Tests if the tracker URL is an empty string when there's no API key.
```
acicovic:
```suggestion
	 * Test if the tracker URL is an empty string when there's no API key.
```

tests/Integration/OtherTest.php
```diff
@@ -245,4 +245,29 @@ public function test_post_has_trackable_status_password_protected_with_filter():
 		$result = Parsely::post_has_trackable_status( $post );
 		self::assertTrue( $result );
 	}
+
+	/**
+	 * Tests if the tracker URL is correctly generated with a set API key.
```
acicovic:
```suggestion
	 * Test if the tracker URL is correctly generated with a set API key.
```

src/Endpoints/class-graphql-metadata.php
```diff
@@ -0,0 +1,159 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Parsely\Endpoints;
+
+use WP_Post;
+
+/**
+ * Injects Parse.ly Metadata to the GraphQL outputs
```
acicovic: Maybe revision of this comment is needed?

src/Endpoints/class-graphql-metadata.php
```diff
@@ -0,0 +1,159 @@
...
+ *
+ * @since 3.2.0
+ */
+class GraphQL_Metadata extends Metadata_Endpoint {
+	private const GRAPHQL_VERSION        = '1.0.0';
+	private const GRAPHQL_CONTAINER_TYPE = 'ParselyMetaContainer';
+
+	/**
+	 * Register fields in WPGraphQL plugin
```
acicovic:
```suggestion
	 * Register fields in WPGraphQL plugin.
```

src/class-parsely.php
```diff
@@ -129,14 +129,30 @@ public function wpparsely_add_cron_interval( array $schedules ): array {
 		return $schedules;
 	}
 
+	/**
+	 * Gets the full URL of the JavaScript tracker file for the site. If not api is set, returns an empty string.
```
```suggestion
	 * Get the full URL of the JavaScript tracker file for the site. If an API key is not set, return an empty string.
```

src/class-parsely.php
```diff
@@ -224,6 +236,18 @@ public function insert_page_header_metadata(): void {
...
+	 *
+	 * @return void
+	 */
+	public function insert_page_header_metadata(): void {
+		$parsely_options = $this->get_options();
+		$this->render_metadata( $parsely_options['meta_type'] );
+	}
+
 	/**
 	 * Deprecated. Echoes the metadata into the page, and returns the inserted values.
```
```suggestion
	 * Deprecated. Echo the metadata into the page, and return the inserted values.
```

tests/Integration/Endpoints/GraphQLMetadataTest.php
```diff
@@ -0,0 +1,84 @@
...
+
+	/**
+	 * Internal Parsely variable
+	 *
+	 * @var Parsely $parsely Holds the Parsely object
+	 */
+	private static $parsely;
+
+	/**
+	 * The setUp run before each test
```
acicovic:
```suggestion
	 * The setup run before each test.
```

tests/Integration/Endpoints/GraphQLMetadataTest.php
```diff
@@ -0,0 +1,84 @@
+<?php
+/**
+ * Parsely GraphQL Metadata tests.
```
acicovic:
```suggestion
 * Parse.ly GraphQL Metadata tests.
```

tests/Unit/Endpoints/GraphQLMetadataTest.php
```diff
@@ -0,0 +1,68 @@
+<?php
+/**
+ * Parsely GraphQL Metadata tests.
```
acicovic:
```suggestion
 * Parse.ly GraphQL Metadata tests.
```

acicovic (dismissed):
This has turned into a comments review for me 😃. None of my suggestions are about something that is blocking, so approving this. ~~The changes are based on our discussion to go forward with using imperative within our comments whenever possible.~~

Edit for future reference: The [Coding Standards Handbook](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#language) actually recommends third person singular, so this is the way to go in the future.

acicovic (approved):

PR #722, opened 2022-03-15
Fix type errors when getting authors

### LINKED ISSUE #721: Uncaught TypeError: Return value of Parsely\Parsely::get_author_name() must be of the type string, bool returned
## Describe the bug
A Parse.ly customer reported that they were receiving the following error when scrolling on their site: 

> PHP Fatal error: Uncaught TypeError: Return value of Parsely\Parsely::get_author_name() must be of the type string, bool returned in /srv/htdocs/wp-content/plugins/wp-parsely/src/class-parsely.php:912

I have not been able to reproduce this currently on my end and have asked for more clarification on whether this is on one specific page or on all of their pages

## Expected behavior

We would ideally not be getting a fatal error causing site crashing. I haven't been able to find a page where the author value would be set to a Boolean as the error seems to be reporting however

## Environment

Looks like they're on 3.1.2 based on the tracker configuration. 

## Additional context
We don't get a lot of plugin error reports. I'm happy to ask for any clarification that would be helpful in finding a solution. I have already asked if this is occurring on multiple pages or on one as I am curious if they have a bad author value on one page or if there is something bigger going on here. Thank you for helping investigate.

pauarge:
## Description

This PR adds additional checks over the author name metadata generation to avoid type errors. We are also improving the type annotations on the function headers to make them more strict and allow static type checkers to better do their job.

## Motivation and Context

Closes #721 

## How Has This Been Tested?

Create some posts in a WordPress install and check that the authors get rendered correctly as metadata. We want to consider multiple cases, including single author and multiple authors (using Coauthors Plus plugin). Those authors should have defined/undefined fields, such as first name, last name, etc.

### DIFF
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -909,7 +909,7 @@ class Parsely {
 	 * authors if coauthors plugin is in use.
 	 *
 	 * @param WP_Post $post The post object.
-	 * @return string[]
+	 * @return array<string>
 	 */
 	private function get_author_names( WP_Post $post ): array {
 		$authors = $this->get_coauthor_names( $post->ID );
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -831,7 +831,7 @@ class Parsely {
 	 * https://github.com/Automattic/Co-Authors-Plus/blob/master/template-tags.php#L3-35
 	 *
 	 * @param int $post_id The id of the post.
-	 * @return array
+	 * @return WP_User[]
 	 */
 	private function get_coauthor_names( int $post_id ): array {
 		$coauthors = array();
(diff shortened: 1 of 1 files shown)


acicovic (dismissed):

src/class-parsely.php
```diff
@@ -879,39 +879,41 @@ private function get_coauthor_names( int $post_id ): array {
...
+		return '';
 	}
 
 	/**
 	 * Retrieve all the authors for a post as an array. Can include multiple
 	 * authors if coauthors plugin is in use.
 	 *
 	 * @param WP_Post $post The post object.
-	 * @return array
+	 * @return string[]
```
```suggestion
	 * @return array<string>
```

src/class-parsely.php
```diff
@@ -831,7 +831,7 @@ private function get_custom_taxonomy_values( WP_Post $post_obj, array $parsely_o
 	 * https://github.com/Automattic/Co-Authors-Plus/blob/master/template-tags.php#L3-35
 	 *
 	 * @param int $post_id The id of the post.
-	 * @return array
+	 * @return WP_User[]
```
```suggestion
	 * @return array<WP_User>
```

GaryJones (commented):
Changes make sense, though I believe we agreed to use the slightly expanded type format for readability and consistency?

acicovic (approved):

PR #723, opened 2022-03-15
Improving type support on categories metadata

pauarge:
## Description

This PR improves the type safety on two private functions we use to generate categories metadata in `class-parsely.php`.  Specifically, we are getting rid of an unused parameter and we're fixing some PHPStan issues by making type definitions stricter.

## Motivation and Context

More predictable code behavior through stricter types.

## How Has This Been Tested?

Enable categories as tags in Parse.ly wp-admin. Check that categories get rendered correctly as metadata.

### DIFF
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -668,7 +668,7 @@ class Parsely {
 	 *
 	 * @param int    $post_id The id of the post you're trying to get categories for.
 	 * @param string $delimiter What character will delimit the categories.
-	 * @return string[] All the child categories of the current post.
+	 * @return array<string> All the child categories of the current post.
 	 */
 	private function get_categories( int $post_id, string $delimiter = '/' ): array {
 		$tags = array();
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -804,7 +804,7 @@ class Parsely {
 	 * Get all term values from custom taxonomies.
 	 *
 	 * @param WP_Post $post_obj The post object.
-	 * @return string[]
+	 * @return array<string>
 	 */
 	private function get_custom_taxonomy_values( WP_Post $post_obj ): array {
 		// filter out default WordPress taxonomies.
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -382,7 +382,7 @@ class Parsely {
 			if ( $parsely_options['cats_as_tags'] ) {
 				$tags = array_merge( $tags, $this->get_categories( $post->ID ) );
 				// add custom taxonomy values.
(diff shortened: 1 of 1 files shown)


src/class-parsely.php
```diff
@@ -804,24 +804,22 @@ private function get_bottom_level_term( int $post_id, string $taxonomy_name ): s
...
-	 * @return array
+	 * @return string[]
 	 */
-	private function get_custom_taxonomy_values( WP_Post $post_obj, array $parsely_options ): array {
+	private function get_custom_taxonomy_values( WP_Post $post_obj ): array {
 		// filter out default WordPress taxonomies.
 		$all_taxonomies = array_diff( get_taxonomies(), array( 'post_tag', 'nav_menu', 'author', 'link_category', 'post_format' ) );
 		$all_values     = array();
 
-		if ( is_array( $all_taxonomies ) ) {
```
pauarge, line 815: `$all_taxonomies` is always an array.

src/class-parsely.php
```diff
@@ -804,24 +804,22 @@ private function get_bottom_level_term( int $post_id, string $taxonomy_name ): s
 	 * Get all term values from custom taxonomies.
 	 *
 	 * @param WP_Post $post_obj The post object.
-	 * @param array   $parsely_options Unused? The parsely options.
```
pauarge, line 807: `$parsely_options` are not used in this method. Since it's a private one, we can change the header comfortably.

acicovic (dismissed):

src/class-parsely.php
```diff
@@ -668,7 +668,7 @@ private function get_tags( int $post_id ): array {
 	 *
 	 * @param int    $post_id The id of the post you're trying to get categories for.
 	 * @param string $delimiter What character will delimit the categories.
-	 * @return array All the child categories of the current post.
+	 * @return string[] All the child categories of the current post.
```
```suggestion
	 * @return array<string> All the child categories of the current post.
```

src/class-parsely.php
```diff
@@ -804,24 +804,22 @@ private function get_bottom_level_term( int $post_id, string $taxonomy_name ): s
 	 * Get all term values from custom taxonomies.
 	 *
 	 * @param WP_Post $post_obj The post object.
-	 * @param array   $parsely_options Unused? The parsely options.
-	 * @return array
+	 * @return string[]
```
```suggestion
	 * @return array<string>
```

GaryJones (commented):
Changes make sense, but thought that we had decided on using the slightly longer type format for readability and consistency?

acicovic (approved):

PR #724, opened 2022-03-15
Add more tests to Recommendations Block

acicovic:
## Description
This PR adds:
- A unit test using the React Testing Library, for testing that the Block correctly displays the API data passed to it. The API data is mocked.
- An E2E test for testing errors generated by the Block.
- An E2E test for testing that any setting changes in the editor get reflected on the Block in real time.
- The `SetSiteId()` function which replaces the `*PluginApiKey()` functions. The benefit of this is that we can now also specify a custom Site ID if needed. 

## Motivation and Context
- Add more tests.
- Use "Site ID" terminology everywhere possible.

## How Has This Been Tested?
- "Manually" tested in interactive mode.
- Tests pass.


tests/blocks/recommendations/block.test.js
```diff
@@ -0,0 +1,155 @@
...
+/**
+ * Internal dependencies
+ */
+import ParselyRecommendations from '../../../src/blocks/recommendations/components/parsely-recommendations';
+import ParselyRecommendationsList from '../../../src/blocks/recommendations/components/parsely-recommendations-list';
+import ParselyRecommendationsTitle from '../../../src/blocks/recommendations/components/parsely-recommendations-title';
+import RecommendationsStore from '../../../src/blocks/recommendations/recommendations-store';
+
+describe( 'Recommendations Block', () => {
+	const apiResultSingle = [
```
pauarge: This could be extracted outside the `describe` block.

acicovic: Done!

tests/Unit/blocks/recommendations/block-structure.test.js
```diff
@@ -0,0 +1,342 @@
...
+		metadata: null,
+		pub_date: '2022-03-14T13:14:00',
+		section: 'Articles',
+		tags: [ 'parsely_smart:entity:Tag 1', 'parsely_smart:iab:Tag 2' ],
+		thumb_url_medium: 'https://example.com/image-thumb.jpg',
+		title: 'My nice article',
+		url: 'https://example.com/my-nice-article/',
+	},
+];
+const apiResultMultiple = [
```
pauarge: Maybe we could construct this one doing something like:

```
const apiResultMultiple = apiResultSingle.concat([
  // new items
  ...
])
```

tests/e2e/specs/blocks/recommendations/block-errors.spec.js
```diff
@@ -0,0 +1,57 @@
...
+		await page.waitForSelector( '.parsely-recommendations-error' );
+		const text = await page.$eval( '.parsely-recommendations-error', ( element ) => element.textContent );
+		expect( text ).toMatch( 'Access denied. Please verify that your Site ID is valid.' );
+	} );
+
+	/**
+	 * Verifies that the block will display an error when an empty Site ID is provided.
+	 */
+	it( 'Should display an error when an empty Site ID is provided', async () => {
+		await setSiteId( '' );
```
pauarge, line 48: Instead of introducing a new function, we could achieve this with current helpers: `deactivatePluginApiKey`

pauarge (changes requested):
Great stuff. Left a couple of potential improvements. Also, let's try to fix the SonarCloud issues.

tests/Unit/blocks/recommendations/block-structure.test.js
acicovic: If you think this is better, I can go with that. In that case I'll have to modify some checks though because the two objects aren't the same. Let me know what you would prefer.

tests/e2e/specs/blocks/recommendations/block-errors.spec.js
acicovic: In the case you mention, we could use `deactivatePluginApiKey()` but I'm unsure this is what we really want to do. Let me explain my reasoning:

The `setSiteId()` function was essentially introduced to:
- Be able to set a custom Site ID (which is not an option in `activatePluginApiKey()` where the ID is hardcoded).
- Start moving away from the `Plugin API Key` terminology towards `Site ID`, and this without doing a full change right now because I wanted to keep the PR focused.

Instead of using `deactivatePluginApiKey()`, I would probably introduce a `setDefaultSiteId()` and `setEmptySiteId()` function to replace `activatePluginApiKey()` and `deactivatePluginApiKey()` respectively.

Alternatively, `setSiteId()` could have the default parameter `e2etest.example.com` if we want to have 2 functions (or even 1 function) instead of three. The bottom line is that for some tests, the Site ID will become important and we should be able to customize it.

This is an easy thing and I can do a PR for it if we want to.

acicovic:
> Also, let's try to fix the SonarCloud issues.

Had not spotted the SonarCloud code smells. Will look into that.

acicovic:
Code smells are now fixed, and merging should calm SonarCloud.

tests/e2e/specs/blocks/recommendations/block-errors.spec.js
pauarge: I believe that we should not introduce a duplicate function. This is internal code that we use for testing, so we don't have to keep strict backwards compatibility on the methods signature as we do in the rest of our code.

I like the proposal of using a default parameter on `setSiteId`, which would avoid having multiple functions. I think this is a tiny refactor, so we could very well do this in this PR if we want to move away from the API Key terminology.

tests/Unit/blocks/recommendations/block-structure.test.js
pauarge: I misread and thought they were identical. Ignore my comment.

pauarge:
@acicovic The SonarCloud alerts come from the `block-structure.test.js` file, which signals many duplicated files: https://sonarcloud.io/component_measures?metric=new_duplicated_lines_density&selected=Parsely_wp-parsely%3Atests%2FUnit%2Fblocks%2Frecommendations%2Fblock-structure.test.js&view=list&pullRequest=724&id=Parsely_wp-parsely

Is there something we could do to avoid duplications on that file?

acicovic:
> Is there something we could do to avoid duplications on that file?

Yes, I'll revisit the file to minimize repetition. We should be able to refactor certain parts into functions.

pauarge (approved):

PR #728, opened 2022-03-17
Remove phased-out boost parameters from Recommended Widget

acicovic:
## Description
We've been informed that the `li_interactions` and `li_referrals` boost parameters have been phased-out and are now unsupported. This PR removes these parameters from the Recommended Widget.

## Motivation and Context
Avoid unexpected/buggy behavior and not showing outdated options to plugin users.

## How Has This Been Tested?
Automated tests pass.

### DIFF
--- a/src/UI/class-recommended-widget.php
+++ b/src/UI/class-recommended-widget.php
@@ -278,7 +278,7 @@ final class Recommended_Widget extends WP_Widget {
 	 *
 	 * @since 2.5.0
 	 *
-	 * @return array Boost parameters values and labels.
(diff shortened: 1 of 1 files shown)


pauarge (changes requested):
Changes look good. May I suggest that now that we're modifying this function, we update the return type as well? Something like `array<string, string>`?

acicovic:
> Changes look good. May I suggest that now that we're modifying this function, we update the return type as well? Something like `array<string, string>`?

Done!

pauarge (approved):
LGTM

Issue #682, opened 2022-02-20
Refactor metadata construction


GaryJones:
## Is your feature request related to a problem?
The `Parsely\Parsely::construct_parsely_metadata()` method is a bit of a mess. I thought that maybe the [Builder pattern](https://refactoring.guru/design-patterns/builder) could be a better way of structuring the logic. This issue is to summarise a potential approach. 

## Describe the solution you'd like
In this pattern, a "builder" object is the context under which metadata should be constructed e.g. single post, single page, author archive, homepage etc. You would have one builder class for each condition currently in the `construct_parsely_metadata()` method. In the examples below, I've only shown a few.

As per the pattern, we start off with an interface for the builder objects. There would be a build method for each of the metadata keys, but I've only shown a few here - two which are required for posts and non-posts, and one that only needs to be set for posts.

```php
interface MetaData {
	public function buildHeadline(): void;
	public function buildUrl(): void;
	public function buildAuthor(): void;
}
```

Then. there's a base abstract builder class, which contains a `$metadata` property, and utility methods.

I've also chosen to add in a noop method for methods which are not required for non-posts. This makes the usage of the classes much easier, but probably goes against the pure Builder pattern:

```php
abstract class AbstractMetaData {
	/**
	 * @var array<string, mixed>
	 */
	protected $metadata = array();

	/**
	 * Sanitize content
	 *
	 * @since 2.6.0
	 *
	 * @param string|null $val The content you'd like sanitized.
	 * @return string
	 */
	protected function clean_value( ?string $val ): string {
		if ( null === $val ) {
			return '';
		}

		return trim( wp_strip_all_tags( str_replace( [ "\n", "\r" ], '', $val ) ) );
	}

	/**
	 * @return array<string, mixed>
	 */
	public function get_metadata(): array {
		return $this->metadata;
	}

	// Include noop functions for optional metadata values.
	public function buildAuthor(): void {
		return; // @codeCoverageIgnore
	}
}
```

Now we create a builder class for each condition / context in `Parsely\Parsely::construct_parsely_metadata()`.
Only the headline and URLs have been added here, other than the class for the Single Post, which has an overriding `buildAuthor()` method as well. The implementation logic is taken/adapted from the current logic (and may not be perfect):

```php
class FrontPage extends AbstractMetaData implements MetaData {

	public function buildHeadline():  void {
		$this->metadata['headline'] = $this->clean_value( get_bloginfo( 'name', 'raw' ) );
	}
[...]

PR #742, opened 2022-03-28
Moving metadata generation to external class

pauarge:
## Description

We are working towards a better way of generating metadata in the plugin. All that logic lies currently on the `Parsely` class, which is a bit problematic. In this initial PR, we are simply moving the metadata-related methods from that class into their own `Metadata` class. 

The method's contents is fairly untouched (minus some minor adjustments due to moving), and we've added a proxy method on `Parsely` for backwards compatibility reasons.

## Motivation and Context

See https://github.com/Parsely/wp-parsely/issues/682

## How Has This Been Tested?

All tests pass and metadata is generated as usual.

### DIFF
--- a/src/class-metadata.php
+++ b/src/class-metadata.php
@@ -1,0 +1,632 @@
+<?php
+/**
+ * Metadata class
+ *
+ * @package Parsely
+ * @since 3.3.0
+ */
+
+declare(strict_types=1);
+
+namespace Parsely;
+
+use WP_User;
+use WP_Post;
+
+/**
+ * Generates and inserts metadata readable by the Parse.ly Crawler.
+ *
+ * @since 1.0.0
+ * @since 3.3.0 Moved from class-parsely to separate file
+ */
+class Metadata {
+	/**
+	 * Instance of Parsely class.
+	 *
+	 * @var Parsely
+	 */
+	private $parsely;
+
+	/**
+	 * Constructor.
+	 *
+	 * @param Parsely $parsely Instance of Parsely class.
+	 */
+	public function __construct( Parsely $parsely ) {
+		$this->parsely = $parsely;
+	}
+
+	/**
+	 * Creates parsely metadata object from post metadata.
+	 *
+	 * @param array<string, mixed> $parsely_options parsely_options array.
+	 * @param WP_Post              $post object.
+	 *
+	 * @return array<string, mixed>
+	 */
+	public function construct_metadata( array $parsely_options, WP_Post $post ): array {
+		$parsely_page      = array(
+			'@context' => 'http://schema.org',
+			'@type'    => 'WebPage',
+		);
+		$current_url       = $this->get_current_url();
+		$queried_object_id = get_queried_object_id();
+
+		if ( is_front_page() && ! is_paged() ) {
+			$parsely_page['headline'] = $this->get_clean_parsely_page_value( get_bloginfo( 'name', 'raw' ) );
+			$parsely_page['url']      = home_url();
+		} elseif ( is_front_page() && is_paged() ) {
+			$parsely_page['headline'] = $this->get_clean_parsely_page_value( get_bloginfo( 'name', 'raw' ) );
+			$parsely_page['url']      = $current_url;
+		} elseif (
+			is_home() && (
+				! ( 'page' === get_option( 'show_on_front' ) && ! get_option( 'page_on_front' ) ) ||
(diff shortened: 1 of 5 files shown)


src/Endpoints/class-rest-metadata.php
```diff
@@ -81,7 +82,7 @@ public function get_callback( array $object ): array {
 		if ( false === $post ) {
 			$meta = '';
 		} else {
-			$meta = $this->parsely->construct_parsely_metadata( $options, $post );
+			$meta = $this->parsely->construct_parsely_metadata;
```
GaryJones: This bit doesn't seem right.

acicovic:
2 questions:
- Would we want to put a test that still calls the old construct_parsely_metadata() function directly?
- Would we want to mark the construct_parsely_metadata() as deprecated?

pauarge:
> Would we want to put a test that still calls the old construct_parsely_metadata() function directly?

That is still the case. All tests call that function.

> Would we want to mark the construct_parsely_metadata() as deprecated?

Absolutely.

src/class-metadata.php
```diff
@@ -0,0 +1,632 @@
...
+namespace Parsely;
+
+use WP_User;
+use WP_Post;
+
+/**
+ * Generates and inserts metadata readable by the Parse.ly Crawler.
+ *
+ * @since 1.0.0
+ * @since 3.3.0 Moved from class-parsely to separate file
```
GaryJones:
```suggestion
 * @since 3.3.0 Logic extracted from Parsely\Parsely class to separate file/class.
```

src/class-metadata.php
```diff
@@ -0,0 +1,632 @@
...
+
+	/**
+	 * Creates parsely metadata object from post metadata.
+	 *
+	 * @param array<string, mixed> $parsely_options parsely_options array.
+	 * @param WP_Post              $post object.
+	 *
+	 * @return array<string, mixed>
+	 */
+	public function construct_metadata( array $parsely_options, WP_Post $post ): array {
```
GaryJones, line 47: Would there be an advantage to splitting this up into smaller methods now, rather than iterate on this later? 177 lines seems a lot to add as a "new" method, even if it is just moving it from a different class.

Could also look at making the second arg optional now, since we know that the only bit of code calling this new method is ourselves, and it's not needed for many of the conditions/contexts.

In fact, we could think about whether the $parsely_options arg is needed, since that should be consistent between requests (and for tests we update the DB anyway).

src/class-parsely.php
```diff
@@ -332,214 +331,21 @@ public static function post_has_trackable_status( $post ): bool {
...
-		$metadata['dateCreated']   = $post_created_gmt;
-		$metadata['datePublished'] = $post_created_gmt;
-		$metadata['dateModified']  = $post_created_gmt;
-
-		$post_modified_gmt = get_post_modified_time( $date_format, true, $post );
-
-		if ( false !== $post_modified_gmt && $post_modified_gmt > $post_created_gmt ) {
-			$metadata['dateModified'] = $post_modified_gmt;
-		}
+		_deprecated_function( 'Parsely::construct_parsely_metadata', '3.3.0' );
```
GaryJones: I think we could soft deprecate (just note in the changelog, documentation and DocBlocks) this in 3.3, and then hard deprecate this (call `_deprecated_function()`) in 3.4. It's a smoother experience for the customer developers, and we can't remove this until 4.0 anyway, so there's little point in rushing for us, if internally it's already using the new code anyway.

GaryJones (commented):
I like this PR as a big first step to extracting the metadata construction logic, but I hope that more changes are planned around this before release, so that we can really architect the code (i.e. [replace conditions with polymorphism](https://refactoring.guru/replace-conditional-with-polymorphism))

src/class-metadata.php
pauarge: I agree on all of your points and those are changes we are going to do. I do have a PR in the pipeline, to be opened right after this one, with these kind of changes.

However, I see value on merging them separately. This way, reviews are easier and clearer. We have just released 3.2.0, so all this changes will be shipped atomically in 3.3.0. I don't see the current state of this PR as the final releasable state of 3.3.0.

acicovic (approved):

src/class-metadata.php
GaryJones: Great that we're in alignment! Thanks for confirming.

PR #743, opened 2022-03-28
Adding tracker URL to REST API

### LINKED ISSUE #732: Add script URL to REST API metadata

We have recently made the script URL field available on the GraphQL Metadata endpoint. We would like to add that same field into the REST API metadata endpoint.

We would like to add the `script_url` field into the REST API endpoint. We can use the `get_tracker_url` function from the `Parsely` class.

We will need to bump the REST API version to `1.1`.

Not implementing the field.

## Additional context

https://github.com/Parsely/wp-parsely/pull/710

pauarge:
## Description

This PR brings feature parity back between the REST API and the GraphQL API. With this update, we're providing the tracking script URL generated according to the API key. This way, consumer sites will be able to fetch it directly from where the API is pointing them.

The feature can be disabled by a filter. This PR also includes additional tests to check the functionality.

## Motivation and Context

Closes https://github.com/Parsely/wp-parsely/issues/732.

## How Has This Been Tested?

We're leveraging the same functionality that GraphQL uses. To test, go to a REST API post or page and check in the `parsely` object. It should contain the `tracker_url` field.

Screen Shot 2022-03-29 at 11 59 40 AM

### DIFF
--- a/src/Endpoints/class-rest-metadata.php
+++ b/src/Endpoints/class-rest-metadata.php
@@ -19,7 +19,7 @@ use WP_Post;
  * @since 3.2.0 Renamed FQCN from `Parsely\Rest` to `Parsely\Endpoints\Rest_Metadata`.
  */
 class Rest_Metadata extends Metadata_Endpoint {
-	private const REST_VERSION = '1.0.0';
+	private const REST_VERSION = '1.1.0';
 
 	/**
 	 * Register fields in WordPress REST API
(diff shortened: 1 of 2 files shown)


src/Endpoints/class-rest-metadata.php
```diff
@@ -95,11 +95,24 @@ public function get_callback( array $object ): array {
...
 		 *
 		 * @param bool $enabled True if enabled, false if not.
+		 * @param WP_Post|false $post Current post object.
 		 */
 		if ( apply_filters( 'wp_parsely_enable_rest_rendered_support', true, $post ) ) {
 			$response['rendered'] = $this->get_rendered_meta( $options['meta_type'] );
 		}
 
+		/**
+		 * Filter whether REST API support in rendered string format is enabled or not.
```
acicovic: This comment needs to be updated?

acicovic (approved):

PR #751, opened 2022-03-31
Extracting metadata rendering to UI class

pauarge:
## Description

This PR extracts the functions responsible for rendering the metadata on the WordPress front-end to its own separate class, `class-metadata-header.php`. The code itself hasn't been modified.

The following methods have been deprecated:

- `render_metadata`
- `insert_page_header_metadata`

The following related functions have been deleted from `class-parsely.php`:

- `insert_parsely_page`
- `filter_empty_and_not_string_from_array`

## Motivation and Context

See #682 

## How Has This Been Tested?

Metadata renders consistently with how it was before the changes.


src/UI/class-metadata-header.php
```diff
@@ -0,0 +1,156 @@
...
+
+declare(strict_types=1);
+
+namespace Parsely\UI;
+
+use Parsely\Metadata;
+use Parsely\Parsely;
+use WP_Post;
+
+use const Parsely\PARSELY_FILE;
```
GaryJones: Thoughts on injecting this into the constructor to avoid a reference to a global constant? It's essentially used to ask "What is the path to the `views/` directory?".

pauarge: I like it. This way it's easier for us to test too.

acicovic: Noting that we use `use const Parsely\PARSELY_FILE` in some places already, so if the change goes forward it would be good to change it everywhere.

Having mixed feelings about this change, as I'm unsure that it really improves anything and I like to avoid extra function arguments whenever possible. I see also limited benefits of this in testing, but I might be missing something of course.

acicovic:
Have you considered alternative class names like `Metadata_Renderer` or other similar names, and do you think `Metadata_Header` is a good fit for that class?

src/UI/class-metadata-header.php
pauarge: I can actually see both points. I think consistency is important, and indeed doing the change here would mean adopt the same pattern somewhere else, which might be tricky.

pauarge:
> Have you considered alternative class names like Metadata_Renderer or other similar names, and do you think Metadata_Header is a good fit for that class?

`Metadata_Renderer` makes more sense. I'm going to change it.

acicovic (dismissed):

acicovic (dismissed):

acicovic (approved):

PR #758, opened 2022-04-05
Improve image URLs in generated metadata

### LINKED ISSUE #703: Recommendations Block: Featured images are limited to 120px width
_**Edit: This issue was a bit broad and a research in progress. It has now been updated to be more specific to the problem encountered, and some other points mentioned here may be moved to other issues.**_

The Recommendations Block shows poor quality images when those are set as a "Featured Image" in a post/page. The root of the issue is that we're passing to them a `?w=120` querystring in the plugin code.

To see this in action, let's take [this API request](https://api.parsely.com/v2/related?apikey=wpvip.com&url=wpvip.com&limit=1) as an example, which returns:
```
image_url: "https://wpvip.com/wp-content/uploads/sites/3/2022/02/StudioSimpatico.jpg?w=120"
thumb_url_medium: "https://images.parsely.com/dtC8RKDnBdyGCAbwBBhELPv7cgI=/85x85/smart/https%3A//wpvip.com/wp-content/uploads/sites/3/2022/02/StudioSimpatico.jpg%3Fw%3D120"
```
From the above we can see that:
- `image_url` is the real source image, but limited in width per the `w=120` parameter.
- `thumb_url_medium` is a (small and cropped) image served by Parse.ly, and it is using the previous image as a source.

This issue does not happen for external images.

acicovic:
## Description
This PR:
- Fixes our metadata generation function that would erroneously limit all featured image URLs to thumbnail size.
- Removes the `get_first_image()` fallback function (which was not working), in favor of letting the Parse.ly crawler select the image URL by itself (should usually be the first image in the page).

## Motivation and Context

Fix #703

See #682 

## How Has This Been Tested?
- Manual testing on local devenv.
- Automated tests pass.

### DIFF
--- a/src/class-metadata.php
+++ b/src/class-metadata.php
@@ -558,18 +558,17 @@ class Metadata {
 	}
 
 	/**
-	 * Get the first image from a post
-	 * https://css-tricks.com/snippets/wordpress/get-the-first-image-from-a-post/
+	 * Gets the first image from a post.
 	 *
 	 * @since 3.3.0 Moved to class-metadata
 	 *
-	 * @param WP_Post $post The post object you're interested in.
+	 * @param WP_Post $post The post object.
 	 * @return string
 	 */
 	private function get_first_image( WP_Post $post ): string {
 		ob_start();
 		ob_end_clean();
-		if ( preg_match_all( '/<img.+src=[\'"]( [^\'"]+ )[\'"].*>/i', $post->post_content, $matches ) ) {
+		if ( preg_match_all( '/\<img.+src\=(?:\"|\')(.+?)(?:\"|\')(?:.+?)\>/i', $post->post_content, $matches ) ) {
 			return $matches[1][0];
 		}
 		return '';
--- a/src/class-metadata.php
+++ b/src/class-metadata.php
@@ -104,11 +104,12 @@ class Metadata {
 
 			// Get featured image and thumbnail. On failure, fall back to the post's first image.
 			$image_url = get_the_post_thumbnail_url( $post, 'full' );
-			if ( false !== $image_url ) {
-				$thumb_url = get_the_post_thumbnail_url( $post, 'thumbnail' );
-			} else {
-				$image_url = $this->get_first_image( $post );
-				$thumb_url = $image_url;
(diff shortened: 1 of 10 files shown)


acicovic:
Based on the results I'm getting from the `/recommended` Parse.ly API endpoint:

If the metadata doesn't contain any image URLs, it seems that the Parse.ly crawler grabs the first image it finds in the page and sets it as the image URL. As such, we could argue that our fallback solution is unneeded, since the crawler will do this by itself. We also currently do this using a regex, so we do incur a "regex penalty" for pages that don't contain a featured image.

src/class-metadata.php
```diff
@@ -102,12 +102,14 @@ public function construct_metadata( array $parsely_options, WP_Post $post ): arr
...
 
-			if ( has_post_thumbnail( $post ) ) {
-				$image_id  = get_post_thumbnail_id( $post );
-				$image_url = wp_get_attachment_image_src( $image_id );
-				$image_url = $image_url[0];
-			} else {
-				$image_url = $this->get_first_image( $post );
+			// Get featured image and thumbnail.
+			$image_url = get_the_post_thumbnail_url( $post, 'full' );
+			if ( false === $image_url ) {
```
GaryJones: `get_the_post_thumbnail_url()` has a late filter, so it could return any data type (like `null`) if someone applies a mis-configured filter.

So that we're definitely dealing with a string, then consider change this check (and the instance below) to:

```suggestion
			if ( ! is_string( $image_url ) ) {
```

That way, we know that `null`, `true`, `5` and other non-strings can't get through, while still guarding against the `false` which is the most likely possibility.

It also makes the code a little more agnostic. My suggestion basically says "If it's not a string, then make it an empty string", where as the code you've got says "If it is a literal false (because `get_the_post_thumbnail_url()` couldn't find a truthy ID from the `$post`), then make it an empty string". i.e. we want to test for the actual thing we're looking for (a string), and not just a specific other data type that we presume the function might return.

acicovic: Exceptional catch. Implemented.

pauarge (approved):
Awesome catch!

PR #761, opened 2022-04-06
Update @uses and @covers comments taking into account Metadata class

acicovic:
## Description
We recently moved some functions to the `Metadata` class, but test comments (`@uses` and `@covers`) did not take this into account, showing broken paths. This PR updates function paths wherever needed.

Note: paths to the `get_first_image()` function were not updated due to its removal in comments within #758.

## Motivation and Context
Have correct function paths into our tests.

## How Has This Been Tested?
- Comment changes only, should not impact any code
- Automated tests pass

### DIFF
--- a/tests/Integration/GetCurrentUrlTest.php
+++ b/tests/Integration/GetCurrentUrlTest.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * \Parsely\Parsely::get_current_url() tests.
+ * \Parsely\Metadata::get_current_url() tests.
  *
  * @package Parsely\Tests
  */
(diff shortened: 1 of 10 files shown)


GaryJones:
Looking at the code coverage step in https://github.com/Parsely/wp-parsely/runs/5851508348?check_suite_focus=true then there are still lots of coverage issues, including with the Metadata class.

acicovic:
> there are still lots of coverage issues, including with the Metadata class.

Agreed. We should probably talk about it to organize, and maybe @pauarge has something in store for this.

pauarge (approved):
I do have some upcoming changes for the Metadata generation, but @acicovic work in this PR is still relevant and doesn't conflict with what I'm doing. Recognizing that there are still some rough edges, I think this is an improvement, so happy to merge it.

PR #757, opened 2022-04-04
Fix Recommendations Block FSE issues

acicovic:
## Description
The [Recommendations Block](https://github.com/Parsely/wp-parsely/pull/642) has some styling issues in FSE and does not work correctly in the back-end when saved in FSE mode.

This PR is a work in progress which aims to eliminate all the issues we're having with the Recommendations Block in FSE mode.

- [x] Fix styling issues
- [ ] Fix Block showing error or disappearing in FSE after save and reload

## Motivation and Context
Fixes #711 for a smooth FSE experience.

PR #762, opened 2022-04-06
Register the Recommendations block using Block MetaData and a block.json file

### LINKED ISSUE #711: Make Recommendations Block work correctly in FSE
## Describe the bug
Currently, the Recommendations Block has issues with the Full Site Editing environment. Notably:

- It displays badly in the editor (distorted styles)
- When the template is saved and we revisit/refresh the editor, the Block vanishes or shows a “This block has encountered an error and cannot be previewed” error, which makes its options non-modifiable.

## To Reproduce
With a Block Theme, open the FSE Editor and drop the Block somewhere.

ryanwelcher:
## Description
This PR is based on #757 and contains additional changes. It changes and simplifies the block registration process by leveraging block metadata and a `block.json` file, and it solves the following issues:
- Styling issues in FSE
- Block showing error or disappearing in FSE after save and reload

## Motivation and Context
- This change stemmed from issues with this block in FSE and after reviewing the code in an attempt to fix the issue, I made recommendations to change both how the block is registered (reflected in this PR) as well as some changes to the experience in the block editor.
- Fixes #711

## How Has This Been Tested?
Tested locally only


src/blocks/recommendations/class-recommendations-block.php
```diff
@@ -49,41 +49,38 @@ public function run(): void {
...
-			'wp-parsely-recommendations-block-editor',
-			$plugin_url . 'build/recommendations-edit.js',
-			$editor_asset_file['dependencies'],
-			$editor_asset_file['version'],
-			true
-		);
+		if ( ! isset( $wp_version ) || version_compare( $wp_version, '5.9' ) < 0 ) {
+			$plugin_url      = plugin_dir_url( PARSELY_FILE );
+			$view_asset_path = plugin_dir_path( PARSELY_FILE ) . 'build/blocks/recommendations/view.asset.php';
+			if ( file_exists( $view_asset_path ) ) {
```
GaryJones: Assuming that the build should always succeed (and we want a hard fail if it doesn't / hasn't been run locally), what does this condition here give us?

src/blocks/recommendations/block.json
```diff
@@ -0,0 +1,44 @@
...
+		"sort": {
+			"type":"string",
+			"default":"score"
+		},
+		"tag": {
+			"type":"string"
+		},
+		"title": {
+			"type":"string",
+			"default": "Related Content"
```
GaryJones, line 37: Can this string be translated at some point?

src/blocks/recommendations/class-recommendations-block.php
```diff
@@ -49,41 +49,38 @@ public function run(): void {
...
+		 */
 
-		wp_register_style(
-			'wp-parsely-recommendations-block',
-			$plugin_url . 'build/style-recommendations-edit.css',
-			array(),
-			$script_asset_file['version']
+		register_block_type(
+			plugin_dir_path( PARSELY_FILE ) . 'build/blocks/recommendations/',
+			self::get_block_registration_args()
```
GaryJones: Let's inline the array returned from the `get_blocks_registration_args()` method, and then delete that method.

src/blocks/recommendations/class-recommendations-block.php
ryanwelcher: It's an extra safeguard against a fatal error if we try to `require` a file that doesn't exist, but that's about it. Happy to remove this check if you feel it's superfluous.

src/blocks/recommendations/block.json
ryanwelcher: Strings in `block.json` should translated automatically during the block registration process - https://github.com/WordPress/wordpress-develop/blob/573581f43f63f12cc37620172432cf93745f568c/src/wp-includes/blocks.php#L289. I didn't test this however.

src/blocks/recommendations/block.json
```diff
@@ -0,0 +1,44 @@
+{
+	"$schema": "https://schemas.wp.org/wp/5.9/block.json",
+	"apiVersion": 2,
+	"name": "wp-parsely/recommendations",
+	"title": "Parse.ly Recommendations",
```
GaryJones, line 5: As per https://core.trac.wordpress.org/changeset/49981 the `title`, `description`, each `keyword`, and each style `label` are automatically wrapped in `translate_with_gettext_context()` with the textdomain passed from the value on line 7 ✅

src/blocks/recommendations/class-recommendations-block.php
GaryJones: Under what conditions could we reasonably expect the file not to exist, and NOT want a fatal?

src/blocks/recommendations/block.json
GaryJones: If I'm reading that right - that's only for top-level keys, and not an attribute's [default] value.

src/blocks/recommendations/class-recommendations-block.php
ryanwelcher: This would only become an issue when/if the build directory was either non-existent or this file was missing. 

From a developer's point of view, it's not really a big deal to white screen the site if the build didn't get run, but in the (edge)-case of a plugin update going sideways on a client site, white screening the admin would be problematic.  At any rate, I have removed the check.

src/blocks/recommendations/block.json
```diff
@@ -0,0 +1,44 @@
...
+		},
+		"title": {
+			"type":"string",
+			"default": "Related Content"
+		}
+	},
+	"editorScript": "file:./edit.js",
+	"style": "file:./style-edit.css",
+	"editorStyle": "file:./edit.css",
+	"viewScript": "file:./view.js"
```
GaryJones, line 43: Noting that using / relying on the block.json means WP 5.8 is the minimum, and `viewScript` was only supported in WP 5.9.

src/blocks/recommendations/block.json
ryanwelcher: Yes, I think you're correct. Update incoming.

src/blocks/recommendations/block.json
jblz: I was trying to remember why I didn't use the json for this in the first place and the minimum WordPress version is it.

ryanwelcher: What is the minimum version this block needs to support?  I probably should have asked that first 😄

acicovic: OK, so the Block in its initial state was supporting WP 5.6 (released December 2020). [That line is still present in class-recommendations-block.php](https://github.com/Parsely/wp-parsely/blob/ed18932ef133a4dd3d8f28492084e43d44e90a78/src/blocks/recommendations/class-recommendations-block.php#L19). Personally, I don’t see a need to support such an old version if this results in difficult problems for us.

The big change with this PR is that _the block now works in FSE_, which wasn’t the case before. I clearly see the point that supporting 5.9 and upwards is only supporting the “bleeding edge”. My suggestion is that if we can support 5.8 with very minor changes that will be easy to update in a few months, we can go that route. _This provided of course that the block will keep working in FSE_.

Else I’m totally fine with supporting 5.9 and upwards, as users who care for their WP installations will sooner or later get there. WordPress 6.0 release is in 24 May and we generally encourage people to upgrade, so 5.9 will not be that new in a couple of months.

acicovic:
I can confirm from my manual testing that this works fine in WP 5.9.3 (pages, posts, FSE). If we agree to stick with the WP 5.8 support and upwards, we can open this for review officially. Automated testing (as well as other block adjustments) will be addressed in different PRs once this gets merged.

src/blocks/recommendations/block.json
acicovic: Regarding the `ViewScript` issue, it seems to me that this should work for 5.8 now with [this commit](https://github.com/Parsely/wp-parsely/pull/762/commits/982a47cf6c785a88af2a91ca01fe217efc588362). WP 5.8 was released on 20 July 2021.

WP 5.9 and 5.8 account for about of 62% of WP installs, with 5.8 being around 11% ([source](https://wordpress.org/about/stats/)). 5.7 and 5.6 follow with a 6% and 3% respectively, so not a huge gain there.

pauarge: We have released the block on version 3.2.0 of the plugin under a feature flag, which gives us the flexibility of doing changes as such. I think Alex is making a very good point about the potential users of the block. 

I am going to argue that the full benefits of the block come with 5.9, so I don't want to commit now to support 5.8, which will probably bring us difficulties down the line. If we all agree, I would strongly push for supporting the block just on 5.9+ and having the least amount of "hacks".

pauarge:
@ryanwelcher The PR looks to me that's on the right path. Do you reckon we could move it out from draft?

acicovic:
> @ryanwelcher The PR looks to me that's on the right path. Do you reckon we could move it out from draft?

@pauarge Based on what you said, I think we can remove 5.8 compatibility code and move out of draft (it is I who had set the PR as draft).

src/blocks/recommendations/block.json
GaryJones: To me, this should be partly driven as a business decision, so much as a "what's easiest for us to maintain" decision.

If we make the whole plugin require WP 5.9, we're potentially cutting off current and potential customers from using it. Folks aren't going to do a whole WP core upgrade just to use a block from wp-parsely - the plugin doesn't have that power.

If we make just the block part require WP 5.9, while the rest of the plugin supports down to 5.0, then it's doable, but it needs clear documentation to set correct expectations.

acicovic: I do agree with the idea that there should be a "decision factor" besides development effort (though effort is definitively a big thing for us as a small team). In cases like this one, I think it should be based on data and goals to help us make correct decisions.

For example, with some statistics on the WP versions used by our users, and (let's say) an intention to keep compatibility with at least 70% of them, this could orient us. Without such rules and data, making correct decisions is harder. Trying to cover almost every user is always a great idea, but it can be costly in practice and limited in returns.

If backwards compatibility becomes problematic for us or for users that do update, I would go for serving the users who update. If we abort the idea of `block.json`, we'll have to find a solution for users who want to have FSE available. But keeping 5.8 compatibility is simple.

All this to say that there is a point in all arguments made here and that we could take this discussion further if needed.

GaryJones:
Seems there's a failing integration test for PHP 7.1 that needs reviewing.

pauarge:
> @GaryJones Seems there's a failing integration test for PHP 7.1 that needs reviewing.

Worth noting that PHP 7.1 is using WordPress 5.8, which might be causing the issue.

src/blocks/recommendations/components/parsely-recommendations.js
```diff
@@ -25,11 +25,23 @@ export default function ParselyRecommendations( {
 		state: { error, isLoaded, recommendations },
 	} = useRecommendationsStore();
 
+	function getErrorMessage() {
+		errorMessage = __( 'Error: ', 'wp-parsely' ) + JSON.stringify( error );
```
pauarge: Could we use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) for this? 

Also, `errorMessage` should be declared as a variable.

README.md
```diff
@@ -87,7 +87,9 @@ A sample `JSON-LD` meta tag and structured data for an article or post:
 
 ## The Parse.ly Recommendations Block (beta)
 
-The plugin includes a Recommendations Block that uses the [Parse.ly API](https://www.parse.ly/help/api/recommendations#get-related) to showcase links to content on your site. More information about enabling the Recommendations Block can be found in its [documentation](https://github.com/Parsely/wp-parsely/blob/trunk/src/blocks/recommendations/README.md). This feature is currently in beta and disabled by default.
+**IMPORTANT:** The Recommendations Block is available only in WordPress 5.9 and later.
```
pauarge, line 90:
```suggestion
The Recommendations Block is available only in WordPress 5.9 and later.
```

src/blocks/recommendations/README.md
```diff
@@ -1,18 +1,10 @@
 # Parse.ly Recommendations Block
 
-The Recommendations Block is designed to showcase links to content on your site as provided by the [Parse.ly `/related` API endpoint](https://www.parse.ly/help/api/recommendations#get-related).
-
-You can add it in `Posts`, `Pages`, or nearly any other custom post type. From WordPress 5.8 and up, the block can also be used as a [Block-based Widget](https://wordpress.org/support/article/block-based-widgets-editor/). Note however, that the Recommendations Block is not available in the Full Site Editor's template editor. This restriction will be lifted in the future.
-
-## Disabled by Default
+**IMPORTANT:** The Recommendations Block is available only in WordPress 5.9 and later.
```
pauarge, line 3:
```suggestion
The Recommendations Block is available only in WordPress 5.9 and later.
```

src/blocks/recommendations/class-recommendations-block.php
```diff
@@ -49,88 +44,19 @@ public function run(): void {
 	 *
 	 * @return void
```
pauarge, line 45: We should update the `@since` attribute because the function has been renamed.

README.md
acicovic: Resolving as agreed.

src/blocks/recommendations/README.md
acicovic: Resolving as agreed.

pauarge (approved):

src/blocks/recommendations/class-recommendations-block.php
acicovic: See https://github.com/Parsely/wp-parsely/pull/762/commits/985fdd2cc6604938216cb985a979b5362f077c44

PR #764, opened 2022-04-08
Recommendations Block: Remove feature flag

acicovic:
## Description
The Recommendations Block has been behind a feature flag. This PR removes the feature flag and related code.

## Motivation and Context
- The Recommendations Block should be stable enough to start being used without a feature flag.
- Preparation for merging #762 which fails in a test (this PR should solve this).

## How Has This Been Tested?
Automated tests pass.


src/Endpoints/class-related-api-proxy.php
```diff
@@ -52,10 +52,6 @@ public function __construct( Parsely $parsely, Proxy $proxy ) {
 	 * @return void
 	 */
 	public function run(): void {
-		if ( ! apply_filters( 'wp_parsely_recommendations_block_enabled', false ) ) {
-			return;
-		}
-
```
pauarge: We should keep a way to disable the REST API.

pauarge (changes requested):

src/Endpoints/class-related-api-proxy.php
acicovic: The Recommendations Block and the API proxy are bound together (no proxy, no recommendations) so in that case we should consider reverting and changing some things. Are there valid practical scenarios in which we would want to do this?

pauarge: The issue I see is exposing an endpoint that a site doesn't necessarily want. If they don't want the block, that's fine, they don't add it on an editor. However, if they have the plugin enabled, that endpoint will still be functional. We wanna provide a way to disable it.

You are correct that having one enabled with the other disabled might lead to unexpected behaviors. In that case, I would opt for leaving the feature flag present, but changing its value to `true` by default.

acicovic: This is now done in https://github.com/Parsely/wp-parsely/pull/764/commits/41432cfb145e63b9b8645dda4583b58dc467af20. I've changed the base to [try/use-block-json-for-block-registration](https://github.com/Parsely/wp-parsely/tree/try/use-block-json-for-block-registration) because this PR should need changes from there in order to work correctly.

pauarge (approved):

Issue #323, opened 2021-05-17
Consider a String class


GaryJones:
## Is your feature request related to a problem?

I noticed a few strings and string comparisons being done within various parts of the codebase and wondered if we could improve the code readability / maintainability.

## Describe the solution you'd like

I started experimenting with a new `Parsely\String` utility class:

```php
<?php

namespace Parsely;

class String {
	public static function is_true( $string ) {
		return 'true' === $string;
	}

	public static function is_false( $string ) {
		return 'false' === $string;
	}

	public static function is_true_or_false( $string ) {
		return self::is_true( $string ) || self::is_false( $string );
	}

	public static function is_not_true_or_false( $string ) {
		return ! self::is_true_or_false( $string );
	}

	public static function contains_a_space( $string ) {
		return strpos( $string, ' ' ) !== false;
	}

	public static function contains_a_period( $string ) {
		return strpos( $string, '.' ) !== false;
	}
}
```
```php
<?php

namespace Parsely\Tests;

use PHPUnit\Framework\TestCase;
use Parsely\String as ParselyString;

/**
 * Sample test case.
 *
 * @category   Class
 * @package    SampleTest
 */
final class StringTest extends TesCase {
	/**
	 * Data provider for test_string_is_true().
	 */
	protected function string_is_true_provider() {
		return array(
			array( 'true', true ),
			array( 'truex', false ),
			array( '.true', false ),
			array( 1, false ),
			array( 2, false ),
			array( -3, false ),
			array( array( 'true' ), false ),
		);
	}

	/**
	 * Test the string is identical to 'true'.
	 *
	 * @dataProvider data_string_is_true
	 *
	 * @param string $string   String to be tested.
	 * @param string $expected Expected assertion result.
	 */
	public function test_string_is_true( $string, $expected ) {
		if ( $expected ) {
			self::assertTrue( ParselyString::is_true( $string ) );
		} else {
			self::assertFalse( ParselyString::is_true( $string ) );
		}
	}

	/**
	 * Data provider for test_string_is_false().
	 */
	protected function string_is_false_provider() {
		return array(
			array( 'false', true ),
			array( 'falsex', false ),
			array( '.false', false ),
			array( 1, false ),
			array( 2, false ),
			array( -3, false ),
			array( array( 'false' ), false ),
		);
	}

	/**
	 * Test the string is identical to 'false'.
	 *
	 * @dataProvider data_string_is_false
	 *
	 * @param string $string   String to be tested.
[...]

PR #766, opened 2022-04-08
Settings page: Improve Site ID handling

### LINKED ISSUE #763: Settings page: Improve site ID validation/sanitization
@ryanwelcher reported that it is possible to insert `https://` in front of the site ID, as well as trailing `/` in the plugin's settings. This passes validation but it's not a valid site ID format. And although we do have the help text below, it can happen that users don't notice.

- Add `yourdomain.com` as a placeholder to the site ID textbox
- Possibly enhance the help text to make it more explicit
- Look out for, or automatically remove `http://`, `https://`, `www.` and possible trailing `/`. I'm currently checking with Parse.ly to see if any of the prefixes could be valid for a site ID.

acicovic:
## Description
Our settings page was allowing to save certain invalid Site IDs (for example strings containing `http(s)://` or trailing slashes) which in turn resulted into problems. Also, the help text's wording had some room for improvement.

This PR improves the help text as to be as precise as possible, but also imposes a bigger restriction on what is considered as a valid Site ID. It does accept a `www` prefix that is used in certain keys, and can accept strings with up to 3 `.` characters.

For example:
- `test.com`: valid
- `test.com/`: **invalid**
- `http://test.com`: **invalid**
- `https://test.com`: **invalid**
- `www.test.com`: valid
- `subdomain.test.com`: valid
- `www.subdomain.test.com`: valid
- `subdomain.subdomain.test.com`: valid
- `www.subdomain.subdomain.test.com`: **invalid**

## Motivation and Context
- Be clearer on how we expect Site IDs to be.
- Avoid problems by further limiting the possibility of accepting invalid site IDs.
- Fix #763.

## How Has This Been Tested?
- Manual testing.
- Unit tests created.


src/UI/class-settings-page.php
```diff
@@ -909,9 +910,9 @@ public function validate_options( array $input ): array {
 				__( 'Please specify the Site ID', 'wp-parsely' )
 			);
 		} else {
-			$input['apikey'] = strtolower( $input['apikey'] );
-			$input['apikey'] = sanitize_text_field( $input['apikey'] );
-			if ( strpos( $input['apikey'], '.' ) === false || strpos( $input['apikey'], ' ' ) !== false ) {
+			$api_key_format  = '/^((\w+)\.)?(([\w-]+)?)(\.[\w-]+){1,2}$/';
+			$input['apikey'] = strtolower( sanitize_text_field( $input['apikey'] ) );
+			if ( 1 !== preg_match( $api_key_format, $input['apikey'] ) ) {
```
pauarge: Can we extract the check of a correct API key to a separate function?

pauarge (changes requested):

GaryJones (dismissed):
This is going to need unit tests for the validation of the string.

src/UI/class-settings-page.php
```diff
@@ -310,14 +310,15 @@ private function initialize_basic_section(): void {
 		);
 
 		// Get the API Key.
-		$h          = __( 'Your Site ID is your own site domain (e.g. `mydomain.com`).', 'wp-parsely' );
+		$h          = __( 'Your Site ID is your own site domain without <code>http(s)://</code> prefixes or trailing <code>/</code> (e.g. <code>mydomain.com</code>).', 'wp-parsely' );
```
GaryJones:
```suggestion
		$h          = __( 'Your Site ID is typically your own site domain without <code>http(s)://</code> prefixes or trailing <code>/</code> (e.g. <code>mydomain.com</code>).', 'wp-parsely' );
```

The string could actually be anything, so while the default is to use the domain, we can allow a little wiggle room by adding a "typically".

acicovic:
I know that [this](https://github.com/Parsely/wp-parsely/blob/0e3b1276425589c2e9249abf9d617c4f0a4f9b8d/tests/Unit/SettingsPage/ValidationsTest.php#L83-L97) is duplicated code. I have created an issue for this in #784.

pauarge (approved):

tests/Unit/SettingsPage/ValidationsTest.php
```diff
@@ -0,0 +1,98 @@
...
+	private static $settings_page;
+
+	/**
+	 * Runs the setup process before each test.
+	 */
+	public function set_up(): void {
+		parent::set_up();
+
+		self::$parsely       = new Parsely();
+		self::$settings_page = new Settings_Page( self::$parsely );
```
GaryJones, line 42: Instantiating a class in the `UI` namespace, to test logic around a string, seems like too much coupling going on. 

Seems like perhaps the validation of the string should be handled by a validation-specific class, outside of the `UI` namespace, which the class in the `UI` namespace can then have injected, and defer to.

acicovic: @GaryJones, thank you. You definitely have a point here. Related to #323 and #485.

acicovic: I've put a mention of this comment in the above issues. Marking this conversation as resolved, as I think that changing this is part of a bigger chunk of work outside of this PR's scope.

PR #767, opened 2022-04-08
Recommedations Block: Disable link clicking in Editor

acicovic:
## Description
We realized that when in the Editor, the ability to click on the links (and be taken to their respective page) in the Recommendations Block might become a user annoyance because of accidental clicking or not understanding how the Block works.

For this reason, this PR removes the ability to click the links and be taken to the link address when we are in the Editor's environment.

## Motivation and Context
Offer a better experience for our users.

## How Has This Been Tested?
Manual testing.


pauarge (approved):
LGTM

GaryJones:
How would this work with screen readers? If a screen reader user tabbed their way into a link, and hit enter, would the link still activate?

Instead of just visually disabling the link with CSS, could we try programmatically disabling it, by removing /emptying the `href` attribute on the link instead?

acicovic:
@GaryJones, would you consider [this](https://github.com/Parsely/wp-parsely/pull/767/commits/5775ba04474ac90cae621f3cea27009315702fba) as an appropriate solution?

When we are in the editor, It changes all URLs to `#` when fetching the data.

src/blocks/recommendations/editor.scss
```diff
@@ -1,3 +1,9 @@
+// Fix wrong element height in FSE.
 .parsely-recommendations-card > div {
 	height: unset;
 }
+
+// Disable clicking on links.
+.parsely-recommendations-link {
+	pointer-events: none;
+}
```
pauarge, line 9: If we're going with the solution of having `#` as a link, we no longer need the CSS changes right?

acicovic: I thought about this, but I decided I preferred to leave it there, as clicking on `#` could still move the viewport which would be an annoyance. So I do think that keeping it makes sense.

PR #783, opened 2022-04-12
Getting rid of construct metadata usages

pauarge:
## Description

This PR continues on the metadata refactor by replacing the usages of the deprecated metadata generation methods by the new ones on `class-parsely.php`. We are also introducing changes to make calling to metadata classes more consistent.

## Motivation and Context

See #682 

## How Has This Been Tested?

The output of the metadata fields is consistent.

### DIFF
--- a/src/Endpoints/class-metadata-endpoint.php
+++ b/src/Endpoints/class-metadata-endpoint.php
@@ -64,10 +64,10 @@ abstract class Metadata_Endpoint {
 	 * @return string The metadata as HTML code.
 	 */
 	public function get_rendered_meta( string $meta_type ): string {
-		$metadata = new Metadata_Renderer( $this->parsely );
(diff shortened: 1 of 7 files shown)


acicovic (commented):
In the changes, I see that we're using `$metadata`/`meta` and `metadata_obj`/`metadata` in order to differentiate between the object and the metadata itself. If we're interested (and the object is needed only there), this could be converted into one-liners. This is untested, but something like:

```php
$metadata = ( new Metadata( $this ) )->construct_metadata( $parsely_options, $post );
```

Else, if achievable I would try to make naming as consistent as possible (one name for the metadata and another for the object) across all functions.

GaryJones:
The object could be called a builder, and we could use the [Builder pattern](https://refactoring.guru/design-patterns/builder/php/example).

acicovic (approved):

PR #789, opened 2022-04-13
Adding e2e tests for metadata

pauarge:
## Description

In this pull request we are adding end-to-end tests that check for the correct metadata insertion in the WordPress front-end.

## Motivation and Context

See #682 

## How Has This Been Tested?

Tests run consistently on CI and locally. No changes to the source code have been introduced.


acicovic (dismissed):
As a bonus, we could also check for the non-existence of metas when in LD JSON and the inverse.

acicovic (approved):

PR #790, opened 2022-04-14
Improving type annotations on metadata

pauarge:
## Description

This PR adds more detail on some time annotations in the metadata class, derived from the static analysis.

## Motivation and Context

See #682 

## How Has This Been Tested?

No significant code changes.

### DIFF
--- a/src/class-metadata.php
+++ b/src/class-metadata.php
@@ -323,9 +323,9 @@ class Metadata {
 		 *
 		 * @since 1.8.0
 		 *
-		 * @param string  $category        Category name.
-		 * @param WP_Post $post_obj        Post object.
-		 * @param array<string, mixed>   $parsely_options The Parsely options.
+		 * @param string  $category    Category name.
+		 * @param WP_Post $post_obj    Post object.
+		 * @param array<string, mixed> $parsely_options The Parsely options.
 		 */
 		$category_name = apply_filters( 'wp_parsely_post_category', $category_name, $post_obj, $parsely_options );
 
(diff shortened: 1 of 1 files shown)


src/class-metadata.php
```diff
@@ -325,7 +325,7 @@ private function get_category_name( WP_Post $post_obj, array $parsely_options ):
 		 *
 		 * @param string  $category        Category name.
 		 * @param WP_Post $post_obj        Post object.
-		 * @param array   $parsely_options The Parsely options.
+		 * @param array<string, mixed>   $parsely_options The Parsely options.
```
acicovic: Spacing can probably be reduced here (1 space instead of 3 in the last line, and align the above items.

src/class-metadata.php
```diff
@@ -338,7 +338,7 @@ private function get_category_name( WP_Post $post_obj, array $parsely_options ):
 	 * @since 3.3.0 Moved to class-metadata
 	 *
 	 * @param int $post_id The id of the post you're trying to get tags for.
-	 * @return array The tags of the post represented by the post id.
+	 * @return array<string> The tags of the post represented by the post id.
```
acicovic: Seems that the first line could use some more spaces to align the 2 "The".

src/class-metadata.php
```diff
@@ -517,7 +517,8 @@ private function get_coauthor_names( int $post_id ): array {
 					if ( ! empty( $post_author ) ) {
 						$coauthors[] = $post_author;
 					}
-				} // the empty else case is because if we force guest authors, we don't ever care what value wp_posts.post_author has.
+				}
+				// The empty else case is because if we force guest authors, we don't ever care what value wp_posts.post_author has.
```
acicovic, line 521: I think I understand what that comment means, but I find it a bit confusing. If it needs to be there, maybe there's a way to simplify it or make it more clear?

pauarge: I just reformatted the line, not changed the comment itself. I agree it could be improved, but I think this is out of scope for this PR.

acicovic (approved):

Issue #35, opened 2016-06-23
Add support for dynamic tracking?


jtotoole:
Once this is implemented we can roll back the changes made in this PR: https://github.com/Parsely/wp-parsely/pull/36

acicovic:
This is a work in progress and scheduled for [3.3.0](https://github.com/Parsely/wp-parsely/milestone/29).

PR #791, opened 2022-04-19
Removing Dynamic Tracking references

pauarge:
## Description

In preparation for the upcoming work on Dynamic Tracking, we are removing the references in the settings page to the plugin not supporting it. At this stage, this comment was already misleading. 

We are also removing a chunk of dead code -- the `print_dynamic_tracking_note` and `print_multiple_checkboxes` functions.

## Motivation and Context

See #35 

## How Has This Been Tested?

See screenshots below.

*Before*

Screen Shot 2022-04-19 at 9 10 19 AM

*After*

Screen Shot 2022-04-19 at 9 10 02 AM

### DIFF
--- a/src/UI/class-settings-page.php
+++ b/src/UI/class-settings-page.php
@@ -229,6 +229,10 @@ final class Settings_Page {
 	 */
 	public function add_help_text(): void {
 		$screen = get_current_screen();
+		if ( null === $screen ) {
+			return;
+		}
+
 		$screen->add_help_tab(
 			array(
 				'id'      => 'overview',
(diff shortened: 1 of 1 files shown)


src/UI/class-settings-page.php
```diff
@@ -229,13 +229,16 @@ public function screen_settings( string $screen_settings, WP_Screen $screen ): s
 	 */
 	public function add_help_text(): void {
 		$screen = get_current_screen();
+		if ( null === $screen ) {
+			return;
+		}
```
acicovic, line 234: Question: I see in docs that, `get_current_screen()` can return null. Can you explain the thinking behind removing this?

pauarge: Yes, that method can return null. That's why we're adding this check. Otherwise, the subsequent calls to `add_help_tab` would fail if `$screen` is null.

acicovic: Sorry, ignore my comment. For some reason I thought we're removing this instead of adding it.

acicovic (approved):

src/UI/class-settings-page.php
GaryJones: FWIW, WP core doesn't have this check: https://github.com/WordPress/WordPress/blob/270f2011f8ec7265c3f4ddce39c77ef5b496ed1c/wp-admin/index.php#L37-L39

`set_current_screen()` is also called right before the `load-*` actions anyway, so the global `$current_screen` returned from `get_current_screen()` would definitely be set: 
https://github.com/WordPress/WordPress/blob/270f2011f8ec7265c3f4ddce39c77ef5b496ed1c/wp-admin/admin.php#L212-L237

PR #792, opened 2022-04-19
Adding disable autotrack option for dynamic tracking

### LINKED ISSUE #35: Add support for dynamic tracking?

pauarge:
## Description

By default, the Parse.ly JavaScript tracker sends page view events automatically. This behavior can be disabled by passing the `autotrack: false` flag to the tracker. This is a feature that's often used when users want to have complete control over the tracking, mainly to implement Dynamic Tracking.

In this PR, we are adding this option as a toggle in the wp-admin plugin's settings page. When it is enabled, the `autotrack: false` JS code will be automatically inserted in the site's front-end. We have left this under the advanced settings, since it's a behavior that also requires some coding on the user's side.

## Motivation and Context

Closes #35 

## How Has This Been Tested?

Enable and disable the toggle. Check that when enabled, a new script appears in the front-end with the autotrack feature disabled.

Screen Shot 2022-04-20 at 9 14 59 AM


src/UI/class-settings-page.php
```diff
@@ -941,6 +956,16 @@ public function validate_options( array $input ): array {
 			$input['disable_javascript'] = 'true' === $input['disable_javascript'];
 		}
 
+		if ( 'true' !== $input['disable_autotrack'] && 'false' !== $input['disable_autotrack'] ) {
+			add_settings_error(
+				Parsely::OPTIONS_KEY,
+				'disable_autotrack',
+				__( 'Value passed for disable_autotrack must be either "true" or "false".', 'wp-parsely' )
```
GaryJones: The visible option are `Yes` and `No`, so having the value set as true/false, and then reporting back on that, may be confusing for the user. Make the labels better match the values, and then update the validation message.

pauarge:
We can ignore the SonarCloud alerts for now. They are about duplicated code. This is not ideal, but it's fine at the moment, since that duplicate code is us adding fields to the settings page.

src/class-scripts.php
```diff
@@ -124,6 +124,11 @@ public function enqueue_js_tracker(): void {
 			$js_api_key = "window.wpParselyApiKey = '" . esc_js( $this->parsely->get_api_key() ) . "';";
 			wp_add_inline_script( 'wp-parsely-loader', $js_api_key, 'before' );
 		}
+
+		if ( isset( $parsely_options['disable_autotrack'] ) && $parsely_options['disable_autotrack'] ) {
```
acicovic: Would it be better to have something like `&& true === && $parsely_options['disable_autotrack']` here? (haven't checked whether we need a boolean or string).

src/js/lib/loader.js
```diff
@@ -10,25 +10,29 @@ export function wpParselyInitCustom() {
 	 */
 	const customOnLoad = () => window.wpParselyHooks.doAction( 'wpParselyOnLoad' );
 
+	// Constructing window.PARSELY object.
```
acicovic:
```suggestion
	// Construct window.PARSELY object.
```

src/js/lib/loader.js
```diff
@@ -10,25 +10,29 @@ export function wpParselyInitCustom() {
...
+		window.PARSELY = {
+			onload: customOnLoad,
 		};
-		return;
 	}
 
-	window.PARSELY = {
-		onload: customOnLoad,
-	};
+	// Disabling autotrack if it was defined as such from PHP.
```
acicovic:
```suggestion
	// Disable autotrack if it was set as such in settings.
```

tests/Integration/ScriptsTest.php
```diff
@@ -169,6 +169,44 @@ public function test_enqueue_js_tracker(): void {
 		self::assertEquals( 1, count( $wp_scripts->registered['wp-parsely-loader']->extra ) );
 	}
 
+	/**
+	 * Test the tracker script enqueue.
```
```suggestion
	 * Tests the tracker script enqueue.
```

src/class-scripts.php
```diff
@@ -124,6 +124,11 @@ public function enqueue_js_tracker(): void {
...
+			$should_load_autotrack = 'window.wpParselyDisableAutotrack = true';
+			wp_add_inline_script( 'wp-parsely-loader', $should_load_autotrack, 'before' );
```
acicovic: The `$should_load_autotrack` naming is a bit contrary to what the variable actually does. It would perhaps be better to name it `$should_disable_autotrack` or even `$disable_autotrack` for brevity.

acicovic (commented):
Great work!

I've left some feedback about minor things.

acicovic (approved):

PR #793, opened 2022-04-19
Recommendations Block: Fix thumbnail images not showing

acicovic:
## Description
The Related API Proxy did not pass thumbnail image paths to the Recommendations Block, which resulted in thumbnails never being shown within the Block. This PR fixes the issue.

## Motivation and Context
User should be able to select thumbnail as an option in the Block.

## How Has This Been Tested?
- Manual testing (data inspection and verifying that thumbnails appear in the block).
- Updated integration test to pass the additional data.


pauarge (approved):
Nice catch!

PR #798, opened 2022-04-20
Make certain functions accessible to all types of PHPUnit tests

### LINKED ISSUE #784: Make getMethod() and setPrivateProperty() functions accessible to unit tests
The [getMethod() and setPrivateProperty() functions](https://github.com/Parsely/wp-parsely/blob/dbe094b4673f504e7f69a7c7676006607ad6db07/tests/Integration/TestCase.php#L151-L181) can also come in handy for unit tests, but calling them from where they are currently implemented results in errors.

Extract those functions somewhere where they are equally accessible by integration and unit tests.

## Additional context
I've stumbled upon this while writing unit tests for #766.

acicovic:
## Description
We were having some duplicate reflection code in our integration and unit tests. This PR:
- Unifies the code into a `trait`.
- Modifies the functions to be more in alignment with WordPress conventions.

## Motivation and Context
- Fix #784.
- This PR is also an experiment/attempt to use `traits`, in order to demonstrate the idea that they could be useful for us in other places as well.

## How Has This Been Tested?
Integration and unit tests still pass!


tests/Integration/bootstrap.php
```diff
@@ -45,7 +45,8 @@
 		exit( 1 );
 	}
 
-	// Include the Parsely custom test cases.
+	// Additional necessary requires.
+	require_once dirname( __DIR__ ) . '/trait-tests-trait.php';
 	require_once __DIR__ . '/TestCase.php';
 	require_once __DIR__ . '/StructuredData/NonPostTestCase.php';
```
pauarge: Why we're surrounding this with `dirname` and not the other requires?

tests/trait-tests-trait.php
```diff
@@ -0,0 +1,48 @@
+<?php
```
pauarge: This file's name looks wrong to me (having `trait` repeated twice). For instance, our `Parsely` class lives in `class-parsely.php` but the class is not named `Parsely_Class`. Maybe we could rethink how we name the class, and then rename the file accordingly?

pauarge (changes requested):
There is some room for improvement, but great usage of traits!

tests/trait-tests-trait.php
acicovic: I've named it like that because I've tried to do what I've seen in general and bridge it with WordPress conventions. Maybe you would prefer `trait-tests.php` and rename the trait to `Trait_Tests`? `Tests` by itself is not semantic enough, so in that case we should go for another name altogether.

It's a matter of how we name things and I have no strong opinion on this. Regarding traits, this is what we have so far (from https://make.wordpress.org/core/2020/03/20/updating-the-coding-standards-for-modern-php/):

>  Trait and interface names should use capitalized words separated by underscores, same as class names. Any acronyms should be all upper case.
> - The word “Trait” is not required in a trait name, but won’t be forbidden either.
> - Trait file names should be based on the trait name with trait- prepended and the underscores in the trait name replaced with hyphens.

tests/Integration/bootstrap.php
acicovic: Because the trait's php file is in the parent directory, while the other two aren't.

acicovic:
Regarding SonarCloud objections, maybe we should add some ignore rules there?

tests/trait-tests-trait.php
pauarge: This trait is not a general trait (for now). What it's doing is providing functions to change the visibility of methods inside a class. What about we rename it so something like `Tests_Visiblity` and then the file name would be `trait-tests-visibility.php`? 

Open to different names, but I don't know if the idea makes sense.

pauarge:
> Regarding SonarCloud objections, maybe we should add some ignore rules there?

I think we can mark the issues as `aknowledged`. In the specific case, it is telling us that we're using `setAccessible`, which is something that we shouldn't use in production code. However, this is fine in tests, especially when that functionality is the core of the trait.

acicovic:
> > Regarding SonarCloud objections, maybe we should add some ignore rules there? 

> I think we can mark the issues as `aknowledged`.

I looked a bit for it but cannot find anything that would allow me to "acknowledge". Maybe you have access to it?
Else, if we just ignored it, would it still appear in the future?

tests/trait-tests-trait.php
acicovic: I had opted for the name to be general (in order to avoid renames in the future), but this makes sense for now so I'll go with this.

pauarge:
> Maybe you have access to it?

You have to log in with your GitHub account. I've just marked them as `Won't fix`.

tests/trait-tests-trait.php
acicovic: I went ahead and renamed this to `Tests_Reflection`, as a placeholder for all reflection methods for tests.

pauarge (approved):

PR #807, opened 2022-04-21
Remove unused variables

pauarge:
## Description

This PR fixes a potential code smell detected by SonarCloud. It removes:

- Unused index from a `foreach` iteration on a function in the metadata generation.
- Unused variable on `NetworkAdminSitesListTest`.
- Simplifies a regex expression by ditching the `new RegExp` syntax.

## Motivation and Context

https://sonarcloud.io/project/issues?resolved=false&severities=MINOR&id=Parsely_wp-parsely&open=AX_erSgEBgpFtPR-wZ0j

## How Has This Been Tested?

Code works as expected as this variable was unused.


src/class-metadata.php
```diff
@@ -603,7 +603,7 @@ private function get_bottom_level_term( int $post_id, string $taxonomy_name ): s
 		$terms_not_parents = array_intersect_key( $terms, $term_ids_not_parents );
 		// remove array index keys.
 		$terms_not_parents_cleaned = array();
-		foreach ( $terms_not_parents as $index => $value ) {
+		foreach ( $terms_not_parents as $value ) {
 			$terms_not_parents_cleaned[] = $value;
 		}
```
GaryJones: If you're trying to simplify it from an associative array to an indexed array, then consider using [`array_values()` function](https://www.php.net/manual/en/function.array-values.php) instead of creating a foreach loop.

GaryJones: Otherwise, this looks like it's looping through one array and assigning them as terms to another, which seems redundant.

tests/Integration/UI/NetworkAdminSitesListTest.php
```diff
@@ -80,8 +80,10 @@ public function test_api_key_column_is_present(): void {
 	 * @return void
 	 */
 	public function test_api_key_column_is_correctly_printed(): void {
-		$blog_id_with_api_key    = $this->factory->blog->create();
-		$blog_id_without_api_key = $this->factory->blog->create();
+		$blog_id_with_api_key = $this->factory->blog->create();
+
+		// Creating a blog without an API key.
```
acicovic:
```suggestion
		// Create a blog without a Site ID.
```

tests/Integration/UI/NetworkAdminSitesListTest.php
```diff
@@ -80,8 +80,10 @@ public function test_api_key_column_is_present(): void {
 	 * @return void
 	 */
 	public function test_api_key_column_is_correctly_printed(): void {
-		$blog_id_with_api_key    = $this->factory->blog->create();
-		$blog_id_without_api_key = $this->factory->blog->create();
+		$blog_id_with_api_key = $this->factory->blog->create();
+
```
acicovic, line 84: I would remove this empty line (L84) so the related lines are together.

tests/Integration/UI/NetworkAdminSitesListTest.php
```diff
@@ -80,8 +80,10 @@ public function test_api_key_column_is_present(): void {
 	 * @return void
 	 */
 	public function test_api_key_column_is_correctly_printed(): void {
-		$blog_id_with_api_key    = $this->factory->blog->create();
-		$blog_id_without_api_key = $this->factory->blog->create();
+		$blog_id_with_api_key = $this->factory->blog->create();
```
acicovic, line 83: Should we consider renaming this to `$blog_id_with_site_id`?

tests/Integration/UI/NetworkAdminSitesListTest.php
pauarge: I don't necessarily agree. Since the second statement has a comment above, I aesthetically prefer to leave one line break before it.

tests/Integration/UI/NetworkAdminSitesListTest.php
pauarge: I see your point towards converging to Site ID. However, we're still using `apikey` in the DB representation, thus changing the variable name would make it more confusing.

acicovic (approved):

PR #821, opened 2022-04-22
Recommendations Block: Update documentation

### LINKED ISSUE #714: Recommendations Block: Update documentation
Our Recommendations Block documentation needs to be re-read and revised.

@GaryJones has left some related comments in #642. They are copy-pasted here along with some other notes.

Notes to consider when updating docs:
- [x] FSE experience is currently unstable (modify current related text, maybe add a known issues section)
- [x] Separate docs to user-facing and dev-facing (src/blocks/recommendations/README.md) - This README isn't going to be surfaced to anyone except developers looking through source code, so customer-facing information from here should go elsewhere and leave this file to be aimed at a developer audience.
- [ ] src/blocks/recommendations/README.md - Feels like some of these entries need more description (and not just "it gets passed to the API"), and maybe a list of available options. We can include links to more detailed documentation elsewhere to keep it canonical.

acicovic:
## Description
- This is a preparatory step (for proof-reading and verification) before adding this to the official Parse.ly documentation.
- We can then keep the README.md file as is, or delete it.

## Motivation and Context
- Keep documentation current.
- Resolve #714.

## How Has This Been Tested?
No code changes.


src/blocks/recommendations/README.md
```diff
@@ -1,44 +1,63 @@
 # Parse.ly Recommendations Block
 
-**IMPORTANT:** The Recommendations Block is available only in WordPress 5.9 and later.
+The Recommendations Block is designed to showcase links to content on your site as provided by the [Parse.ly `/related` API endpoint](https://www.parse.ly/help/api/recommendations#get-related). You can add it in Posts, Pages, or nearly any other custom post type. The Block can also be used in Full Site Editing (FSE) mode or as a [Block-based Widget](https://wordpress.org/support/article/block-based-widgets-editor/).
```
pauarge, line 3: Should we mention that this replaces the legacy widget?

src/blocks/recommendations/README.md
```diff
@@ -1,44 +1,63 @@
...
 
-The Recommendations Block is designed to showcase links to content on your site as provided by the [Parse.ly `/related` API endpoint](https://www.parse.ly/help/api/recommendations#get-related).
+## Requirements
 
-You can add it in `Posts`, `Pages`, or nearly any other custom post type. The Block can also be used in Full Site Editing (FSE) mode or as a [Block-based Widget](https://wordpress.org/support/article/block-based-widgets-editor/).
+The Block is available only in WordPress 5.9 and higher.
```
pauarge:
```suggestion
The Block is available only in WordPress 5.9 or later.
```

src/blocks/recommendations/README.md
```diff
@@ -1,44 +1,63 @@
...
+
+### Additional CSS class(es) (Advanced section)
+
+For styling purposes, you can specify one or more CSS classes that will be assigned to the block. 
+
+## FAQ
+
+### I can't find the Block.
+
+Ensure that you're running WordPress version 5.9.0 or higher, as the Block is unavailable in lower WordPress versions.
```
pauarge: Make sure they don't have the Proxy disabled?

pauarge (changes requested):
Looking good, left some minor improvements. Once this gets merged, let's delete it and put it up on Parse.ly's docs.

src/blocks/recommendations/README.md
acicovic: We could argue that the Block replaces the Widget, but currently for this to be true the following must apply:
- Running WP 5.9 or later.
- Not having disabled Block widgets through a plugin like Classic Widgets.

So the Block is a replacement, but not in all cases currently. We could add a mention like:
> The Block is targeted to completely replace the `Recommended Widget` which will be removed in wp-parsely version 4.0.0. If you're using the Widget, we highly recommend to consider the Block and start using it instead.

src/blocks/recommendations/README.md
acicovic: Good catch. I had thought about it but then forgot.

acicovic: When the proxy is disabled, the block itself will display: `The REST route is unavailable. To use it, wp_parsely_enable_related_api_proxy should be true.`.

I think that this should be enough, but let me know if you would also like to have it in the docs.

src/blocks/recommendations/README.md
acicovic: I've added a note about this.

src/blocks/recommendations/README.md
pauarge: I agree it's enough, no need to include it in the docs.

pauarge (approved):

PR #824, opened 2022-04-22
Fix SonarCloud's "errorMessage is already declared in the upper scope"

acicovic:
## Description
SonarCloud would complain with the message `'errorMessage' is already declared in the upper scope`. This PR attempts to fix this.

## Motivation and Context
Having less SonarCloud warnings and better code.

## How Has This Been Tested?
- Quick manual testing without site ID.
- Tests still pass (specifically the `block-errors.spec.js` test).


src/blocks/recommendations/components/parsely-recommendations.js
```diff
@@ -26,7 +26,7 @@ export default function ParselyRecommendations( {
 	} = useRecommendationsStore();
 
 	function getErrorMessage() {
-		let errorMessage = `${ __( 'Error:', 'wp-parsely' ) } ${ JSON.stringify( error ) }`;
+		errorMessage = `${ __( 'Error:', 'wp-parsely' ) } ${ JSON.stringify( error ) }`;
```
pauarge: This is indeed fixing the SonarCloud complaint, but making the problem worse. What we're trying to achieve here is to have two separate variables so we don't accidentally override values. By removing the `let`, we are indeed overriding whatever was defined in `errorMessage` in the scope above.

This should be solved by naming the variable on line 29 something different than `errorMessage`, so it's just used in the scope of the `getErrorMessage` function.

pauarge (changes requested):

pauarge (approved):

PR #825, opened 2022-04-24
Recommendations Block: Handle failed HTTP request

GaryJones:
## Description
When the Parse.ly Recommendations API was not accessible to the proxy endpoint, such as when developing offline or if the Parse.ly API was down, and recommendations were not able to be retrieved, then the block editor displayed:

> Error: {"errors":{"http_request_failed":["A valid URL was not provided."]},"error_data":[]}

The error message comes from WP_HTTP class in WordPress core, and in this case was slightly misleading, as the attempted URL was correct.

By adding some `errorMesssage` handling to the block's JavaScript, we can improve the message and avoid JSON being shown, so it reads as:

> The Parse.ly Recommendations API is not accessible. You may be offline.

_Note that this solution follows the existing model of handling errors, which means including all of the JSON in the `errorMessage.includes`, including a string in English. A site running in a different locale may not end up triggering the same error handling, so this should be checked and reviewed in a new issue/PR._

## Motivation and Context
Friendlier error handling for authors.

## How Has This Been Tested?
I was on an aeroplane with no internet connection and running a wp-parsely test site locally.

My test site before:
Screenshot 2022-04-23 at 17 01 23

My test site after (though I added one further word to the error message after this screenshot was taken):
Screenshot 2022-04-23 at 17 00 40


acicovic:
Good catch!

The code now also checks for errors with a `fetch_error` code. While writing a test for this with puppeteer, I stumbled upon this and integrated it for greater error coverage.

acicovic:
Was going to merge develop into this, but I have to go. If anyone wants to complete this, please go ahead.

pauarge (approved):

PR #827, opened 2022-04-24
Settings page: Remove media months query

GaryJones:
## Description
Disable a potentially expensive query on the settings page.

## Motivation and Context
The settings page calls `wp_enqueue_media()`, and this always runs a query to capture the month-year combinations that attachments were uploaded. This is only potentially useful when the logo is being selected, and even then, it's more likely that the user would upload a new logo or choose a recently uploaded one, than use the month-year filter to try and find an older upload (search by file name is also available anyway).

The filter hook to disable this query has a note that this query can be expensive for large media libraries, so let's disable this query by default. Since we control the page, and won't be calling `wp_enqueue_media()` a second time, there's no benefit to re-enabling the month-year query after `wp_enqueue_media()` has been called.

## How Has This Been Tested?
Run Query Monitor on the settings page, and filter to queries coming from the plugin. Apply the change, and see the query disappear.

The query in Query Monitor before:
Screenshot 2022-04-23 at 17 26 03

The populated month-year filter in the modal before:
Screenshot 2022-04-23 at 17 31 50

The populated month-year filter in the modal after:
Screenshot 2022-04-23 at 17 31 37


pauarge (approved):
🚢 🚢 🚢

acicovic (approved):

PR #836, opened 2022-04-27
Remove Parse.ly options as a parameter for Metadata generation

pauarge:
## Description

We keep moving towards improving our metadata generation. We will culminate this using a builder as described on the issue linked below. 

We aim to release the next version of the plugin with a common interface, in which the client code will pass a `$post` object and the function will return the metadata object. We want to keep this interface within the `Metadata` class for days to come. We will be calling the builder methods from within this class. 

In this PR, we are modifying that public interface. Until now, it has been required that users provide a `$parsely_options` object to generate the metadata. Since when constructing the `Metadata` class a `Parsely` instance is already provided, we can extract the plugin's configuration from there. Therefore, we are moving this redundant parameter.

## Motivation and Context

See https://github.com/Parsely/wp-parsely/issues/682.

## How Has This Been Tested?

Metadata generation works as it did before.

### DIFF
--- a/src/class-metadata.php
+++ b/src/class-metadata.php
@@ -37,14 +37,15 @@ class Metadata {
 	}
 
 	/**
-	 * Creates parsely metadata object from post metadata.
+	 * Creates Parse.ly metadata object from post metadata.
 	 *
-	 * @param array<string, mixed> $parsely_options parsely_options array.
-	 * @param WP_Post              $post object.
+	 * @param WP_Post $post object.
 	 *
 	 * @return array<string, mixed>
 	 */
-	public function construct_metadata( array $parsely_options, WP_Post $post ): array {
+	public function construct_metadata( WP_Post $post ): array {
+		$options = $this->parsely->get_options();
+
 		$parsely_page      = array(
 			'@context' => 'https://schema.org',
 			'@type'    => 'WebPage',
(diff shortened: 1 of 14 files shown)


acicovic (approved):

src/class-metadata.php
```diff
@@ -37,14 +37,15 @@ public function __construct( Parsely $parsely ) {
...
+	 * Creates Parse.ly metadata object from post metadata.
 	 *
-	 * @param array<string, mixed> $parsely_options parsely_options array.
-	 * @param WP_Post              $post object.
+	 * @param WP_Post $post object.
 	 *
 	 * @return array<string, mixed>
 	 */
-	public function construct_metadata( array $parsely_options, WP_Post $post ): array {
+	public function construct_metadata( WP_Post $post ): array {
```
GaryJones, line 46: This is a breaking change - it's changing the signature of a public method. Why is it in a 3.3.0 release? (unless this method was added after the 3.2 release?)

If you left the tests alone, would this still pass? I suspect it wouldn't. This is a good indicator that custom code in user's code would no longer work correctly.

pauarge: Yes, this method has been added on 3.3.0. Notice that we keep the public method for backwards compatibility: https://github.com/Parsely/wp-parsely/blob/5acb745bdb89d45205bdeb764653d49d9ab75631/src/class-parsely.php#L239

There is an issue to remove this method in 4.0.0

PR #837, opened 2022-04-27
Fixing coverage annotations in tests

pauarge:
## Description

We have had outdated PHPUnit documentation for a while, which hurts our coverage tests. This PR updates all comments to properly represent what the tests are covering. From now on, it is very important that every newly introduced PR checks that tests are correctly annotated.

## Motivation and Context

We want the coverage tests to provide relevant metrics.

## How Has This Been Tested?

See tests runs on GitHub Actions 8.0, where they run with coverage. No coverage warnings should appear.

## Screenshots (if appropriate)

We went from:

```
Classes: 12.50% (3/24)    
Methods: 19.40% (26/134)  
Lines:   10.74% (138/1285)
```

To:

```
Classes: 25.00% (6/24)    
Methods: 36.57% (49/134)  
Lines:   36.50% (469/1285)
```

Issue #730, opened 2022-03-18
Apply Inline Documentation Standards for our code's comments


acicovic:
## Is your feature request related to a problem?
Not all of our code's comments are in compliance with the [Inline Documentation Standards](https://developer.wordpress.org/coding-standards/inline-documentation-standards/). For example, the standard suggests that PHP function comments should be in _third person_, which is not always the case in our code.

## Describe the solution you'd like
Read [the standards](https://developer.wordpress.org/coding-standards/inline-documentation-standards/), see where we aren't in compliance with them and make any corrections needed.

acicovic:
This is a constant work in progress, and something we are taking care of in new code. I don't feel there's need for an issue to be associated to this work, so I'm closing this.

PR #838, opened 2022-04-28
Improve PHP comments conformity to Inline Documentation Standards

acicovic:
## Description
As described in #730, many of our comments do not conform to the [Inline Documentation Standards](https://developer.wordpress.org/coding-standards/inline-documentation-standards/). This PR is an attempt to quickly skim through the codebase and correct the most obvious comment issues.

Please note that:
- This is not a comment rewriting PR, where comments are rewritten to be better or provide more context.
- Changes to function comments that are in tests were avoided in order to avoid clashes with #837. Another PR will be made for tests.
- Minimal effort has been put in parts of the code that are deprecated or will be removed in version 4.0.0 of the plugin (Recommended Widget and some deprecated functions).

## Motivation and Context
- Solve as many comment issues as fast as possible.
- Related to #730.
- Related PRs for tests and JS will come in the future.

## How Has This Been Tested?
No code changes, automated tests pass.

### DIFF
--- a/src/Endpoints/class-related-api-proxy.php
+++ b/src/Endpoints/class-related-api-proxy.php
@@ -18,7 +18,7 @@ use WP_REST_Server;
 use WP_REST_Request;
 
 /**
- * Configure a REST API endpoint for use e.g. by the Recommendations Block.
+ * Configures a REST API endpoint for use e.g. by the Recommendations Block.
  */
 final class Related_API_Proxy {
 	/**
@@ -36,7 +36,7 @@ final class Related_API_Proxy {
 	private $proxy;
 
 	/**
-	 * Used to inject dependencies.
+	 * Constructor.
 	 *
 	 * @param Parsely $parsely Instance of Parsely class.
 	 * @param Proxy   $proxy   Proxy object which does the actual calls to the Parse.ly API.
@@ -47,7 +47,7 @@ final class Related_API_Proxy {
 	}
 
 	/**
-	 * Entrypoint to register the endpoint and otherwise initialize this class.
+	 * Registers the endpoint and initializes this class.
 	 *
 	 * @return void
 	 */
@@ -79,7 +79,7 @@ final class Related_API_Proxy {
 	}
 
 	/**
-	 * Determine if the endpoint can be called.
+	 * Determines if the endpoint can be called.
 	 *
 	 * @return bool
 	 */
--- a/src/Endpoints/class-related-api-proxy.php
+++ b/src/Endpoints/class-related-api-proxy.php
@@ -39,7 +39,8 @@ final class Related_API_Proxy {
 	 * Constructor.
 	 *
 	 * @param Parsely $parsely Instance of Parsely class.
-	 * @param Proxy   $proxy   Proxy object which does the actual calls to the Parse.ly API.
+	 * @param Proxy   $proxy   Proxy object which does the actual calls to the
+	 *                         Parse.ly API.
 	 */
 	public function __construct( Parsely $parsely, Proxy $proxy ) {
 		$this->parsely = $parsely;
--- a/src/Endpoints/class-related-api-proxy.php
+++ b/src/Endpoints/class-related-api-proxy.php
@@ -1,9 +1,9 @@
 <?php
 /**
- * Parsely Related REST API Endpoint
+ * Endpoints: Parse.ly `/related` API proxy endpoint class
  *
  * @package Parsely
- * @since 3.2.0
+ * @since   3.2.0
  */
 
 declare(strict_types=1);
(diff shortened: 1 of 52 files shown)


pauarge (changes requested):
Do we want to remove `return void` from phpdoc? What does WP core do?

acicovic:
> Do we want to remove `return void` from phpdoc? What does WP core do?

See the last bullet point [here](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#1-functions-class-methods). My understanding is that we should remove them (for void only). I think we should refer to this instead of inspecting core code, since the core itself might not be updated in certain parts.

pauarge:
I agree that we should look at the standards vs particular code. So yes, it looks like we should not use it. It's a bit counter intuitive to me, because if you are not putting it, it means that either you have forgot about it or it actually returns void. 

Having said that, I'd rather be consistent with the standards.

tests/Integration/Integrations/IntegrationsTest.php
```diff
@@ -69,8 +69,6 @@ function( $integrations ) {
 // phpcs:disable Generic.Files.OneObjectStructurePerFile.MultipleFound
 /**
  * Class FakeIntegration2
- *
- * @package Parsely\Tests\Integrations
```
pauarge, line 73: Maybe we shouldn't remove this one

wp-parsely.php
```diff
@@ -1,11 +1,10 @@
...
- * Parse.ly
+ * Parse.ly plugin loader
  *
- * @package      Parsely\wp-parsely
- * @author       Parse.ly
- * @copyright    2012 Parse.ly
- * @license      GPL-2.0-or-later
+ * Requires any necessary files and loads the plugin.
+ *
+ * @package Parsely
```
pauarge: Would this modifications have some effects on how WordPress.org represents our plugin?

src/Endpoints/class-related-api-proxy.php
```diff
@@ -36,20 +36,19 @@ final class Related_API_Proxy {
 	private $proxy;
 
 	/**
-	 * Used to inject dependencies.
+	 * Constructor.
 	 *
 	 * @param Parsely $parsely Instance of Parsely class.
-	 * @param Proxy   $proxy   Proxy object which does the actual calls to the Parse.ly API.
+	 * @param Proxy   $proxy   Proxy object which does the actual calls to the
+	 *                         Parse.ly API.
```
pauarge, line 43: Is there any line character limitation that we're following for this line breaks?

acicovic: [The standards](https://developer.wordpress.org/coding-standards/inline-documentation-standards/) state:

> DocBlock text should wrap to the next line after 80 characters of text. If the DocBlock itself is indented on the left 20 character positions, the wrap could occur at position 100, but should not extend beyond a total of 120 characters wide.

To conclude:

- I go for 80 chars anywhere possible.
- In some special cases I've gone for 100 and in one one case for 120, as the comments would take a lot of lines.
- In general, when we have long variable names and types (for example arrays), it's OK to go with 100 chars if this will save us an extra line.
- 120 chars is a last resort when having comments that will just take up too many lines.
- I also look at the code which is close to the comment. Having short lines of code with a really long comment doesn't look coherent.

Let me know if this answers the question.

pauarge: That makes sense. Again, I personally don't like this rule (given we're in the age of ultra hdpi and wide displays), but if that's the standard, happy to adhere to it.

wp-parsely.php
acicovic: Theoretically no, (although I suspect line 3 could be a potential problem) as WP looks at [header fields](https://developer.wordpress.org/plugins/plugin-basics/header-requirements/) and Packagist uses the [composer.json file](https://packagist.org/#how-to-submit-packages).

However, I've restored this just to be sure we're safe.

tests/Integration/Integrations/IntegrationsTest.php
acicovic: Can you tell me what's your reasoning for this?

In the beginning I did start subcategorizing tests and it quickly became involved and too detailed (I think this is why I rolled back). If we categorize a type of test, my reasoning is that we should probably do it for all of them for consistency.

Let me know of what you think.

acicovic: Just thinking out loud, but this could also change if we restructure our tests? So we may indeed want to look at it at that time?

pauarge: That makes sense, we can leave it as is.

pauarge (approved):

GaryJones:
Nice work @acicovic!

acicovic:
> Nice work @acicovic!

Thank you! 😃

PR #843, opened 2022-05-03
Refactor metadata generation to use builders

### LINKED ISSUE #682: Refactor metadata construction

pauarge:
## Description

This PR breaks down the metadata generation into separate classes. The current selects a builder class depending on what page we're currently in and instantiates that builder. Then, the builder's `get_metadata` method is called.

All builders inherit from a single abstract class. They must implement the `get_metadata` function, which should return the `$metadata` object after calling the necessary `build_*` function. Which functions to call depend on the actual page we're building metadata for.

Hat tip to @GaryJones for the initial design proposal.

## Motivation and Context

Closes #682


src/Meta/class-author-archive-builder.php
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	/**
+	 * Populates the `headline` field in the metadata object.
+	 *
+	 * @since 3.4.0
+	 */
+	private function build_headline(): void {
+		$author                     = ( get_query_var( 'author_name' ) ) ? get_user_by( 'slug', get_query_var( 'author_name' ) ) : get_userdata( get_query_var( 'author' ) );
+		$this->metadata['headline'] = $this->clean_value( 'Author - ' . $author->data->display_name );
```
GaryJones: Strings like this will need i18n with placeholders.

src/Meta/class-date-builder.php
```diff
@@ -0,0 +1,56 @@
...
+	private function build_headline(): void {
+		if ( is_year() ) {
+			/* translators: %s: Archive year */
+			$this->metadata['headline'] = sprintf( __( 'Yearly Archive - %s', 'wp-parsely' ), get_the_time( 'Y' ) );
+		} elseif ( is_month() ) {
+			/* translators: %s: Archive month, formatted as F, Y */
+			$this->metadata['headline'] = sprintf( __( 'Monthly Archive - %s', 'wp-parsely' ), get_the_time( 'F, Y' ) );
+		} elseif ( is_day() ) {
+			/* translators: %s: Archive day, formatted as F jS, Y */
+			$this->metadata['headline'] = sprintf( __( 'Daily Archive - %s', 'wp-parsely' ), get_the_time( 'F jS, Y' ) );
```
GaryJones: This is a very US-centric date format - so we should consider how we allow these formats to be amended? Perhaps we can also tap into the General Settings site date format?

src/Meta/class-metadata-builder.php
```diff
@@ -0,0 +1,136 @@
...
+declare(strict_types=1);
+
+namespace Parsely\Meta;
+
+use Parsely\Parsely;
+
+/**
+ * Abstract class that implements modular builders for Metadata. It should be
+ * extended in subclasses that implement a particular page view to render
+ * metadata for. Specifically, the `get_metadata` method should be implemented.
```
GaryJones:
```suggestion
 * Abstract class that implements modular builders for Metadata.
 *
 * It should be extended in subclasses that implement a particular page view to render
 * metadata for. Specifically, the `get_metadata()` method should be implemented.
```

src/Meta/class-post-builder.php
```diff
@@ -0,0 +1,571 @@
...
+			$lowercase_callback = 'strtolower';
+		}
+		if ( $options['lowercase_tags'] ) {
+			$tags = array_map( $lowercase_callback, $tags );
+		}
+
+		/**
+		 * Filters the post tags that are used as metadata keywords.
+		 *
+		 * @param string[] $tags Post tags.
```
GaryJones:
```suggestion
		 * @param array<string> $tags Post tags.
```

src/Meta/class-post-builder.php
```diff
@@ -0,0 +1,571 @@
...
+	/**
+	 * Returns the top-most category/taxonomy value in a hierarcy given a
+	 * taxonomy value's ID.
+	 *
+	 * (WordPress calls taxonomy values 'terms').
+	 *
+	 * @since 3.3.0 Moved to class-metadata
+	 *
+	 * @param int    $term_id The id of the top level term.
+	 * @param string $taxonomy_name The name of the taxonomy.
```
GaryJones:
```suggestion
	 * @param int    $term_id       The ID of the top level term.
	 * @param string $taxonomy_name The name of the taxonomy.
```

src/Meta/class-post-builder.php
```diff
@@ -0,0 +1,571 @@
...
+		 * @param WP_Post $post_obj    Post object.
+		 * @param array<string, mixed> $parsely_options The Parsely options.
+		 */
+		$category_name = apply_filters( 'wp_parsely_post_category', $category_name, $post_obj, $parsely_options );
+
+		return $this->clean_value( $category_name );
+	}
+
+	/**
+	 * Returns the top-most category/taxonomy value in a hierarcy given a
```
GaryJones:
```suggestion
	 * Returns the top-most category/taxonomy value in a hierarchy given a
```

src/Meta/class-post-builder.php
```diff
@@ -0,0 +1,571 @@
...
+	/**
+	 * Returns the bottom-most category/taxonomy value in a hierarchy given a
+	 * post ID.
+	 *
+	 * (WordPress calls taxonomy values 'terms').
+	 *
+	 * @since 3.3.0 Moved to class-metadata
+	 *
+	 * @param int    $post_id The post id you're interested in.
+	 * @param string $taxonomy_name The name of the taxonomy.
```
GaryJones:
```suggestion
	 * @param int    $post_id       The post id you're interested in.
	 * @param string $taxonomy_name The name of the taxonomy.
```

src/Meta/class-post-builder.php
```diff
@@ -0,0 +1,571 @@
...
+			// if you assign multiple child terms in a custom taxonomy, will only return the first.
+			return $terms_not_parents_cleaned[0]->name ?? '';
+		}
+
+		return '';
+	}
+
+	/**
+	 * Retrieves all the authors for a post as an array. Can include multiple
+	 * authors if coauthors plugin is in use.
```
GaryJones:
```suggestion
	 * authors if the Co-Authors Plus plugin is in use.
```

src/Meta/class-post-builder.php
```diff
@@ -0,0 +1,571 @@
...
+		// Getting the author name for each author.
+		$authors = array_map( array( $this, 'get_author_name' ), $authors );
+
+		/**
+		 * Filters the list of author names for a post.
+		 *
+		 * @since 1.14.0
+		 *
+		 * @param string[] $authors One or more author names.
+		 * @param WP_Post  $post    Post object.
```
GaryJones:
```suggestion
		 * @param array<string> $authors One or more author names.
		 * @param WP_Post       $post    Post object.
```

src/Meta/class-post-builder.php
```diff
@@ -0,0 +1,571 @@
...
+			}
+		}
+
+		/**
+		 * Filters the list of author WP_User objects for a post.
+		 *
+		 * @since 1.14.0
+		 *
+		 * @param WP_User[] $authors One or more authors as WP_User objects.
+		 * @param WP_Post   $post    Post object.
```
GaryJones:
```suggestion
		 * @param array<WP_User> $authors One or more authors as WP_User objects.
		 * @param WP_Post        $post    Post object.
```

src/Meta/class-post-builder.php
```diff
@@ -0,0 +1,571 @@
...
+	}
+
+	/**
+	 * Determines author name from display name, falling back to firstname
+	 * lastname, then nickname and finally the nicename.
+	 *
+	 * @since 3.3.0 Moved to class-metadata
+	 *
+	 * @param ?WP_User $author The author of the post.
+	 * @return string
```
GaryJones:
```suggestion
	 * @return string An author name.
```

src/Meta/class-post-builder.php
```diff
@@ -0,0 +1,571 @@
...
+	 * Returns a list of coauthors for a post assuming the Co-Authors Plus plugin
+	 * is installed.
+	 *
+	 * Borrowed from
+	 * https://github.com/Automattic/Co-Authors-Plus/blob/master/template-tags.php#L3-35
+	 *
+	 * @since 3.3.0 Moved to class-metadata
+	 *
+	 * @param int $post_id The id of the post.
+	 * @return array<WP_User>
```
GaryJones:
```suggestion
	 * @return array<WP_User> List of coauthors, or an empty array if the Co-Authors Plus plugin is not active.
```

src/Meta/class-post-builder.php
```diff
@@ -0,0 +1,571 @@
...
+		return $tags;
+	}
+
+	/**
+	 * Returns an array of all the child categories for the current post.
+	 *
+	 * @since 3.3.0 Moved to class-metadata
+	 *
+	 * @param int    $post_id The id of the post you're trying to get categories for.
+	 * @param string $delimiter What character will delimit the categories.
```
GaryJones:
```suggestion
	 * @param int    $post_id   The id of the post you're trying to get categories for.
	 * @param string $delimiter What character will delimit the categories.
```

src/Meta/class-post-builder.php
```diff
@@ -0,0 +1,571 @@
...
+	 */
+	private function get_categories( int $post_id, string $delimiter = '/' ): array {
+		$tags = array();
+		foreach ( get_the_category( $post_id ) as $category ) {
+			$hierarchy = get_category_parents( $category->term_id, false, $delimiter );
+			if ( ! is_wp_error( $hierarchy ) ) {
+				$tags[] = rtrim( $hierarchy, '/' );
+			}
+		}
+		// take last element in the hierarchy, a string representing the full parent->child tree,
```
GaryJones:
```suggestion
		// Take last element in the hierarchy, a string representing the full parent->child tree,
```

src/Meta/class-post-builder.php
```diff
@@ -0,0 +1,571 @@
...
+		return array_diff( $tags, array( $default_category_name ) );
+	}
+
+	/**
+	 * Gets all term values from custom taxonomies.
+	 *
+	 * @since 3.3.0 Moved to class-metadata
+	 *
+	 * @param WP_Post $post_obj The post object.
+	 * @return array<string>
```
GaryJones:
```suggestion
	 * Gets all term names from all custom taxonomies assigned to a post.
	 *
	 * @since 3.3.0 Moved to class-metadata
	 * @since 3.4.0 Moved to class-post-builder
	 *
	 * @param WP_Post $post_obj The post object to find the terms for.
	 * @return array<string> Term names.
```

GaryJones (dismissed):

src/Meta/class-date-builder.php
pauarge: I agree and it's a good point. I'd say that this would change the metadata output, which is not what we want with this PR. 

I've opened an issue to track this and change it separately: https://github.com/Parsely/wp-parsely/issues/850

src/Meta/class-author-archive-builder.php
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Author Archive Page Metadata Builder class
+ *
+ * @package Parsely
+ * @since 3.4.0
+ */
+
+declare(strict_types=1);
+
+namespace Parsely\Meta;
```
acicovic: Is there a specific reason for going with `Meta` instead of `Metadata` for the namespace?

src/Meta/class-metadata-builder.php
```diff
@@ -0,0 +1,137 @@
...
+	 * Populates the `url` field in the metadata object by getting the current page's URL.
+	 *
+	 * @since 3.4.0
+	 */
+	protected function build_url(): void {
+		$this->metadata['url'] = $this->get_current_url();
+	}
+
+	/**
+	 * Sanitize string content.
```
acicovic:
```suggestion
	 * Sanitizes string content.
```

src/Meta/class-metadata-builder.php
```diff
@@ -0,0 +1,137 @@
...
+	 * returns the value.
+	 *
+	 * @since 3.4.0
+	 *
+	 * @return array<string, mixed>
+	 */
+	abstract public function get_metadata(): array;
+
+	/**
+	 * Build basic metadata present in all pages.
```
acicovic:
```suggestion
	 * Builds basic metadata present in all pages.
```

src/Meta/class-date-builder.php
GaryJones: Resolving for this PR.

src/Meta/class-post-builder.php
```diff
@@ -0,0 +1,572 @@
...
+				$term_name = $this->get_bottom_level_term( $post_obj->ID, $parsely_options['custom_taxonomy_section'] );
+			}
+
+			if ( is_string( $term_name ) && 0 < strlen( $term_name ) ) {
+				$category_name = $term_name;
+			}
+		}
+
+		/**
+		 * Filters the constructed category name that are used as metadata keywords.
```
acicovic:
```suggestion
		 * Filters the constructed category name.
```

src/Meta/class-post-builder.php
```diff
@@ -0,0 +1,572 @@
...
+	/**
+	 * Returns a list of coauthors for a post assuming the Co-Authors Plus plugin
+	 * is installed.
+	 *
+	 * Borrowed from
+	 * https://github.com/Automattic/Co-Authors-Plus/blob/master/template-tags.php#L3-35
+	 *
+	 * @since 3.3.0 Moved to class-metadata
+	 *
+	 * @param int $post_id The id of the post.
```
acicovic:
```suggestion
	 * @param int $post_id The ID of the post.
```

src/Meta/class-post-builder.php
```diff
@@ -0,0 +1,572 @@
...
+
+		return '';
+	}
+
+	/**
+	 * Returns the tags associated with this page or post.
+	 *
+	 * @since 3.3.0 Moved to class-metadata
+	 *
+	 * @param int $post_id   The id of the post you're trying to get tags for.
```
acicovic:
```suggestion
	 * @param int $post_id   The ID of the post you're trying to get tags for.
```

src/Meta/class-post-builder.php
```diff
@@ -0,0 +1,572 @@
...
+		}
+		return $tags;
+	}
+
+	/**
+	 * Returns an array of all the child categories for the current post.
+	 *
+	 * @since 3.3.0 Moved to class-metadata
+	 *
+	 * @param int    $post_id   The id of the post you're trying to get categories for.
```
acicovic:
```suggestion
	 * @param int    $post_id   The ID of the post you're trying to get categories for.
```

src/Meta/class-post-builder.php
```diff
@@ -0,0 +1,572 @@
...
+	 * Gets all term names from all custom taxonomies assigned to a post.
+	 *
+	 * @since 3.3.0 Moved to class-metadata
+	 * @since 3.4.0 Moved to class-post-builder
+	 *
+	 * @param WP_Post $post_obj The post object to find the terms for.
+	 * @return array<string> Term names.
+	 */
+	private function get_custom_taxonomy_values( WP_Post $post_obj ): array {
+		// filter out default WordPress taxonomies.
```
acicovic:
```suggestion
		// Filter out default WordPress taxonomies.
```

src/Meta/class-post-builder.php
```diff
@@ -0,0 +1,572 @@
...
+
+		// Get array of IDs of terms which are not parents.
+		$term_ids_not_parents = array_diff( $term_ids, $parents );
+		// Get corresponding term objects, which are mapped to array index keys.
+		$terms_not_parents = array_intersect_key( $terms, $term_ids_not_parents );
+		// remove array index keys.
+		$terms_not_parents_cleaned = array_values( $terms_not_parents );
+
+		if ( ! empty( $terms_not_parents_cleaned ) ) {
+			// if you assign multiple child terms in a custom taxonomy, will only return the first.
```
acicovic:
```suggestion
			// If you assign multiple child terms in a custom taxonomy, will only return the first.
```

src/Meta/class-author-archive-builder.php
pauarge: We do have a `Metadata` class, so it might have created conflicts or at least confusion. I wanted to keep them separate without having to do many breaking changes to the existing public API.

What do you have in mind?

acicovic: I would have gone for naming this `Metadata`, but have no objections if you think that this improves code clarity or evades possible conflicts.

pauarge: I have renamed it to `Metadata`. It looks like it doesn't introduce any conflicts.

acicovic (approved):

PR #846, opened 2022-05-06
Preparation for release 3.3.1

pauarge:
Prepares `develop` for the 3.3.1 release. We have updated the CHANGELOG and bumped the version numbers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Parse.ly
 
-Stable tag: 3.3.0  
+Stable tag: 3.3.1
 Requires at least: 5.0  
 Tested up to: 5.9.3  
 Requires PHP: 7.1  
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Parse.ly
 
-Stable tag: 3.3.1
+Stable tag: 3.3.1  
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -1,6 +1,6 @@
 # Parse.ly
 
-Stable tag: 3.3.0  
+Stable tag: 3.3.1
```
acicovic:
```suggestion
Stable tag: 3.3.1  
```

pauarge: Actually, this was on purpose. Is there a reason to keep the whitespace?

acicovic: Yes, the lines below also do have it. It results in a line break. Else, the lines appear joined together.

acicovic: I do find a bad idea to use it though, as it is invisible and can be removed by mistake. I would prefer using `\` for this, but don't know why double space was used in the first place, and if replacing it could have side effects.

pauarge: I would be very careful with this, since it is also read by WordPress / WordPress.org.

acicovic (approved):

README.md
acicovic: Yes, this is why I haven't touched it 😃

PR #851, opened 2022-05-09
Add Dependency Review GitHub Action

### LINKED ISSUE #834: Consider using GitHub's dependency review action
We are not using GitHub's [dependency review enforcement action](https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-dependency-review#dependency-review-enforcement), but perhaps we would want to.

Decide if we would like to implement this and act accordingly.

## Additional context
- https://github.blog/2022-04-06-prevent-introduction-known-vulnerabilities-into-your-code/
- https://github.com/Automattic/vip-go-mu-plugins/pull/3042

pauarge:
## Description

This PR adds the Dependency Review GitHub action by GitHub, which checks pull requests for dependencies with known vulnerabilities.

Props to @sjinks for the original PR: https://github.com/Automattic/vip-go-mu-plugins/pull/3042

## Motivation and Context

Closes https://github.com/Parsely/wp-parsely/issues/834

https://github.blog/2022-04-06-prevent-introduction-known-vulnerabilities-into-your-code/

## How Has This Been Tested?

The action is correctly enqueued and executed on PRs.


acicovic (dismissed):

.github/workflows/dependency-review.yml
```diff
@@ -0,0 +1,19 @@
+name: Dependency Review
+
+on:
+  pull_request:
```
GaryJones, line 4: Like we limit it to paths [here](https://github.com/Parsely/wp-parsely/blob/693763741ec1a65a5bb8b1dcc40e83beb3d97a9a/.github/workflows/node.js.yml#L9-L12), can we limit this workflow from running unless `composer.json`, `package.json` and any other dependency-defining configuration gets updated? That would greatly cut down on the amount of time this had to run?

(Alternatively, maybe the `path` is a bad approach, because a semver constraint may not change, but one release that falls within that constraint may be affected, and it wouldn't get flagged.)

pauarge: I agree that limiting it two those files would greatly limit the number of executions, which is a good thing. I would tie it to `package-lock.json` too.

On the topic of versions, we now have a very flexible `composer.json`, as we don't lock to specific versions (something that we do on `package.json`). Maybe this action would be more effective in that latter case, but I think this is not a change that affects the PR itself.

acicovic (approved):

PR #864, opened 2022-05-10
Signal filter overrides on settings page

### LINKED ISSUE #713: Settings page: Consider signaling filter overrides
A recent discussion made me realize that certain features can be controlled both by the Settings page and WordPress filters.

An example of this is the `wp_parsely_load_js_tracker` filter (Disable JS tracking setting). When this filter is set, it overrides any setting in the Settings page. This is the correct behavior, but it brings the followup question of whether we should somehow signal this to the user. For example, a user might change a setting and expect it to work, but due to a filter in the code, this setting change might not do anything at all.

If the above **untested** hypothesis is correct, we could disable the setting in the settings page or/and provide a notification under the setting along the lines of "This setting is controlled by a filter in your code, and setting it here will have no effect". This way, the user would know that they have to check their code or notify their developer.

Leaving as is.

pauarge:
## Description

This PR adds a text that warns a user about a filter hook being defined somewhere in their code. We are adding this feature to help users debug behaviors in which those filters might override the setting defined in the UI.

## Motivation and Context

Closes #713 

## How Has This Been Tested?

Add some of the following filters in your code:

- `wp_parsely_load_js_tracker`
- `wp_parsely_trackable_statuses`
- `wp_parsely_metadata`

Then go to the settings page and you should see the notices under the appropriate settings.

Screen Shot 2022-05-11 at 11 27 57 AM

### DIFF
--- a/src/UI/class-settings-page.php
+++ b/src/UI/class-settings-page.php
@@ -645,10 +645,8 @@ Once you have changed a value and saved, please contact support@parsely.com to r
 	 */
 	private function print_filter_text( array $args ): void {
 		if ( isset( $args['filter'] ) && has_filter( $args['filter'] ) ) {
-			echo '<p><b>';
-			echo '<code>' . esc_html( $args['filter'] ) . '</code>';
-			echo esc_html__( 'filter hook is in use!', 'wp-parsely' );
-			echo '</b> ';
+			echo '<p>';
+			echo '<b><code>' . esc_html( $args['filter'] ) . '</code>' . esc_html__( 'filter hook is in use!', 'wp-parsely' ) . '</b>';
 			echo esc_html__( 'A callback is attached to the filter hook that might interfere and override this setting.', 'wp-parsely' );
 			echo '</p>';
 		}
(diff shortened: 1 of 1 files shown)


src/UI/class-settings-page.php
```diff
@@ -633,14 +634,27 @@ private function initialize_advanced_section(): void {
...
+	/**
+	 * Prints out a warning if the filter for the setting is defined, if any.
+	 *
+	 * @since 3.4.0
+	 *
+	 * @param array $args The arguments for the form field. May contain 'filter'.
+	 */
+	public function print_filter_text( array $args ): void {
+		if ( isset( $args['filter'] ) && has_filter( $args['filter'] ) ) {
+			echo '<p><b>The <code>' . esc_html( $args['filter'] ) . '</code> filter is defined!</b> You have defined it somewhere in your code, therefore it might interfere and override this setting.</p>';
```
GaryJones:
```suggestion
			echo '<p><b>The <code>' . esc_html( $args['filter'] ) . '</code> filter hook is in use!</b> A callback is attached to the filter hook that might interfere and override this setting.</p>';
```

acicovic (approved):
I'm thinking that maybe it would be nice to have tests for this.

pauarge:
> I'm thinking that maybe it would be nice to have tests for this.

It's quite hard to test this class as-is. We are currently not testing UI-based methods in integration tests for the settings page. E2E are not great either because this function requires a filter to be added. 

I'd say we investigate how can we improve testability of this class instead of trying to fix it here.

acicovic:
> I'd say we investigate how can we improve testability of this class instead of trying to fix it here.

I agree.

PR #866, opened 2022-05-11
Add debug information in the Site Health page

### LINKED ISSUE #638: Add a Site Health Status item if the API Key is not set
@GaryJones has proposed the idea of adding a Site Health Status item if the website API Key is not set.

Here are some additional considerations that I can think of:
- Should we consider adding an option that would disable the above warning? (in my experience some plugins use this to promote premium features or don't allow you to disable the warning, which can be frustrating).
- Is there any special work/adjustment that would be needed so this can work correctly on multisite setups?

pauarge:
## Description

This PR adds debugging tools into the Site Health page in wp-admin. In _Site Health Status_, it adds a check that passes if the Site ID is defined and fails critically otherwise. In _Site Health Info_, it provides a section that prints the plugin's options stored in the database (intended for support purposes).

## Motivation and Context

Closes #638

## How Has This Been Tested?

Remove the Site ID from your site and see the status check failing. See that it's not showing on the options printed in Info. Add it back again and see the check back to passing.

*Debug information*

Screen Shot 2022-05-12 at 11 23 27 AM

*Test passing*

Screen Shot 2022-05-12 at 11 22 43 AM

*Test failing* 

Screen Shot 2022-05-12 at 11 24 03 AM

### DIFF
--- a/src/UI/class-site-health.php
+++ b/src/UI/class-site-health.php
@@ -74,7 +74,7 @@ final class Site_Health {
 			if ( $this->parsely->api_key_is_missing() ) {
 				$result['status']  = 'critical';
 				$result['label']   = __( 'You need to provide the Site ID', 'wp-parsely' );
-				$result['actions'] = __( 'The site id can be provided in the <a href="/wp-admin/options-general.php?page=parsely">Parse.ly Settings Page</a>.', 'wp-parsely' );
+				$result['actions'] = __( 'The site ID can be set in the <a href="/wp-admin/options-general.php?page=parsely">Parse.ly Settings Page</a>.', 'wp-parsely' );
 			}
 
 			return $result;
--- a/src/UI/class-site-health.php
+++ b/src/UI/class-site-health.php
@@ -107,12 +107,10 @@ final class Site_Health {
 		);
 
 		foreach ( $options as $name => $value ) {
-			if ( $value ) {
-				$args['parsely']['fields'][ $name ] = array(
-					'label' => $name,
-					'value' => $value,
-				);
-			}
+			$args['parsely']['fields'][ $name ] = array(
+				'label' => $name,
+				'value' => $value,
+			);
 		}
 
 		return $args;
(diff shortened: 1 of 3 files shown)


acicovic:
Will this work correctly for multisite?
I'm thinking of subsites where the user might not necessarily desire Parse.ly, but would be nagged by the health notification anyway. In this case, we should introduce a setting so they can disable the health item.

pauarge:
> Will this work correctly for multisite?

Site Health is per individual site. There's no Site Health screen in the network admin. If you don't want to have the plugin in a site of your network, what you should do is disable the plugin for that specific site. In that case, the changes on this PR would not appear.

src/UI/class-site-health.php
```diff
@@ -0,0 +1,118 @@
...
+					'<p>%s</p>',
+					__( 'Your Site ID uniquely represents your site within Parse.ly. It is required for the tracking to work correctly.', 'wp-parsely' )
+				),
+				'test'        => 'loopback_requests',
+			);
+
+			if ( $this->parsely->api_key_is_missing() ) {
+				$result['status']  = 'critical';
+				$result['label']   = __( 'You need to provide the Site ID', 'wp-parsely' );
+				$result['actions'] = __( 'The site id can be provided in the <a href="/wp-admin/options-general.php?page=parsely">Parse.ly Settings Page</a>.', 'wp-parsely' );
```
acicovic:
```suggestion
				$result['actions'] = __( 'The site ID can be set in the <a href="/wp-admin/options-general.php?page=parsely">Parse.ly Settings Page</a>.', 'wp-parsely' );
```

tests/Integration/UI/SiteHealthTest.php
```diff
@@ -0,0 +1,66 @@
...
+final class SiteHealthTest extends TestCase {
+	/**
+	 * Internal variable.
+	 *
+	 * @var Site_Health $site_health Holds the Admin_Bar object
+	 */
+	private static $site_health;
+
+	/**
+	 * The setUp run before each test
```
acicovic:
```suggestion
	 * The setup run before each test.
```

acicovic (approved):

PR #876, opened 2022-05-16
Add Content Helper feature

pauarge:
## Description
This PR introduces a new feature called **Content Helper**, which is the main feature of the [3.4.0 milestone](https://github.com/Parsely/wp-parsely/milestone/34).

### How it works:
- The Content Helper adds a Parse.ly icon in the Gutenberg editor.
- Clicking the icon presents a Gutenberg sidebar, displaying the top performing posts based on the current post's `tag` or `category` (in order of priority). From the sidebar, it is also possible to visit the specific posts, or their respective Parse.ly Dash pages.
- If the current post does not have any tag or category assigned, the Content Helper displays the top posts by the current `author` (WP user).

### Things of note:
- This is an MVP and the first iteration on this feature (possible feature requests should be posted in the repo's issues).
- If the post contains multiple categories/tags, only the first category/tag (as returned by WordPress) is taken into account.
- Currently, only one filter (tag, category, author) is taken into account. That means that the Content Helper doesn’t support queries of `and` type (for example, author + tag).
- Posts are considered as "Top posts" based on their number of views.
- To get the list of posts, we use Parse.ly's `/analytics/posts` API endpoint.
- This PR also contains a related small refactoring to reduce code duplication.

## Motivation and Context
Start implementing tools within the Gutenberg editor that will be providing help and insights to our users, so they can get even more value from their Parse.ly integration.

## How Has This Been Tested?
At this point, manual testing in a local devenv only. Another PR adding tests will be be posted in the coming days. The reasoning behind this was to make the current PR available for review as soon as possible, since the intent is to release it next week if possible.

## Screenshot

image


GaryJones:
What's the goal of this new feature? What do we want editors to know or what actions do we went them to take? How does this help them with their content writing?

acicovic:
Hi,

The basic idea behind this is to present writers with similar posts that are the most successful for their website.

This way writers can use successful posts as an inspiration/reference while writing new posts or editing existing ones, and might be able to find patterns of what makes a post successful for their website. For those who haven't thought about this, it opens a new possibility. For writers that have thought of going through this manually, this gives them a shortcut and in-editor way to accomplish this.

src/Endpoints/class-base-api-proxy.php
```diff
@@ -0,0 +1,144 @@
...
+		$filter_key = trim( str_replace( '/', '_', $endpoint ), '_' );
+		if ( ! apply_filters( 'wp_parsely_enable_' . $filter_key . '_api_proxy', true ) ) {
+			return;
+		}
+
+		$get_items_args = array(
+			'query' => array(
+				'default'           => array(),
+				'sanitize_callback' => function ( $query ) {
+					// question: how should we sanitize these?
```
chriszarate: I know this is copypasta, but sanitization here should probably map over the array and call `sanitize_text_field` for each member

src/Endpoints/class-base-api-proxy.php
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	/**
+	 * Determines if there are enough permissions to call the endpoint.
+	 *
+	 * @return bool
+	 */
+	public function permission_callback(): bool {
+		// Unauthenticated.
+		return true;
```
chriszarate: Consider making this an abstract method, so that implementors do not accidentally inherit unauthenticated endpoints.

src/Endpoints/class-base-api-proxy.php
```diff
@@ -0,0 +1,144 @@
...
+				'error' => new WP_Error( 400, __( 'A Parse.ly API Key must be set in site options to use this endpoint', 'wp-parsely' ) ),
+			);
+		}
+
+		if ( true === $require_api_secret && false === $this->parsely->api_secret_is_set() ) {
+			return (object) array(
+				'data'  => array(),
+				'error' => new WP_Error( 400, __( 'A Parse.ly API Secret must be set in site options to use this endpoint', 'wp-parsely' ) ),
+			);
+		}
```
chriszarate, line 135: The code in this `if` block can never be reached

src/Endpoints/class-base-api-proxy.php
```diff
@@ -0,0 +1,144 @@
...
+
+		if ( null !== $param_item ) {
+			$params = $request->get_params()[ $param_item ];
```
chriszarate: use `get_param()`

https://developer.wordpress.org/reference/classes/wp_rest_request/get_param/

src/Endpoints/class-base-api-proxy.php
```diff
@@ -0,0 +1,144 @@
...
+		return (object) array( 'data' => $this->generate_data( $response ) );
+	}
+
+	/**
+	 * Generates the final data from the passed response.
+	 *
+	 * @param array<string, mixed> $response The response received by the proxy.
+	 * @return array<stdClass> The generated data.
+	 */
+	protected function generate_data( array $response ): array {
```
chriszarate: An abstract method seems useful here, too.

src/blocks/content-helper/components/post-card.tsx
```diff
@@ -0,0 +1,38 @@
...
+						<li>{ __( 'Published:', 'wp-parsely' ) } { post.date }</li>
+						<li>{ __( 'Author:', 'wp-parsely' ) } { post.author }</li>
+					</ul>
+					<p>
+						<Button href={ post.url } target="_blank" variant="primary">{ __( 'Open Post', 'wp-parsely' ) }</Button> { ' ' }
+						<Button href={ post.statsUrl } target="_blank" variant="secondary">{ __( 'Post Stats', 'wp-parsely' ) }</Button>
+					</p>
+				</CardBody>
+			</Card>
+			<br />
```
chriszarate, line 33: Is it possible to style this component and avoid literal line breaks?

src/blocks/content-helper/components/post-list.tsx
```diff
@@ -0,0 +1,55 @@
...
+	const [ posts, setPosts ] = useState<SuggestedPost[]>( [] );
+
+	const fetchPosts = async ( retries: number ) => {
+		ContentHelperProvider.getTopPosts()
+			.then( ( result ) => {
+				setPosts( result.posts );
+				setMessage( result.message );
+				setLoading( false );
+			} )
+			.catch( async ( err: string ) => {
```
chriszarate: Is `err` a `string`? You might want to access `err.message`

src/blocks/content-helper/components/post-list.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+const FETCH_RETRIES = 3;
+
+function PostList() {
+	const [ loading, setLoading ] = useState<boolean>( true );
+	const [ error, setError ] = useState<string>( null );
+	const [ message, setMessage ] = useState<string>( null );
+	const [ posts, setPosts ] = useState<SuggestedPost[]>( [] );
+
+	const fetchPosts = async ( retries: number ) => {
```
chriszarate: It's generally better practice to define this function inside `useEffect`:

1. You don't need it to be defined every render.
2. You avoid accidentally calling it outside the effect, which would cause hard-to-debug state issues.

src/blocks/content-helper/components/post-list.tsx
```diff
@@ -0,0 +1,55 @@
...
+	useEffect( () => {
+		setLoading( true );
+		fetchPosts( FETCH_RETRIES );
+	}, [] );
+
+	const body = error ? <p>{ error }</p> : posts.map( ( post ) => <PostCard key={ post.id } post={ post } /> );
+
+	return (
+		<>
+			<p>{ message }</p>
```
chriszarate, line 52: Should `message` be rendered even in an error state?

src/blocks/content-helper/content-helper-provider.ts
```diff
@@ -0,0 +1,118 @@
...
+	error?: string;
+	data?: SuggestedPost[];
+}
+
+class ContentHelperProvider {
+	static async getTopPosts(): Promise<GetTopPostsResult> {
+		const editor = select( 'core/editor' );
+
+		// Get post's author.
+		const currentPost = editor.getCurrentPost() as Post;
```
chriszarate, line 30: Which type is being overridden here?

src/blocks/content-helper/content-helper-provider.ts
```diff
@@ -0,0 +1,118 @@
...
+		const author = select( 'core' ).getEntityRecord( 'root', 'user', currentPost.author ) as User;
+
+		// Get post's first category.
+		const categoryId = editor.getEditedPostAttribute( 'categories' ) as Array<number>[0];
```
chriszarate: Isn't `Array<number>[0]` equivalent to `number`?

src/blocks/content-helper/content-helper-provider.ts
```diff
@@ -0,0 +1,118 @@
...
+		}
+
+		return response?.data || [];
+	}
+
+	private static processData( data: SuggestedPost[] ): SuggestedPost[] {
+		return data.map( ( p ) => {
+			// @ts-ignore
+			p.statsUrl = `${ window.wpParselyContentHelperPrefix }?url=${ p.url }`;
+			return p;
```
chriszarate: - What error is being ignored?
- Encode the URL

```suggestion
			const statsUrl = `${ window.wpParselyContentHelperPrefix }?url=${ window.encodeURIComponent( p.url ) }`;
			return Object.assign( p, { statsUrl } );
```

src/blocks/content-helper/content-helper-provider.ts
chriszarate: In general I think these `as` operators should be avoided. Work with the existing return type of these functions, inspect and validate the result, and perform type narrowing to arrive at the expected type.

src/blocks/content-helper/content-helper.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { Panel, PanelBody, PanelHeader } from '@wordpress/components';
+import { PluginSidebar } from '@wordpress/edit-post';
+import { registerPlugin } from '@wordpress/plugins';
+
+/**
+ * Internal dependencies
+ */
+import PostList from './components/post-list';
+// @ts-ignore
+import { ReactComponent as LeafIcon } from '../../img/parsely-logo.svg';
```
chriszarate: https://duncanleung.com/typescript-module-declearation-svg-img-assets/

src/Endpoints/class-base-api-proxy.php
acicovic: Done.

src/Endpoints/class-base-api-proxy.php
acicovic: Done.

src/Endpoints/class-base-api-proxy.php
acicovic: Done.

src/Endpoints/class-base-api-proxy.php
acicovic: Done.

src/Endpoints/class-base-api-proxy.php
acicovic: I've stepped through the code and it is reachable (notably when trying to use the Content Helper but having no secret set in settings). The problem that I noticed with this code (and the block above it) is that it was crashing the WordPress Editor without showing the error message when using the Content Helper.

I've applied a fix for this.

src/blocks/content-helper/components/post-list.tsx
acicovic: Yes. It can contain details about the error.

src/blocks/content-helper/components/post-card.tsx
acicovic: What's the issue we would be trying to avoid?

src/blocks/content-helper/components/post-list.tsx
acicovic: This seems correct to me.

src/blocks/content-helper/content-helper-provider.ts
acicovic: We're getting a "Property 'wpParselyContentHelperPrefix' does not exist on type 'Window & typeof globalThis" error.

src/Endpoints/class-base-api-proxy.php
chriszarate: My fault, I was conflating `api_key_is_set` and `api_secret_is_set`. Glad to see I accidentally uncovered a different problem! :)

src/blocks/content-helper/components/post-card.tsx
chriszarate: Line breaks are a semantic element and are imprecise at applying vertical spacing. It would be better to provide spacing via CSS.

acicovic: Oh, got it. Didn't realize you were talking about the `<br />`. Will look into that.

src/blocks/content-helper/components/post-list.tsx
chriszarate: Unless code somewhere is literally throwing a string (`throw 'a string'`), then `err` is not a string. In the end, via React, you are essentially coercing the error to a string. In other words, something like this is happening:

```
`${ new Error( 'leveraging .toString() method' ) }`
```

It's probably an `Error`, but this is one of those situations where JavaScript is very permissive and you need to do some type narrowing. 

https://kentcdodds.com/blog/get-a-catch-block-error-message-with-typescript

That's a lot of work for a little benefit, and it could be totally fine to coerce to string, but just explaining what's going on under the hood.

src/blocks/content-helper/components/post-list.tsx
chriszarate: How, since `setMessage` is only called in the `.then` block / happy path? If an error is set, it seems `message` will always be `null`, which results in an empty `<p/>` being rendered.

It seems we might benefit from an early return in an error case:

```
if ( error ) {
   return <p>{ error }</p>;
}
```

src/blocks/content-helper/content-helper-provider.ts
chriszarate: Ideally we could import types provided by Parse.ly's JavaScript library. If those don't exist, we can declare our own (alongside others, like for SVG imports).

https://www.jamestharpe.com/typescript-extend-window/

src/blocks/content-helper/components/post-list.tsx
acicovic: Thanks for the info, this is something to keep in mind for me.

In this case, [fetchData()](https://github.com/Parsely/wp-parsely/blob/3207db79d69e06711e1e4fc0c1eba9a61e802f34/src/blocks/content-helper/content-helper-provider.ts#L62-L79) returns an array of results or a string error.

Maybe not an ideal solution (?). For the time being I propose to leave as is and work on it in the future if you think it may be worth it?

acicovic: Upon further examination what I said above is partly true, because there's a case where an error is returned.

src/blocks/content-helper/content-helper-provider.ts
acicovic: The type returned is a `Record<string, any>`. In that specific case, using `as` simplifies the code and allows to get correct Intellisense quickly. I would go for keeping for the time being as it also should not have side effects as per the [docs](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions):

> Like a type annotation, type assertions are removed by the compiler and won’t affect the runtime behavior of your code.

Having said that, I do agree that manipulating the original type would be the best practice in general.

src/blocks/content-helper/components/post-list.tsx
acicovic: The errors we return are strings and they are presented within the message. However, if other errors were to occur what you say is correct. So I've implemented your suggestion.

src/blocks/content-helper/components/post-card.tsx
acicovic: I've tried this but Gutenberg somehow overrides the style and I haven't found a way around this yet. I propose to tackle this in another PR so we can go forward with the release.

src/blocks/content-helper/content-helper-provider.ts
chriszarate: Sounds good, we can revisit as we get more familiar with Gutenberg dev. :)

chriszarate (approved):
Looks good. Did you implement a guard to protect against `undefined` values in the data loader? Or will that come in a later PR?

acicovic:
> Looks good. Did you implement a guard to protect against `undefined` values in the data loader? Or will that come in a later PR?

[This](https://github.com/Parsely/wp-parsely/blob/f26ed34cfb74b90f48cb5532f160f5d2ee070681/src/blocks/content-helper/content-helper-provider.ts#L43-L45) acts as the guard, as [buildFetchDataQuery()](https://github.com/Parsely/wp-parsely/blob/f26ed34cfb74b90f48cb5532f160f5d2ee070681/src/blocks/content-helper/content-helper-provider.ts#L88-L96) returns with an error message and `null` query when all of the needed information is `undefined`.

acicovic (approved):

PR #892, opened 2022-05-19
Add Analytics Proxy and Endpoint for Content Helper (moved to #894)

acicovic:
Moved to #894

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -27,6 +27,7 @@ declare(strict_types=1);
 namespace Parsely;
 
 use Parsely\Endpoints\Related_API_Proxy;
+use Parsely\Endpoints\Analytics_API_Proxy;
 use Parsely\Endpoints\GraphQL_Metadata;
 use Parsely\Endpoints\Rest_Metadata;
 use Parsely\Integrations\Amp;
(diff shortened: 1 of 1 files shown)


src/RemoteAPI/class-base-proxy.php
```diff
@@ -63,7 +63,10 @@ public function get_api_url( array $query ): string {
 		}
 
 		$query['apikey'] = $this->parsely->get_api_key();
-		$query           = array_filter( $query );
+		if ( $this->parsely->api_secret_is_set() ) {
+			$query['api_secret'] = $this->parsely->get_api_secret();
+		}
```
pauarge: Do we want to unconditionally add the api secret? Even if it's defined, it might not be necessary in all the cases.

acicovic: Per https://www.parse.ly/help/api/endpoint#request

> this parameter is required for every request unless otherwise noted and should contain the secret token obtainable from the [API Settings](http://dash.parsely.com/to/settings/api) page.

However, in my experience, it was not required for the Recommendations Block.

I'd rather include it everywhere instead of needing it in an unknown edge case and not having it. I'm also thinking about Parse.ly updates that could change this behavior. However, if there's a good reason, we could change this during refactoring.

PR #894, opened 2022-05-19
Add Analytics Proxy and Endpoint for Content Helper

acicovic:
## Description
This PR introduces an `Analytics` Proxy and Endpoint, which hits the Parsely `/analytics` API. It is based on our existing code for the Parse.ly `/related` API. It currently returns the following fields:

```
author
pub_date
title
url
```

## Motivation and Context
This is based on PR #876. It is posted as a separate PR to make its review easier, but should be merged into #876 as soon as the Content Helper can accept data so we can see it in action and perform any adjustments. We can then also refactor this to reduce code duplication if needed.

## How Has This Been Tested?
This code is in draft status and totally untested, waiting to be merged into #876.

### DIFF
--- a/src/Endpoints/class-analytics-api-proxy.php
+++ b/src/Endpoints/class-analytics-api-proxy.php
@@ -1,0 +1,137 @@
+<?php
+/**
+ * Endpoints: Parse.ly `/analytics` API proxy endpoint class
+ *
+ * @package Parsely
+ * @since   3.4.0
+ */
+
+declare(strict_types=1);
+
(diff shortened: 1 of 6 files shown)


pauarge:
This looks good overall. Could we add some tests?

acicovic:
> This looks good overall. Could we add some tests?

We'll definitely add tests. The reason I haven't done so yet is that a possible refactoring could result in rewriting parts of tests. So I'd wait for any possible refactoring first.

Issue #471, opened 2021-10-27
Move Javascript code to TypeScript


pauarge:
## Is your feature request related to a problem?

Currently we don't type check our Javascript codebase (we do for the PHP side). This leads to more bug-prone code. 

## Describe the solution you'd like

We could move the Javascript files to TypeScript. Since we're using built code already, adding an additional step that transpiles TS to JS wouldn't be much of an issue.

## Describe alternatives you've considered

Not type check the JS code because it's relatively small, anyway. This might be valid now, but it the future it will probably grow on size.

jimmywarting:
you can have type safety without typescript syntax if you use jsdoc... most time you don't even have to annotate your code if you do it right

Imo typescript only slows down the development processes, you will just spend more time upfront.
...Deno removed typescript
...esm + typescript is no good at the moment, its a #developerPain

PR #895, opened 2022-05-20
Add TypeScript support

pauarge:
## Description

This PR adds TypeScript support into our build pipeline. It is compatible with the current JavaScript code, and we expect to have the two languages alongside for a bit.

## Motivation and Context

See #471 

## How Has This Been Tested?

No changes on built files.


GaryJones:
> This PR also partially migrates the Recommendations Block to TypeScript.

I would suggest separating this out into a new PR.

acicovic (approved):

PR #896, opened 2022-05-20
Remove legacy Recommended Widget

pauarge:
## Description

In this PR we are removing the legacy Recommended Widget. The widget wasn't widely used and some parts of it wouldn't work in certain scenarios. What's more, it has already been replaced by the Recommendations Block, which is what we recommend to transition to.

NOTE: This is currently tagged as version 3.4.0, but since it's a breaking change, we'll rename 3.4.0 to 4.0.0.

## Motivation and Context

Clear engineering debt.

## How Has This Been Tested?

The plugin functions as expected but the Recommended Widget is not present anymore.


acicovic:
I've posted an add-on to this PR in #899. Let's review and merge it here.

Issue #922, opened 2022-06-10
Fix remaining SonarCloud warnings


acicovic:
## Is your feature request related to a problem?
We still have some SonarCloud warnings. Some may be irrelevant to us (e.g. class naming warnings), but we should fix the rest if possible. 

## Describe the solution you'd like
Inspect the warnings, put the irrelevant warnings to an ignore list if possible, and fix the rest.

PR #923, opened 2022-06-11
Fix some SonarCloud issues

GaryJones:
## Description
Fixes a few SonarCloud issues.

I also marked some issues as wontfix, where it wouldn't help code readability.

## Motivation and Context
Having a base of 0 issues makes it easier to see when new ones get added.
See #922 

## How Has This Been Tested?
Ran unit and integration tests locally.

### DIFF
--- a/src/class-scripts.php
+++ b/src/class-scripts.php
@@ -137,7 +137,7 @@ class Scripts {
 	 */
 	public function script_loader_tag( string $tag, string $handle, string $src ): string {
 		$parsely_options = $this->parsely->get_options();
-		if ( in_array(
+		if ( \in_array(
 			$handle,
 			array(
 				'wp-parsely-loader',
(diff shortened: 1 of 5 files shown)


src/class-scripts.php
```diff
@@ -137,7 +137,7 @@ public function enqueue_js_tracker(): void {
 	 */
 	public function script_loader_tag( string $tag, string $handle, string $src ): string {
 		$parsely_options = $this->parsely->get_options();
-		if ( in_array(
+		if ( \in_array(
```
acicovic, line 140: Do you remember what kind of warning this resolves?

acicovic (approved):
❤️ Thank you so much for giving some time for this!

src/class-scripts.php
GaryJones: By forcing to use the global namespace, it enables some opcode optimizations. 

- https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/
- https://github.com/cakephp/cakephp/issues/12654
- [PHP master showing specially handled namespace functions](https://github.com/php/php-src/blob/0ae6a6755071c42964d979a81e7800819c9d2261/Zend/zend_compile.c#L4300-L4364)

There are a couple of tools around that will allow this to be added at compile time.

PR #985, opened 2022-07-16
Switch to Composer autoloader

### LINKED ISSUE #961: Use an autoloader for the plugin's files
Recently, we had to [fix a bug](https://github.com/Parsely/wp-parsely/pull/959) due to a forgotten `require`. @GaryJones, @chriszarate and I agree that moving to an autoloader is a good move. It will help prevent similar issues and allow for cleaner code.

Implement an autoloader solution for the plugin's files.

GaryJones:
## Description
Switch to Composer autoloader

Built plugin (i.e. a Zip on GitHub, or what is pushed to WordPress.org, will need to ensure that `composer install --no-dev -a` is run, so that Authoritative class maps are created.

## Motivation and Context
Fixes #961.

## How Has This Been Tested?
- Load the front and back ends with and without this patch.
- Tests pass.


GaryJones:
As noted in the description - we may want to add further enhancements/changes before we close the original issue.

acicovic:
@GaryJones, thank you very much for the contribution. Would you have any suspicion about why E2E tests fail the way they do?

GaryJones:
All failures seem to stop at `No node found for selector: #user_login` - but not sure why that is.

acicovic:
I went ahead in order to make tests pass and apply some autoload file optimizations for E2E tests and the deployment workflow. I think that as things stand, `composer install --no-dev -a` is not needed.

chriszarate (approved):
:+1:

PR #1005, opened 2022-08-03
Add init_content_helper() action

acicovic:
## Description
This PR adds an `init_content_helper()` action, allowing to disable the Content Helper by invoking `remove_action()`.

I'm aware that this could be considered a breaking change. However:
- The Content Helper has been out only for a short time.
- The previous action was not adequately named and it was located in the `Script` class (which made it all that harder to use and possibly having to write bad code to access it). Someone would have to go out of their way to find the action and try to use it.
- To my knowledge, 99.99% of cases don't have a reason to disable the Content Helper.
- We needed to introduce this feature as soon as possible, and creating a major release for this would be expensive but with minimal actual user impact.

As such, I've decided to integrate it into 3.5 which will be released very soon.

## Motivation and Context
We have a specific use case where it is desirable to disable the Content Helper.

## How Has This Been Tested?
Tests pass, and an integration test has also been added as part of this PR.

### DIFF
--- a/src/blocks/content-helper/class-content-helper.php
+++ b/src/blocks/content-helper/class-content-helper.php
@@ -1,0 +1,44 @@
+<?php
+/**
+ * Parse.ly Content Helper class
+ *
+ * @package Parsely
+ * @since 3.5.0
+ */
+
+declare(strict_types=1);
+
+namespace Parsely;
+
+/**
+ * Parse.ly Content Helper.
+ *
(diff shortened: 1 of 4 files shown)


chriszarate (approved):
:+1: Longer-term, i'd like to strategize about how to improve the ergonomics of these hooks. This doesn't feel like it will scale very well.

PR #1077, opened 2022-09-12
Content Helper: Fix broken links in "Post Stats" button

acicovic:
## Description
The 3.5.0 release introduced a bug which resulted in incorrect links in the Content Helper's "Post Stats" button. We are planning on releasing a 3.5.1 release that will include this fix. Thanks to @chriszarate for spotting this.

## Motivation and Context
Bug fix.

## How Has This Been Tested?
- Manual testing works.
- The related Integration Test has been adjusted and now checks for URL corectness.


src/blocks/content-helper/class-content-helper.php
```diff
@@ -39,6 +49,12 @@ public function run(): void {
 			array(),
 			$content_helper_asset['version']
 		);
+
+		// Set wpParselyContentHelperPrefix for the "Post Stats" button links
+		// to point to the correct location.
+		$prefix                = trailingslashit( 'https://dash.parsely.com/' . esc_js( $this->parsely->get_api_key() ) ) . 'find';
+		$analytics_link_prefix = 'window.wpParselyContentHelperPrefix = "' . $prefix . '";';
+		wp_add_inline_script( 'wp-parsely-block-content-helper', $analytics_link_prefix, 'before' );
```
chriszarate: Summary of what's happening here:

1. Here, we inject a browser global containing the Parse.ly API prefix.
2. Client-side, we [fetch posts data](https://github.com/Parsely/wp-parsely/blob/develop/src/blocks/content-helper/content-helper-provider.ts#L64) from our REST API endpoint.
3. We process that data and [mix-in a `scriptUrl` property](https://github.com/Parsely/wp-parsely/blob/develop/src/blocks/content-helper/content-helper-provider.ts#L80), which uses this browser global. It is the only reference in the code to this global.

Instead of this, I'd recommend just defining the correct `scriptUrl` property server-side and adding it the API response here:

https://github.com/Parsely/wp-parsely/blob/557d0ddbebe9bbff4dd2d5aba33fb12568163efa/src/Endpoints/class-analytics-posts-api-proxy.php#L46

Then you can remove the `processData` function and with it the need for this global.

acicovic: Let me know if what I've done is close to what you were thinking.

src/Endpoints/class-analytics-posts-api-proxy.php
```diff
@@ -41,16 +41,19 @@ public function get_items( WP_REST_Request $request ): stdClass {
 	 * @return array<stdClass> The generated data.
 	 */
 	protected function generate_data( array $response ): array {
-		$date_format = get_option( 'date_format' );
-		$result      = array_map(
-			static function( stdClass $item ) use ( $date_format ) {
+		$date_format    = get_option( 'date_format' );
+		$stats_base_url = trailingslashit( 'https://dash.parsely.com/' . esc_js( $this->parsely->get_api_key() ) ) . 'find';
```
chriszarate: no real need to escape this here since it is not because used in output

```suggestion
		$stats_base_url = trailingslashit( 'https://dash.parsely.com/' . $this->parsely->get_api_key() ) . 'find';
```

chriszarate (dismissed):
:+1: one suggested change, looks good

PR #1093, opened 2022-09-15
Content Helper: Add "Current post details" panel

### LINKED ISSUE #965: Content Helper: Add current post information
The Content Helper doesn't currently include any information about the current post being edited.

Add a panel to the Content Helper that will show certain useful information about the post that is being currently edited. To do that, we should:
- Get the post's published URL.
- Use the Parse.ly API to get information about the post.

From a quick research, here is the information we can currently show about a post using the API:
- https://www.parse.ly/help/api/analytics-endpoint#get-analyticspostdetail
- https://www.parse.ly/help/api/referrers#get-referrerspostdetail
- https://www.parse.ly/help/api/shares#get-sharespostdetail

Some questions to ask ourselves here would be:
- What qualifies as "useful information"?
- Would any of the above information provided by the API be considered as useful?
- What would be some useful information to show that is not currently provided by the API, and can the API be extended to provide it?

### LINKED ISSUE #1204: Content Helper Current Post Details: Add direct visits to the Referrers section
Our [Current Post Details PR](https://github.com/Parsely/wp-parsely/pull/1093) does not include direct traffic under its Referrers section.

Not providing this may look like a bug because the referral traffic will not add up to the total traffic displayed.

Add direct traffic under the Referrers section.

acicovic:
Note: I apologize in advance for the big PR.

## Description
This PR:
- Adds a "Current post details" panel to the Content Helper.
- Sets the panel as the default to be opened on sidebar load.
- Introduces a theme for the panel that resembles the Parse.ly dashboard style. (thanks go to @garrettvercoe for his excellent mock-up work!).
- Implements all the back-end code for implementing the API calls needed.
- Implements a function to load our API endpoints more easily, as the list is growing.

## Reviewing this PR
I still feel that I'm in the beginning of my TypeScript/React journey and that I may not be following certain best practices or conventions.

Please feel free to pinpoint those so we can improve the code and my skills together.

## Motivation and Context
- Closes #965.
- Closes #1204.
- Present useful/helpful information to users within the context of the Content Helper. 
- Bring some Parse.ly flavor to the Content Helper!

## How Has This Been Tested?
- Manual testing (getting data for a few different URLs).
- Fixed current Content Helper tests so they pass.
- Automated tests specific to the new panel are to be iomplemented as part of another PR.

## Screenshots
current-post-details-screenshot


GaryJones:
One quick observation to test - if the referer URL is long (`mywonderfulbutabnormallylongurltoseeifitbreaksthelayout.com`), how does the layout look in the sidebar?

src/Endpoints/class-analytics-post-detail-api-proxy.php
```diff
@@ -0,0 +1,84 @@
...
+	 * @param WP_REST_Request $request The request object.
+	 */
+	public function get_items( WP_REST_Request $request ): stdClass {
+		return $this->get_data( $request );
+	}
+
+	/**
+	 * Generates the final data from the passed response.
+	 *
+	 * @param array<string, mixed> $response The response received by the proxy.
```
GaryJones, line 40: Line 47 treats the items of the `$response` array as `stdClass` - is that right? If so, could `stdClass` be used here instead of `mixed`?

src/Endpoints/class-analytics-post-detail-api-proxy.php
```diff
@@ -0,0 +1,84 @@
...
+	 * @return array<stdClass> The generated data.
+	 */
+	protected function generate_data( array $response ): array {
+		$stats_base_url = trailingslashit( 'https://dash.parsely.com/' . $this->parsely->get_api_key() ) . 'find';
```
GaryJones, line 44: Minor suggestion for another ticket - consider having `dash.parsely.com` defined in a constant that could then be adjusted during development to test against the beta dashboard.

src/Endpoints/class-analytics-post-detail-api-proxy.php
```diff
@@ -0,0 +1,84 @@
...
+	 *
+	 * @since 3.6.0
+	 *
+	 * @param float $time The time as a float number.
+	 * @return string The resulting formatted time duration.
+	 */
+	private function get_duration( float $time ) {
+		$decimal_value = round( fmod( $time, 1 ) * 60 );
+		return sprintf( '%2d:%02d', (int) $time, $decimal_value );
+	}
```
GaryJones: I think this could benefit from unit tests - for instance, a `$time` of `999.999` returns a string of `999:60`, which doesn't make a lot of sense. If you're going to format it into `mm:ss` then it needs some more guards.

Also, updating the docs to indicate an example of the incoming format, and what the returned format should look like (and that it is meant to be `mm:ss`) would be useful for understanding the intent of the code.

src/Endpoints/class-referrers-post-detail-api-proxy.php
```diff
@@ -0,0 +1,217 @@
...
+	 * internationalized percentage format.
+	 *
+	 * @since 3.6.0
+	 *
+	 * @param int $number The number to be calculated as a percentage.
+	 * @param int $total The total number to compare against.
+	 * @return string The resulting internationalized percentage.
+	 */
+	private function get_i18n_percentage( int $number, int $total ) {
+		return number_format_i18n( $number / $total * 100, 2 );
```
GaryJones, line 226: If the `$total` is 0, then this will generate a Division by Zero fatal. Add a guard clause for that condition.

src/blocks/content-helper/current-post-details/component.tsx
```diff
@@ -0,0 +1,309 @@
...
+	const periodStart = Intl.DateTimeFormat(
+		document.documentElement.lang,
+		{ month: 'short', day: 'numeric' }
+	).format( new Date( period.start ) );
+
+	return (
+		<div className="section period">
+			{ __( 'Last ', 'wp-parsely' ) + period.days + __( ' Days ', 'wp-parsely' ) }
+			<span>
+				{ __( '(', 'wp-parsely' ) + periodStart + __( ' - Today)', 'wp-parsely' ) }
```
GaryJones: Is it possible to make these strings combined, with an interpolated / placeholder? It will make the translations easier on translators.

src/blocks/content-helper/current-post-details/component.tsx
```diff
@@ -0,0 +1,309 @@
...
+					Object.entries( data.referrers.top ).map( ( [ key, value ] ) => {
+						if ( key === 'totals' ) {
+							totalViewsPercentage = Number( value.viewsPercentage );
+							return null;
+						}
+
+						return (
+							<tr key={ key }>
+								<th title={ key }>{ key }</th>
+								<td title={ value.viewsPercentage + __( '%', 'wp-parsely' ) }>
```
GaryJones: The number should be added as a placeholder / interpolated value within the string, as for some locales, the percent sign comes before the number.
(This will apply to anywhere that a percent sign is being displayed)

You'll need to import and use `sprintf` from `@wordpress/i18n`.

src/blocks/content-helper/current-post-details/component.tsx
```diff
@@ -0,0 +1,309 @@
...
+								</td>
+								<td>{ ImpreciseNumber( value.views ) }</td>
+							</tr>
+						);
+					} )
+				}
+				</tbody>
+			</table>
+			<div>
+				{ totalViewsPercentage + __( '% of views came from top referrers.', 'wp-parsely' ) }
```
GaryJones: Example instance of needing to include the number with the percent sign via sprintf placeholder.

src/blocks/content-helper/current-post-details/component.tsx
```diff
@@ -0,0 +1,309 @@
...
+ *
+ * Note: This function is not made to process float numbers.
+ *
+ * @param {string} value          The number to process. It can be formatted.
+ * @param {number} fractionDigits The number of desired fraction digits.
+ * @param {string} glue           A string to put between the number and unit.
+ * @return {string} The number formatted as an imprecise number.
+ */
+function ImpreciseNumber( value: string, fractionDigits = 1, glue = '' ): string {
+	const number = parseInt( value.replace( /\D/g, '' ) );
```
GaryJones: `parseInt()` should always include the second (radix) argument since it may not always default to `10`.

tests/e2e/specs/blocks/content-helper/top-bar-icon.spec.js
```diff
@@ -21,7 +21,7 @@ const pluginButton = 'button[aria-label="Parse.ly Content Helper"]';
  * Tests for the Content Helper's top bar icon.
  */
 describe( 'Content Helper top bar icon in the WordPress Post Editor', () => {
-	const relatedPostsTitle = 'Related top-performing post';
+	const contentHelperTitle = 'Current Post DetailsRelated Top-Performing Posts';
```
GaryJones, line 24: This doesn't seem quite right?

wp-parsely.php
```diff
@@ -188,3 +202,18 @@ function parsely_integrations( $parsely = null ): Integrations {
...
+ * Instantiates and runs the specified API endpoint.
+ *
+ * @since 3.6.0
+ *
+ * @param string $proxy_class_name The proxy class to instantiate.
+ * @param string $api_proxy_class_name The API proxy class to instantiate and run.
+ */
+function parsely_run_rest_api_endpoint( string $proxy_class_name, string $api_proxy_class_name ): void {
+	$proxy_instance        = new $proxy_class_name( $GLOBALS['parsely'] );
+	$cached_proxy_instance = new Cached_Proxy( $proxy_instance, new WordPress_Cache() );
```
GaryJones: I wonder if we could avoid spinning up new instances of this WP cache each time.

```php
static $cache = new WordPress_Cache();
...
$cached_proxy_instance = new Cached_Proxy( $proxy_instance, $cache );
...
```

GaryJones (changes requested):

src/Endpoints/class-analytics-post-detail-api-proxy.php
acicovic: This is probably what should be done. Since this is an overridden function present in a few classes, I'll address this in another PR to keep this PR focused.

src/Endpoints/class-analytics-post-detail-api-proxy.php
acicovic: We are defining this in 2-3 places within our codebase. Noting this to see if it must be a function or constant.

tests/e2e/specs/blocks/content-helper/top-bar-icon.spec.js
acicovic: This isn't pretty, but it is the expected value (titles of all Content Helper panels).

src/Endpoints/class-referrers-post-detail-api-proxy.php
acicovic: This opens-up the broader question of what we would need to do even in the UI side in case of zero views/totals. I'll need to dig a bit more into this.

src/Endpoints/class-analytics-post-detail-api-proxy.php
GaryJones: A function would be fine too, or a public constant/method in a class that could then be mocked / overridden locally.

garrettvercoe:
Placing my thoughts here for visibility on the UI:
image

src/blocks/content-helper/current-post-details/component.tsx
acicovic: Thanks for catching this!

src/Endpoints/class-referrers-post-detail-api-proxy.php
acicovic: I've added a guard against this in that function, and noted about the UI so I can look into it later. I'm marking this as resolved, but feel free to comment if needed.

src/Endpoints/class-analytics-post-detail-api-proxy.php
acicovic: I've improved the function and its docs. I did not write unit tests for it, but marking this as resolved.

GaryJones:
re: UI tooltip suggestion - let's please not do that,  as that becomes an accessibility issue.

If the colours are important, then there should be a way to confer the same information without just colour alone.

Since the numbers are already present, maybe the coloured bar doesn't add much for anyone?

src/Endpoints/class-analytics-post-detail-api-proxy.php
acicovic: Now tracked in #1208.

src/Endpoints/class-analytics-post-detail-api-proxy.php
acicovic: Now tracked in #1209.

wp-parsely.php
acicovic: Now tracked in #1210.

src/blocks/content-helper/content-helper.scss
```diff
@@ -1,7 +1,226 @@
...
+		}
+
+		// Actions section (Visit Post, View in Parse.ly buttons).
+		div.section.actions {
+			display: inline-flex;
+			justify-content: space-between;
+			width: 100%;
+
+			a.components-button {
+				border-radius: 0.4rem;
```
garrettvercoe: This should be nudged down to 4px (.25rem)

src/blocks/content-helper/current-post-details/component.tsx
acicovic: I have improved our i18n in [this commit](https://github.com/Parsely/wp-parsely/pull/1093/commits/cadb922761d967adb5f5af6196fd2bd4e548d508). Marking this and your other related remarks as resolved, but feel free to comment if something else should be needed.

acicovic:
@GaryJones and @garrettvercoe, regarding tooltips and color bars:

I don't think that color bars should be removed, as they are an integral part of what makes this design look good and they remind of what is in the Parse.ly dashboard. They also give you a sense of what is going on without needing to read the numbers in detail.

For the color bars, I use the `title` attribute to add in the percentage. This way, the user gets a percentage tooltip when hovering the percentage bars. This works in the multlcolor bar as well, for every distinct color. In this way, I think we have the best of both worlds. We do get to show a tooltip that is not an accessibility concern. Having said that, we could tweak the title to hold more information.

Following the same logic, in the Top 5 Referrers section, every URL holds the URL as the title. So even if the displayed URL gets [truncated by CSS](https://github.com/Parsely/wp-parsely/blob/7df14b1ad9c7542f8ba174a9f59c4e37a7b230d3/src/blocks/content-helper/content-helper.scss#L141-L151) (@GaryJones , this should answer your first question), the user can see the full URL by hovering on the truncated URL.

image

src/blocks/content-helper/content-helper.scss
acicovic: Good catch. I've updated this to `4px`, since all our `border-radius` values are specified in `px` units.

acicovic:
@garrettvercoe, regarding your "image review" 😃:

1. Font sizes are all the same.

2. About the "Last 7 days" `top-margin`, I have no strong opinions about that. We can remove it if you think that it's OK when there is an outline on the panel title. The first image is with the margin, the second without.

image

image

3. I've renamed "Engaged Time" to "Avg. Time".

4. For the time being, I've left all color bars just display the percentage on hover (using their `title` attribute), but we can iterate on that.

5. Regarding referrer group spacing, here is a sample (hacked up using dev tools). It is a bit tight but I think it passes. You can let me know if you think this is not acceptable and have other ideas.
image

6. Adding even `1px` of `letter-spacing` looks nice but makes the buttons too big in width. Would you consider stacking them in order to have the `letter-spacing`?

image

Thank you for your work and let me know of anything else!

GaryJones:
@acicovic Please demonstrate your tooltip solution in a screen recording using:
- keyboard only
- a screen reader (VoiceOver is fine).

src/Endpoints/class-referrers-post-detail-api-proxy.php
```diff
@@ -37,6 +43,8 @@ public function run(): void {
 	 * @param WP_REST_Request $request The request object.
 	 */
 	public function get_items( WP_REST_Request $request ): stdClass {
+		$this->total_views = (int) $request->get_param( 'total_views' );
```
GaryJones: Instead of casting as `int`, consider wrapping `absint()` around it instead, to then ensure it is a positive integer.

src/blocks/content-helper/current-post-details/component.tsx
```diff
@@ -173,8 +184,8 @@ function ReferrerTypesSection( data: PostPerformanceData ) {
 					return (
 						<div className={ 'bar-fill ' + key } key={ key }
 							title={
-								/* translators: %s: Percentage value, %%: Escaped percent sign */
-								sprintf( __( '%s%%', 'wp-parsely' ), value.viewsPercentage ) // eslint-disable-line @wordpress/valid-sprintf
+								/* translators: 1: Referrer type, 2: Percentage value, %%: Escaped percent sign */
+								sprintf( __( '%s: %s%%', 'wp-parsely' ), getKeyTitle( key ), value.viewsPercentage ) // eslint-disable-line @wordpress/valid-sprintf
```
GaryJones: Strings with multiple placeholders should use numbered placeholders.
```suggestion
								sprintf( __( '%1$s: %2$s%%', 'wp-parsely' ), getKeyTitle( key ), value.viewsPercentage ) // eslint-disable-line @wordpress/valid-sprintf
```

src/blocks/content-helper/current-post-details/component.tsx
```diff
@@ -225,9 +236,14 @@ function TopReferrersSection( data: PostPerformanceData ) {
...
 
+						let referrerUrl = key;
+						if ( key === 'direct' ) {
+							referrerUrl = __( 'Direct', 'wp-parsely' );
+						}
+
 						return (
 							<tr key={ key }>
-								<th title={ key }>{ key }</th>
+								<th title={ referrerUrl }>{ referrerUrl }</th>
```
GaryJones:
```suggestion
								<th scope="row">{ referrerUrl }</th>
```

There's no need to repeat the content of the table header cell, as the title attribute.

The `scope` attribute is worth adding to the row `th`, since the table is non-regular - that is, the column heading `th` spans two columns, and the column heading `th` should have a `scope="col"` as well.

src/blocks/content-helper/current-post-details/component.tsx
```diff
@@ -225,9 +236,14 @@ function TopReferrersSection( data: PostPerformanceData ) {
...
 								<td title={
```
GaryJones: OK, I can see what's trying to be done, but it's not going to be accessible.

Right now you have a `td` element with a title attribute (not typically read out by screen readers, and touch devices can't access it either since it assumes hovering with a pointer), and inside the cell is an empty div (screen readers won't see anything), followed by a visible number in a new cell.

Instead, what's needed is the `td` (no attribute), followed by the empty div with an `aria-hidden` attribute (so it is definitely ignored by screen readers), and a screen-reader-text span with the percentage, then the number in the next cell.

The resulting HTML of that middle cell would be something like:

```html
<td>
	<div aria-hidden="true" style="..."></div>
	<span class="screen-reader-text">27%</span>
</td>
```

Alternatively, you could use an aria-label instead:

```html
<td>
	<div aria-label="27%" style="..."></div>
</td>
```

src/blocks/content-helper/current-post-details/component.tsx
```diff
@@ -0,0 +1,341 @@
...
+	const periodStart = Intl.DateTimeFormat(
+		document.documentElement.lang,
+		{ month: 'short', day: 'numeric' }
+	).format( new Date( period.start ) );
+
+	return (
+		<div className="section period">
+			{
+				/* translators: Number of days for which data is being shown */
+				sprintf( __( 'Last %d Days', 'wp-parsely' ), period.days )
```
GaryJones, line 114: Even in English, this string is going to be wrong if the number of days is exactly 1. 

Use https://developer.wordpress.org/block-editor/reference-guides/packages/packages-i18n/#n to allow singular and plural strings, and pass in the `period.days`, and let the i18n system choose the right string for the right number.

src/blocks/content-helper/current-post-details/component.tsx
```diff
@@ -0,0 +1,341 @@
...
+ */
+function TopReferrersSection( data: PostPerformanceData ) {
+	let totalViewsPercentage = 0;
+	return (
+		<div className="section top-referrers">
+			<table>
+				<thead>
+					<tr>
+						<th>{ __( 'Top Referrers', 'wp-parsely' ) }</th>
+						<th colSpan={ 2 }>{ __( 'Page Views', 'wp-parsely' ) }</th>
```
GaryJones:
```suggestion
						<th scope="col" colSpan={ 2 }>{ __( 'Page Views', 'wp-parsely' ) }</th>
```

src/blocks/content-helper/current-post-details/component.tsx
```diff
@@ -0,0 +1,341 @@
...
+								<td>{ ImpreciseNumber( value.views ) }</td>
+							</tr>
+						);
+					} )
+				}
+				</tbody>
+			</table>
+			<div> {
+				/* translators: %s: Percentage value, %%: Escaped percent sign */
+				sprintf( __( '%s%% of views came from top referrers.', 'wp-parsely' ), totalViewsPercentage ) // eslint-disable-line @wordpress/valid-sprintf
```
GaryJones: In other locales, the string may change depending on whether the thing being counted is 0, 1, or some other special number, so use https://developer.wordpress.org/block-editor/reference-guides/packages/packages-i18n/#n to allow singular and plural strings, and pass in the `totalViewsPercentage`, and let the i18n system choose the right string for the right number.

src/blocks/content-helper/current-post-details/component.tsx
```diff
@@ -0,0 +1,341 @@
...
+/**
+ * Outputs the "Actions" section.
+ *
+ * @param {PostPerformanceData} data The data needed to populate the section.
+ */
+function ActionsSection( data: PostPerformanceData ) {
+	return (
+		<div className="section actions">
+			<Button href={ data.url } rel="noopener" target="_blank" variant="secondary">{ __( 'Visit Post', 'wp-parsely' ) }</Button>
+			<Button href={ data.statsUrl } rel="noopener" target="_blank" variant="primary">{ __( 'View in Parse.ly', 'wp-parsely' ) }</Button>
```
GaryJones: We need there to be a `<span className="screen-reader-text"> { __( '(opens in new tab)', 'wp-parsely' ) }</span>` as part of the button text for both of these buttons. It's OK to keep these as a separately translated string in this case since they are self-contained.

Any link with a target of `_blank` should tell screen reader users that it will open a new tab so they understand why they've just been shifted away from the current tab.

wp-parsely.php
```diff
@@ -26,17 +26,21 @@
...
+use Parsely\Endpoints\Referrers_Post_Detail_API_Proxy;
 use Parsely\Endpoints\Rest_Metadata;
 use Parsely\Integrations\Amp;
 use Parsely\Integrations\Facebook_Instant_Articles;
 use Parsely\Integrations\Google_Web_Stories;
 use Parsely\Integrations\Integrations;
+use Parsely\RemoteAPI\Analytics_Post_Detail_Proxy;
 use Parsely\RemoteAPI\Cached_Proxy;
 use Parsely\RemoteAPI\Related_Proxy;
 use Parsely\RemoteAPI\Analytics_Posts_Proxy;
```
GaryJones, line 40: Might as well move this one into the correct alphabetical order position.

GaryJones (dismissed):

src/blocks/content-helper/current-post-details/component.tsx
acicovic: We currently get the data for 7 days in a hardcoded manner, so as things stand it cannot be 1.

As such, I won't proceed with that change, but your point is clear and correct. Given we ever allow 1 to be an option, we will revise this.

acicovic:
@GaryJones, thanks for all the feedback as well as giving some explanations. In the commits https://github.com/Parsely/wp-parsely/pull/1093/commits/07aedc66e692a7682b3cdb717a44296dc1b0a75a and https://github.com/Parsely/wp-parsely/pull/1093/commits/a35f0610f385c80a8c087ab462dbac88942460fc, I'm attempting to address all your accessibility feedback (thus marking the related items as resolved). The things I've done a bit differently are:

1. I've left the `title` attributess in the code. Granted, it's not available to everyone. However, if they don't create any accessibility issues, it's actually nice to have them as they show the built-in browser tooltip. This is not something we rely on for fundamental functionality, but rather a nice bonus.
2. For the `scope` attributes, I've used `colgroup` instead of `col` where we have the `colspan=2` after reading [this](https://www.w3.org/WAI/EO/Drafts/tutorials/tables/scope/#table-with-headers-spanning-multiple-rows-or-columns).

Let me know if you have additional feedback regarding any of the above.

acicovic:
https://github.com/Parsely/wp-parsely/pull/1216 has just been merged, so we now have some conflicts appearing here. I'll work on them to resolve them.

src/blocks/content-helper/current-post-details/component.tsx
GaryJones: If the data is always 7 days, why is there even a `period.days` variable?

GaryJones:
> I've left the title attributess in the code. Granted, it's not available to everyone. However, if they don't create any accessibility issues, it's actually nice to have them as they show the built-in browser tooltip. This is not something we rely on for fundamental functionality, but rather a nice bonus.

Please, remove all of them.

(And use the className="screen-reader-text" approach I suggested.)

> Granted, it's not available to everyone.

If information is worth including, then include it for everyone. There should be no exceptions.

Mobile device users, screen reader users, keyboard-only, braille device users, and those with physical impairments who can't keep a pointing device hovered over the element long enough to trigger the tooltop, all can't access the `title` attribute.

https://core.trac.wordpress.org/ticket/24766 is the ticket for WP core, where it removed all of the title attributes. As such, hovering over an element inside wp-admin isn't an expected behaviour, even if it's in a browser that supports showing title values as tooltips. 

Please read https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title - at no point does it say that the title attribute should be used primarily as a way to display as tooltips. 

Some screen reader users will have their screen reader software set to read out `title` attributes. In that case, they will hear duplicate information since the percentage is (should) already included as screen reader text i.e. it DOES create an accessibility issue.

acicovic:
> (And use the className="screen-reader-text" approach I suggested.)

Are you referring to https://github.com/Parsely/wp-parsely/pull/1093#discussion_r1032699423?
I've used the second approach you mentioned. Should I actually use the first one?

src/blocks/content-helper/current-post-details/component.tsx
acicovic: It caters for the possibility of adding multiple predefined periods in a future iteration of the feature. This is why we also have the `setDataPeriod()` function for example.

acicovic:
> Please, remove all of them.

Since they might be an issue for certain users, I've removed the title attributes.

acicovic:
Something unrelated to code, but very important that I want to address, since this was posted publicly:

>> Granted, it's not available to everyone.

> If information is worth including, then include it for everyone. There should be no exceptions. Your statement is ringing so many alarm bells for me.

My point was that while the tooltip resulting from the `title` attribute would not be available to everyone, this would not be an issue since the accessibility enhancements commit would make up for it. As such, everyone would have the percentage value data, regardless of device or browsing method.

I could have been right or wrong in my assumption, but I'm worried about "ringing alarm bells".

To put it another way:
- The fact that I'm not an accessibility expert is pretty obvious, as can be seen by the comments of this PR.
- The fact that any accessibility concerns have been taken very seriously and promptly addressed by all the current coders of this project is also obvious from the related PRs.
- The fact that I'm open to constructive feedback/criticism is obvious from many PRs.
- My personality and values are not obvious from any PR. Publishing phrases like "Your statement is ringing so many alarm bells for me" is open to wide misinterpretation and could raise unfavorable and harmful thought against me, which I'm sure is not what you meant and want.

So to conclude, I would say that alarm ringing of any sort is not necessary. I'll continue focusing on addressing feedback the best I can, and as always I value your contributions to the project.

src/blocks/content-helper/current-post-details/component.tsx
```diff
@@ -0,0 +1,359 @@
...
+ * Outputs the current post's details or shows an error message on failure.
+ */
+function CurrentPostDetails() {
+	const [ loading, setLoading ] = useState<boolean>( true );
+	const [ error, setError ] = useState( null );
+	const [ postDetailsData, setPostDetails ] = useState<PostPerformanceData>( null );
+	const provider = new CurrentPostDetailsProvider();
+
+	useEffect( () => {
+		const fetchPosts = async ( retries: number ) => {
```
luiztiago, line 34: I think you can move the `fetchPosts` outside the `useEffect` hook. The idea is to have the `fetchPosts` as a `useEffect` dependency and call to the async function inside the useEffect function.

luiztiago:
```
const fetchSomething = async () => {
  setLoading( true );

  provider.getSomething()
    .then(...)
    .catch(...)
};

useEffect(() => fetchPosts(FETCH_RETRIES), [fetchPosts])
```

luiztiago (commented):
Just a comment on useEffect usage, but I'm not sure about the usage here.

src/blocks/content-helper/current-post-details/component.tsx
chriszarate: Effects are tricky, especially [their dependencies](https://reactjs.org/docs/hooks-reference.html#conditionally-firing-an-effect)! This suggestion will change the behavior of the hook.

In the original code, the dependency array is empty, which means that the effect will run only once, on mount (and again if the component re-mounts).

In the suggested code, the dependency array contains a function that is re-created on each render. Since these functions will not be strictly equal, the effect will run on every render, which is not desired. 

You could navigate this issue by using `useCallback` to ensure the functions are strictly equal, but in this case keeping the logic inside the effect is probably preferable. In general, `useCallback` is most useful when passing a function as a prop to a child component, as a way to prevent unnecessary rerenders.

src/blocks/content-helper/current-post-details/component.tsx
```diff
@@ -0,0 +1,359 @@
...
+interface PostDetailsSectionProps {
+	data: PostPerformanceData;
+}
+
+/**
+ * Outputs the current post's details or shows an error message on failure.
+ */
+function CurrentPostDetails() {
+	const [ loading, setLoading ] = useState<boolean>( true );
+	const [ error, setError ] = useState( null );
```
chriszarate, line 29:
```suggestion
	const [ error, setError ] = useState<Error>( null );
```

src/blocks/content-helper/current-post-details/component.tsx
```diff
@@ -0,0 +1,359 @@
...
+		setLoading( true );
+		fetchPosts( FETCH_RETRIES );
+	}, [] );
+
+	if ( error ) {
+		// "Soft" error for which we do not want to show an "Error:" prefix.
+		if ( 'string' === typeof error ) {
+			return <p>{ error }</p>;
+		}
+
```
chriszarate, line 60: This is a shaky approach to handling different error types. Better to set a property on an `Error` object that governs this behavior (`error.type = 'soft'`). Or just always make sure the message reflects what you want to display to users.

Throwing anything other than an `Error` object should be considered bad practice. It is tricky to enforce with TypeScript, but there may be a lint rule we can use to help enforce this.

src/blocks/content-helper/current-post-details/component.tsx
```diff
@@ -0,0 +1,359 @@
...
+		// Error coming from apiFetch.
+		if ( error?.message ) {
```
chriszarate, line 62: Type annotation for `useState` should allow confidence that we have an Error object
```suggestion
		if ( error.message ) {
```

src/blocks/content-helper/current-post-details/component.tsx
```diff
@@ -0,0 +1,359 @@
...
+		// Error coming from the WordPress REST API.
+		const errorMessage = JSON.stringify( error ).match( /\[\"(.*?)\"\]/ )[ 1 ];
+		return <p>{ __( 'Error:', 'wp-parsely' ) } { errorMessage }</p>;
+	}
+
+	return (
+		<>
+			{ loading ? <Spinner /> : <CurrentPostDetailsSections data={ postDetailsData } /> }
+		</>
+	);
```
chriszarate: fragment unneeded here
```suggestion
	return (
		loading
			? <Spinner />
			: <CurrentPostDetailsSections data={ postDetailsData } />
	);
```

src/blocks/content-helper/current-post-details/component.tsx
```diff
@@ -0,0 +1,359 @@
...
+function CurrentPostDetailsSections( props: PostDetailsSectionProps ) {
+	const data = props.data;
+
+	return (
+		<div className="current-post-details-panel">
+			{ DataPeriodSection( data ) }
+			{ GeneralPerformanceSection( data ) }
+			{ ReferrerTypesSection( data ) }
+			{ TopReferrersSection( data ) }
+			{ ActionsSection( data ) }
```
chriszarate: These are React components, because they are functions that return JSX. Update them to use props.
```suggestion
			<DataPeriodSection {...props} />
			<GeneralPerformanceSection {...props} />
			<ReferrerTypesSection {...props} />
			<TopReferrersSection {...props} />
			<ActionsSection {...props} />
```

src/blocks/content-helper/current-post-details/component.tsx
```diff
@@ -0,0 +1,359 @@
...
+	);
+}
+
+/**
+ * Outputs the "Period" section, which denotes the period for which data is
+ * shown.
+ *
+ * @param {PostPerformanceData} data The data needed to populate the section.
+ */
+function DataPeriodSection( data: PostPerformanceData ) {
```
chriszarate: Update for all these components
```suggestion
function DataPeriodSection( props: PostPerformanceData ) {
```

src/blocks/content-helper/current-post-details/component.tsx
```diff
@@ -0,0 +1,359 @@
...
+ * Implements the "Imprecise Number" functionality of the Parse.ly dashboard.
+ *
+ * Note: This function is not made to process float numbers.
+ *
+ * @param {string} value          The number to process. It can be formatted.
+ * @param {number} fractionDigits The number of desired fraction digits.
+ * @param {string} glue           A string to put between the number and unit.
+ * @return {string} The number formatted as an imprecise number.
+ */
+function ImpreciseNumber( value: string, fractionDigits = 1, glue = '' ): string {
```
chriszarate: PascalCase should be reserved for classes, types, and React components. Use camelCase
```suggestion
function impreciseNumber( value: string, fractionDigits = 1, glue = '' ): string {
```

src/blocks/content-helper/current-post-details/component.tsx
```diff
@@ -0,0 +1,359 @@
...
+						/* translators: %s: Percentage value, %%: Escaped percent sign */
+						const ariaLabel = sprintf( __( '%s%%', 'wp-parsely' ), value.viewsPercentage ); // eslint-disable-line @wordpress/valid-sprintf
+
+						return (
+							<tr key={ key }>
+								<th scope="row">{ referrerUrl }</th>
+								<td>
+									<div aria-label={ ariaLabel }
+										className="percentage-bar"
+										style={ { '--bar-fill': value.viewsPercentage + '%' } as React.CSSProperties }>
```
chriszarate, line 260: Neat! I didn't know you could do this. I thought you had to render a `<style>` element for this. I guess React doesn't know how to type this though. :) 👍

src/blocks/content-helper/current-post-details/post-performance-data.ts
```diff
@@ -0,0 +1,29 @@
+export interface PostPerformanceData {
+	author: string;
+	avgEngaged: string;
+	date: string;
+	id: number;
+	period: {
+		start: string,
+		end: string,
+		days: number,
```
chriszarate: Maybe a lint rule to enforce consistency
```suggestion
		start: string;
		end: string;
		days: number;
```

src/blocks/content-helper/current-post-details/provider.ts
```diff
@@ -0,0 +1,205 @@
...
+	PostPerformanceData,
+	PostPerformanceReferrerData,
+} from './post-performance-data';
+
+/**
+ * Specifies the form of the response returned by the `/analytics/post/detail`
+ * WordPress REST API endpoint.
+ */
+ interface analyticsApiResponse {
+	error?: object;
```
chriszarate, line 22:
```suggestion
	error?: Error;
```

src/blocks/content-helper/current-post-details/provider.ts
```diff
@@ -0,0 +1,205 @@
...
+	data: PostPerformanceData[];
+}
+
+/**
+ * Specifies the form of the response returned by the `/referrers/post/detail`
+ * WordPress REST API endpoint.
+ */
+interface referrersApiResponse {
+	error?: object;
```
chriszarate, line 31:
```suggestion
	error?: Error;
```

src/blocks/content-helper/current-post-details/provider.ts
```diff
@@ -0,0 +1,205 @@
...
+ interface analyticsApiResponse {
+	error?: object;
+	data: PostPerformanceData[];
+}
+
+/**
+ * Specifies the form of the response returned by the `/referrers/post/detail`
+ * WordPress REST API endpoint.
+ */
+interface referrersApiResponse {
```
chriszarate:
```suggestion
interface ReferrersApiResponse {
```

src/blocks/content-helper/current-post-details/provider.ts
```diff
@@ -0,0 +1,205 @@
...
+import {
+	PostPerformanceData,
+	PostPerformanceReferrerData,
+} from './post-performance-data';
+
+/**
+ * Specifies the form of the response returned by the `/analytics/post/detail`
+ * WordPress REST API endpoint.
+ */
+ interface analyticsApiResponse {
```
chriszarate:
```suggestion
 interface AnalyticsApiResponse {
```

src/blocks/content-helper/current-post-details/provider.ts
```diff
@@ -0,0 +1,205 @@
...
+	 * API.
+	 *
+	 * @param {string} postUrl
+	 * @return {Promise<PostPerformanceData> } The current post's details.
+	 */
+	private async fetchPerformanceDataFromWpEndpoint( postUrl: string ): Promise<PostPerformanceData> {
+		let response;
+
+		try {
+			response = await apiFetch( {
```
chriszarate: Not sure, but does `apiFetch` accept a type parameter? I.e., maybe you could do `apiFetch<ReferrersApiResponse>()` instead of using `as`

src/blocks/content-helper/current-post-details/provider.ts
```diff
@@ -0,0 +1,205 @@
...
+		}
+
+		if ( response?.error ) {
+			return Promise.reject( response.error );
+		}
+
+		// No data was returned.
+		if ( response.data.length === 0 ) {
+			return Promise.reject(
+				sprintf(
```
chriszarate, line 113: create Error here and for other Promise rejections

src/blocks/content-helper/variables.scss
```diff
@@ -0,0 +1,37 @@
+/**
+ * This is a subset of the CSS variables defined in the Parse.ly dashboard. It
+ * is sourced from 1.44/src/styles/base.scss and defines some additional
+ * variables in the end of the file.
+ */
+
+div.wp-parsely-content-helper {
```
chriszarate: https://stackoverflow.com/questions/5797014/why-do-browsers-match-css-selectors-from-right-to-left/5813672#5813672

```suggestion
.wp-parsely-content-helper {
```

src/blocks/content-helper/current-post-details/component.tsx
acicovic: After considering the feedback, I will leave things as they are and mark this as resolved.

src/blocks/content-helper/current-post-details/component.tsx
acicovic: Please let me know if you are OK with the changes in https://github.com/Parsely/wp-parsely/pull/1093/commits/55539492dad4e4652b524a0b0d8c9dec952c34a2.

acicovic:
@chriszarate, I've noted all your comments about errors and I'm already working on something.

If this is OK with you I'd like to address the error-related comments in a separate PR to keep this one focused, as well as to be able to merge it sooner.

acicovic:
@chriszarate, I've attempted to address all your error-related feedback in #1232. Let's take any related conversation there.

chriszarate (dismissed):

acicovic:
@chriszarate, sorry, I had to address a merge conflict. The nitty-gritty is in https://github.com/Parsely/wp-parsely/pull/1093/commits/d64be8e5ea6242d26a2e0b8811193f7962e23af6.

chriszarate (approved):

PR #1098, opened 2022-09-21
Use Parsely->get_options() in the whole codebase

acicovic:
## Description
We had some places in our code where we were using the built-in WordPress `get_option()` function instead of `Parsely->get_options()`, which caters for safe defaults when we call plugin settings. This PR introduces the usage of `Parsely->get_options()` everywhere.

Things of note:
- The `Integration` Interface has been converted to an Abstract Class.
- All built-in integrations must be now be passed a `Parsely` object into their constructor. Custom integrations added by users with the `wp_parsely_add_integration` filter should be unaffected per [our test](https://github.com/Parsely/wp-parsely/blob/8a1f2fcd6401365bdceb1ddef022656ad9d6cb43/tests/Integration/Integrations/IntegrationsTest.php#L36).
- We should  probably back-port this into the 5.5 branch as we have an internal usage scenario for it.
- We should review this carefully, as in case of an issue it could break all internal integrations.

## Motivation and Context
Have safe defaults everywhere.

## How Has This Been Tested?
The changes resulted in test breakages. The tests were fixed and pass now.


GaryJones:
We should be getting rid of the Parsely object, not making it a requirement!

acicovic:
@GaryJones I totally agree with your point and this is the direction we want to take when we find the bandwidth for this type of refactoring.

The reasoning of this PR is that as `Parsely` is already all over the code, adding it in a few more places to have safe defaults everywhere is much less expensive that trying to remove `Parsely` and reworking an implementation.

src/Integrations/class-integration.php
```diff
@@ -10,17 +10,36 @@
...
+	 */
+	protected static $parsely;
+
+	/**
+	 * Constructor.
+	 *
+	 * @param Parsely $parsely Instance of Parsely class.
+	 */
+	public function __construct( Parsely $parsely ) {
+		self::$parsely = $parsely;
```
chriszarate, line 35: why static?

wp-parsely.php
```diff
@@ -168,10 +168,15 @@ function parsely_recommended_widget_register(): void {
  *
  * @since 2.6.0
  *
+ * @param Parsely|string|null $parsely The Parsely object to pass to the integrations.
  * @return Integrations
  */
-function parsely_integrations(): Integrations {
-	$parsely_integrations = new Integrations();
+function parsely_integrations( $parsely = null ): Integrations {
+	if ( empty( $parsely ) || get_class( $parsely ) !== Parsely::class ) {
```
chriszarate, line 178: when would `$parsely` _not_ be null? this function is only called by the init action, correct?

chriszarate (dismissed):

PR #1159, opened 2022-11-02
Metadata Author Archive Builder: Add support for the Co-Authors Plus plugin

acicovic:
## Description
We recently got a bug report about the plugin generating PHP 8 warnings, but from what I could tell this was probably due to our `Author_Archive_Builder` class not supporting the [Co-Authors Plus](https://github.com/Automattic/Co-Authors-Plus) plugin.

In this PR:
- We're adding support for the [Co-Authors Plus](https://github.com/Automattic/Co-Authors-Plus) plugin in our `Author_Archive_Builder` class.
- We make the class more resistant/flexible to things that could go wrong.
- If everything else fails, we use the author's username in the metadata. Previously, we would have an empty string, resulting in `Author -` metadata which was not helpful.

Credit: Thanks to @GaryJones for conducting the initial debugging investigation.

## Motivation and Context
- Support the Co-Authors Plus plugin properly.
- Improve problematic/incomplete metadata that can be an issue for our users.

## How Has This Been Tested?
Manual local testing was conducted, including:
- Testing with the Co-Authors Plus plugin disabled.
- Testing with the Co-Authors Plus plugin enabled:
  - Multiple authors (WP Users)
  - Multiple authors (Guest Authors)
  - `add_filter( 'coauthors_guest_authors_force', '__return_true' );`
  - `add_filter( 'coauthors_guest_authors_enabled', '__return_false' );`

### DIFF
--- a/src/Metadata/class-author-archive-builder.php
+++ b/src/Metadata/class-author-archive-builder.php
@@ -10,6 +10,8 @@ declare(strict_types=1);
 
 namespace Parsely\Metadata;
 
+use WP_User;
+
 /**
  * Implements abstract Metadata Builder class to generate the metadata array
  * for an author archive page.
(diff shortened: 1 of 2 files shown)


GaryJones (commented):
I'd love to see some tests that set up the test environment like CAP creates, to properly check that this bug is actually fixed. This will then also serve when the CAP integration is refactored to its own integration class. 

The changes currently present look sensible but I've not tested, and with no automated tests, I can't approve.

chriszarate (approved):
> I'd love to see some tests that set up the test environment like CAP creates, to properly check that this bug is actually fixed. This will then also serve when the CAP integration is refactored to its own integration class.

I'd love those tests, too. @acicovic Can you please add an issue to record this as a task? Given our limited resources and the impact of this issue, let's get this merged for now. Thanks, all.

acicovic:
I totally agree about the tests. This is now being tracked as part of #1161.

Issue #850, opened 2022-05-09
Allow flexibility on metadata date format


pauarge:
## Is your feature request related to a problem?

We currently render the headline for date page's metadata in a US format (see `class-date-builder.php`). This data is used by the Parse.ly Dashboard to render the page's title.

## Describe the solution you'd like

We'd like to provide more flexibility on how the headline is generated. Specifically, we're targeting the `build_headline` function in `class-date-builder.php`. Instead of using hardcoded formats in `get_the_time` calls, we should use something that takes into account the site's configuration. 

Something like: 

```
get_option('date_format');
get_option('time_format');
```

## Describe alternatives you've considered

Sticking to less flexible date format. The implementation is simpler, but might confuse some non-US users.

## Additional context

See https://github.com/Parsely/wp-parsely/pull/843#discussion_r865807430

acicovic:
For some reason this was not auto-closed, so closing it now.

PR #1179, opened 2022-11-11
Archive Pages: Improve headline metadata by using date format provided by site settings

mehmoodak:
## Description
We used to render the headline for date page's metadata in US format but now its dependent on the `date` and `time` format provided by site

**Closes** https://github.com/Parsely/wp-parsely/issues/850

## How Has This Been Tested?
- Tested manually by visiting the date archive page with different params and verify parsely provided `application/ld+json` attributes.
- Wrote unit tests that verifies every possible headline case for date archive page.

### DIFF
--- a/src/Metadata/class-date-builder.php
+++ b/src/Metadata/class-date-builder.php
@@ -39,6 +39,8 @@ class Date_Builder extends Metadata_Builder {
 	 * @since 3.4.0
 	 */
 	private function build_headline(): void {
+		$site_date_format = get_option( 'date_format' );
+
 		if ( is_year() ) {
 			/* translators: %s: Archive year */
 			$this->metadata['headline'] = sprintf( __( 'Yearly Archive - %s', 'wp-parsely' ), get_the_time( 'Y' ) );
@@ -47,10 +49,11 @@ class Date_Builder extends Metadata_Builder {
 			$this->metadata['headline'] = sprintf( __( 'Monthly Archive - %s', 'wp-parsely' ), get_the_time( 'F, Y' ) );
 		} elseif ( is_day() ) {
 			/* translators: %s: Archive day, formatted as F jS, Y */
-			$this->metadata['headline'] = sprintf( __( 'Daily Archive - %s', 'wp-parsely' ), get_the_time( 'F jS, Y' ) );
+			$this->metadata['headline'] = sprintf( __( 'Daily Archive - %s', 'wp-parsely' ), get_the_time( $site_date_format ) );
 		} elseif ( is_time() ) {
+			$site_time_format = get_option( 'time_format' );
 			/* translators: %s: Archive time, formatted as F jS g:i:s A */
-			$this->metadata['headline'] = sprintf( __( 'Hourly, Minutely, or Secondly Archive - %s', 'wp-parsely' ), get_the_time( 'F jS g:i:s A' ) );
+			$this->metadata['headline'] = sprintf( __( 'Hourly, Minutely, or Secondly Archive - %s', 'wp-parsely' ), get_the_time( "{$site_date_format}:{$site_time_format}" ) );
 		}
 	}
 }
--- a/src/Metadata/class-date-builder.php
+++ b/src/Metadata/class-date-builder.php
@@ -53,7 +53,7 @@ class Date_Builder extends Metadata_Builder {
 		} elseif ( is_time() ) {
 			$site_time_format = get_option( 'time_format' );
 			/* translators: %s: Archive time, formatted as F jS g:i:s A */
-			$this->metadata['headline'] = sprintf( __( 'Hourly, Minutely, or Secondly Archive - %s', 'wp-parsely' ), get_the_time( "{$site_date_format}:{$site_time_format}" ) );
+			$this->metadata['headline'] = sprintf( __( 'Hourly, Minutely, or Secondly Archive - %s', 'wp-parsely' ), get_the_time( "{$site_date_format} : {$site_time_format}" ) );
 		}
 	}
 }
--- a/src/Metadata/class-date-builder.php
+++ b/src/Metadata/class-date-builder.php
@@ -53,7 +53,7 @@ class Date_Builder extends Metadata_Builder {
 		} elseif ( is_time() ) {
 			$site_time_format = get_option( 'time_format' );
 			/* translators: %s: Archive time, formatted as F jS g:i:s A */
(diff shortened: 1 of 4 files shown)


GaryJones:
Logic looks good, but the addition of the new Composer script should be in a separate PR.

tests/Integration/Metadata/DateArchiveTest.php
```diff
@@ -0,0 +1,154 @@
...
+	public function test_daily_archive(): void {
+		$this->go_to( home_url( '/2022/10/31/' ) );
+
+		$parsely_metadata = self::$metadata->construct_metadata( get_post() );
+
+		self::assertEquals( 'Daily Archive - October 31, 2022', $parsely_metadata['headline'] );
+	}
+
+	/**
+	 * Verifies headline of Daily Archive page with users specified date format.
```
acicovic:
```suggestion
	 * Verifies headline of Daily Archive page with user's specified date format.
```

tests/Integration/Metadata/DateArchiveTest.php
```diff
@@ -0,0 +1,154 @@
...
+	public function test_time_archive(): void {
+		$this->go_to( home_url( '/2022/10/31/23' ) );
+
+		$parsely_metadata = self::$metadata->construct_metadata( get_post() );
+
+		self::assertEquals( 'Hourly, Minutely, or Secondly Archive - October 31, 2022:11:59 pm', $parsely_metadata['headline'] );
+	}
+
+	/**
+	 * Verifies headline of Time Archive page with users specified time format.
```
acicovic:
```suggestion
	 * Verifies headline of Time Archive page with user's specified time format.
```

acicovic (changes requested):
This is great, thank you for your contribution and attention to detail!

- Left a couple of minor comment suggestions.
- We should indeed move the `cbf`-related code to another PR for separation of concerns, as well as logging it better into our changelog (which is generated by our PRs).
- To keep code coverage happy, ([see here](https://github.com/Parsely/wp-parsely/actions/runs/3445997202/jobs/5750368877#step:14:149)) we should add the following as `@uses` clauses to every test:
  ```
  Parsely\Metadata::construct_metadata
  Parsely\Metadata\Date_Builder::get_metadata
  Parsely\Metadata\Metadata_Builder::__construct
  Parsely\Metadata\Metadata_Builder::build_basic
  Parsely\Metadata\Metadata_Builder::build_url
  Parsely\Metadata\Metadata_Builder::get_current_url
  Parsely\Parsely::get_options
  ```
- I would also suggest to add `@group metadata` to tests (see other related tests).

mehmoodak:
@acicovic can you please explain the logic behind why we need to add ` @group metadata` in our doc comments of tests?

mehmoodak:
Feedback incorporated please have a look again. Thanks

src/Metadata/class-date-builder.php
```diff
@@ -47,10 +49,11 @@ private function build_headline(): void {
 			$this->metadata['headline'] = sprintf( __( 'Monthly Archive - %s', 'wp-parsely' ), get_the_time( 'F, Y' ) );
 		} elseif ( is_day() ) {
 			/* translators: %s: Archive day, formatted as F jS, Y */
-			$this->metadata['headline'] = sprintf( __( 'Daily Archive - %s', 'wp-parsely' ), get_the_time( 'F jS, Y' ) );
+			$this->metadata['headline'] = sprintf( __( 'Daily Archive - %s', 'wp-parsely' ), get_the_time( $site_date_format ) );
 		} elseif ( is_time() ) {
+			$site_time_format = get_option( 'time_format' );
 			/* translators: %s: Archive time, formatted as F jS g:i:s A */
-			$this->metadata['headline'] = sprintf( __( 'Hourly, Minutely, or Secondly Archive - %s', 'wp-parsely' ), get_the_time( 'F jS g:i:s A' ) );
+			$this->metadata['headline'] = sprintf( __( 'Hourly, Minutely, or Secondly Archive - %s', 'wp-parsely' ), get_the_time( "{$site_date_format}:{$site_time_format}" ) );
```
GaryJones: Whilst it seems that WP core no longer has a default title for `is_time()`, I don't think the use of the `:` between the date and time makes sense here. You might get, for instance, `November 14th:12:40:00`. Too many `:` for my liking. Instead of the `:` between the date and time, consider a space instead.

```suggestion
			$this->metadata['headline'] = sprintf( __( 'Hourly, Minutely, or Secondly Archive - %s', 'wp-parsely' ), get_the_time( "{$site_date_format} {$site_time_format}" ) );
```

The corresponding test(s) will need updating as well.

tests/Integration/Metadata/DateArchiveTest.php
```diff
@@ -0,0 +1,193 @@
...
+		self::factory()->post->create( array( 'post_date' => '2022-10-31 23:59:59' ) );
+	}
+
+	/**
+	 * Setup method called before each test.
+	 */
+	public function set_up(): void {
+		parent::set_up();
+
+		$this->set_permalink_structure( '/%year%/%monthnum%/%day%/%hour%/%minute%/%second%/%postname%/' );
```
GaryJones, line 44: I'd suggest explicitly setting the date and time format here, so you aren't relying on WP core never changing its defaults. It will also make the tests easier to follow.

acicovic:
> @acicovic can you please explain the logic behind why we need to add `@group metadata` in our doc comments of tests?

I suggested this for coherence since it has been done in other tests that test metadata. The idea here is to have `@group metadata` for metadata-related tests (not for all tests).

If you search our codebase, you will see a few other test groups like `ui`, `filters`, `scripts`, etc. I don't think we have a practical use case for them, at least yet.

acicovic (dismissed):

src/Metadata/class-date-builder.php
acicovic: I think that @GaryJones was saying is to add a space instead of a `:` (not adding a space before and after the `:`). But I'll let him clarify on that.

acicovic: @mehmoodak, if there is a specific reason you consider `:` to be a better choice, let us know.

mehmoodak: You are right. I misread this. Making this change

mehmoodak: Change incorporated.

acicovic (approved):
This looks good to me! 👍

PR #1182, opened 2022-11-14
Integration Tests: Replace the deprecated this->factory with self::factory()

### LINKED ISSUE #1121: Consistently use self::factory() in tests
## Describe the bug
The use of the non-static property of `$this->factory->...` in integration tests is deprecated. See https://core.trac.wordpress.org/changeset/54087. The use of the static method `self::factory()->...` is the correct approach.

While the patch includes a back-compat fix for populating `$this->factory`, it's set in the `WP_UnitTestCase_Base::set_up()`, so as of PHP 8.2 and WP trunk, any attempt to use `$this->factory->...` in test suite `set_up()` will fail as `factory` will be `null`.

This plugin doesn't seem to suffer from the `set_up()` issue, but it does have some `$this->factory->...` uses that should be switched over.

mehmoodak:
## Description
Closes: #1121 

The use of the non-static property of `$this->factory->`... in integration tests is deprecated. See https://core.trac.wordpress.org/changeset/54087. The use of the static method `self::factory()->`... is the correct approach.

## How Has This Been Tested?
- Replaced the method and verified that after changes all tests are passing successfully.

### DIFF
--- a/tests/Integration/Metadata/SinglePostTest.php
+++ b/tests/Integration/Metadata/SinglePostTest.php
@@ -663,7 +663,7 @@ final class SinglePostTest extends TestCase {
 		$parsely = new Parsely();
 
 		// Create a single post.
-		$post_id = $this->factory->post->create();
+		$post_id = self::factory()->post->create();
 		$post    = get_post( $post_id );
 
 		unset( $post->post_date );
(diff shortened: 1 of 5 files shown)


acicovic:
Hello,
I see that an empty `factory` file has been added as part of the PR. Is there a reason we need this file, or should it be removed?

mehmoodak:
> I see that an empty factory file has been added as part of the PR. Is there a reason we need this file, or should it be removed?

@acicovic we don't need it .... not sure how it landed there 🤔 but removed it now

acicovic (approved):

PR #1190, opened 2022-11-17
README.md: Update "Tested up to" to 6.1.1

acicovic:
## Description
This PR updates the "Tested up to" field of `README.md` to `6.1.1`.

## Motivation and Context
Keep our `README.md` updated with the latest information.

## How Has This Been Tested?
No code changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 Stable tag: 3.5.2  
 Requires at least: 5.0  
-Tested up to: 6.1.0  
+Tested up to: 6.1.1  
 Requires PHP: 7.1  
 License: GPLv2 or later  
 License URI: https://www.gnu.org/licenses/gpl-2.0.html  


GaryJones (changes requested):
> **Tested up to** – The version of WordPress that the plugin has been tested against. This field ignores minor versions, as plugins shouldn’t break with a minor update. This means a plugin only need to define the major version it is tested against and the WordPress.org plugin directory will automatically add the minor version.
-- https://developer.wordpress.org/plugins/wordpress-org/how-your-readme-txt-works/

acicovic:
Nice find, I was totally unaware of that. So far we've been doing this manually.

acicovic:
@GaryJones, have you seen this working in practice?

GaryJones:
@acicovic If you look at the [readme for 3.5.2 in SVN](https://plugins.trac.wordpress.org/browser/wp-parsely/tags/3.5.2/README.md), you'll see it says tested up to `6.0.2`, but when you look at the sidebar of https://wordpress.org/plugins/wp-parsely/ then it shows `6.0.3`.

i.e. the readme only needed to say `6.0`, and it doesn't need PRs to bump the final extra digit.

acicovic:
Closing in favor of #1197.

PR #1195, opened 2022-11-18
Avoid tracking page views when post is opened in preview mode

### LINKED ISSUE #1155: Do not track post previews
We have reports from Parse.ly support that some times the plugin injects tracking code within post previews. This should not be the case for drafts theoretically, but seems to be happening in practice.

Here are some parts of related communications, to highlight why this is important:

> A user creates a post, previews it while in Draft mode and that preview triggers a page view. Since the URL isn't live yet, we can't collect metadata, and that eventually creates some issues on our side.

> Customers will effectively experience a “duplicate” issue where there’s two versions of a piece surfaced in their dashboard--one without metadata (that showcases the earliest traffic) and one with metadata (sparked after a subsequent automated re-crawl).

> Even more importantly, a post often won't appear on the post list for some time when published. That's because in this scenario, we have a special schedule for recrawl attempts.

Unless there is a valid scenario for tracking post previews, make it so post previews never get tracked (even when their status is published).

mehmoodak:
## Description
When published post is opened in preview mode then we were counting it as page viewed on Parsely so now we fixed it and don't count the posts opened in preview mode.

Closes: #1155 

## How Has This Been Tested?
1. Added integration tests
2. Tested manually

### DIFF
--- a/tests/Integration/ScriptsTest.php
+++ b/tests/Integration/ScriptsTest.php
@@ -207,14 +207,12 @@ final class ScriptsTest extends TestCase {
 		self::$scripts->register_scripts();
 		self::$scripts->enqueue_js_tracker();
 
-		// Verify that tracker script is registered and enqueued.
 		$this->assert_script_statuses(
 			'wp-parsely-tracker',
 			array( 'registered' ),
 			array( 'enqueued' )
 		);
 
-		// Verify that loader script is registered and enqueued.
 		$this->assert_script_statuses(
 			'wp-parsely-loader',
 			array( 'registered' ),
@@ -245,14 +243,12 @@ final class ScriptsTest extends TestCase {
 		self::$scripts->register_scripts();
 		self::$scripts->enqueue_js_tracker();
 
-		// Verify that tracker script is registered and enqueued.
 		$this->assert_script_statuses(
 			'wp-parsely-tracker',
 			array( 'registered' ),
 			array( 'enqueued' )
 		);
 
-		// Verify that loader script is registered and enqueued.
 		$this->assert_script_statuses(
 			'wp-parsely-loader',
 			array( 'registered' ),
--- a/tests/Integration/ScriptsTest.php
+++ b/tests/Integration/ScriptsTest.php
@@ -187,6 +187,80 @@ final class ScriptsTest extends TestCase {
 	}
 
 	/**
+	 * Verifies that tracker scripts are not loading for drafted posts
+	 *
+	 * @covers \Parsely\Scripts::enqueue_js_tracker
+	 *
+	 * @uses Parsely\Parsely::api_key_is_set
+	 * @uses Parsely\Parsely::get_api_key
+	 * @uses Parsely\Parsely::get_options
+	 * @uses Parsely\Parsely::get_tracker_url
+	 * @uses Parsely\Scripts::__construct
+	 * @uses Parsely\Scripts::register_scripts
+	 *
+	 * @group scripts
+	 */
+	public function test_should_not_enqueue_tracker_scripts_for_drafted_posts(): void {
+		$this->set_admin_user();
+		$this->go_to_new_post( 'draft' );
+		
+		self::$scripts->register_scripts();
+		self::$scripts->enqueue_js_tracker();
+
+		// Verify that tracker script is registered and enqueued.
+		$this->assert_script_statuses(
+			'wp-parsely-tracker',
+			array( 'registered' ),
+			array( 'enqueued' )
+		);
+
+		// Verify that loader script is registered and enqueued.
+		$this->assert_script_statuses(
+			'wp-parsely-loader',
+			array( 'registered' ),
+			array( 'enqueued' )
+		);
+	}
+
+	/**
+	 * Verifies that tracker scripts are not loading for published posts in preview mode
+	 *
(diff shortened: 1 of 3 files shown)


tests/Integration/ScriptsTest.php
```diff
@@ -186,6 +186,80 @@ public function test_enqueue_js_tracker(): void {
...
+		$this->go_to_new_post( 'draft' );
+		
+		self::$scripts->register_scripts();
+		self::$scripts->enqueue_js_tracker();
+
+		// Verify that tracker script is registered and enqueued.
+		$this->assert_script_statuses(
+			'wp-parsely-tracker',
+			array( 'registered' ),
+			array( 'enqueued' )
```
GaryJones, line 213: If the test method is to check the tracking script is not enqueued, why does this line suggest it's looking for an enqueued state?

mehmoodak: It is not looking for enqueued state instead it is looking for assert false on `array( 'enqueued' )`. Please have look at [assert_script_statuses](https://github.com/Parsely/wp-parsely/blob/changes/avoid-tracking-views-of-published-post-in-preview-mode/tests/Integration/ScriptsTest.php#L566) function signature.

GaryJones: Oh, of course 🤦🏻‍♂️

GaryJones: In which case the comments above these calls in the various test methods need reviewing and fixing.

i.e.
```diff
- // Verify that tracker script is registered and enqueued.
+ // Verify that tracker script is registered but not enqueued.
```

mehmoodak: I think we shouldn't have those comments because those are just explaining the already good written code. So removing these.

chriszarate: I had the same confusion as Gary. Boolean arguments (or arguments that have Boolean behavior, in this case) are inherently confusing. No need to tackle in this PR, but we should try to avoid them in the future.

chriszarate (dismissed):
+3 lines of code, +100 lines of test code

My kind of PR! 👏

tests/Integration/ScriptsTest.php
mehmoodak: @chriszarate here I am using the already available `assert_script_statuses` function ...... so in future how can we avoid it if we need to use the same function?

Should we create new functions with more clear names?

tests/Integration/ScriptsTest.php
```diff
@@ -186,6 +186,76 @@ public function test_enqueue_js_tracker(): void {
 		self::assertEquals( 1, count( $wp_scripts->registered['wp-parsely-loader']->extra ) );
 	}
 
+	/**
+	 * Verifies that tracker scripts are not loading for drafted posts
```
```suggestion
	 * Verifies that tracker scripts are not loading for drafted posts.
```

tests/Integration/ScriptsTest.php
```diff
@@ -186,6 +186,76 @@ public function test_enqueue_js_tracker(): void {
...
+
+		$this->assert_script_statuses(
+			'wp-parsely-loader',
+			array( 'registered' ),
+			array( 'enqueued' )
+		);
+	}
+
+	/**
+	 * Verifies that tracker scripts are not loading for published posts in preview mode
```
```suggestion
	 * Verifies that tracker scripts are not loading for published posts in preview mode.
```

tests/Integration/TestCase.php
```diff
@@ -154,15 +155,42 @@ public function create_test_taxonomy( string $taxonomy_key, string $term_name )
 	}
 
 	/**
-	 * Creates a new post and navigates to it.
+	 * Creates a new post
```
acicovic:
```suggestion
	 * Creates a new post.
```

tests/Integration/TestCase.php
```diff
@@ -154,15 +155,42 @@ public function create_test_taxonomy( string $taxonomy_key, string $term_name )
...
+	 */
+	public function go_to_new_post( string $post_status = 'publish' ): int {
+		$post_id = $this->create_test_post( $post_status );
 		$this->go_to( '/?p=' . $post_id );
 
 		return $post_id;
 	}
+
+	/**
+	 * Set current user as admin
```
acicovic:
```suggestion
	 * Sets current user as admin.
```

acicovic (commented):
Thank you for your work! Left some minor suggestions for comments.

tests/Integration/ScriptsTest.php
acicovic: We can create another function if we find this one confusing, but let's leave this for another PR. When we create the new function, we can also replace the old calls with the new ones everywhere in our code.

acicovic (approved):

tests/Integration/ScriptsTest.php
GaryJones: I agree with Chris and Alex (and the removal of the comments). Another PR is the better place to address this.

Move the boolean aspect to the assertion name:
```
$this->assert_script_is_registered( 'wp-parsely-tracker' );
$this->assert_script_is_not_registered( 'wp-parsely-tracker' );
$this->assert_script_is_enqueued( 'wp-parsely-tracker' );
$this->assert_script_is_not_enqueued( 'wp-parsely-tracker' );
```

PR #1216, opened 2022-11-23
Content Helper: Improve UI of Related Top Performing Posts

### LINKED ISSUE #1206: Redesign Content Helper's "Related Top-Performing Posts" panel
The original content that was within the Content Helper is now moving to a dedicated panel called "Related Top-Performing Posts", which will need some redesign. We already have a mock-up from @garrettvercoe for that, but we should also take into account the implementation of #940 which is scheduled to start soon.

- Given the upcoming implementation of #940, see if this has any impacts on design.
- Decide/agree on a new mock-up if needed.
- Implement the redesign.

mehmoodak:
## Description
Updated UI of related top performing posts

Closes: #1206 

## How Has This Been Tested?
- Tested manually and make sure that existing tests are running fine.
- Added structure tests for the updated layout

image


tests/js/blocks/content-helper/structure.test.js
```diff
@@ -4,20 +4,227 @@
...
+			.mockImplementation( () => Promise.reject( {
+				errors: {
+					parsely_site_id_not_set: 'error msg',
+				},
+			} ) );
+
+		render( <RelatedTopPostList /> );
+		expect( getSpinner() ).toBeInTheDocument();
+
+		await waitFor( () => screen.findByTestId( 'parsely-contact-us' ), { timeout: 3000 } );
```
mehmoodak: Can we do this without passing timeout and just look for when the element appears on screen? ..... I tired multiple things but none of them worked

tests/js/blocks/content-helper/structure.test.js
```diff
@@ -4,20 +4,227 @@
...
+
+		expect( getRelatedTopPostsMock ).toHaveBeenCalled();
```
mehmoodak: I wanted to use `toHaveBeenCalledTimes` function which also verifies how much time it is called but it seems like mock counts are persisting between each tests so how can we clear these mock counts between tests?

Tried: `clearAllMocks, resetAllMocks, restoreAllMocks` in `breforeEach and afterEach` callbacks but none of them worked

tests/js/blocks/content-helper/structure.test.js
chriszarate: Does it display on initial render or does it wait for a state update to render? If the latter you can look into using `act`

tests/js/blocks/content-helper/structure.test.js
chriszarate: Sounds like, because you are `await`ing, the component renders multiple times and the function is called multiple times.

mehmoodak: Yes it does call multiple times because of the retry logic present in catch part of the API ...... but the counter of calling mocks should reset if I am doing `clearAllMocks, resetAllMocks, restoreAllMocks` at the start of the test but it isn't happening

tests/js/blocks/content-helper/structure.test.js
mehmoodak: By using act you mean this?

```
await act( async () => {
	waitFor( () => screen.findByTestId( 'parsely-contact-us' ) );
} );
```

It is not working while the above statement is working fine.

src/blocks/content-helper/components/related-top-post-list-item.tsx
```diff
@@ -2,35 +2,35 @@
...
-					<p>
-						<Button href={ post.url } target="_blank" variant="primary">{ __( 'Open Post', 'wp-parsely' ) }</Button> { ' ' }
-						<Button href={ post.statsUrl } target="_blank" variant="secondary">{ __( 'Post Stats', 'wp-parsely' ) }</Button>
-					</p>
-				</CardBody>
-			</Card>
-		</>
+		<div className="parsely-top-post" data-testid="parsely-top-post">
+			<div className="parsely-top-post-title">
+				<a className="parsely-top-post-stats-link" href={ post.statsUrl } target="_blank" rel="noreferrer" title={ __( 'View in Parse.ly', 'wp-parsely' ) }>
```
GaryJones:
```suggestion
				<a className="parsely-top-post-stats-link" href={ post.statsUrl } target="_blank" rel="noreferrer" title={ __( 'View in Parse.ly (opens new tab)', 'wp-parsely' ) }>
```

src/blocks/content-helper/components/related-top-post-list-item.tsx
```diff
@@ -2,35 +2,35 @@
...
+					{ post.title }
+				</a>
+				<a className="parsely-top-post-link" href={ post.url } target="_blank" rel="noreferrer" title={ __( 'View Published Post', 'wp-parsely' ) }>
```
GaryJones:
```suggestion
				<a className="parsely-top-post-link" href={ post.url } target="_blank" rel="noreferrer" title={ __( 'View Published Post (opens new tab)', 'wp-parsely' ) }>
```

src/blocks/content-helper/components/related-top-post-list-item.tsx
```diff
@@ -2,35 +2,35 @@
...
+					<PublishedLinkIcon />
+				</a>
+			</div>
+			<ul className="parsely-top-post-info">
+				<li className="parsely-top-post-date">{ post.date }</li>
+				<li className="parsely-top-post-author">{ post.author }</li>
+				<li className="parsely-top-post-views"><ViewsIcon />{ post.views }</li>
+			</ul>
```
GaryJones: This will be read out be screen readers as something like "List. 3 items. Item 1. 1st January 2022. Item 2. John Smith. Item 3: 84." That's quite verbose.

I'd rather see this marked up as just `span` elements but with some screen reader text to emulate the visual separations.

```html
<p className="parsely-top-post-info">
	<span className="parsely-top-post-date"><span className="screen-reader-text">Date </span>{ post.date }</span>
	<span className="parsely-top-post-author"><span className="screen-reader-text">Author </span>{ post.author }</span>
	<span className="parsely-top-post-views"><span className="screen-reader-text">Number of views </span><ViewsIcon />{ post.views }</span>
</p>
```

Alternatively, `dl` with `dt` (screen-reader-text) and `dd` could be used.

src/blocks/content-helper/components/related-top-post-list-item.tsx
```diff
@@ -2,35 +2,35 @@
...
-						<li>{ __( 'Author:', 'wp-parsely' ) } { post.author }</li>
-					</ul>
-					<p>
-						<Button href={ post.url } target="_blank" variant="primary">{ __( 'Open Post', 'wp-parsely' ) }</Button> { ' ' }
-						<Button href={ post.statsUrl } target="_blank" variant="secondary">{ __( 'Post Stats', 'wp-parsely' ) }</Button>
-					</p>
-				</CardBody>
-			</Card>
-		</>
+		<div className="parsely-top-post" data-testid="parsely-top-post">
```
GaryJones: The component is called a list item, but it's not marked up as a list item? This should be `li`, and with the `ol` wrapping it in the RelatedTopPostList component since it's an ordered list.

src/blocks/content-helper/components/related-top-post-list.tsx
```diff
@@ -56,31 +73,44 @@ function RelatedTopPostList() {
...
 		// Error coming from the WordPress REST API.
 		const errorMessage = JSON.stringify( error ).match( /\[\"(.*?)\"\]/ )[ 1 ];
-		return <p>{ __( 'Error:', 'wp-parsely' ) } { errorMessage }</p>;
+		return <p data-testid="wp-api-error">{ __( 'Error:', 'wp-parsely' ) } { errorMessage }</p>;
 	}
 
 	// Show related top posts list.
-	const postList = posts.map( ( post ) => <RelatedTopPostListItem key={ post.id } post={ post } /> );
+	const postList: JSX.Element = (
+		<div className="parsely-top-posts">
```
GaryJones: I think this should be the `ol`.

src/blocks/content-helper/content-helper-provider.ts
```diff
@@ -80,7 +83,7 @@ class ContentHelperProvider {
 			return Promise.reject( error );
 		}
 
-		let message = `${ __( 'Top-performing posts', 'wp-parsely' ) } ${ apiQuery.message }.`;
+		let message = `${ __( 'Top-performing posts', 'wp-parsely' ) } ${ apiQuery.message } ${ __( `in last`, 'wp-parsely' ) } ${ RELATED_POSTS_DEFAULT_TIME_RANGE } ${ __( `days.`, 'wp-parsely' ) }`;
```
GaryJones: Instead of three separate strings:
- `Top-performing posts`
- `in last`
- `days.`
... this should be a single string with placeholders and values that are sprintf'd in as needed. This makes it a far easier string for translators to translate correctly.

For PHP, that would be something like
- `Top-performing posts %s in last %s days.`

src/blocks/content-helper/icons/published-link-icon.tsx
```diff
@@ -0,0 +1,20 @@
+/**
+ * WordPress dependencies
+ */
+import { SVG, Path } from '@wordpress/components';
+
+export const ViewsIcon = () => (
+	<SVG
```
GaryJones, line 7: Please see https://css-tricks.com/accessible-svgs/#aa-example-2-standalone-icon-decorative and add the `aria-hidden` attribute as necessary.

src/blocks/content-helper/icons/views-icon.tsx
```diff
@@ -0,0 +1,21 @@
+/**
+ * WordPress dependencies
+ */
+import { SVG, Path } from '@wordpress/components';
+
+export const ViewsIcon = () => (
+	<SVG
```
GaryJones, line 7: As above - `aria-hidden` is needed since this is purely visual decoration.

GaryJones (changes requested):
Some accessibility considerations to address please.

src/blocks/content-helper/icons/published-link-icon.tsx
mehmoodak: Added in next line.

src/blocks/content-helper/icons/views-icon.tsx
mehmoodak: Added in next line.

mehmoodak:
@GaryJones thanks for your valuable feedback. Implemented the feedback please have a look again. Thanks

package.json
```diff
@@ -68,7 +68,8 @@
 		"@wordpress/scripts": "^24.6.0",
 		"@wordpress/url": "^3.23.0",
 		"concurrently": "^7.6.0",
-		"eslint-plugin-jest": "^27.1.6",
+		"dayjs": "^1.11.6",
+		"eslint-plugin-jest": "^27.1.5",
```
GaryJones: Any reason this constraint has dropped down a patch version?

src/blocks/content-helper/components/related-top-post-list.tsx
```diff
@@ -26,8 +27,17 @@ function RelatedTopPostList() {
...
 		const fetchPosts = async ( retries: number ) => {
 			ContentHelperProvider.getRelatedTopPosts()
-				.then( ( result ) => {
-					setPosts( result.posts );
+				.then( ( result ): void => {
+					const mappedPosts: RelatedTopPostData[] = result.posts.map(
+						( post: RelatedTopPostData ): RelatedTopPostData => (
+							{
+								...post,
+								date: dayjs( post.date ).format( 'MMM D, YYYY' ),
```
GaryJones: This is a very US-centric date format. Is there an option to use either the site format, or a pre-existing user-defined / locale format?

src/blocks/content-helper/content-helper-provider.ts
```diff
@@ -80,7 +83,8 @@ class ContentHelperProvider {
 			return Promise.reject( error );
 		}
 
-		let message = `${ __( 'Top-performing posts', 'wp-parsely' ) } ${ apiQuery.message }.`;
+		/* translators: %s: message from api, %d: default time range of related posts api */
```
GaryJones:
```suggestion
		/* translators: %s: message such as "from category Foo", %d: number of days */
```

Translators won't care where the placeholder values come from; they only need an example of what it could look like.

src/blocks/content-helper/content-helper-provider.ts
```diff
@@ -138,22 +140,22 @@ class ContentHelperProvider {
 		if ( tag ) {
 			return ( {
 				query: { limit, tag },
-				message: `${ __( 'with the tag', 'wp-parsely' ) } "${ tag.name }"`,
+				message: `${ __( 'with tag', 'wp-parsely' ) } "${ tag.name }"`,
```
GaryJones: Needs sprintf() - the placeholder may not be the last word in other locales.

src/blocks/content-helper/content-helper-provider.ts
```diff
@@ -138,22 +140,22 @@ class ContentHelperProvider {
...
 			} );
 		}
 
 		// A category exists.
 		if ( category?.name ) {
 			return ( {
 				query: { limit, section: category.name },
-				message: `${ __( 'in the category', 'wp-parsely' ) } "${ category.name }"`,
+				message: `${ __( 'from category', 'wp-parsely' ) } "${ category.name }"`,
```
GaryJones: Needs sprintf() - the placeholder may not be the last word in other locales.

src/blocks/content-helper/content-helper-provider.ts
```diff
@@ -138,22 +140,22 @@ class ContentHelperProvider {
...
 			} );
 		}
 
 		// Only the post author exists.
 		return ( {
 			query: { limit, author: author.name },
-			message: `${ __( 'by the author', 'wp-parsely' ) } "${ author.name }"`,
+			message: `${ __( 'by author', 'wp-parsely' ) } "${ author.name }"`,
```
GaryJones: Needs sprintf() - the placeholder may not be the last word in other locales.

GaryJones (changes requested):
The new markup makes sense, but I've not pulled it down to test it.

Some other items...

package.json
mehmoodak: My bad .... will revert it

src/blocks/content-helper/components/related-top-post-list.tsx
mehmoodak: We have a very limited space on sidebar so seems like using anything else than this can move things to second line which will not look from UI perspective .... I think for now its better to stick with this and in future we can look into this if there are any feedback from customer to change this.

BTW for different languages can I use `__` function here? 🤔

acicovic: Hey @mehmoodak, just FYI, I had a similar issue and here is the solution I came up with:

https://github.com/Parsely/wp-parsely/blob/a87b21b95c6435206c479ad12373c1e5f9996e6d/src/blocks/content-helper/current-post-details/component.tsx#L106-L110

This outputs a short date, taking into account the current language of the page. It also does not introduce any external dependencies. And it is customizable so you can format the date the way you want. You can give it a try if you want.

GaryJones: > We have a very limited space on sidebar so seems like using anything else than this can move things to second line which will not look from UI perspective

Let me introduce you to those using German as their locale in wp-admin... :D 

(German is notorious for using longer words without breaks, so if you're short of space for anything in English, it needs a re-think.)

src/blocks/content-helper/content-helper-provider.ts
```diff
@@ -80,7 +83,8 @@ class ContentHelperProvider {
 			return Promise.reject( error );
 		}
 
-		let message = `${ __( 'Top-performing posts', 'wp-parsely' ) } ${ apiQuery.message }.`;
+		/* translators: %s: message such as "from category Foo", %d: number of days */
+		let message = sprintf( __( 'Top-performing posts %1$s in last %2$d days.', 'wp-parsely' ), apiQuery.message, RELATED_POSTS_DEFAULT_TIME_RANGE );
```
GaryJones, line 87: In English, this string will be broken if the number of days is 1.

Use https://developer.wordpress.org/block-editor/reference-guides/packages/packages-i18n/#n to provide a singular and plural string, and let the i18n system use the correct [translated] string.

GaryJones: Because of the extra string placeholder, it makes it a little more complex, so this may need to be of the form:

```js
let message = _n( 
	sprintf(
		__('Top-performing posts %1$s in last %2$d day.', 'wp-parsely' ),
		apiQuery.message,
		RELATED_POSTS_DEFAULT_TIME_RANGE
	),
	sprintf(
		__('Top-performing posts %1$s in last %2$d days.', 'wp-parsely' ),
		apiQuery.message,
		RELATED_POSTS_DEFAULT_TIME_RANGE
	),
	RELATED_POSTS_DEFAULT_TIME_RANGE
);
```

That is (working from the inside out), it creates a singular and plural translatable string, which after translation then has its placeholders replaced, and then the correct string is assigned to `message` depending on the value of the time range constant.

src/blocks/content-helper/content-helper-provider.ts
```diff
@@ -138,22 +140,25 @@ class ContentHelperProvider {
...
 			} );
 		}
 
 		// A category exists.
 		if ( category?.name ) {
 			return ( {
 				query: { limit, section: category.name },
-				message: `${ __( 'in the category', 'wp-parsely' ) } "${ category.name }"`,
+				/* translators: %s: message such as "from category Foo" */
+				message: sprintf( __( 'from category "%1$s"', 'wp-parsely' ), category.name ),
```
GaryJones: Since posts could be in multiple categories, would this be better as:

```suggestion
				message: sprintf( __( 'in category "%1$s"', 'wp-parsely' ), category.name ),
```

GaryJones (changes requested):

src/blocks/content-helper/content-helper-provider.ts
mehmoodak: I think this

```
Top-performing posts from category "Developers" in last 3 days.
```

is better than this

```
Top-performing posts in category "Developers" in last 3 days.
```

but please do confirm if you think change should be done.

src/blocks/content-helper/content-helper-provider.ts
mehmoodak: Actually I was not making these changes because by default API returns data for last 3 days and have a plan to do these changes while adding API params filter (next ticket in which I will work).

BTW based on your feedback I did tried to do this but facing following error which I was unable to solve

image

Please do let me know the proper way to solve this and then I will make the changes in the next PR.

src/blocks/content-helper/components/related-top-post-list.tsx
acicovic: We should strive to keep things short (which promotes our best design and may work on most languages). For languages that will naturally be too long, there should be a fallback that is not perfect, but is at least viewable. If this is not possible (it looks bad when long), then maybe we would need feedback from @garrettvercoe.

src/blocks/content-helper/content-helper-provider.ts
acicovic: I don't know which is better.

src/blocks/content-helper/content-helper-provider.ts
acicovic: I don't think that we have to look into this in this PR, since returning data for 1 day is not possible. When/if returning data for 1 day becomes available, we can look into plural/singular options.

src/blocks/content-helper/components/related-top-post-list.tsx
mehmoodak: Thanks @acicovic for sharing the example. Implemented it and removed the `dayjs` dependency which is now extra.

As far as long names are concerned the UI will not break so I think we are good here :)

src/blocks/content-helper/content-helper-provider.ts
GaryJones: "In" sounds better to me and the sample survey of my 11yo twins!

mehmoodak: Feedback Implemented.

tests/js/blocks/content-helper/structure.test.js
```diff
@@ -4,20 +4,213 @@
...
-		const spinner = screen.getByRole( 'presentation' );
+		expect( getRelatedTopPostsFn ).toHaveBeenCalled();
+
+		const spinner = getSpinner();
+		expect( spinner ).toBeInTheDocument();
 		expect( spinner ).toBeVisible();
-		expect( spinner ).toHaveClass( 'components-spinner' );
 	} );
+
+	test( 'should show contact us message when parsely site id is not set', async () => {
```
acicovic: For consistency, let's stick with "Parse.ly" (with the proper capitalization and the period). The only exception to that would be when referring to our `Parsely` classes, namespaces, etc, since these do not contain the period.

Also, it would be generally better to capitalize "Site ID", as it is in our options page.

tests/js/blocks/content-helper/structure.test.js
```diff
@@ -4,20 +4,213 @@
...
+		const getRelatedTopPostsFn = getRelatedTopPostsMockFn( () => Promise.reject( {
+			errors: {
+				parsely_site_id_not_set: 'error msg',
+			},
+		} ) );
+
+		expect( await verifyContactUsMessage( getRelatedTopPostsFn ) ).toBeTruthy();
+	} );
+
+	test( 'should show contact us message when parsely secret is not set', async () => {
```
acicovic: For clarity and consistency, it is better to refer to this as "Parse.ly API Secret" as in our options page.

tests/js/blocks/content-helper/structure.test.js
```diff
@@ -4,20 +4,213 @@
...
+		expect( getRelatedTopPostsFn ).toHaveBeenCalled();
+		expect( getSpinner() ).toBeNull();
+
+		const wpApiError = screen.queryByTestId( 'wp-api-error' );
+		expect( wpApiError ).toBeInTheDocument();
+		expect( wpApiError ).toBeVisible();
+		expect( wpApiError.textContent ).toEqual( 'Error: fake error from WP api' );
+	} );
+
+	test( 'should show not data message when there is no tag, category or author in the post', async () => {
```
acicovic:
```suggestion
	test( 'should not show data message when there is no tag, category or author in the post', async () => {
```

tests/js/blocks/content-helper/structure.test.js
```diff
@@ -4,20 +4,213 @@
...
+			},
+		} ) );
+
+		expect( await verifyContactUsMessage( getRelatedTopPostsFn ) ).toBeTruthy();
+	} );
+
+	test( 'should show contact us message when parsely secret is not set', async () => {
+		const getRelatedTopPostsFn = getRelatedTopPostsMockFn( () => Promise.reject( {
+			errors: {
+				parsely_api_secret_not_set: 'error msg',
```
acicovic: Let's always start with a capital letter in messages. Also, let's try to avoid word abbreviations ("msg" instead of "message") unless there is a good reason for it.

tests/js/blocks/content-helper/structure.test.js
```diff
@@ -4,20 +4,213 @@
...
+			},
+		} ) );
+
+		expect( await verifyContactUsMessage( getRelatedTopPostsFn ) ).toBeTruthy();
+	} );
+
+	test( 'should show error message when API returns the error', async () => {
+		const getRelatedTopPostsFn = getRelatedTopPostsMockFn( () => Promise.reject( {
+			message: 'fake error from api',
```
acicovic: In some places, "API" is written as "api". Let's capitalize this.

src/blocks/content-helper/components/related-top-post-list.tsx
```diff
@@ -56,31 +73,44 @@ function RelatedTopPostList() {
 
 		// Error coming from apiFetch.
 		if ( error?.message ) {
-			return <p>{ __( 'Error:', 'wp-parsely' ) } { error.message }</p>;
+			return <p className="parsely-top-posts-desc" data-testid="api-error">{ __( 'Error:', 'wp-parsely' ) } { error.message }</p>;
```
acicovic: "desc" is usually associated to descending sorting order, and this is what came to mind when I saw this. If we mean "description" here, we could go with "descr" or the whole word.

src/blocks/content-helper/content-helper-provider.ts
```diff
@@ -41,6 +41,9 @@ interface GetRelatedTopPostsResult {
 	posts: RelatedTopPostData[];
 }
 
+export const RELATED_POSTS_DEFAULT_LIMIT = 5;
+export const RELATED_POSTS_DEFAULT_TIME_RANGE = 3; // in days
```
acicovic:
```suggestion
export const RELATED_POSTS_DEFAULT_TIME_RANGE = 3; // In days.
```

Let's attempt to respect WP commenting guidelines wherever we can. This gives consistency across all of our comments.

src/blocks/content-helper/components/related-top-post-list-item.tsx
```diff
@@ -2,35 +2,35 @@
...
+			<div className="parsely-top-post-title">
+				<a className="parsely-top-post-stats-link" href={ post.statsUrl } target="_blank" rel="noreferrer" title={ __( 'View in Parse.ly (opens new tab)', 'wp-parsely' ) }>
+					{ post.title }
+				</a>
+				<a className="parsely-top-post-link" href={ post.url } target="_blank" rel="noreferrer" title={ __( 'View Published Post (opens new tab)', 'wp-parsely' ) }>
+					<PublishedLinkIcon />
+				</a>
+			</div>
+			<p className="parsely-top-post-info">
+				<span className="parsely-top-post-date"><span className="screen-reader-text">Date </span>{ post.date }</span>
```
acicovic, line 29: I would guess that the answer is yes, but just wanted to verify that the internal `<span>`s are needed here?

acicovic (commented):
Thank you so much for your work 😃

I've left a question and some remarks mainly about comments and messages. I apologize if this seems like nitpicking. The reason is that I once had to revise the whole codebase's comments, so my intention is just to maintain this as well as we can 😃

Another thing you might be aware of but I wanted to point out just in case, is that using test IDs should be our last resort. However, it is fine to use them if there is no alternatives or if alternative solutions would become too complicated.

acicovic:
For completeness, noting here that there will be some conflicts between this PR and the one I'm currently working on. This is fine and once both PRs are merged we can do some quick refactorings so we don't duplicate code or have many files for the same purpose.

src/blocks/content-helper/components/related-top-post-list-item.tsx
mehmoodak: Yes they are needed .... 1 for styling and 1 for a11y.

mehmoodak:
@acicovic feedback is implemented. Please have a look again. Thanks

acicovic (approved):
@mehmoodak, thank you for the changes. From my side, I'm approving this.

If your questions regarding `structure.test.js` are still open, let's see if we get some feedback from @chriszarate.

mehmoodak:
> If your questions regarding structure.test.js are still open, let's see if we get some feedback from @chriszarate.

Questions are still opened but not a blocker for this PR

acicovic:
@mehmoodak in that case and unless you're waiting for specific feedback, feel free to merge.

PR #1220, opened 2022-11-29
Content Helper: Fix tag parameter while requesting Parse.ly Analytics API

mehmoodak:
## Description
Instead of passing just tag slug to the Parsely Analytics API we were sending the whole tag object due to which the url with URL params will become following

```
http://parsely.test/index.php?rest_route=/wp-parsely/v1/analytics/posts&limit=5&tag[id]=5&tag[count]=1&tag[description]=&tag[link]=http://parsely.test/?tag=bigwp&tag[name]=BigWP&tag[slug]=bigwp&tag[taxonomy]=post_tag&tag[_links][self][0][href]=http://parsely.test/index.php?rest_route=/wp/v2/tags/5&tag[_links][collection][0][href]=http://parsely.test/index.php?rest_route=/wp/v2/tags&tag[_links][about][0][href]=http://parsely.test/index.php?rest_route=/wp/v2/taxonomies/post_tag&tag[_links][wp:post_type][0][href]=http://parsely.test/index.php?rest_route=%2Fwp%2Fv2%2Fposts&tags=5&tag[_links][curies][0][name]=wp&tag[_links][curies][0][href]=https://api.w.org/{rel}&tag[_links][curies][0][templated]=true&_locale=user
```

This does show us results but those were just top performing posts not related to the tag. Now after the fix the url will become like following

```
http://parsely.test/index.php?rest_route=/wp-parsely/v1/analytics/posts&limit=5&tag=bigwp&_locale=user
```

## How Has This Been Tested?
- Make sure that existing tests are running fine.


src/blocks/content-helper/content-helper-provider.ts
```diff
@@ -137,9 +137,9 @@ class ContentHelperProvider {
 		}
 
 		// A tag exists.
-		if ( tag ) {
+		if ( tag?.slug ) {
 			return ( {
-				query: { limit, tag },
+				query: { limit, tag: tag.slug },
```
mehmoodak, line 142: Using slug here because passing `name` doesn't show anything unlike category case.

acicovic (approved):
Thank you!

PR #1224, opened 2022-12-01
Integration Tests: Fix flaky tests related to Parse.ly metadata

### LINKED ISSUE #1205: Fix failing integration tests (Undefined index/array key)
## Describe the bug
Recently, our integration tests have been randomly failing more and more, with error messages such as these:

```
Single Post (Parsely\Tests\Integration\StructuredData\SinglePost)
✘ Custom taxonomy as tags in single post
│
│ Undefined index: keywords
│
│ /home/runner/work/wp-parsely/wp-parsely/tests/Integration/Metadata/SinglePostTest.php:330
│ phpvfscomposer:///home/runner/work/wp-parsely/wp-parsely/vendor/phpunit/phpunit/phpunit:97
```
```
Single Post (Parsely\Tests\Integration\StructuredData\SinglePost)
✘ Post featured image urls in metadata are correct
│
│ Undefined index: image
│
│ /home/runner/work/wp-parsely/wp-parsely/tests/Integration/Metadata/SinglePostTest.php:990
│ phpvfscomposer:///home/runner/work/wp-parsely/wp-parsely/vendor/phpunit/phpunit/phpunit:97
```
```
Single Post (Parsely\Tests\Integration\StructuredData\SinglePost)
✘ Post date with modified after created date in metadata
│
│ Undefined index: dateCreated
│
│ /home/runner/work/wp-parsely/wp-parsely/tests/Integration/Metadata/SinglePostTest.php:860
│ phpvfscomposer:///home/runner/work/wp-parsely/wp-parsely/vendor/phpunit/phpunit/phpunit:97
```
```
Single Post (Parsely\Tests\Integration\StructuredData\SinglePost)
✘ Custom taxonomy as section in single post
│
│ Undefined array key "articleSection"
│
│ /home/runner/work/wp-parsely/wp-parsely/tests/Integration/Metadata/SinglePostTest.php:482
│
```
I've seen other sporadic errors, but we should fix this at first and then go from there. We can focus on the `SinglePostTest.php` file where most (if not all) of the errors seem to occur. Fixing those will reveal whether there are also other tests to work on.

## To Reproduce
Posting a PR often results in such errors. A comprehensive log of those exists in our [Actions](https://github.com/Parsely/wp-parsely/actions) page.

## Expected behavior
We should not be getting this type of errors.

mehmoodak:
## Description
Our integeration tests fails rarely before but after merging #1179 the frequency increases but the tests usually pass on the second run.

In almost all cases the root cause of the function is the implementation of [construct_metadata function](https://github.com/Parsely/wp-parsely/blob/fixes/flaky-integeration-tests/src/class-metadata.php#L54) in which we were using the conditional functions which requires that right `WP_QUERY` is called but in our tests the value from last running test were interferring with the current running test and failing it. The thing to note here is that the tests in question were flaky and happens when only certain race coditions are met and due to this it took a lot of time to find the root cause.

Fixes: #1205 

## How does the PR addresses the issue?
In real scenario we navigate to the page and make sure that right metadata is present but the tests which were failing were not navigating to the post due to which `WP_QUERY` was not getting updated and conditional values from last tests are causing issues. To fix the tests we navigated to the post before comparing metadata values.

## How Has This Been Tested?
- Just fixes the tests and ran it `1000+` times to be sure that there is no flakiness now before fix the issue were appearing in almost every `25` tries.

### DIFF
--- a/tests/Integration/Endpoints/RestMetadataTest.php
+++ b/tests/Integration/Endpoints/RestMetadataTest.php
@@ -195,7 +195,7 @@ final class RestMetadataTest extends TestCase {
 		self::set_options( array( 'apikey' => 'testkey' ) );
 		$post_id = self::factory()->post->create();
 
-		// Navigate to post which setup conditional functions that refers current post.
+		// Go to current post to update WP_Query with correct data.
 		$this->go_to( get_permalink( $post_id ) );
 
 		$meta_object = self::$rest->get_callback( get_post( $post_id, 'ARRAY_A' ) );
@@ -307,7 +307,7 @@ final class RestMetadataTest extends TestCase {
 		self::set_options( array( 'apikey' => 'testkey' ) );
 		$post_id = self::factory()->post->create();
 
-		// Navigate to post which setup conditional functions that refers current post.
+		// Go to current post to update WP_Query with correct data.
 		$this->go_to( get_permalink( $post_id ) );
 
 		$meta_object = self::$rest->get_callback( get_post( $post_id, 'ARRAY_A' ) );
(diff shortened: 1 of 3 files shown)


acicovic (dismissed):
Excellent work, thank you for taking the time to dig into this!

This might just be me, but I find the current comment a bit confusing. Maybe it could be something like this?
```
// Go to current post to update WP_Query with correct data.
```

I'm approving as-is, in case you prefer to merge it as-is.

Thank you!

acicovic (approved):

PR #1225, opened 2022-12-02
Content Helper: Fix blocked API requests caused by ad blockers

### LINKED ISSUE #1203: Content Helper is being blocked by certain ad blockers
The Content Helper is being blocked by certain ad blockers (I've seen uBlock Origin doing that, but there may be others).

This is unwanted since:
1. No actual tracking takes place in the Content Helper.
2. The block cripples our functionality and displays an "Error: You are probably offline" message, which is totally uninformative/misleading.

1. Test with 2-3 popular ad blockers to see which of those exhibit the blocking behavior.
2. Investigate what triggers the block, and if there are any reasonable ways to work around it.
3. When the "Error: You are probably offline" message occurs, include some text where we explain that certain ad blockers can cause this, so users are aware. If everything else fails, this will at least give some information to our users so they can mitigate the issue.

mehmoodak:
## Description
Due to the presence of `analytics` keyword in our WP REST API endpoint ad blockers were blocking the request. Replaced `analytics` keyword with `stats` to fix the issue.

Fixes: #1203 

Error message is also updated to include the ad blocker hint when Parse.ly API returns error.

## How Has This Been Tested?
- Updated tests where needed.
- Deployed to test website hosted on internet (not on localhost) to verify the changes.


acicovic:
I haven't tested this, but two things that come to mind about the error hint are:
- With the current solution, would it be easy to integrate it in other parts of our codebase? (CH Current Post Details panel immediately comes to mind, but this could be anything using the API like a Dashboard Widget or a Post List enhancement). If not, we probably need to implement this at a higher level than the final component.
- Has this been tested with WP in a language other than English? Will the hint still display?

mehmoodak:
> With the current solution, would it be easy to integrate it in other parts of our codebase? (CH Current Post Details panel immediately comes to mind, but this could be anything using the API like a Dashboard Widget or a Post List enhancement). If not, we probably need to implement this at a higher level than the final component.

If by current solution you mean the actual fix then we cannot move it to any higher level component because that was just a url change and for others also we can change the url to fix the issue but as of now as far as I know the issue was only appearing in "Content Helper: Top Performing Posts" so therefore only fixes it on this screen.

If you mean by hint then for now we are only using it in 1 place i.e. Top Performing Posts but if needed we can refactor it i.e. export the `<ErrorHint />` component and then display it anywhere we like

> Has this been tested with WP in a language other than English? Will the hint still display?

Not tested but it will definitely display because the hint is just a string is HTML that we are translating by `__` function .... didn't done anything in the API because as per our current needs seems like static text can work fine.

acicovic:
Hi,

I'm uniquely speaking about the error hint part. I'll number the list to make things easier.

1. The first point was that we may need to display the hint in places other than Gutenberg (like in a Dashboard Widget or the Post List for example). To the extent possible we should have a single central solution to cover all those cases. Else we'll have code duplication and maintenance in many places.

2. The second point was not about the translation of the hint itself. It seems to me that the current code appends the hint to every error message that comes from apiFetch. However, what we want is append the hint only if the error is "You are probably offline". However, if WP API error messages get translated, instead of receiving "You are probably offline" you may receive "Vous êtes probablement hors ligne" in French for example. If this is the case, we might need to rely on other information such as an error number to know when we need to apply the hint.

Let me know if this clarifies things or if more information is needed.

mehmoodak:
> The first point was that we may need to display the hint in places other than Gutenberg (like in a Dashboard Widget or the Post List for example). To the extent possible we should have a single central solution to cover all those cases. Else we'll have code duplication and maintenance in many places.

Okay I will make it as an `ErrorHint` component as described above 

> The second point was not about the translation of the hint itself. It seems to me that the current code appends the hint to every error message that comes from apiFetch. However, what we want is append the hint only if the error is "You are probably offline". However, if WP API error messages get translated, instead of receiving "You are probably offline" you may receive "Vous êtes probablement hors ligne" in French for example. If this is the case, we might need to rely on other information such as an error number to know when we need to apply the hint.

Improvement is implemented to only show error if api fetch is failed i.e. offline case. For supported languages concern what are your suggestions?

1. We should support all default languages of WordPress in the plugin? (ideal but not feasible).
2. Don't show hint on other languages.

If it is `2` then I think it is better to show something than nothing considering English is a universal language and can convey useful information in this case.

acicovic:
My research suggests that `apiFetch()` can fail with many different error messages. One example is [this one](https://github.com/WordPress/WordPress/blob/25ea9680aecaf4791c870658fd8085c10622ad05/wp-includes/rest-api/class-wp-rest-server.php#L1097-L1101). The error code in this case will be `rest_no_route`, and it triggers when we're requesting an endpoint that does not exist. With the current code, we're still showing the hint, although the hint is inaccurate for this specific case.

Also, see [this](https://github.com/WordPress/WordPress/blob/1d29cb663289795c6aca5cf8c624db6ac2475b49/wp-includes/js/dist/api-fetch.js#L720-L723), which is the error we get when the ad blockers block the request. The error code in this case should be `fetch_error`. This (and the previous example) suggest that if we look at the error code, we can get a better pinpoint of the error regardless of the message and its language.

So a first suggestion would be to show the hint only when the error code is `fetch_error`. This might not be 100% accurate either, but it is more precise than triggering the hint whenever `apiFetch()` fails.

In my opinion this will do it for now since [we should revisit error handling in the Content Helper anyway](https://github.com/Parsely/wp-parsely/pull/1093#discussion_r1036449562), and since the possible addition of Dashboard Widgets or Post List enhancements may force us to revise our approach in general.

@chriszarate, feel free to comment if you have a different opinion.

mehmoodak:
> So a first suggestion would be to show the hint only when the error code is fetch_error. This might not be 100% accurate either, but it is more precise than triggering the hint whenever apiFetch() fails.

@acicovic this has been already done (also mentioned in above comment) and PR is good to review without any changes.

acicovic:
> @acicovic this has been already done (also mentioned in above comment) and PR is good to review without any changes.

Sorry, this is my bad. The code I was testing yesterday did not include your latest commits for some reason. I've re-tested it now with the new commits and it works as you said, and I see the related check in the code. I'll give this a final review later today.

tests/js/blocks/content-helper/structure.test.js
```diff
@@ -49,21 +50,24 @@ describe( 'Content Helper', () => {
 
 	test( 'should show error message when API returns the error', async () => {
 		const getRelatedTopPostsFn = getRelatedTopPostsMockFn( () => Promise.reject( {
-			message: 'fake error from API',
+			message: 'fake error from API.',
```
acicovic:
```suggestion
			message: 'Fake error from API.',
```

tests/js/blocks/content-helper/structure.test.js
```diff
@@ -49,21 +50,24 @@ describe( 'Content Helper', () => {
...
-		render( <RelatedTopPostList /> );
-		expect( getSpinner() ).toBeInTheDocument();
+		expect( await verifyApiErrorMessage( getRelatedTopPostsFn ) ).toBeTruthy();
+	} );
 
-		await waitFor( () => screen.findByTestId( 'api-error' ), { timeout: 3000 } );
+	test( 'should show error message and hint when API fetch is failed', async () => {
+		const getRelatedTopPostsFn = getRelatedTopPostsMockFn( () => Promise.reject( {
+			code: 'fetch_error',
+			message: 'fake error from API.',
```
acicovic:
```suggestion
			message: 'Fake error from API.',
```

src/blocks/content-helper/content-helper.scss
```diff
@@ -13,12 +13,19 @@
 
 .parsely-top-posts {
 	list-style-type: none;
-	margin: 0;
+	margin: 22px 0 0;
```
acicovic: Do you have a specific reason for choosing `px` instead of a unit like `rem`s?

acicovic (commented):
Left a question and a couple of minor spelling-related feedback.

src/blocks/content-helper/content-helper.scss
mehmoodak: No specific reason, let me know if there is a perferred choice in the codebase.

acicovic: Usage of `rem` is preferred whenever this possible.

src/blocks/shared/functions.scss
```diff
@@ -0,0 +1,12 @@
+@use "sass:math";
+@import "./variables";
+
+@function stripUnit($value) {
+
+	@return math.div($value, $value * 0 + 1);
+}
+
+@function to_rem($pxValue) {
+
```
mehmoodak, line 5: These empty lines are due to our linting specs .... should we change it?

acicovic: What's the linting spec causing this?

mehmoodak: Here is the error:

```
src/blocks/shared/functions.scss
 5:2  ✖  Expected empty line before at-rule  at-rule-empty-line-before
 9:2  ✖  Expected empty line before at-rule  at-rule-empty-line-before
 ```

**Comment by acicovic** [comment]
Regarding https://github.com/Parsely/wp-parsely/pull/1225/commits/b22cfc2f8a401c17375b9637748a8cca4ba76508, I'm unsure it gives us the solution we really want.

I would opt in for adding the `rem` values directly into the CSS, my reasoning being:
- The main point of having `rem`s in our code is to have a relative unit that will be flexible and offer correct proportions if the user happens to have different defaults (for example what happens if the user has a default font size of 24px). To my understanding, converting `px` to `rem` does not lead us to this goal?
- It will make the code simpler, shorter, and easier to understand for us as well.

Having said that, if you have a specific reason for doing this this way, let us know of it.

Thank you!

File: src/blocks/shared/functions.scss (hunk 4 above)
Comment (acicovic) [reply src/blocks/shared/functions.scss:5]: Let's leave this as is for now. We may revisit to fix it if it becomes an annoyance.

**Comment by mehmoodak** [comment]
> if you have a specific reason for doing this this way, let us know of it.

- The added function are just syntactic sugar and there will be no difference in the final output. The `rem` units that are declared are still dependent on user defaults.

- For me the current code does seems simpler, shorter and easier to understand. The complex part is hidden behind the functions and those are just making the dev work fast i.e. no manual conversions of `px` to `rem` while developing from designs and since design tools shows `px` by default so I think having this function will make our dev work a bit fast which is better 🙂

**Comment by acicovic** [comment]
@chriszarate, any feedback/opinions on the 2 above posts?

**Comment by acicovic** [comment]
Besides the px/rem question, everything else seems fine to me and once we get final feedback from @chriszarate, we are ready to merge.

File: src/blocks/content-helper/components/related-top-post-list.tsx [hunk 5]
```diff
@@ -73,7 +74,18 @@ function RelatedTopPostList() {
 
 		// Error coming from apiFetch.
 		if ( error?.message ) {
-			return <p className="parsely-top-posts-descr" data-testid="api-error">{ __( 'Error:', 'wp-parsely' ) } { error.message }</p>;
+			return (
+				<>
+					<p className="parsely-top-posts-descr" data-testid="api-error">
+						{ __( 'Error:', 'wp-parsely' ) } { error.message }
```
chriszarate, line 80: this will probably need an update once #1232 is merged

src/blocks/shared/components/error-hint.tsx
```diff
@@ -0,0 +1,20 @@
+/**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+
+export const AD_BLOCKER_HINT: string = __(
+	'This error can be sometimes caused by ad-blockers or browser tracking protections. Please whitelist this website where needed and try again.',
```
chriszarate:
```suggestion
	'This error can be sometimes caused by ad-blockers or browser tracking protections. Please add this site to any applicable allow lists and try again.',
```

src/blocks/shared/components/error-hint.tsx
```diff
@@ -0,0 +1,20 @@
...
+	'wp-parsely',
+);
+
+export function ErrorHint(): JSX.Element {
```
chriszarate, line 11: Return types can be omitted from React components, even when they are functions.

src/blocks/shared/functions.scss
```diff
@@ -0,0 +1,12 @@
+@use "sass:math";
+@import "./variables";
+
+@function stripUnit($value) {
+
+	@return math.div($value, $value * 0 + 1);
+}
```
chriszarate: x * 0 + 1 = 1

```suggestion
@function stripUnit($value) {

	@return math.div($value, 1);
}
```

src/blocks/shared/functions.scss
```diff
@@ -0,0 +1,12 @@
...
+
+@function to_rem($pxValue) {
+
+	@return math.div(stripUnit($pxValue), stripUnit($html-font-size)) + rem;
```
chriszarate: `stripUnit` not even needed since math.div ignores units

```suggestion
	@return math.div($pxValue, $html-font-size) + rem;
```

chriszarate: looking at the docs, i am probably wrong about this. but in testing it worked 🤷

acicovic:
@mehmoodak, FYI: We've merged https://github.com/Parsely/wp-parsely/pull/1093, so there may be a couple of other endpoints to change to `stats`.

src/blocks/content-helper/components/related-top-post-list.tsx
acicovic: I'll be working on that as part of #1232, so we can merge this as is for the moment.

src/blocks/shared/components/error-hint.tsx
mehmoodak: Yes but I prefer to have types ... it seems easy for me to read the code and know the type 🙂

src/blocks/shared/functions.scss
mehmoodak: `math.div` only ignores the unit if both numerator and denominator have same units (src: https://sass-lang.com/documentation/modules/math#div) and since this matches with our case so ignoring the `stripUnit` function.

mehmoodak:
> FYI: We've merged https://github.com/Parsely/wp-parsely/pull/1093, so there may be a couple of other endpoints to change to stats.

@acicovic changes are done and PR is ready for review.

acicovic (approved):
I've looked at this and as far as I can tell, the important feedback points by @chriszarate have been addressed. @chriszarate, feel free to take a look again, and we can communicate for any further refinements. As things stand, I think I can approve this PR.

src/blocks/shared/components/error-hint.tsx
chriszarate: Makes sense, mind the `null` case

acicovic: In other words, could this fail?

chriszarate: Not as written here. But a component that sometimes `return`s null would need its return type adjusted:

```
function ConditionalComponent ( props: Props ): JSX.Element | null {
  if ( props.foo ) {
    return null;
  }

  return <div>Hello</div>;
}
```

This is nothing surprising, it's just something you have to pay attention to in TypeScript. But sometimes React-provided types feel a bit mysterious, especially surrounding JSX. I personally think explicit typing is always fine, but it can sometimes get burdensome to the point where you start looking around for abstractions. React component return types should always be pretty simple, though.

PR #1229, opened 2022-12-06
Settings: Update radio options text to be more descriptive and easy to understand

### LINKED ISSUE #1196: Improve plugin settings UI

The UX of the plugin settings page needs improvement.

- The "Metadata format" option is currently implemented as a drop-down, even though there are only two options. This field is better implemented as a radio button, which gives better visibility into the available options.
- There are multiple "Yes / No" radio buttons that are difficult to parse as yes / no questions. Instead, present two options with descriptions inline. For example, "Use Top-Level Categories for Section" could be better represented as "Top-Level Categories" with two options describing the different behavior.

mehmoodak:
## Description
There were many radio buttons on plugin settings page with simple Yes/No options which are a bit hard to understand so we updated the text to make it descriptive and more easy to understand.

Fixes: #1196 

## How this change is done?
We were using `print_binary_radio_tag` function to display the radio buttons with simple `Yes/No` option and now we have updated this function to `print_radio_tags` to be more generic so that it can print all the radio options passed to it with custom `Yes/No` text.

## How Has This Been Tested?
- Since these were just copy changes so didn't added any new tests and ran existing tests to make sure everything is working fine.
- Apart from it also compare the different in two functions with diff-checker to make sure that  only text has been changed and no attribute has been missed while updating the function


src/UI/class-settings-page.php
```diff
@@ -303,11 +303,10 @@ private function initialize_basic_section(): void {
 		);
 
 		// Site ID.
-		$h          = __( 'Your Site ID is typically your own site domain without <code>http(s)://</code> prefixes or trailing <code>/</code> (e.g. <code>mydomain.com</code>).', 'wp-parsely' );
```
mehmoodak, line 306: Used the help text directly and avoid the declaration of `$h` variable. To me it seems easy to read when every info which is only used once is available in the `$field_args`

src/UI/class-settings-page.php
```diff
@@ -751,27 +762,43 @@ public function print_select_tag( array $args ): void {
...
-		$name    = Parsely::OPTIONS_KEY . "[$id]";
+	public function print_radio_tags( array $args ): void {
+		$name          = $args['option_key'];
+		$id            = esc_attr( $name );
+		$input_name    = Parsely::OPTIONS_KEY . "[$id]";
+		$options       = $this->parsely->get_options();
+		$selected      = $options[ $name ];
+		$radio_options = $args['radio_options'] ?? array(
+			'true'  => __( 'Yes', 'wp-parsely' ),
+			'false' => __( 'No', 'wp-parsely' ),
```
mehmoodak: These are present for `AMP` setting field, will update it after getting feedback from @danielabloch

tests/e2e/specs/front-end-metadata.spec.js
```diff
@@ -20,10 +20,11 @@ const setMetadataFormat = async ( format ) => {
 	await visitAdminPage( '/options-general.php', '?page=parsely' );
 	await waitForWpAdmin();
 
-	await page.select( 'select', format );
+	const selectedMetadataFormat = await page.$( `#meta_type_${ format }`, format );
+	await selectedMetadataFormat.click();
 
-	const [ input ] = await page.$x( '//p[contains(@class, \'submit\')]//input[contains(@name, \'submit\')]' );
```
mehmoodak, line 25: Just making the selector a bit easier.

tests/e2e/specs/front-end-metadata.spec.js
```diff
@@ -33,6 +35,9 @@ describe( 'Front end metadata insertion', () => {
 		await startUpTest();
 		await setSiteKeys();
 		await selectScreenOptions( { recrawl: true, advanced: false } );
+
+		// Reset display name to compare metadata with default values.
+		await setUserDisplayName( 'admin', '' );
```
mehmoodak, line 40: If `filters.spec.js` file ran before this file then some tests throws error because inside `filter.spec.js` file we are changing the display names so here we are resetting it to default to make sure that tests don't fail no matter the order.

src/UI/class-settings-page.php
acicovic: I agree.

tests/e2e/specs/front-end-metadata.spec.js
acicovic: OK, let's keep an eye on this though, just in case it starts failing more often. In this case, we should revert. I don't know why certain selectors have been written this way, and if there is actually any advantage to that.

tests/e2e/specs/front-end-metadata.spec.js
acicovic: Great find!

acicovic (dismissed):
Manually tested this branch on a local installation, confirming that:
- Existing settings are correctly kept.
- Saving settings for the first time works without side effects.

acicovic:
I've manually tested and approved as above, but we have an E2E test that has failed 3 times for me, so we should look into that. Once E2E tests pass, we should be able to merge this.

PR #1230, opened 2022-12-06
Avoid extra WordPress_Cache instances while proxying the Parse.ly API

### LINKED ISSUE #1210: Avoid uneeded `WordPress_Cache` instantiations
I wonder if we could avoid spinning up new instances of this WP cache each time.

```php
static $cache = new WordPress_Cache();
...
$cached_proxy_instance = new Cached_Proxy( $proxy_instance, $cache );
...
```

_Originally posted by @GaryJones in https://github.com/Parsely/wp-parsely/pull/1093#discussion_r1025304594_

mehmoodak:
## Description
A suggestion has been made in one of the PR to avoid instantiating multipe `WordPress_Cache` objects so this PR implemented that suggestion.

Fixes: #1210 

## How Has This Been Tested?
- Ran existing tests after making changes to make sure nothing breaks.


acicovic:
The original suggestion as I understood it was to move `new WordPress_Cache()` in a static variable outside the function.

Have you tried this, or do you think it could be problematic?

mehmoodak:
> The original suggestion as I understood it was to move new WordPress_Cache() in a static variable outside the function.

As far as I understood the original suggestion was just to use the WP_Cache once so using the variable seems like a simple solution to me instead of using static variable.

acicovic:
Will see if I can check with Gary on this.

acicovic:
@mehmoodak, are you OK if we move this to the 3.7 milestone?

I'm proposing this so we have enough time to test it instead of throwing it into a very near release, as I'm not really familiar yet on how `WordPress_Cache()` works. I do think this should be OK, but we'd rather be sure, especially if we end up using a static variable or such.

mehmoodak:
From the code seems like it is a just refactoring change. But completely fine if we are going to move it to 3.7 milestone.

acicovic:
OK, moving to 3.7 for now.

GaryJones (approved):
The changes currently in the PR is what I had in mind ✅

acicovic:
OK @mehmoodak, feel free to merge this. I've reverted it back for 3.6. Sorry for the back and forth!

acicovic (approved):

PR #1231, opened 2022-12-06
Move Parse.ly API base url to constant for easier mocking during development

### LINKED ISSUE #1208: Consider defining `dash.parsely.com` in a constant
Consider having `dash.parsely.com` defined in a constant that could then be adjusted during development to test against the beta dashboard.

_Originally posted by @GaryJones in https://github.com/Parsely/wp-parsely/pull/1093#discussion_r1025240406_

mehmoodak:
## Description
Created a `const` for Parse.ly API url in both PHP and JS files so that it is easier to mock during development e.g. using beta dashboard API

Fixes: #1208 

## How Has This Been Tested?
- Ran existing tests to make sure nothing is breaking.


acicovic:
I think we should rename the constant to something like DASH_BASE_URL or DASHBOARD_BASE_URL since this is the basic URL of the dashboard, not of the API.

acicovic:
I still see API in const names. Is there a reason you think we should keep it?

mehmoodak:
> I still see API in const names. Is there a reason you think we should keep it?

@acicovic made the changes  please have a look again. Thanks

acicovic (approved):

PR #1232, opened 2022-12-07
Content Helper: Improve error handling

acicovic:
## Description
@chriszarate pointed out in #1093 that the Content Helper's error handling could use some improvements.

This PR addresses the feedback received and takes into account error handling scenarios like the [issue we recently had with ad blockers](https://github.com/Parsely/wp-parsely/pull/1203). It also centralizes functionality related to errors in a single place, so it can be shared between the Content Helper's panels.

## Motivation and Context
Write better code, handle errors better.

## How Has This Been Tested?


chriszarate (dismissed):
Yesssss :)

acicovic:
Since we got there so fast, I'll:
- Extend this solution to the whole Content Helper.
- Integrate with the error hint for all Content Helper panels.
- Schedule this for 3.6.

acicovic:
@mehmoodak and @chriszarate, I'll wait for us to merge #1225 before working more on this.

src/blocks/content-helper/content-helper-error.ts
```diff
@@ -0,0 +1,37 @@
...
+ * Extends the standard JS Error class with an error code field and an optional
+ * prefix to the error message.
+ *
+ * @see https://github.com/microsoft/TypeScript/wiki/FAQ#why-doesnt-extending-built-ins-like-error-array-and-map-work
+ */
+export class ContentHelperError extends Error {
+	public code: string;
+	public prefix: string;
+
+	constructor( message: string, code: string, prefix = __( 'Error: ', 'wp-parsely' ) ) {
```
mehmoodak: Should we improve `code` type here by using `ContentHelperErrorCode` enum?

src/blocks/content-helper/current-post-details/component.tsx
```diff
@@ -53,19 +54,7 @@ function CurrentPostDetails() {
...
-
-		// Error coming from apiFetch.
-		if ( error?.message ) {
-			return <p>{ __( 'Error:', 'wp-parsely' ) } { error.message }</p>;
-		}
-
-		// Error coming from the WordPress REST API.
-		const errorMessage = JSON.stringify( error ).match( /\[\"(.*?)\"\]/ )[ 1 ];
-		return <p>{ __( 'Error:', 'wp-parsely' ) } { errorMessage }</p>;
+		return <p>{ error.message }</p>;
```
mehmoodak: How are we going to include `ErrorHint` here?

mehmoodak (dismissed):

src/blocks/content-helper/content-helper-error.ts
acicovic: I've done this on purpose, since in some parts of the code we pass `wpError.code` which might not be in our collection.

src/blocks/content-helper/current-post-details/component.tsx
acicovic: The updated code should take care of that.

acicovic:
I will be marking this PR as ready for review. From a functionality perspective it works, but our structure test fails.

I'm surely doing something I shouldn't, so can somebody suggest the correct approach to do this?

mehmoodak:
> I will be marking this PR as ready for review. From a functionality perspective it works, but our structure test fails.
>
> I'm surely doing something I shouldn't, so can somebody suggest the correct approach to do this?

From a quick look I am not able to find anything major, it will need some more debugging to pin point the issue so should I help you on this or are you already taking care of this? .... e2e tests are passing so seems like mostly things are fine.

acicovic:
> From a quick look I am not able to find anything major, it will need some more debugging to pin point the issue so should I help you on this or are you already taking care of this? .... e2e tests are passing so seems like mostly things are fine.

If you can, please help me 😃
I'm not looking into this currently.

tests/js/blocks/content-helper/structure.test.js
```diff
@@ -29,38 +30,37 @@ describe( 'Content Helper', () => {
...
 	test( 'should show contact us message when Parse.ly Site ID is not set', async () => {
-		const getRelatedTopPostsFn = getRelatedTopPostsMockFn( () => Promise.reject( {
-			errors: {
-				parsely_site_id_not_set: 'Error message.',
-			},
-		} ) );
+		const getRelatedTopPostsFn = getRelatedTopPostsMockFn( () => Promise.reject( new ContentHelperError(
+			'Error message.',
+			ContentHelperErrorCode.PluginSettingsSiteIdNotSet
+		) ) );
```
mehmoodak, line 36: Return `ContentHelperError` instance instead of simple object to fix the tests.

tests/js/blocks/content-helper/structure.test.js
```diff
@@ -72,25 +72,6 @@ describe( 'Content Helper', () => {
 		);
 	} );
 
-	test( 'should show error message when WordPress REST API returns the error', async () => {
```
mehmoodak, line 74: Removed this test because as of now this is extra i.e. we are treating parsely api errors and wp api errors in same way.

src/blocks/content-helper/content-helper-error.tsx
```diff
@@ -26,14 +26,14 @@ export enum ContentHelperErrorCode {
  * @see https://github.com/microsoft/TypeScript/wiki/FAQ#why-doesnt-extending-built-ins-like-error-array-and-map-work
  */
 export class ContentHelperError extends Error {
-	protected code: string;
+	protected code: ContentHelperErrorCode;
```
mehmoodak, line 29: Used more stricter typings.

mehmoodak (commented):

mehmoodak (approved):
Looks Good, approving the PR.

@acicovic it will be great if you can also look at my last commit.

src/blocks/content-helper/content-helper-error.tsx
acicovic: Will this still work if we pass a string that is not within the `ContentHelperErrorCode` enum, or will this produce an error?

acicovic: Asking because we have code like this:

https://github.com/Parsely/wp-parsely/blob/cea8d2418823feaf7ee3bf566c06fefe9ec13933/src/blocks/content-helper/current-post-details/provider.ts#L109-L113

`wpError.code` can be something that is not within our enumeration.

mehmoodak: During development we can only pass the strings defined in the enum.

But if the code is running in browser then it can work with any string because mainly TS types are for development purposes.

Having stricter types is better because in future if we pass anything to the function which isn't defined in the `enum` then we can take appropriate action.

acicovic: Thank you!

chriszarate (approved):
:+1: x 2 :)

PR #1237, opened 2022-12-08
Content Helper: Fix Parse.ly Posts API call for pages

### LINKED ISSUE #1236: Hide Content Helper Top Performing panel for pages
## Describe the bug
We currently show the “Related Top-Performing Posts” for pages, but it fails with strange errors as pages are not supported. We should not load/show this panel at all when the user opens pages. Maybe we can use [this](https://developer.wordpress.org/block-editor/reference-guides/data/data-core-editor/#getcurrentposttype) for checking the post type.

## Findings
The errors that we are facing are happening due to the bug in the code so instead of hiding the panel we are fixing the code and after that we will show the content to the user based on tag, categories or author.

mehmoodak:
## Description
Our top performing posts panel were showing error in pages and the root cause of this is the absense of `null` check so added that to fix the issue.

Fixes: #1236 

## How Has This Been Tested?
- Just a minor fix so ran existing tests.


src/blocks/content-helper/components/related-top-post-list.tsx
```diff
@@ -48,6 +48,8 @@ function RelatedTopPostList() {
 					} else {
 						setLoading( false );
 						setError( err );
+
+						throw err;
```
mehmoodak: I think we should throw the error even after catching it so that the browser console can shows the error instead of showing nothing.

mehmoodak: Not a good idea for now because we are showing error messages based on this.

acicovic (dismissed):

acicovic:
@mehmoodak, did anything change in the latest push?

mehmoodak:
> did anything change in the latest push?

No, there were conflicts in build files so just resolved those.

acicovic (approved):

PR #1239, opened 2022-12-09
Add TypeScript support to all remaining JavaScript files

### LINKED ISSUE #471: Move Javascript code to TypeScript

mehmoodak:
## Description
We have TypeScript support present in our project but not all of our files are migrated so in this PR we have migrated the remaining JS files to TS and adds `strict` mode in `tsconfig.json` to improve our code quality.

As a part of this work we haven't improved all of the existing typing i.e. just added support so our future code will have strong types and fixes any TS issues in our existing code.

Fixes: #471 

## How Has This Been Tested?
- Ran existing tests to make sure nothing is breaking


.eslintrc
```diff
@@ -1,28 +1,46 @@
...
+			{
+				"allowedTextDomain": [
+					"wp-parsely"
+				]
+			}
+		],
+		"@typescript-eslint/ban-ts-comment": "off",
+		// Enabling ts and rule and disabling base rule as it can report incorrect errors.
+		"no-shadow": "off",
+		"@typescript-eslint/no-shadow": ["error"]
```
mehmoodak: Just added this, rest of the changes are just formatting.

src/@types/assets/puppeteer.d.ts
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+declare const page: any;
```
mehmoodak: Can we improve typing here? 

For testing we are not using puppeteer directly but using `@wordpress/e2e-test-utils` package which doesn't have exported types so what is the best way to improve the typing of `page` variable in our test files .... considering how much it is used I think we should have strong typing of this.

(At same time shouldn't be a blocker for this PR)

src/@types/assets/wp-e2e-test-utils.d.ts
```diff
@@ -0,0 +1 @@
+declare module '@wordpress/e2e-test-utils';
```
mehmoodak, line 1: Any way of using strong typing for this also?

src/blocks/content-helper/current-post-details/provider.ts
```diff
@@ -44,8 +44,10 @@ class CurrentPostDetailsProvider {
 	 * Constructor.
 	 */
 	constructor() {
-		// Return data for the last 7 days (today included).
-		this.setDataPeriod( 7 );
```
mehmoodak, line 52: In TS strict mode assignment of the variables should be either while declaration or inside constructor so in this case we have two possible fixes.

1. Moved the function content inside the constructor.
2. Use `!:` in typing during declaration.

Chose first approach because we just have to define 3 lines in the constructor.

src/blocks/recommendations/actions.ts
```diff
@@ -1,4 +1,4 @@
-import { RECOMMENDATIONS_BLOCK_ERROR, RECOMMENDATIONS_BLOCK_LOADED, RECOMMENDATIONS_BLOCK_RECOMMENDATIONS } from './constants';
```
mehmoodak, line 1: Moved separate constants to single enum which I think is better.

src/@types/assets/window.d.ts
```diff
@@ -0,0 +1,26 @@
...
+				parsely_site_uuid: string,
+			},
+			autotrack?: boolean,
+			onload?: () => unknown,
+			onReady?: () => unknown,
+		},
+		wpParselyApiKey: string,
+		wpParselyDisableAutotrack?: boolean;
+		wpParselyHooks?: _Hooks;
+	}
```
mehmoodak, line 28: Instead of defining things as global inside `.eslintrc` chose this approach so that we have proper context of each variable which we are using from `window` object.

tsconfig.json
```diff
@@ -1,17 +1,22 @@
...
+		"esModuleInterop": true,
+		"jsx": "react-jsx",
+		"module": "es6",
+		"moduleResolution": "node",
+		"noImplicitAny": true,
+		"outDir": "./build/",
+		"resolveJsonModule": true,
+		"target": "es5",
+		"typeRoots": [
+			"./src/@types/**",
```
mehmoodak, line 14: - Changed this line.
- Added strict mode.
- Format the file.

webpack.config.js
```diff
@@ -18,7 +18,7 @@ module.exports = {
 			path.resolve( __dirname, 'src', 'js', 'lib', 'uuid-profile-call' ),
 		],
 		'recommended-widget': [
-			path.resolve( __dirname, 'src', 'js', 'widgets', 'recommended.js' ),
```
mehmoodak, line 21: Changed because the file is now in ts. If we don't have any extension defined then the build tool looks for both js and ts file extension

tests/js/blocks/recommendations/structure.test.tsx
```diff
@@ -128,22 +129,15 @@ function verifyBlockStructure( resultCount, blockTitle, imageStyle = '' ) {
...
 		results.push(
 			{
-				author: 'John Doe',
-				authors: [ 'John Doe', 'Jane Doe' ],
-				full_content_word_count: 0,
 				image_url: `https://example.com/original-${ i }.jpg`,
-				metadata: null,
-				pub_date: '2022-03-14T13:14:00',
-				section: 'Articles',
-				tags: [ 'parsely_smart:entity:Tag 1', 'parsely_smart:iab:Tag 2' ],
```
mehmoodak, line 146: Extra fields so removed it.

tests/e2e/specs/browse-logo-button.spec.ts
```diff
@@ -46,7 +46,8 @@ describe( 'Browse for logo button', () => {
 
 		// Confirm image deletion.
 		// Note: Dialog handling must be placed before the click event.
-		await page.on( 'dialog', async ( dialog ) => {
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
mehmoodak: Didn't have an idea what to use on other than `any` so choose this for now and we can review later for strong typing.

src/@types/assets/puppeteer.d.ts
mehmoodak: One possible option is to manually get the page interface from the puppeteer repo and used that but looking for any non-custom approach.

mehmoodak (commented):

src/@types/assets/puppeteer.d.ts
chriszarate: delete this file and
`npm i --save-dev @types/jest-envi
ronment-puppeteer`

Better typing, you will have some additional type errors to fix but should be easy

src/@types/assets/window.d.ts
```diff
@@ -0,0 +1,29 @@
...
+ * External dependencies.
+ */
+import { _Hooks } from '@wordpress/hooks/build-types/createHooks';
+
+export {};
+
+declare global {
+	interface Window {
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any
+		wp: any;
```
chriszarate, line 11: Would be nice to have a real type here, but I guess Gutenberg does this too :/
https://github.com/WordPress/gutenberg/blob/2c570001ff9dda80b588989accdec221bb8fcc6b/packages/e2e-test-utils-playwright/src/types.ts

src/blocks/content-helper/models/related-top-post-data.ts
```diff
@@ -7,3 +7,12 @@ export interface RelatedTopPostData {
 	url: string;
 	views: number;
 }
+
+export interface RelatedTopPostApiError {
+	code: string;
+	message :string;
```
chriszarate:
```suggestion
	message: string;
```

src/blocks/recommendations/components/parsely-recommendations-fetcher.tsx
```diff
@@ -58,7 +58,7 @@ const ParselyRecommendationsFetcher = ( { boost, limit, sort, isEditMode } : Par
 		}
 
 		if ( error ) {
-			dispatch( setError( { error } ) );
+			dispatch( setError( { error: error as string } ) );
```
chriszarate, line 61: is it an Error or a string?

```suggestion
			dispatch( setError( error ) );
```

src/@types/assets/window.d.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * External dependencies.
+ */
+import { _Hooks } from '@wordpress/hooks/build-types/createHooks';
```
chriszarate:
```suggestion
import type { _Hooks } from '@wordpress/hooks/build-types/createHooks';
```

src/blocks/recommendations/recommendations-store.tsx
```diff
@@ -6,17 +6,29 @@ import { createContext, useContext, useReducer } from '@wordpress/element';
...
  * Internal dependencies
  */
-import { RECOMMENDATIONS_BLOCK_ERROR, RECOMMENDATIONS_BLOCK_LOADED, RECOMMENDATIONS_BLOCK_RECOMMENDATIONS } from './constants';
+import { RecommendationsAction } from './constants';
+import { Recommendation } from './models/Recommendation';
 
-const RecommendationsContext = createContext();
+interface RecommendationState {
+	isLoaded: boolean;
+	recommendations: Recommendation[] | undefined;
```
chriszarate: Is the undefined state needed? If not, maybe use an empty array in the default state. Would simplify logic.

src/blocks/recommendations/recommendations-store.tsx
```diff
@@ -6,17 +6,29 @@ import { createContext, useContext, useReducer } from '@wordpress/element';
...
+	uuid: string | null;
+	clientId: string | null;
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	error: any;
```
chriszarate: Can use `Error` here?

src/blocks/recommendations/recommendations-store.tsx
```diff
@@ -37,12 +49,18 @@ const reducer = ( state, action ) => {
 	}
 };
 
-const RecommendationsStore = ( props ) => {
-	const defaultState = {
+interface RecommendationStore {
+	clientId?: string;
+	children: JSX.Element | JSX.Element[];
```
chriszarate: `ReactNode` is almost always the correct type for `children`
```suggestion
	children: ReactNode;
```

src/js/admin-settings.ts
```diff
@@ -0,0 +1,25 @@
+document.querySelector( '.media-single-image button.browse' )?.addEventListener( 'click', selectImage );
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function selectImage( this: any ) {
+	const optionName = this.dataset.option;
```
chriszarate: Lots of ways to go about this, but recommend using event param instead of this and injecting the correct DOM type:
```suggestion
function selectImage( evt: Event ) {
	const optionName = ( evt.target as HTMLButtonElement ).dataset.option;
```

src/js/widgets/recommended.ts
```diff
@@ -19,9 +46,9 @@ function constructUrl( apiUrl, permalink, personalized ) {
 	return `${ apiUrl }&url=${ encodeURIComponent( permalink ) }`;
 }
 
-function constructWidget( widget ) {
-	const apiUrl = widget.getAttribute( 'data-parsely-widget-api-url' );
-	const permalink = widget.getAttribute( 'data-parsely-widget-permalink' );
+function constructWidget( widget: Element ): WidgetOptions {
+	const apiUrl = widget.getAttribute( 'data-parsely-widget-api-url' ) as string;
+	const permalink = widget.getAttribute( 'data-parsely-widget-permalink' ) as string;
```
chriszarate: The `as string` is dangerous here because it suppresses the `null` case. We should save `as` for last-resort cases where we're certain it will be the type we inject. Better to acknowledge that it could be null and set a default value. Or maybe it should throw an error.
```suggestion
	const apiUrl = widget.getAttribute( 'data-parsely-widget-api-url' ) || '';
	const permalink = widget.getAttribute( 'data-parsely-widget-permalink' ) || '';
```

tests/e2e/specs/blocks/recommendations/settings.spec.ts
```diff
@@ -119,17 +119,17 @@ async function resultsContainImage() {
...
  * Returns the "target" and "rel" attribute of all links contained within the Block.
  *
  * @return {Promise<Array<string, string>>} The "target" and "rel" attributes of all links contained within the Block.
  */
-async function getLinkTargets() {
-	return page.$$eval( '.parsely-recommendations-link', ( links ) => links.map(
-		( link ) => [ link.getAttribute( 'target' ), link.getAttribute( 'rel' ) ] )
+async function getLinkTargets(): Promise<string[][]> {
+	return page.$$eval( '.parsely-recommendations-link', ( links: HTMLElement[] ) => links.map(
+		( link: HTMLElement ): string[] => [ link.getAttribute( 'target' ) as string, link.getAttribute( 'rel' ) as string ] )
```
chriszarate: same as above

tests/e2e/specs/browse-logo-button.spec.ts
chriszarate: type might magically appear with the types package installed, or import from 'puppeteer-core'

tests/js/blocks/recommendations/structure.test.tsx
```diff
@@ -27,7 +28,17 @@ describe( 'Recommendations Block', () => {
 	it( 'should display loading text when starting', () => {
 		render(
 			<RecommendationsStore>
-				<ParselyRecommendations />
+				<ParselyRecommendations
+					// Passing default values taken from block.json
+					boost={ 'views' }
```
chriszarate: opinion alert :)
```suggestion
					boost="views"
```

src/@types/assets/puppeteer.d.ts
mehmoodak: Thanks a lot. Having this will make our tests a lot healthier 🙂

src/@types/assets/window.d.ts
mehmoodak: Any idea how can I get the real type for this?

src/blocks/recommendations/components/parsely-recommendations-fetcher.tsx
mehmoodak: In the code at one place we are assigning type `unknown` to it and at one place type `string` ..... so just to make minimal changes I forced the typing here else we could have make changes at other places which I think doesn't like in the scope of this PR 🤔 

Implementing the suggested change is giving following error so leaving it as it is

```
Argument of type '{}' is not assignable to parameter of type 'SetErrorPayload'.
  Property 'error' is missing in type '{}' but required in type 'SetErrorPayload'.
```

src/@types/assets/window.d.ts
chriszarate: Not sure, exploring in the gutenberg repo would be your best bet. But the `wp` global has been around for a long time, so maybe it is untyped.

mehmoodak: I did explored it via Search Engine but found nothing.

src/blocks/recommendations/components/parsely-recommendations-fetcher.tsx
chriszarate: Ok, makes sense. Maybe a future PR could clean up error typing everywhere. Thanks for looking!

chriszarate (dismissed):
Looks good, nice work, and thanks for breaking things up nicely in commits.

.eslintrc
```diff
@@ -1,28 +1,46 @@
...
+		"@wordpress/i18n-text-domain": [
+			"error",
+			{
+				"allowedTextDomain": [
+					"wp-parsely"
+				]
+			}
+		],
+		"@typescript-eslint/ban-ts-comment": "off",
+		// Enabling ts and rule and disabling base rule as it can report incorrect errors.
```
acicovic: Can this comment be improved? I'm not clear about what is said in the first part.

tests/js/blocks/recommendations/structure.test.tsx
acicovic: I think that this was done to fully emulate an API response, but I guess it is OK to remove them.

acicovic (commented):
Excellent work!

Also, it was very useful that you've left comments in various places to explain the reasoning behind the code change. It was great for me to have those explanations without having to dig into the "why" and make my own assumptions. This was super helpful!

I've left a couple of minor comments. Feel free to mark my second comment as resolved without doing anything.

The only downside is that this PR will need to be revisited after we've merged #1232.

.eslintrc
mehmoodak: There was a typo, now fixed it.

acicovic (approved):

src/blocks/content-helper/content-helper-provider.ts
```diff
@@ -109,9 +109,9 @@ class ContentHelperProvider {
 
 		try {
 			response = await apiFetch( {
-				path: addQueryArgs( '/wp-parsely/v1/stats/posts', query.query ),
+				path: addQueryArgs( '/wp-parsely/v1/stats/posts', query.query as object ),
 			} ) as RelatedTopPostsApiResponse;
-		} catch ( wpError ) {
+		} catch ( wpError: any ) { // eslint-disable-line @typescript-eslint/no-explicit-any
```
acicovic, line 114: I guess this `any` is required now?

acicovic (approved):

src/blocks/content-helper/content-helper-provider.ts
mehmoodak: Yes it is.

PR #1241, opened 2022-12-09
Settings Page: Rename "Track as Page" to "Track as Non-Post"

acicovic:
## Description
This PR updates the "Track as Page" label to "Track as Non-Post" within the plugin's settings page.

## Motivation and Context
The "Track as Page" label could create confusion during integrations using the plugin, since the Parse.ly terminology for this is Non-Post, as outlined [here](https://docs.parse.ly/metadata-jsonld/#distinguishing-between-posts-and-non-posts-pages). This is a small but meaningful change to bring our terminologies in sync to avoid needless friction or misunderstandings.

## How Has This Been Tested?
Existing E2E tests pass.


mehmoodak (dismissed):

mehmoodak:
Should we also change this line from Readme?

```
Define how to track every Post Type (as Page, as Post or no tracking).
```

acicovic:
Yes, great catch 👍
I've updated this.

mehmoodak (approved):

Issue #1242, opened 2022-12-09
Harmonize some options with their UI label


acicovic:
## Is your feature request related to a problem?
As time passes, our terminologies might change, and with it our settings page gets updates. Examples of this are "API Key" that has been renamed to "Site ID" and "Track as Page" that has been renamed to "Track as Non-Post".

In many of those cases, we quickly update the UI but not the underlying options. For example, the value of "Track as Non-Post" is still `page`. In large part, this is because renaming the options themselves can have undesired effects if not handled properly. However, this leads to out-of-sync code that can become harder to understand as time goes by.

## Describe the solution you'd like
Find the affected options and give them a more representative name, making sure that:
- Nothing breaks.
- Upgrade path works.
- Downgrade path works.

Related PRs/issues:
- https://github.com/Parsely/wp-parsely/issues/799
- https://github.com/Parsely/wp-parsely/pull/1241

GaryJones:
Renaming public methods and property names could be done on a major semantic versioning release.

PR #1244, opened 2022-12-13
Rename API Key to Site ID to improve consistency

### LINKED ISSUE #799: Rename "API Key" to "Site ID"
In different places within the plugin and its documentation, we may refer to the `Site ID` as `API Key`. To avoid confusion, everything should be renamed to `Site ID`. 

Find all occurrences of `API Key` and change them to `Site ID` within the plugin's code and the documentation.

mehmoodak:
## Description
In our codebase we are using both terms `Site ID` and `API Key` so in this PR we are just making changes to only use the term `Site ID`

Fixes: #799
Related: #1242

## How Has This Been Tested?
- Ran existing tests to make sure nothing is breaking


src/Integrations/class-amp.php
```diff
@@ -161,11 +159,11 @@ public static function construct_amp_json(): string {
 	public static function construct_amp_config(): array {
 		$options = self::$parsely->get_options();
 
-		if ( isset( $options['apikey'] ) && is_string( $options['apikey'] ) && '' !== $options['apikey'] ) {
+		if ( self::$parsely->site_id_is_set() ) {
 			return array(
 				'vars' => array(
 					// This field will be rendered in a JS context.
-					'apikey' => esc_js( $options['apikey'] ),
+					'site_id' => esc_js( $options['site_id'] ),
```
mehmoodak: Is it safe to replace this. I think it is not but I also don't have enough knowledge of AMP.

From here it seems like we are setting the key in JS context and it will used somehow by the users so we shouldn't changed it but please do confirm. Thanks

cc: @GaryJones

mehmoodak: Got the info from docs that it is necessary (https://docs.parse.ly/google-amp/)

tests/e2e/specs/recommended-widget.spec.js
```diff
@@ -51,7 +51,7 @@ describe( 'Recommended widget', () => {
 		await activateTheme( 'twentytwentytwo' );
 	} );
 
-	it( 'Widget should be available but inactive without api key and secret', async () => {
+	it( 'Widget should be available but inactive without site ID and secret', async () => {
```
acicovic:
```suggestion
	it( 'Widget should be available but inactive without Site ID and secret', async () => {
```

tests/Unit/SettingsPage/ValidationsTest.php
```diff
@@ -46,16 +46,16 @@ public function set_up(): void {
...
-	 * @covers \Parsely\UI\Settings_Page::validate_api_key
+	 * @covers \Parsely\UI\Settings_Page::validate_site_id
 	 */
-	public function test_validate_api_keys(): void {
-		$validate_api_key = self::get_method( 'validate_api_key', Settings_Page::class );
+	public function test_validate_site_ids(): void {
+		$validate_site_id = self::get_method( 'validate_site_id', Settings_Page::class );
 
-		// Test valid API keys.
+		// Test valid site_ids.
```
acicovic:
```suggestion
		// Test valid Site IDs.
```

tests/Unit/SettingsPage/ValidationsTest.php
```diff
@@ -64,10 +64,10 @@ public function test_validate_api_keys(): void {
 			'subdomain.subdomain.test.com',
 		);
 		foreach ( $valid_values as $value ) {
-			self::assertTrue( $validate_api_key->invoke( self::$settings_page, $value ) );
+			self::assertTrue( $validate_site_id->invoke( self::$settings_page, $value ) );
 		}
 
-		// Test invalid API keys.
+		// Test invalid site_ids.
```
acicovic:
```suggestion
		// Test invalid Site IDs.
```

tests/Integration/ProxyEndpointTest.php
```diff
@@ -133,10 +133,10 @@ public function test_do_not_register_route_when_proxy_is_disabled(): void {
 
 	/**
 	 * Verifies that calls return an error and do not perform a remote call when
-	 * the apikey is not populated in site options.
+	 * the site_id is not populated in site options.
```
acicovic:
```suggestion
	 * the Site ID is not populated in site options.
```

tests/Integration/Endpoints/RelatedProxyEndpointTest.php
```diff
@@ -69,7 +69,7 @@ public function test_do_not_register_route_when_proxy_is_disabled(): void {
 
 	/**
 	 * Verifies that calling `GET /wp-parsely/v1/related` returns an error and
-	 * does not perform a remote call when the apikey is not populated
+	 * does not perform a remote call when the site_id is not populated
```
acicovic:
```suggestion
	 * does not perform a remote call when the Site ID is not populated
```

tests/Integration/Endpoints/AnalyticsPostsProxyEndpointTest.php
```diff
@@ -70,22 +70,22 @@ public function test_do_not_register_route_when_proxy_is_disabled(): void {
 
 	/**
 	 * Verifies that calling `GET /wp-parsely/v1/stats/posts` returns an
-	 * error and does not perform a remote call when the apikey is not populated
+	 * error and does not perform a remote call when the site_id is not populated
```
acicovic:
```suggestion
	 * error and does not perform a remote call when the Site ID is not populated
```

src/class-scripts.php
```diff
@@ -116,11 +116,11 @@ public function enqueue_js_tracker(): void {
...
 		wp_enqueue_script( 'wp-parsely-tracker' );
 
-		// If we don't have an API secret, there's no need to set the API key.
-		// Setting the API key triggers the UUID Profile Call function.
+		// If we don't have an API secret, there's no need to set the Site ID.
+		// Setting the Site ID triggers the UUID Profile Call function.
 		if ( isset( $parsely_options['api_secret'] ) && is_string( $parsely_options['api_secret'] ) && '' !== $parsely_options['api_secret'] ) {
-			$js_api_key = "window.wpParselyApiKey = '" . esc_js( $this->parsely->get_api_key() ) . "';";
-			wp_add_inline_script( 'wp-parsely-loader', $js_api_key, 'before' );
+			$js_site_id = "window.wpParselySiteID = '" . esc_js( $this->parsely->get_site_id() ) . "';";
```
acicovic: Should this be `wpParselySiteID` or `wpParselySiteId`?
No strong opinions, just asking.

src/class-scripts.php
```diff
@@ -40,14 +40,14 @@ public function __construct( Parsely $parsely ) {
...
-		if ( $this->parsely->api_key_is_set() && true !== $parsely_options['disable_javascript'] ) {
+		if ( $this->parsely->site_id_is_set() && true !== $parsely_options['disable_javascript'] ) {
 			add_action( 'init', array( $this, 'register_scripts' ) );
 			add_action( 'wp_enqueue_scripts', array( $this, 'enqueue_js_tracker' ) );
 		}
 	}
 
 	/**
-	 * Registers scripts, if there's an API key value saved.
+	 * Registers scripts, if there's an Site ID value saved.
```
acicovic:
```suggestion
	 * Registers scripts, if there's a Site ID value saved.
```

src/class-parsely.php
```diff
@@ -406,44 +416,44 @@ public function convert_jsonld_to_parsely_type( string $type ): string {
 	}
 
 	/**
-	 * Determines if an API key is saved in the options.
+	 * Determines if an Site ID is saved in the options.
```
```suggestion
	 * Determines if a Site ID is saved in the options.
```

src/class-parsely.php
```diff
@@ -406,44 +416,44 @@ public function convert_jsonld_to_parsely_type( string $type ): string {
...
-				'' !== $options['apikey']
+				isset( $options['site_id'] ) &&
+				is_string( $options['site_id'] ) &&
+				'' !== $options['site_id']
 		);
 	}
 
 	/**
-	 * Determines if an API key is not saved in the options.
+	 * Determines if an Site ID is not saved in the options.
```
```suggestion
	 * Determines if a Site ID is not saved in the options.
```

tests/Integration/Integrations/FacebookInstantArticlesTest.php
```diff
@@ -81,13 +81,13 @@ public function test_integration_only_runs_when_FBIA_plugin_is_active(): void {
 	}
 
 	/**
-	 * Verifies that the integration is active only if an API key is set.
+	 * Verifies that the integration is active only if an Site ID is set.
```
acicovic:
```suggestion
	 * Verifies that the integration is active only if a Site ID is set.
```

src/class-parsely.php
```diff
@@ -359,6 +359,16 @@ public function get_options(): array {
...
 		}
 
+		/**
+		 * Makes the usage of `apikey` backward compatible while replacing it with `site_id`.
+		 *
+		 * @since 3.7.0.
+		 */
+		if ( isset( $options['apikey'] ) ) {
+			$options['site_id'] = $options['apikey'];
+			unset( $options['apikey'] );
```
acicovic: Could this result in removal of the `apikey` setting from the database?

If yes, let's not remove the `apikey` option yet, in case a plugin downgrade is needed for any reason. We could add a comment for this to remove it in wp-parsely version 4 for example.

src/Integrations/class-amp.php
```diff
@@ -154,18 +152,19 @@ public static function construct_amp_json(): string {
...
 	public static function construct_amp_config(): array {
 		$options = self::$parsely->get_options();
 
-		if ( isset( $options['apikey'] ) && is_string( $options['apikey'] ) && '' !== $options['apikey'] ) {
+		if ( self::$parsely->site_id_is_set() ) {
 			return array(
 				'vars' => array(
 					// This field will be rendered in a JS context.
-					'apikey' => esc_js( $options['apikey'] ),
+					'apikey' => esc_js( $options['site_id'] ),
```
acicovic: We're still using `apikey` here in the left part of the assignment. Is this on purpose?

tests/Integration/Integrations/AmpTest.php
```diff
@@ -149,24 +149,24 @@ public function test_can_register_Parsely_for_AMP_native_analytics(): void {
 		$amp       = new Amp( self::$parsely );
 		$analytics = array();
 
-		// If apikey is empty, $analytics are returned.
+		// If site_id is empty, $analytics are returned.
```
acicovic:
```suggestion
		// If Site ID is empty, $analytics are returned.
```

tests/Integration/Integrations/AmpTest.php
```diff
@@ -122,16 +122,16 @@ public function test_can_register_Parsely_for_AMP_analytics(): void {
...
```
acicovic:
```suggestion
		// If Site ID is empty, $analytics are returned.
```

src/class-parsely.php
```diff
@@ -406,44 +416,44 @@ public function convert_jsonld_to_parsely_type( string $type ): string {
 	}
 
 	/**
-	 * Determines if an API key is saved in the options.
+	 * Determines if an Site ID is saved in the options.
 	 *
 	 * @since 2.6.0
 	 *
-	 * @return bool True is API key is set, false if it is missing.
+	 * @return bool True is Site ID is set, false if it is missing.
```
acicovic, line 473: For all the functions that have been renamed, it would be nice to have a `@since` tag, stating that the function has been renamed. Let's leave this for last, as we'll need to determine if we will release this in 3.7 or 3.8 (to be discussed).

acicovic (commented):
Thanks for taking the time to wrangle this. I've included some questions, comments and requests for changes.

As this PR touches pretty much the core of the plugin and everything it includes, we should be extra careful about compatibility and side-effects at they may have a global repercussion. Any changes should be forward and backwards compatible. For example:

- User upgrades.
- User upgrades, then downgrades.
- User upgrades, changes the Site ID and then downgrades.

Any of the above scenarios should not present breakages or issues. Both `apikey` and `site_id` should work together for some time, just to make sure that we got our users covered in case of needed downgrades.

As such, if this has not been already done, I would ask for some local testing, taking into account the above scenarios (and more if there is anything I've missed).

src/Integrations/class-amp.php
mehmoodak: Yes it is on purpose. Ref: https://github.com/Parsely/wp-parsely/pull/1244#discussion_r1046622380

mehmoodak:
> User upgrades.
User upgrades, then downgrades.
User upgrades, changes the Site ID and then downgrades.

Thanks for this info, was not aware of this. Now I have updated the code to address this and tested these cases.

**Side Note:**
We do have functionality which calls the function with format `upgrade_plugin_to_version_[PARSELY_VERSION]` which is good and I think we should extend it to manage these kind of migrations because as of now it seems very limited. Let me know if this sounds good so we can create ticket for it.

src/class-parsely.php
```diff
@@ -475,4 +489,21 @@ public function get_api_secret(): string {
 
 		return $this->api_secret_is_set() ? $options['api_secret'] : '';
 	}
+
+	/**
+	 * Callback which called before updating parsely options
```
```suggestion
	 * Callback which is called before updating plugin options.
```

src/class-parsely.php
```diff
@@ -475,4 +489,21 @@ public function get_api_secret(): string {
...
+	 *
+	 * @param mixed $options   The value of the option now that it's been updated.
```
acicovic: Do we need the three spaces here?

src/class-parsely.php
```diff
@@ -475,4 +489,21 @@ public function get_api_secret(): string {
...
+	 */
+	public function pre_update_parsely_options( $options ) {
+		/**
+		 * Makes option `apikey` backward compatible.
```
```suggestion
		 * Keeps the deprecated `apikey` option for backward compatibility.
```

acicovic (commented):
I'll test this out but the code seems good to me. Left some minor suggestions/questions regarding comments.

acicovic:
> We do have functionality which calls the function with format upgrade_plugin_to_version_[PARSELY_VERSION] which is good and I think we should extend it to manage these kind of migrations because as of now it seems very limited. Let me know if this sounds good so we can create ticket for it.

I don't think we have a strong use case for this yet, but if we have in the future, we could consider giving more attention to this. If you want to post an issue to keep track of this, you are welcome to do so.

src/class-parsely.php
```diff
@@ -359,6 +360,16 @@ public function get_options(): array {
...
 		}
 
+		/**
+		 * Makes option `site_id` compatible with `apikey`.
+		 *
+		 * @since 3.7.0
+		 * @todo Should be removed after 4.0.0 upgrade
+		 */
+		if ( isset( $options['apikey'] ) ) {
+			$options['site_id'] = $options['apikey'];
```
chriszarate: I would probably just avoid changing the name of the option to navigate the upgrade/downgrade issues Alex mentioned. You can put a nice, long comment here that references `site_id` so that its clear and searchable.

mehmoodak: Apologies didn't get the suggestion can you please explain it a bit. There are multiple approaches to do this task so please let me know what you are suggesting.

1. Don't do anything and keep the code as it is.
2. Change all functions and all occurrances of `API Key` string but don't change the option key.
3. Change everything possible with backward compatibility.

I have gone with 3 and I think you are suggesting 2 ..... right? @acicovic please also share your thoughts on it. Thanks

acicovic: I think that @chriszarate is suggesting 2, the reason being that it is guaranteed not to break. With 3, we're introducing a chance of breakage even if our code is good. My preference for stability is 2. My preference for the "most updated code" is 3.

I've encountered an edge case where this can break, and will be posting it about below. Given we decide to go to solution 2, would the change need to happen in many places?

mehmoodak: @acicovic changes will be done in couple of places and I can do it.

So updating the PR with suggested change and will let you know when it is ready for review again.

acicovic:
Normal upgrade/downgrade paths work well for me, but I've managed to make this malfunction in an edge case scenario.

To reproduce (upgrade path):
1. Checkout the current develop branch
2. Go to the settings page and change the Site ID value without saving it
3. Checkout this branch
4. Save the setting from the settings page

This will result in a `Notice: Undefined index: site_id in /var/www/html/wp-content/plugins/wp-parsely/src/class-parsely.php` message. After this, the Site ID field will be empty.

Doing this the reverse way of checking out branches (downgrade path) will produce no notice, but will still result in a lost site ID.

In both cases, the plugin will display a `"The Parse.ly plugin is not active. You need to provide your Parse.ly Dash Site ID before things get cooking"` message in every admin page after the fact. However, until the key is re-added by a sufficiently authorized user, the plugin is disable and will not function.

This shows us that if the Site ID is being changed while the plugin gets upgraded or downgraded, it can result in a lost Site ID. Although this has a pretty small chance of happening, it is still an edge case to account for.

mehmoodak:
I am making changes as suggested above but the found edge case is going to be very very rare.

mehmoodak:
Feedback Implemented. Please have a look. Thanks

src/UI/class-recommended-widget.php
```diff
@@ -350,28 +350,16 @@ private function get_boost_params(): array {
...
 
 	/**
-	 * Checks if both the API key and API secret settings are populated with
+	 * Checks if both the Site ID and API secret settings are populated with
 	 * non-empty values.
 	 *
 	 * @since 2.5.0
 	 *
-	 * @return bool True if apikey and api_secret settings are not empty strings.
+	 * @return bool True if site_id and api_secret settings are not empty strings.
```
acicovic: Maybe add `apikey` here since we're referring to the setting itself?

```suggestion
	 * @return bool True if apikey and api_secret settings are not empty strings.
```

acicovic (commented):
This looks good to me overall. However, as I feel some tiredness creeping in, if @chriszarate can have an additional look into this, I would appreciate it.

Thanks!

acicovic:
> I am making changes as suggested above but the found edge case is going to be very very rare.

To be totally fair, this would present itself with any renamed setting. So it's not really a problem with this specific implementation, but with how settings work and how renaming them can have this side-effect.

tests/Integration/UI/NetworkAdminSitesListTest.php
```diff
@@ -50,63 +50,63 @@ public function set_up(): void {
...
 
 		self::$sites_list->run();
 
-		update_blog_option( $blog_id_with_api_key, Parsely::OPTIONS_KEY, array( 'apikey' => 'parselyrocks.example.com' ) );
+		update_blog_option( $blog_id_with_site_id, Parsely::OPTIONS_KEY, array( 'apikey' => 'parselyrocks.example.com' ) );
 
 		$this->table->prepare_items();
 
-		self::assertCount( 3, $this->table->items, 'There should be the main site, the subsite with the apikey set, and a subsite without.' );
+		self::assertCount( 3, $this->table->items, 'There should be the main site, the subsite with the site_id set, and a subsite without.' );
```
acicovic:
```suggestion
		self::assertCount( 3, $this->table->items, 'There should be the main site, the subsite with the Site ID set, and a subsite without.' );
```

acicovic (approved):

PR #1245, opened 2022-12-13
Fix SonarCloud errors in develop branch

acicovic:
## Description
I just realized that SonarCloud has raised a couple of errors for our `develop` branch. This PR fixes that by replacing a couple of `require` with `require_once`.

## Motivation and Context
Pet SonarCloud and make it feel really nice.

### DIFF
--- a/src/blocks/content-helper/class-content-helper.php
+++ b/src/blocks/content-helper/class-content-helper.php
@@ -23,7 +23,7 @@ class Content_Helper {
 	 * @since 3.5.0
 	 */
 	public function run(): void {
-		$content_helper_asset = require plugin_dir_path( PARSELY_FILE ) . 'build/content-helper.asset.php';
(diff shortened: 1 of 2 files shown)


acicovic:
It seems that fixing this code triggers a new complaint from SonarCloud, which I find a bit annoying and maybe a bit unnecessary to dig into.

I guess that we should flag those or the previous errors as `Won't fix`. @mehmoodak, asking for you opinion on this since I know you were working on SonarCloud lately.

acicovic:
By the way, the original complaint was to use `require_once` instead of `require`, but when we change this we get the errors as seen in the checks of this PR.

mehmoodak:
On SonarCloud the suggestion is to implement the namespace/modules but on our code we are importing data from build file which means no modules so seems like we should ignore it.

Apart from this I am working on fixing all SonarCloud issues and PR will be up soon so feel free to ignore this warning 🙂

mehmoodak (approved):

PR #1252, opened 2022-12-18
Fix PHPStan Errors

### LINKED ISSUE #921: Fix remaining PHPStan warnings
We still have PHPStan warnings in our code.

Take a look at the warnings and fix where applicable.

mehmoodak:
## Description
- Fixes PHPStan errors across whole codebase.
- PHPStan is like TypeScript and have same benefits, I think we should use this more in codebase so that it can have positive impact on developer productivity and increase reliability of the code.

Fixes: #921 

## How Has This Been Tested?
- Ran existing tests.
- For local testing PHPStan extension must be installed in the code editor (for VSCode this [one](https://marketplace.visualstudio.com/items?itemName=SanderRonde.phpstan-vscode))

## Suggestions:
- I think we should add PHPStan in our `pre-commit` hook and in our GitHub Actions to make defining types mandatory ... please share your thoughts 🙂


src/Endpoints/class-analytics-post-detail-api-proxy.php
```diff
@@ -29,8 +30,8 @@ public function run(): void {
 	 * Cached "proxy" to the Parse.ly `/analytics/post/detail` API endpoint.
 	 *
 	 * @param WP_REST_Request $request The request object.
-	 * @return stdClass|WPError stdClass containing the data or a WP_Error
-	 *                          object on failure.
+	 *
+	 * @return stdClass|WP_Error stdClass containing the data or a WP_Error object on failure.
```
mehmoodak, line 34: `WPError` is the wrong type therefore replaced it with `WP_Error`

src/Endpoints/class-analytics-post-detail-api-proxy.php
```diff
@@ -39,10 +40,10 @@ public function get_items( WP_REST_Request $request ) {
 	/**
 	 * Generates the final data from the passed response.
 	 *
-	 * @param array<string, mixed> $response The response received by the proxy.
+	 * @param array<stdClass> $response The response received by the proxy.
```
mehmoodak, line 43: Fixed type (same change has been made in other files related to API Endpoints)

src/Endpoints/class-analytics-posts-api-proxy.php
```diff
@@ -12,8 +12,11 @@
 
 use stdClass;
 use WP_REST_Request;
+use WP_Error;
 use Parsely\Parsely;
 
+use function Parsely\Utils\get_date_format;
```
mehmoodak, line 18: Many WordPress functions doesn't have proper typing so PHPStan gives errors if we try to do type casting so there were two ways to overcome this issue.

1. Do manual casting via PHPStan comments anywhwere we use these WP functions.
2. Make a util file that contains all of the WP functions that we are using and need typecasting.

Chose `2` because of maintainability reasons i.e. just one file which contains all of our typed functions.

src/Endpoints/class-base-api-proxy.php
```diff
@@ -148,12 +148,18 @@ protected function get_data( WP_REST_Request $request, bool $require_api_secret
 		}
 
 		// A proxy with caching behavior is used here.
-		$response = $this->proxy->get_items( $params );
+		$response = $this->proxy->get_items( $params ); // @phpstan-ignore-line.
```
mehmoodak, line 151: Used `// @phpstan-ignore-line.` in some places where a bit more refactoring was needed to fix the code and therefore have chances to break something.

src/Endpoints/class-base-api-proxy.php
```diff
@@ -148,12 +148,18 @@ protected function get_data( WP_REST_Request $request, bool $require_api_secret
...
+
+		if ( false === $response ) {
+			return new stdClass();
+		}
```
mehmoodak, line 155: This `false` case was missing and PHPStan were complaining about it so added it.

src/Endpoints/class-graphql-metadata.php
```diff
@@ -114,6 +114,7 @@ private function register_object_types(): void {
 				),
 			),
 		);
+		// @phpstan-ignore-next-line
 		register_graphql_object_type( self::GRAPHQL_CONTAINER_TYPE, $container_type );
```
mehmoodak: These functions are from `WPGraphQL` plugin and we don't have proper typing for this therefore ignoring this (same reason for other ignore comments in this file)

src/Endpoints/class-graphql-metadata.php
```diff
@@ -123,16 +124,16 @@ private function register_object_types(): void {
...
+			$post_id = $graphql_post->ID; // @phpstan-ignore-line
 			$post    = WP_Post::get_instance( $post_id );
 
 			if ( false === $post ) {
 				return null;
 			}
 
-			$options             = $this->parsely->get_options();
-			$object_types        = array_unique( array_merge( $options['track_post_types'], $options['track_page_types'] ) );
+			$object_types        = $this->parsely->get_all_track_types();
```
mehmoodak, line 134: We were using same logic in a couple of places so moved it to a function that we can use everywhere.

src/Endpoints/class-referrers-post-detail-api-proxy.php
```diff
@@ -126,6 +128,7 @@ private function generate_referrer_types_data( array $response ): stdClass {
 		}
 
 		// Set percentage values and format numbers.
+		// @phpstan-ignore-next-line.
```
mehmoodak, line 131: `$result` is `stdClass` here and not iterateable so therefore PHPStan were complaining about it. Instead of changing code I have added ignore comment to quickly make PHPStan happy. (same reason for next ignore comment)

src/Endpoints/class-referrers-post-detail-api-proxy.php
```diff
@@ -175,7 +178,7 @@ private function generate_referrers_data(
 		}
 
 		// If applicable, add the direct views.
-		if ( $direct_views >= $referrer_views ) {
+		if ( isset( $referrer_views ) && $direct_views >= $referrer_views ) {
```
mehmoodak, line 181: `$referrer_views` couldn't set in some cases so added this check as suggested by PHPStan.

src/Endpoints/class-rest-metadata.php
```diff
@@ -67,13 +66,18 @@ public function register_meta(): void {
...
 	 * @return array<string, mixed> The `parsely` object to be rendered in the REST API. Contains a
 	 *                              version number describing the response and the `meta` object
 	 *                              containing the actual metadata.
 	 */
 	public function get_callback( array $object ): array {
+		/**
+		 * Variable.
+		 *
+		 * @var int
+		 */
```
mehmoodak, line 81: `$post_id` was of type `mixed` before so added this comment to let PHPStan treat this as `int`.

Used same syntax in other places for better typing (mostly in util functions file)

src/Integrations/class-amp.php
```diff
@@ -55,7 +53,7 @@ public function is_amp_request(): bool {
 	public function can_handle_amp_request(): bool {
 		$options = self::$parsely->get_options();
 
-		return $this->is_amp_request() && is_array( $options ) && ! $options['disable_amp'];
```
mehmoodak, line 58: Removed `is_array` check because `$options` will always be array here as we are merging values with defaults so this check is extra.

src/Integrations/class-amp.php
```diff
@@ -112,7 +110,7 @@ public function register_parsely_for_amp_native_analytics( ?array $analytics ):
 
 		$options = self::$parsely->get_options();
 
-		if ( isset( $options['disable_amp'] ) && true === $options['disable_amp'] ) {
```
mehmoodak, line 115: Extra here because it will always be set either with the chosen option or with default option.

src/Integrations/class-amp.php
```diff
@@ -159,13 +157,11 @@ public static function construct_amp_json(): string {
 	 * @return array<string, array<string, string>>
 	 */
 	public static function construct_amp_config(): array {
-		$options = self::$parsely->get_options();
-
-		if ( isset( $options['apikey'] ) && is_string( $options['apikey'] ) && '' !== $options['apikey'] ) {
```
mehmoodak, line 164: Used already available functions which are better.

src/Metadata/class-category-builder.php
```diff
@@ -39,7 +39,10 @@ public function get_metadata(): array {
 	 * @since 3.4.0
 	 */
 	private function build_headline(): void {
-		$category                   = get_queried_object();
-		$this->metadata['headline'] = $this->clean_value( $category->name );
```
mehmoodak, line 43: `$category->name` could be null here so added check.

src/Metadata/class-post-builder.php
```diff
@@ -14,11 +14,15 @@
...
 
+use function Parsely\Utils\get_default_category;
+
 /**
  * Implements abstract Metadata Builder class to generate the metadata array
  * for a post page.
  *
  * @since 3.4.0
+ *
+ * @phpstan-import-type ParselyOptions from Parsely
```
mehmoodak, line 25: Defined custom type (similar to TypeScript interface). PHPStan uses this to improved typing of $options array in the codebase.

To view full type definition see `class-parsely.php` file.

Same thing is done for metadata attributes and recommended widget options.

PHPStan Docs: https://phpstan.org/writing-php-code/phpdoc-types#local-type-aliases

src/Metadata/class-post-builder.php
```diff
@@ -271,25 +275,28 @@ private function build_metadata_post_times(): void {
...
-	private function get_category_name( WP_Post $post_obj, array $parsely_options ): string {
+	private function get_category_name( WP_Post $post_obj, $parsely_options ): string {
 		$taxonomy_dropdown_choice = get_the_terms( $post_obj->ID, $parsely_options['custom_taxonomy_section'] );
 		// Get top-level taxonomy name for chosen taxonomy and assign to $parent_name; it will be used
 		// as the category value if 'use_top_level_cats' option is checked.
 		// Assign as the default category name if no value is checked for the chosen taxonomy.
-		$category_name = get_cat_name( get_option( 'default_category' ) );
-		if ( ! empty( $taxonomy_dropdown_choice ) && ! is_wp_error( $taxonomy_dropdown_choice ) ) {
+		$category_name = get_cat_name( get_default_category() );
+		if ( false !== $taxonomy_dropdown_choice && ! is_wp_error( $taxonomy_dropdown_choice ) ) {
```
mehmoodak, line 289: PHPStan gives error on the usage of `empty` functions due to many gotchas so therefore replaced it in the codebase wherever possible.

Here `get_the_terms` function can return `WP_Term[]|false|WP_Error` so therefore inside if we are checking for no false and no error means if we get WP_Terms then go inside if function

src/Metadata/class-post-builder.php
```diff
@@ -271,25 +275,28 @@ private function build_metadata_post_times(): void {
...
-				$term_name  = $this->get_top_level_term( $first_term->term_id, $first_term->taxonomy );
+				if ( null !== $first_term ) {
+					$term_name = $this->get_top_level_term( $first_term->term_id, $first_term->taxonomy );
+				}
 			} else {
 				$term_name = $this->get_bottom_level_term( $post_obj->ID, $parsely_options['custom_taxonomy_section'] );
 			}
 
-			if ( is_string( $term_name ) && 0 < strlen( $term_name ) ) {
+			if ( isset( $term_name ) && is_string( $term_name ) && 0 < strlen( $term_name ) ) {
```
mehmoodak, line 299: Just a null check. Anywhere we are adding these checks are because PHPStan complains.

Will skip these type of comments below so please keep this reason in mind 🙂

src/Metadata/class-post-builder.php
```diff
@@ -442,15 +451,15 @@ private function get_coauthor_names( int $post_id ): array {
 						$coauthor_slug = preg_replace( '#^cap-#', '', $coauthor->slug );
 						$post_author   = $coauthors_plus->get_coauthor_by( 'user_nicename', $coauthor_slug );
```
mehmoodak, line 452: `get_coauthor_by` can return `object|false`.

Source: https://github.com/Automattic/Co-Authors-Plus/blob/3a4cc3693e42a3fbea31a40a3edc367ddd1beacd/co-authors-plus.php#L301

src/Metadata/class-post-builder.php
```diff
@@ -475,7 +484,7 @@ private function get_author_name( ?WP_User $author ): string {
 			return '';
 		}
 
-		if ( ! empty( $author->display_name ) ) {
+		if ( '' !== $author->display_name ) {
```
mehmoodak, line 487: If there is no value then by default we have `''` so therefore just checking this.

src/UI/class-metadata-renderer.php
```diff
@@ -107,7 +107,7 @@ public function render_metadata( string $meta_type ): void {
 		$metadata = ( new Metadata( $this->parsely ) )->construct_metadata( $parsed_post );
 
 		// Something went wrong - abort.
-		if ( 0 === count( $metadata ) || ! isset( $metadata['headline'] ) ) {
```
mehmoodak, line 110: Only one check is enough here i.e. can check for both conditions.

src/UI/class-metadata-renderer.php
```diff
@@ -116,13 +116,13 @@ public function render_metadata( string $meta_type ): void {
 			echo '<script type="application/ld+json">' . wp_json_encode( $metadata ) . '</script>';
 		} else {
 			// Assume `meta_type` is `repeated_metas`.
-			$parsely_post_type = $this->parsely->convert_jsonld_to_parsely_type( $metadata['@type'] );
+			$parsely_post_type = $this->parsely->convert_jsonld_to_parsely_type( $metadata['@type'] ?? '' );
```
mehmoodak, line 119: All attributes of metadata are defined as optional in the interface there we need these extra values in case of `null`. Same reason for other places where we are using `??` with metadata.

src/UI/class-network-admin-sites-list.php
```diff
@@ -21,6 +20,12 @@
  */
 final class Network_Admin_Sites_List {
 	const COLUMN_NAME = 'parsely-api-key';
+	/**
+	 * Instance of Parsely class.
+	 *
+	 * @var Parsely
+	 */
+	private $parsely;
```
mehmoodak, line 28: Not sure why it wasn't defined before. Since we were using this in the codebase so defined it.

src/UI/class-recommended-widget.php
```diff
@@ -18,6 +18,18 @@
...
+ *   title: string,
+ *   return_limit: int,
+ *   display_direction: string,
+ *   published_within: int,
+ *   sort: string,
+ *   boost: string,
+ *   personalize_results: bool,
+ *   img_src: string,
+ *   display_author: bool,
+ * }
```
mehmoodak, line 32: Define structure for array so that PHPStan can enforce typings.

Source: https://phpstan.org/writing-php-code/phpdoc-types#array-shapes

src/UI/class-recommended-widget.php
```diff
@@ -27,6 +39,23 @@ final class Recommended_Widget extends WP_Widget {
...
+		'return_limit'        => 5,
+		'display_direction'   => 'vertical',
+		'published_within'    => 0,
+		'sort'                => 'score',
+		'boost'               => 'views',
+		'personalize_results' => false,
+		'img_src'             => 'parsely_thumb',
+		'display_author'      => false,
+	);
+
```
mehmoodak, line 58: In the code below we were using these as defaults if not defined so created a separate array for this and a merge function (same like parsely options)

src/UI/class-recommended-widget.php
```diff
@@ -88,14 +117,15 @@ private function get_api_url( string $api_key, ?int $published_within, ?string $
 	/**
 	 * This is the widget function.
 	 *
-	 * @param array $args Widget Arguments.
-	 * @param array $instance Values saved to the db.
+	 * @param array<string, string> $args            Widget Arguments.
+	 * @param array<mixed>          $widget_settings Values saved to the db.
 	 */
-	public function widget( $args, $instance ): void {
+	public function widget( $args, $widget_settings ): void /* @phpstan-ignore-line */ {
```
mehmoodak, line 123: - Changed var name because we have to use `get_widget_settings` function below which will either use our values or default. So used `instance` there so that we will have less changes in the code (if we use a separate name then every occurrance of `instance` will need to be changed.
- Uses `/* @phpstan-ignore-line */` here because our function signature have better types than `WP_Widget::widget` functions and compiler were complaining so going with more type safe option

src/UI/class-recommended-widget.php
```diff
@@ -109,27 +139,26 @@ public function widget( $args, $instance ): void {
...
 
 		// Set up the variables.
-		$options = $this->parsely->get_options();
 		$api_url = $this->get_api_url(
-			$options['apikey'],
+			$this->parsely->get_api_key(),
 			$instance['published_within'],
 			$instance['sort'],
 			$instance['boost'],
-			(int) $instance['return_limit']
```
mehmoodak, line 118: No need as our defined array strucutre already knows that `return_limit` is of type `int`

src/UI/class-recommended-widget.php
```diff
@@ -109,27 +139,26 @@ public function widget( $args, $instance ): void {
...
+			$instance['return_limit']
 		);
 
 		$recommended_widget_script_asset = require plugin_dir_path( PARSELY_FILE ) . 'build/recommended-widget.asset.php';
 
 		?>
 
 		<div class="parsely-recommended-widget"
-			data-parsely-widget-display-author="<?php echo esc_attr( wp_json_encode( isset( $instance['display_author'] ) && $instance['display_author'] ) ); ?>"
```
mehmoodak, line 126: No need of null checks here because these will either have our defined values or default ones.

src/UI/class-recommended-widget.php
```diff
@@ -172,19 +201,21 @@ public function form( $instance ): void {
...
 		// editable fields: title.
-		$title               = ! empty( $instance['title'] ) ? $instance['title'] : '';
-		$return_limit        = ! empty( $instance['return_limit'] ) ? (int) $instance['return_limit'] : 5;
-		$display_direction   = ! empty( $instance['display_direction'] ) ? $instance['display_direction'] : 'vertical';
-		$published_within    = ! empty( $instance['published_within'] ) ? $instance['published_within'] : 0;
-		$sort                = ! empty( $instance['sort'] ) ? $instance['sort'] : 'score';
-		$boost               = ! empty( $instance['boost'] ) ? $instance['boost'] : 'views';
-		$personalize_results = ! empty( $instance['personalize_results'] ) ? $instance['personalize_results'] : false;
-		$img_src             = ! empty( $instance['img_src'] ) ? $instance['img_src'] : 'parsely_thumb';
-		$display_author      = ! empty( $instance['display_author'] ) ? $instance['display_author'] : false;
```
mehmoodak, line 187: Since we are assigning defaults above so no need here.

src/UI/class-recommended-widget.php
```diff
@@ -259,26 +290,30 @@ class="tiny-text" aria-labelledby="<?php echo esc_attr( $this->get_field_id( 'pu
...
-	 * @param array $new_instance The new values for the db.
-	 * @param array $old_instance Values saved to the db.
-	 * @return array
+	 * @param WidgetSettings $new_instance The new values for the db.
+	 * @param WidgetSettings $old_instance Values saved to the db.
+	 *
+	 * @return WidgetSettings
 	 */
-	public function update( $new_instance, $old_instance ): array {
+	public function update( $new_instance, $old_instance ) /* @phpstan-ignore-line */ {
```
mehmoodak, line 305: Same as above i.e. enforcing types instead of using WP default function signature.

src/UI/class-recommended-widget.php
```diff
@@ -326,21 +361,36 @@ private function get_boost_params(): array {
 	private function api_key_and_secret_are_populated(): bool {
 		$options = $this->parsely->get_options();
 
-		// No options are saved, so API key is not available.
-		if ( ! is_array( $options ) ) {
-			return false;
-		}
```
mehmoodak, line 332: `$options` always be `array` and have keys defined with at least default values.

src/UI/class-settings-page.php
```diff
@@ -19,6 +19,20 @@
...
+ *   option_key: string,
+ *   label_for: string,
+ *   title?: string,
+ *   help_text?: string,
+ *   yes_text?: string,
+ *   filter?: string,
+ *   optional_args?: array<string, string>,
+ *   select_options?: array<string, string>,
+ *   radio_options?: array<string, string>,
+ * }
```
mehmoodak, line 35: Define setting arguments that we need to print controls of each setting. Here optional fields are marked as `?:` and these will also need null check below.

src/UI/class-settings-page.php
```diff
@@ -892,17 +932,17 @@ public function print_track_post_types_table( array $args ): void {
...
 	 */
 	public function get_tracking_values_for_display(): array {
 		$options = $this->parsely->get_options();
 		$types   = array( 'post', 'page' );
 		$result  = array();
 
 		foreach ( $types as $type ) {
-			$array_key = "track_{$type}_types";
-			if ( array_key_exists( $array_key, $options ) ) {
-				foreach ( $options[ $array_key ] as $post_type ) {
```
mehmoodak, line 905: Second line was giving PHPStan error on `$options` so changed the logic a bit to avoid the error without defining ignore comment.

src/UI/class-settings-page.php
```diff
@@ -941,22 +982,24 @@ public function validate_options( array $input ): array {
...
+		if ( '' !== $input['metadata_secret'] ) {
 			if ( strlen( $input['metadata_secret'] ) !== 10 ) {
 				add_settings_error(
 					Parsely::OPTIONS_KEY,
 					'metadata_secret',
 					__( 'Metadata secret is incorrect. Please contact Parse.ly support!', 'wp-parsely' )
 				);
-			} elseif ( isset( $input['parsely_wipe_metadata_cache'] ) && 'true' === $input['parsely_wipe_metadata_cache'] ) {
+			} elseif (
+				isset( $input['parsely_wipe_metadata_cache'] ) && 'true' === $input['parsely_wipe_metadata_cache'] // @phpstan-ignore-line
```
mehmoodak, line 993: `parsely_wipe_metadata_cache` is defined as bool in our array structure but here we are using true as string so added ignore comment to avoid this error (just a quicker and faster solution)

src/class-parsely.php
```diff
@@ -53,6 +78,7 @@ class Parsely {
 		'metadata_secret'             => '',
 		'parsely_wipe_metadata_cache' => false,
 		'disable_autotrack'           => false,
+		'plugin_version'              => '',
```
mehmoodak, line 81: Added `plugin_version` as non optional field while defining array structure so therefore it also needs the default value.

src/class-parsely.php
```diff
@@ -416,8 +459,7 @@ public function api_key_is_set(): bool {
 		$options = $this->get_options();
 
 		return (
-				isset( $options['apikey'] ) &&
-				is_string( $options['apikey'] ) &&
+				is_string( $options['apikey'] ) && // @phpstan-ignore-line.
```
mehmoodak: We can remove this condition because `apikey` will always be string here .... but just keeping it for safety purpose i.e. for edge cases where we have some other type in the database.

Let me know if you think it is good to remove.

mehmoodak (commented):

wp-parsely.php
```diff
@@ -194,7 +194,7 @@ function parsely_integrations( $parsely = null ): Integrations {
 	// If $parsely value is "", then this function is being called by the init
 	// hook and we can get the value from $GLOBALS. If $parsely is an instance
 	// of the Parsely object, then this function is being called by a test.
-	if ( empty( $parsely ) || get_class( $parsely ) !== Parsely::class ) {
+	if ( '' === $parsely || get_class( (object) $parsely ) !== Parsely::class ) {
```
chriszarate:
```suggestion
	if ( ! is_object( $parsely ) || get_class( $parsely ) !== Parsely::class ) {
```

src/Endpoints/class-analytics-posts-api-proxy.php
chriszarate: @mehmoodak What about something like https://github.com/szepeviktor/phpstan-wordpress

(Never used it, just googled. Since you mentioned PHPStan is like TypeScript, it seemed like their might be community typings out there.)

src/Endpoints/class-graphql-metadata.php
chriszarate: https://github.com/AxeWP/wp-graphql-stubs ?

src/Endpoints/class-analytics-posts-api-proxy.php
mehmoodak: @chriszarate we are already using it in our project and it does help a lot while writing code related to WordPress.

But there is a catch here like some WordPress functions does return `mixed` value which PHPStan complains about so for that we do need an additional doc comment to specify the correct type for our usecase.

src/Endpoints/class-graphql-metadata.php
mehmoodak: Thanks @chriszarate, added this and now it doesn't require ignore comments.

chriszarate (approved):

src/Endpoints/class-base-api-proxy.php
acicovic: I wonder if this should return an error instead of an empty `stdClass`. Thoughts?

src/Endpoints/class-rest-metadata.php
```diff
@@ -67,13 +66,18 @@ public function register_meta(): void {
 	 * Function to get hooked into the `get_callback` property of the `parsely`
 	 * REST API field. It generates the `parsely` object in the REST API.
 	 *
-	 * @param array $object         The WordPress object to extract to render the metadata for,
+	 * @param array<string, mixed> $object         The WordPress object to extract to render the metadata for,
```
acicovic: Is spacing correct here?

src/Metadata/class-category-builder.php
acicovic: Maybe we're handling this later in other parts of the code, but I wonder if we should we have a default empty vale for `$this->metadata['headline']`?

src/Metadata/class-post-builder.php
```diff
@@ -92,13 +96,13 @@ private function build_type(): void {
...
 		 *
-		 * @param array $jsonld_type JSON-LD @type value, default is NewsArticle.
+		 * @param string $jsonld_type JSON-LD @type value, default is NewsArticle.
 		 * @param int $id Post ID.
 		 * @param string $post_type The Post type in WordPress.
 		 *
 		 * @since 2.5.0
 		 */
-		$type = (string) apply_filters( 'wp_parsely_post_type', 'NewsArticle', $this->post->ID, $this->post->post_type );
+		$type = apply_filters( 'wp_parsely_post_type', 'NewsArticle', $this->post->ID, $this->post->post_type );
```
acicovic, line 105: Are we certain that `(string)` is not needed here?

src/blocks/recommendations/class-recommendations-block.php
```diff
@@ -75,15 +75,15 @@ public static function render_callback( array $attributes ): string {
 		wp_enqueue_script( 'wp-parsely-recommendations-view-script' );
 		ob_start();
 		?>
-<section
+		<section
```
acicovic, line 78: I think this will inject unneeded tabs into the markup. This is probably not an issue, but wanted to pinpoint it.

src/class-parsely.php
```diff
@@ -475,4 +517,17 @@ public function get_api_secret(): string {
 
 		return $this->api_secret_is_set() ? $options['api_secret'] : '';
 	}
+
+	/**
+	 * Get all track post types.
```
```suggestion
	 * Gets all tracked post types.
```

src/class-parsely.php
acicovic: I think that we can remove this, but have no issue to leave it as is for extra safety. I'll leave the final decision to you.

acicovic (commented):
This is really excellent work and I greatly appreciate the time and effort you've put into it, not only for the code but also for the comments that make the review so much easier.

acicovic:
@mehmoodak, since you've invested the time and effort to do this, I agree that we should integrate PHPStan as a part of our standard workflow in order to keep things consistent as we go forward. Feel free to open PRs or set any settings needed for this.

Also, if you think that some lines where you've put ignore comments have a strong reason to be refactored, feel free to track this in an issue so we can tackle it in the future.

src/Endpoints/class-base-api-proxy.php
mehmoodak: I think returning empty `stdClass` is better ...... because in reality it is not an error so we also shouldn't treat it like error.

src/Metadata/class-category-builder.php
mehmoodak: I think we shouldn't have default value assigned to it because of [this condition](https://github.com/Parsely/wp-parsely/blob/fix/phpstan-errors/src/UI/class-metadata-renderer.php#L110) which decides whether something went wrong or not.

src/Metadata/class-post-builder.php
mehmoodak: Personally I am confident because of two reasons.

1. We are treating this variable as string going forward.
2. Our test also indicates that we returns string here (like [this one](https://github.com/Parsely/wp-parsely/blob/fix/phpstan-errors/tests/Integration/OtherTest.php#L156))

If you want I can put it back with an ignore PHPStan comment.

src/blocks/recommendations/class-recommendations-block.php
mehmoodak: It was causing issue while folding/unfolding the functions in the code editor. Let me know if we need to revert this otherwise I think this one is better.

src/class-parsely.php
mehmoodak: Okay removing this then since user can only give this input from settings which indicates only strings can enter.

mehmoodak:
Thanks @acicovic for the review. I have implemented the feedback and left comments on others so please have a look again. Thanks 🙂

acicovic (approved):

PR #1268, opened 2022-12-26
Add VSCode Workspace Settings

### LINKED ISSUE #1435: Add VSCode Workspace Settings
Since VSCode is widely used among developers so we should add support for recommended VSCode extensions and the needed settings.

mehmoodak:
## Description
- Add VSCode Workspace settings which includes
  - Recommended Extensions.
  - Default settings of extensions.
  - Adding XDebug task to debug the code.

Closes: #1435 

## How Has This Been Tested?
- Ran existing tests to make sure nothing is breaking.


.vscode/settings.json
```diff
@@ -0,0 +1,116 @@
...
+	"phpCodeSniffer.autoExecutable": true,
+	"phpCodeSniffer.standard": "Custom",
+	"phpCodeSniffer.standardCustom": "WordPress-VIP-Go,WordPress-Docs",
+	// PHP Code Beautifier
+	"phpcbf.standard": "WordPress-VIP-Go,WordPress-Docs",
+	"[php]": {
+		"editor.defaultFormatter": "obliviousharmony.vscode-php-codesniffer"
+	},
+	// PHPStan
+	"phpstan.timeout": 30000,
```
mehmoodak: I have tested the settings on my other Macbook and due to timeout phpstan was not working so I increased the timeout from default 10s to 30s just to make sure that it works for most devs without any configs.

.vscode/settings.json
```diff
@@ -0,0 +1,116 @@
...
+	"phpstan.ignoreErrors": [
+		".*The Xdebug PHP extension is active.*", ".*slow down performance.*", ".*breakpoints.*",
```
mehmoodak: If `Xdebug` is setup in local environment then PHPStan raises this error and stops the execution so we have added this error in ignore list so that PHPStan will complete its analyzation no matter what is the state of Xdebug.

.vscode/settings.json
```diff
@@ -0,0 +1,116 @@
...
+		"xmlreader",
+		"xmlrpc",
+		"xmlwriter",
+		"xsl",
+		"Zend OPcache",
+		"zip",
+		"zlib",
+		"wordpress",
+		// Custom
+		"vendor/php-stubs/wordpress-tests-stubs/"
```
mehmoodak: WordPress tests stubs library didn't have types library for PHPStan so added it via composer and included it in the list.

mehmoodak (commented):

.vscode/launch.json
```diff
@@ -0,0 +1,14 @@
+{
+	// Use IntelliSense to learn about possible attributes.
+	// Hover to view descriptions of existing attributes.
+	// For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
```
GaryJones: Are comments allowed in these JSON files?

src/UI/class-admin-bar.php
```diff
@@ -67,7 +67,7 @@ public function run(): void {
 	 */
 	public function admin_bar_parsely_stats_button( WP_Admin_Bar $admin_bar ): void {
 		/**
-		 * Internal Variable.
+		 * Variable.
```
GaryJones: This comment adds nothing but noise now.

.vscode/launch.json
mehmoodak: Comments are not allowed in JSON but VSCode doesn't complian about it if added. Infact here VSCode added it on its own so maybe its better to keep them as they can convery more meaning.

src/UI/class-admin-bar.php
mehmoodak: It gives proper type to the variable by following PHPStan type-casting convention. If we remove this comment then type of the variable will come as `mixed` and PHPStan on `level: 9` complains about variable having mixed types.

.vscode/launch.json
acicovic: For all its settings, VSCode employs `json` files described as "JSON with Comments", that allow for the use of such comments.

acicovic:
Hey @mehmoodak, thank you for you work.

What approach do you suggest for anyone who already has their own settings files, or anyone who wants to further customize the ones you're providing here?

mehmoodak:
> What approach do you suggest for anyone who already has their own settings files, or anyone who wants to further customize the ones you're providing here?

@acicovic  It depends whether the dev have user level settings or workspace level settings.

- If they have User Level Settings then the Workspace Level Settings will have more priority and they can extend their User Level Settings in case of customization (not prefereable case)

- If the have WorkSpace Level Settings then those should match with the ones defined here if not then we should address the mismatch issue i.e. in most cases Workspace Level Settings should have everything that dev needs for their setup.

acicovic:
I think that having the settings ready is very important. At the same time, I do not think we should force IDE settings, since everyone should be free to use their own set of extensions, removing and adding as they wish.

For example, I use a spellchecking extension. However, as it does generate noise, I do know that not everyone would want it, and so I would not include it in the default configuration. However, I still want to be able to use it personally. I also have an additional debugging configuration using the Brave browser, which not everyone will have. I also have additional settings that are personal preference.

If this solution overwrites personal settings and does not give a choice to have them, then maybe I would change the approach by renaming the `.vscode` folder to something else. Then, in the dev instructions I would add a step for renaming the folder to `.vscode` if the developer desires to have the predefined settings. Finally, I would exclude the `.vscode` folder from git. This should allow every developer to have their own settings, while also taking advantage of the ready settings as a starting point if they wish.

What do you think?

chriszarate:
@acicovic This page explains how workspace settings (what are being proposed here) override user settings:

https://code.visualstudio.com/docs/getstarted/settings#_settings-precedence

acicovic:
Since I have workspace settings, they will be overwritten by the current solution. Personally, I think we should definitely propose these settings, but not force them.

However, @mehmoodak and @chriszarate, if you both think that the current approach should be the way to go, let me know and we can get this forward. I guess I can git ignore those locally, so everyone should still be able to function as desired.

mehmoodak:
> Since I have workspace settings, they will be overwritten by the current solution. 

@acicovic is it possible to move these settings to User Level instead of Workspace?

> if you both think that the current approach should be the way to go

I think it is the better way and should be the default one. For personal preferences we can utilise user level settings or maybe there is some other option (I didn't explored this other option because that is all I need 🙂)

acicovic:
I'll need some time to test this to see if there is a reasonable workaround. If there isn't, we should not go ahead with this, as forcing settings on others without any other choice is not something I can endorse.

For clarity and ease of use, whichever solution is chosen should be noted in the CONTRIBUTING.md file.

chriszarate:
It doesn't seem that this is solvable:
https://github.com/microsoft/vscode/issues/40233

Alex, are the workspace settings you use broadly useful? Could they be contributed here? Alternatively, it seems like this should be documented instead of committed.

acicovic:
I think I found a way to commit default settings while also giving the choice to override them. It involves moving all those default settings to a `.code-workspace` file and actually letting the user use `settings.json` (which always has the last word) for any overriding. `settings.json` should be added to `.gitignore`.

The caveat is that for this scheme to work, the user must open the project as a workspace (not a folder). A small compromise for the benefits IMO, but we should make it clear in our docs in order to avoid confusion.

@mehmoodak, @chriszarate, what do you think? I can do the work if needed.

mehmoodak:
@acicovic Fine with the recommended approach. Please go ahead with the changes and Thanks for taking it to the finish line.

src/UI/class-admin-bar.php
GaryJones: The `@var` below does that. The summary of `Variable` is redundant.

mehmoodak: Removing description and just having `@var` in doc comment gives following error from PHPCS

```
Missing short description in doc comment
```

acicovic:
Now that VSCode supports profiles (see [here](https://code.visualstudio.com/updates/v1_75#_profiles) and [here](https://www.youtube.com/watch?v=PCDfH84cvCQ&t=250s)), this is the approach I would recommend for tackling this.

mehmoodak:
> Now that VSCode supports profiles (see [here](https://code.visualstudio.com/updates/v1_75#_profiles) and [here](https://www.youtube.com/watch?v=PCDfH84cvCQ&t=250s)), this is the approach I would recommend for tackling this. 

I have tried it and it seems like there is fundamental difference between Workspace Settings and Profiles.

- `Workspace Settings` are defined on top of `User Settings`
- `Profiles` are not defined on top of `User / Workspace Settings` which means if any of the user wanted to customise something then they will have to make changes in the actual profile which will update that profile for other users also (to me this profile feature seems like a per user feature or a team feature where full team agrees on one setup)

Now I have tried using `Multi-root Workspaces` feature of VSCode that shouldn't conflict with already available workspace settings.

.vscode/wp-parsely.code-workspace
```diff
@@ -0,0 +1,152 @@
...
+			"editor.defaultFormatter": "obliviousharmony.vscode-php-codesniffer"
+		},
+		// PHPStan
+		"phpstan.timeout": 30000,
+		"phpstan.ignoreErrors": [
+			".*The Xdebug PHP extension is active.*",
+			".*slow down performance.*",
+			".*breakpoints.*",
+		],
+		// PHP Intelephense
```
acicovic, line 70: I'm using PHP Intelephense as well. Strangely, some WordPress-related classes and functions do not get recognized, unless I copy this section to my own `settings.json`.

So, might be worth to add a note like:
```suggestion
		// PHP Intelephense - Copy this section to your settings.json if certain WordPress functions/classes do not get recognized
```

.vscode/wp-parsely.code-workspace
```diff
@@ -0,0 +1,152 @@
...
+		},
+		"[typescript]": {
+			"editor.defaultFormatter": "dbaeumer.vscode-eslint",
+		},
+		"[javascriptreact]": {
+			"editor.defaultFormatter": "dbaeumer.vscode-eslint",
+		},
+		"[typescriptreact]": {
+			"editor.defaultFormatter": "dbaeumer.vscode-eslint",
+		},
```
acicovic: This can be shortened to the following if we want to:

```suggestion
"[javascript][typescript][javascriptreact][typescriptreact]": {
	"editor.defaultFormatter": "dbaeumer.vscode-eslint"
},
```

Also wondering if we would like to add something like `"editor.maxTokenizationLineLength": 2500` in there.

.vscode/wp-parsely.code-workspace
```diff
@@ -0,0 +1,152 @@
...
+		]
+	},
+	"extensions": {
+		"recommendations": [
+			"dbaeumer.vscode-eslint",
+			"usernamehw.errorlens",
+			"xdebug.php-debug",
+			"bmewburn.vscode-intelephense-client",
+			"obliviousharmony.vscode-php-codesniffer",
+			"persoderlind.vscode-phpcbf",
```
acicovic, line 29: I'm a bit skeptical about those 2 extensions. I'm using [PHP Sniffer & Beautifier](https://marketplace.visualstudio.com/items?itemName=ValeryanM.vscode-phpsab) instead (also as my PHP formatter). Not a blocker though as I can work as things are.

Have you tried the above extension and opted using the 2 others instead? If yes, were they offering something extra?

.vscode/wp-parsely.code-workspace
```diff
@@ -0,0 +1,152 @@
...
+				"type": "php",
+				"request": "launch",
+				"port": 9003
+			},
+		]
+	},
+	"extensions": {
+		"recommendations": [
+			"dbaeumer.vscode-eslint",
+			"usernamehw.errorlens",
```
acicovic, line 25: I'm using Error Lens as well, bu I understand it can be noisy/overwhelming for some folks. But since this is just suggested, I'm fine with leaving it.

.vscode/wp-parsely.code-workspace
```diff
@@ -0,0 +1,152 @@
...
+			"xdebug.php-debug",
+			"bmewburn.vscode-intelephense-client",
+			"obliviousharmony.vscode-php-codesniffer",
+			"persoderlind.vscode-phpcbf",
+			"sanderronde.phpstan-vscode"
+		]
+	},
+	"settings": {
+		// VSCode Custom Settings
+		"typescript.tsdk": "node_modules/typescript/lib",
```
acicovic, line 35: > "typescript.tsdk": "node_modules/typescript/lib",

Why do we need this line? What does it do for us?

CONTRIBUTING.md
```diff
@@ -26,6 +26,17 @@ The plugin uses strong types, so be sure to declare `strict_types=1` on new file
 
 For JavaScript we recommend installing ESLint. This plugin includes a [.eslintrc](https://github.com/Parsely/wp-parsely/blob/develop/.eslintrc) file that defines our coding standards.
 
+### Setting up IDE / Code Editor
```
acicovic:
```suggestion
### Setting up your IDE / Code Editor
```

CONTRIBUTING.md
```diff
@@ -26,6 +26,17 @@ The plugin uses strong types, so be sure to declare `strict_types=1` on new file
...
+
+#### VSCode
+
+This plugin uses the [Multi-root Workspaces](https://code.visualstudio.com/docs/editor/multi-root-workspaces) feature provided by VSCode which setup the editor
```
acicovic: We have a line break here? Is it needed? Also, we might need some better wording here but we can go ahead for now.

acicovic (commented):
I'll be able to use this solution, so extra feedback might come with usage over time. Some additional settings we could add to the workspace file are:
```
"intelephense.environment.phpVersion": "7.1",
"search.exclude": {
	"build/": true
},
```

- The first setting should make Intelephense flag code that is incompatible with PHP 7.1 syntax. Noting that while our plugin is PHP 7.2 and up, we are currently keeping 7.1 syntax.
- The second setting prevents searches in the `build` folder, avoiding noise.

Also, as part of the PR I think we should git ignore all files in the `.vscode` folder except `wp-parsely.code-workspace`, so devs can have their customized `settings.json`, `launch.json` etc.

.vscode/wp-parsely.code-workspace
acicovic: What I said above is not the case if you open this as a workspace, so let's probably ignore this.

acicovic:
Some related settings that might be of interest (I have them in my global settings file).
```
"intelephense.format.enable": false,
"php.suggest.basic": false,
"[markdown]": {
	"files.trimTrailingWhitespace": false
},
```

.vscode/wp-parsely.code-workspace
mehmoodak: > Also wondering if we would like to add something like "editor.maxTokenizationLineLength": 2500 in there.

I think as of now we don't need this and default is fine.

.vscode/wp-parsely.code-workspace
mehmoodak: Not sure whether I tried [PHP Sniffer & Beautifier](https://marketplace.visualstudio.com/items?itemName=ValeryanM.vscode-phpsab) extension or not but I did try couples and only this combination of two extensions worked for me.

.vscode/wp-parsely.code-workspace
mehmoodak: Sounds Good. I also think we should keep/suggest it and devs can then take their own decision.

.vscode/wp-parsely.code-workspace
mehmoodak: Instead of using the VSCode provided TS version it uses the one from `node_modules` folder on which project is dependent.

CONTRIBUTING.md
mehmoodak: Deleted line break.

mehmoodak:
Implemented feedback. Please have a look again. Thanks

acicovic (approved):
I think there are a couple of points we can improve or research more, but this works, so good to merge for now. 👍

PR #1269, opened 2022-12-26
Add type coverage

GaryJones:
## Description
With PHPStan now passing without error, we can check how much parameter, return, and property types we now have.

The configuration values are set to not produce a PHPStan run failure - increasing the percentages would do. This could be checked each week / month and the config upped accordingly, but at least there shouldn't be an unknown decrease in percentages now.

## Motivation and Context
Property types aren't supported until PHP 7.4, so they can't be used yet, hence the 0%.

Param types may be `mixed` - not officially supported until PHP 8, which means that they aren't given a type in the code at the moment, so 100% coverage won't be possible until PHP 8 is supported or other refactorings are done.

Uses https://github.com/TomasVotruba/type-coverage.

## How Has This Been Tested?
1. Run `composer static analysis` - it will pass.
2. Apply patch, run `composer static-analysis`, and it will continue to pass.
3. Temporarily increase config percentages, run `composer static-analysis`, and it will now fail.
4. Optionally change `true` to `fail` to print suggestions on possible change.

Twi test runs - the first with failure percentages high enough to create test fails, the second with the percentages set at the exact level of coverage there is:
Screenshot 2022-12-26 at 12 52 25


GaryJones:
Looks like we'd need PHP 7.2 as the min version to get a version of the type coverage package to install. Now that PHP 8.2 has released, and there are efforts to [drop earlier versions of PHP for WP core](https://core.trac.wordpress.org/ticket/57345), then I'll leave it to the wp-parsely team to judge which step forward to take here.

Alternatively, since this is a development-only package, then a `php` constraint can be added to `require-dev`, and the Workflows adjusted accordingly.

phpstan.neon
```diff
@@ -7,3 +7,8 @@ parameters:
     - uninstall.php
   scanFiles:
     - vendor/axepress/wp-graphql-stubs/wp-graphql-stubs.php
+  type_coverage:
+      return_type: 88
+      param_type: 77
```
mehmoodak: Can we use PHP Docs to define the param types and return types ..... since we are using PHPStan so these comments should help in coverage?

https://phpstan.org/writing-php-code/phpdocs-basics#properties

mehmoodak (commented):
I also think we should deprecate lower versions of PHP and updgrade our composer packages. But how many versions we should deprecate needs a discussion .... maybe @acicovic can comment better here.

acicovic:
I wasn't aware that there is discussion to update the minimum version in core. For me this is good news.

I'm in favor of supporting a newer PHP release as minimum, but I'm wondering if there should be any planning for it (for example the need to release as a new major version). If this is the case, we could use the workaround as mentioned by @GaryJones to harvest the benefits more quickly, and update to the newer version when we feel the time is right.

Having said that, if everyone thinks that updating to 7.2 is a good call and that we should do it ASAP, I have no issue with this.

@chriszarate for opinions/awareness.

chriszarate:
I recommend consulting with VIP WP Engineering team for advice.

phpstan.neon
GaryJones: The idea would be to move away from just documenting them (which, you're right, is also read by PHPStan), but actually let the PHP engine make use of them (and if necessary, complain about them) by moving out of the DocBlocks and into the parsed code as param and return types; PHPStan and other tools read them from there too, as do IDEs.

GaryJones:
> I recommend consulting with VIP WP Engineering team for advice.

I'd say that the Parse.ly Sales/Marketing and Support teams are the ones to ask, since a bump to a higher min PHP version potentially excludes customers stuck on PHP 7.1, and the Support team would need to handle any subsequent tickets (with the `Requires PHP` header, then automated updates would not happen for those customers, so low chance of unexpected breakage).

If only we had some in-plugin opt-in telemetry to find out what PHP versions the users were using...

phpstan.neon
acicovic: @mehmoodak, is this actionable in this PR, and can you demonstrate or explain what you're proposing?
Thanks!

acicovic:
@GaryJones
- I initially had trouble running this due to limited memory. What do you think is a sensible default to add to `--memory-limit`? The value we have in our `composer.json` is `536870912`.
- PHPStan now runs on many more files (101 - this is why it crashed as above I guess), so we're getting `Out of 305 possible param types, only 60 % actually have it. Add more param types to get over 77%`. I guess we should just lower the limit for now?

acicovic:
I've pushed a commit to make this work/pass with the project's current conditions.

composer.json
```diff
@@ -87,7 +88,7 @@
 			"@php ./vendor/bin/phpunit --no-coverage --order-by=random"
 		],
 	   	"static-analysis": [
-			"@php ./vendor/bin/phpstan analyze --memory-limit 536870912"
+			"@php ./vendor/bin/phpstan analyze --memory-limit 1074000000"
```
mehmoodak, line 91: If we don't want any memory limit then we can do following which I think is better

```
@php ./vendor/bin/phpstan analyze --memory-limit=-1
```

mehmoodak (approved):

composer.json
acicovic: I'll leave as-is, but we can revisit this if needed. Good tip by the way!

PR #1270, opened 2022-12-26
Rector: Add rector

GaryJones:
## Description
Add Rector (https://getrector.org/) as a tool to help catch missed opportunities to improve code, particularly when upgrading to newer minimum PHP versions, or choosing to refactor to or from particularly coding styles.

## Motivation and Context
The default configuration here is to apply all of the rules for the set "Up To PHP 7.1". It skips the long array to short array rule, since that conflicts with the WordPress Coding Style that the plugin follows.

Having played with some of the other SetLists (type declaration, early return, dead code, privatization, etc.) then I think there are other rules from those sets which would be useful to add in if desired. I've not included them here.

## How Has This Been Tested?
1. Run `vendor/bin/rector process` with optional path(s) to PHP files. By default, it will use the directory and file paths given in the `rector.php` config file. No changes will happen as the changes have already been made.
2. To demonstrate what would change, add a `Rector\Set\ValueObject\SetList::TYPE_DECLARATION,` to the sets() array in `rector.php`, then run `rector bin process --dry-run` - after a few seconds, you'll see a list of changes.

Showing the end of a run that made changes, and a run of Rector that didn't:
Screenshot 2022-12-26 at 14 43 27


GaryJones:
As with #1269, the tooling requires PHP 7.2 minimum to install. Since this is a development-only package, then a `php` constraint can be added to `require-dev`, and the Workflows adjusted accordingly.

tests/Integration/ScriptsTest.php
```diff
@@ -291,7 +291,7 @@ public function test_wp_parsely_load_js_tracker_filter(): void {
 
 		// Since no secret is provided, the extra fields (inline scripts) on the
 		// loader should not be populated.
-		self::assertEquals( 1, count( $wp_scripts->registered['wp-parsely-loader']->extra ) );
+		self::assertEquals( 1, \is_array( $wp_scripts->registered['wp-parsely-loader']->extra ) || $wp_scripts->registered['wp-parsely-loader']->extra instanceof \Countable ? count( $wp_scripts->registered['wp-parsely-loader']->extra ) : 0 );
```
mehmoodak: I think we should create a function for it because we are using it in multiple places. I think SonarCloud will also complain about this 🤔

rector.php
```diff
@@ -0,0 +1,38 @@
...
+		array(
+			__DIR__ . '/src',
+			__DIR__ . '/tests',
+			__DIR__ . '/views',
+			__DIR__ . '/uninstall.php',
+			__DIR__ . '/wp-parsely.php',
+		)
+	);
+
+	// Define sets of rules.
```
mehmoodak, line 25: Comment seems extra. By the way how about if we change the filename to `rector.config.php` .... think it conveys more meaning 🤔

mehmoodak (approved):

mjangda:
We currently have phpcs, phpcbf, phpstan, and Sonar Cloud all doing some form of static analysis and/or cleanup on the codebase here.

What benefits does Rector give us here that the others don't? Can we replace the others with rector? Or extend the others to do what rector can do? My caution here is adding in too many overlapping code quality-related tools that then need to be maintained.

GaryJones:
Rector provides automated refactoring - which is particularly useful for updating code to comply with minimum or maximum PHP versions. That's not something the other tools can do.

rector.php
GaryJones: The default output from `vendor/bin/rector init` is to create a `rector.php`. Looking at https://getrector.org/documentation/config-configuration, then you can indeed name it what you want, but then you'll need to pass a `--config` option with the new name for every `rector` command. That seems like something that would be forgotten and get frustrating than using the default config file name.

> Comment seems extra.

For someone not overly familiar with Rector or its config file, I'll take whatever guidance I can for now :-)

acicovic:
This PR has been updated and I think we can merge it with a review.

GaryJones:
How is commit [`d2f9a40` (#1270)](https://github.com/Parsely/wp-parsely/pull/1270/commits/d2f9a406bc86cde0e06c2011cdd7eae7ec1ad76d) relevant to adding Rector? Seems like this should be in a seperate PR.

acicovic:
> How is commit [d2f9a40 (#1270)](https://github.com/Parsely/wp-parsely/pull/1270/commits/d2f9a406bc86cde0e06c2011cdd7eae7ec1ad76d) relevant to adding Rector? Seems like this should be in a seperate PR.

It's an adjustment I've made in multiple PRs just to make merging between them easier. I also have it in the type coverage PR for example.

mehmoodak (approved):
Changes looks good but my local machine is crashing when I run `vendor/bin/rector process --dry-run`.

Still approving the PR as it seems like Alex is able to run this command fine.

Issue #265, opened 2021-04-30
Add Support to Trigger Article Crawls & Hook into Appropriate Actions


jblz:
There's an [API endpoint](https://www.parse.ly/help/integration/trigger-crawl) to instruct the service to re-crawl an Article:

> Normally, articles are crawled during the first 24 hours after they are published when the metadata is most likely to change. It is possible to notify our servers when posts get updated. To do that, submit a POST request to the following address, appending the URL of the post to the end of it. The <API_KEY> should be replaced with your respective Site ID (apikey), and <API_SECRET> with the token accessible to the account administrators via API Settings.

`https://dash.parsely.com/<API_KEY>/ping_crawl?secret=<API_SECRET>&url=<URL>`

If an article is changed after that period, it's not apparent how long it would be before it gets recrawled.

WordPress does not have built-in Webhook support (though there exist plugins to add support).  However, it does have built in support for running a function when various actions occur. Should we consider adding such support?

A potential (untested, somewhat pseudo-coded) approach to this might look like:

```
function wp_parsely_trigger_page_crawl( $post_id ) {
  if ( $POST_IS_NEWER_THAN_THRESHOLD ) {
     return;
  }
  $permalink = get_post_permalink( $post );
  wp_remote_post( get_ping_crawl_url( $permalink ), [ 'blocking' => false ] );
}
add_action( 'post_updated', 'wp_parsely_trigger_page_crawl' );
```

We should probably add a filter that lets users at least opt out of this behavior (if not opt in)

mehmoodak:
Find more info in #1073

PR #1271, opened 2023-01-02
Content Helper: Add Parse.ly Stats List Column

### LINKED ISSUE #1215: Include Parse.ly data on the Posts page
One of the most important pages visited by content creators in a CMS is the Posts page. This is where they see all the posts being published, edited, etc. 

By adding a column to the Posts section/page in VIP/wp-admin, we can provide quick context on published posts from the primary area in the CMS where creators review, search for, sort, and create posts.

As a content creator, I want to see how content on the site has been performing so I can make better decisions about what to write about, what articles to revisit and improve, and what to deprioritize in our editorial strategy. I want to see this on the main page that I access to look at all posts on my site. 

**Screen options**
Add a column option called Parse.ly Content Helper. When the column is selected and applied to the page, the below metrics will display in the table view of posts. 

**Parse.ly metrics to include in the column**
Screenshot 2022-11-29 at 5 21 17 PM

- page views
- visitors
- avg. time

mehmoodak:
## Description
Shows Parsely Stats on List Screen of Posts, Pages and Custom Post Types

Closes #1215, #1318

## How Has This Been Done?
Here is the brief summary of the approach:

  - Adding Parse.ly Stats column on list post and pages screens.
  - Enqueued styles for Parsely Stats column.
  - While displaying content on added Parse.ly Stats column we show the placeholder and store the publish date (needed to limit the API).
  - Before enqueuing script on footer we call Parse.ly Analytics API using the publish params stored above.
  - While enqueuing script pass the recieved data from API so that we can show it on frontend.
  
 **Notes:**
 - Ideally we should have avoided the Frontend script approach but we have chosen this approach due to the limitation of hooks on WordPress side i.e. for posts we have `the_posts` hook which we could have utlised for our purpose but for pages `the_posts` hook don't get called and we also don't have any alternative.

## How Has This Been Tested?
- Added Unit and Integration Tests for the new functionality.

- Sorting is not implemented (discussed with product and we are fine for now).
- We can't show Parse.ly Stats on post types which are marked as `Non-Post` in Parse.ly Settings because Parse.ly API don't include Non-Post data in `analytics` endpoint.
- Parse.ly API response don't have `POST_ID` so we have come up with the custom approach for identifying the post and as of now we take the `url` path coming from Analytics API with constraint that it will only show Parse.ly Stats if users Permalink Settings matches with the site on production.

## Screenshots

1. Posts screen with data from Parse.ly API

image

2. Posts screen with error from Parse.ly API

image


src/UI/class-admin-columns-analytics.php
```diff
@@ -0,0 +1,248 @@
...
+	 * @param Analytics_Post $analytics_post Post analytics obj returned from Parse.ly API.
+	 *
+	 * @return string
+	 */
+	private function get_unique_stats_key_from_analytics( $analytics_post ): string {
+		if ( ! isset( $analytics_post['title'] ) || ! isset( $analytics_post['pub_date'] ) ) {
+			return '';
+		}
+
+		return $analytics_post['title'] . '-' . $analytics_post['pub_date'];
```
mehmoodak: There is nothing unique in Parsely API so I have come up with this formula which is not 100% correct. Please do let me know if you have any better suggestion.

Due to this I have also moved #265 to current milestone to decrease the chances of mismatch between parsely api response and WP posts.

mehmoodak (commented):
Added comments.

src/RemoteAPI/class-analytics-posts-api.php
```diff
@@ -11,13 +11,68 @@
...
+	 * @param Analytics_Post_API_Params $api_params Parameters of the API.
+	 *
+	 * @return WP_Error|Analytics_Post[]
+	 */
+	public function get_post_analytics( $api_params ) {
+		$response = $this->get_items( $api_params, true );
+
+		if ( is_wp_error( $response ) ) {
+			return $response;
+		}
```
acicovic: Is this needed here?

src/Utils/utils.php
```diff
@@ -129,3 +133,29 @@ function get_time_format(): string {
...
+ *
+ * @param mixed $obj Input object.
+ *
+ * @return array<string, mixed>|WP_Error
+ */
+function convert_to_associate_array( $obj ) {
+	$encoded = wp_json_encode( $obj );
+
+	if ( false === $encoded ) {
+		return new WP_Error( 400, __( 'Unable to encode API response for associative array', 'wp-parsely' ) );
```
acicovic: Consider using another error code.
https://developer.wordpress.org/reference/classes/wp_error/#error-codes

src/UI/class-admin-columns-analytics.php
```diff
@@ -0,0 +1,248 @@
+<?php
+/**
+ * UI: Class for adding `Parse.ly Stats` on admin columns.
```
acicovic:
```suggestion
 * UI: Class for adding `Parse.ly Stats` on admin columns
```
File headers do not end with a period per commenting standards.
https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#6-file-headers

src/UI/class-admin-columns-analytics.php
```diff
@@ -0,0 +1,248 @@
...
+		if ( $this->parsely->site_id_is_set() && $this->parsely->api_secret_is_set() ) {
+			add_filter( 'the_posts', array( $this, 'set_parsely_stats' ) );
+			// TODO: Add columns for pages and custom post types.
+			add_filter( 'manage_posts_columns', array( $this, 'add_parsely_stats_column_on_list_view' ) );
+			add_action( 'manage_posts_custom_column', array( $this, 'show_parsely_stats' ) );
+		}
+	}
+
+	/**
+	 * Calculate Min and Max Publish date from all the found posts.
```
acicovic:
```suggestion
	 * Calculates Min and Max Publish date from all the found posts.
```

src/UI/class-admin-columns-analytics.php
```diff
@@ -0,0 +1,248 @@
...
+
+	/**
+	 * Adds the `Parse.ly Stats` on admin columns.
+	 *
+	 * @param array<string, string> $columns Columns array which contain keys and labels.
+	 *
+	 * @return array<string, string>
+	 */
+	public function add_parsely_stats_column_on_list_view( array $columns ): array {
+		$columns['parsely-stats'] = 'Parse.ly Stats';
```
acicovic: I think that this column is to be named "Parse.ly Content Helper" or "Content Helper". You can verify this with @danielabloch.

src/UI/class-admin-columns-analytics.php
```diff
@@ -0,0 +1,248 @@
...
+use DateTime;
+use WP_Post;
+use Parsely\Parsely;
+use Parsely\RemoteAPI\Analytics_Posts_API;
+
+use const Parsely\Utils\WP_MAX_POSTS_PER_PAGE;
+use const Parsely\Utils\DATE_TIME_UTC_FORMAT;
+
+/**
+ * Shows `Parse.ly Stats` on Admin Columns
```
acicovic:
```suggestion
 * Shows `Parse.ly Stats` on Admin Columns.
```

acicovic (commented):
Left a preliminary review, mostly minor things.

src/RemoteAPI/class-analytics-posts-api.php
mehmoodak: Yes handling this in my future commits to show admin notice in case of any error while getting data from Parse.ly Analytics API.

src/Utils/utils.php
mehmoodak: I think 400 is the right error code in this scenario .... please do let me know if you have any suggestions?

src/UI/class-admin-columns-analytics.php
mehmoodak: Picked this from Figma designs and the name looks good to me, will take feedback from design or product on this.

webpack.config.js
```diff
@@ -5,9 +5,12 @@ module.exports = {
 	...defaultConfig,
 	entry: {
 		...defaultConfig.entry(),
+		'admin-parsely-stats': [
+			path.resolve( __dirname, 'src', 'css', 'admin-parsely-stats.scss' ),
+		],
 		'admin-settings': [
 			path.resolve( __dirname, 'src', 'js', 'admin-settings' ),
-			path.resolve( __dirname, 'src', 'css', 'admin-settings.css' ),
+			path.resolve( __dirname, 'src', 'css', 'admin-settings.scss' ),
```
mehmoodak, line 14: Changing extension so that we can benefit from SCSS.

src/blocks/content-helper/content-helper.scss
```diff
@@ -1,6 +1,6 @@
 @import "variables";
-@import "../shared/variables";
-@import "../shared/functions";
+@import "../../css/shared/functions";
+@import "../../css/shared/variables";
```
mehmoodak, line 1: Moved shared files from inside `blocks` folder to more generic place.

mehmoodak (commented):
Added more comments.

src/RemoteAPI/class-remote-api-base.php
```diff
@@ -80,11 +82,14 @@ public function get_api_url( array $query ): string {
...
 	 * @since 3.2.0
+	 * @since 3.7.0 Added $associative param.
 	 *
 	 * @param array<string, mixed> $query The query arguments to send to the remote API.
+	 * @param bool                 $associative When TRUE, returned objects will be converted into associative arrays.
+	 *
 	 * @return WP_Error|array<string, mixed>
 	 */
-	public function get_items( array $query ) {
+	public function get_items( $query, $associative = false ) {
```
chriszarate, line 98: I'm not seeing the value of this new arg. Is conversion to associative array truly needed? Even if it is, can this method be left simple and that conversion done in the spot where it is needed?

src/UI/class-admin-columns-analytics.php
```diff
@@ -0,0 +1,322 @@
...
+	 * @return bool
+	 */
+	private function is_post_list_screen(): bool {
+		if ( ! isset( $_SERVER['REQUEST_URI'] ) ) {
+			return false;
+		}
+
+		$page_url = esc_url_raw( $_SERVER['REQUEST_URI'] );
+
+		return '/wp-admin/edit.php' === $page_url || strpos( $page_url, 'post_type=post' ) !== false;
```
chriszarate: You can't depend on this hardcoded URL check. WordPress can be installed to any path. Use `get_current_screen`

src/Utils/utils.php
```diff
@@ -12,7 +12,32 @@
...
+ */
+function get_utc_date( int $days = 0 ): string {
+	if ( 0 === $days ) {
+		$utc_date = gmdate( DATE_UTC_FORMAT );
+	} else {
+		$utc_date = gmdate( DATE_UTC_FORMAT, (int) strtotime( "{$days} days" ) );
+	}
+
+	return $utc_date;
+}
```
chriszarate: 1. This function name could be improved to better describe what the function does
2. Early returns can make this easier to read:

```suggestion
function get_utc_date( int $days = 0 ): string {
	if ( 0 === $days ) {
		return gmdate( DATE_UTC_FORMAT );
	}

	return gmdate( DATE_UTC_FORMAT, (int) strtotime( "{$days} days" ) );
}
```

src/Utils/utils.php
chriszarate: The possibility of this error (e.g., as a result of object references) should encourage us to avoid this conversion

src/RemoteAPI/class-remote-api-base.php
mehmoodak: Actually on PHPStan I have looked on how to give types to `stdClass` but found nothing so therefore now I tried to turn `stdClass` to `array` and then use proper typing for it.

Code wise we don't need this but for strong typing we need this so I chose the second one.

src/UI/class-admin-columns-analytics.php
mehmoodak: I tried using `get_current_screen` but it returns `null` ..... still experimenting with this.

src/Utils/utils.php
mehmoodak: @chriszarate apologies didn't get it .... can you please explain a bit more 🙂

src/RemoteAPI/class-analytics-posts-api.php
mehmoodak: Misunderstood before. Not needed, will remove it.

src/UI/class-settings-page.php
```diff
@@ -941,10 +941,9 @@ public function get_tracking_values_for_display(): array {
 
 		foreach ( $types as $type ) {
 			$array_value = $options[ "track_{$type}_types" ];
-			if ( is_array( $array_value ) ) {
```
mehmoodak, line 944: PHPStan suggestion since `$array_value` will always be array here.

tests/Integration/ScriptsTest.php
```diff
@@ -476,76 +477,11 @@ public function test_tracker_markup_has_attribute_when_cfasync_filter_is_used():
 		wp_print_scripts();
 		$output = ob_get_clean();
 
+		$loader_asset = require_once plugin_dir_path( PARSELY_FILE ) . 'build/loader.asset.php';
+
 		// phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript
-		self::assertStringContainsString( "<script data-cfasync=\"false\" type='text/javascript' src='http://example.org/wp-content/plugins/wp-parsely/tests/Integration/../../build/loader.js?ver=123456.78.9' id='wp-parsely-loader-js'></script>", $output );
+		self::assertStringContainsString( "<script data-cfasync=\"false\" type='text/javascript' src='http://example.org/wp-content/plugins/wp-parsely/tests/Integration/../../build/loader.js?ver=" . is_bool( $loader_asset ) ? Parsely::VERSION : $loader_asset['version'] . "' id='wp-parsely-loader-js'></script>", $output );
```
mehmoodak, line 483: This test was failing a lot recently so fixed it.

tests/Integration/ScriptsTest.php
```diff
@@ -476,76 +477,11 @@ public function test_tracker_markup_has_attribute_when_cfasync_filter_is_used():
...
 		// phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript
 		self::assertStringContainsString( "<script data-cfasync=\"false\" type='text/javascript' data-parsely-site=\"blog.parsely.com\" src='https://cdn.parsely.com/keys/blog.parsely.com/p.js?ver=123456.78.9' id=\"parsely-cfg\"></script>", $output );
 	}
-
```
mehmoodak, line 484: Moved to `TestCase` because now we are not only using it in `class ScriptTest`

tests/Integration/TestCase.php
```diff
@@ -168,6 +177,81 @@ public function create_test_post( string $post_status = 'publish' ): int {
...
+	 *
+	 * @param int[] $post_ids IDs of the posts.
+	 *
+	 * @return WP_Post[]
+	 */
+	private function get_test_posts( $post_ids = array() ) {
+		$posts = array();
+
+		foreach ( $post_ids as $post_id ) {
+			array_push( $posts, get_post( $post_id ) );
```
mehmoodak, line 244: Was not able to get draft posts using `get_posts` function so chose this approach.

tests/Integration/TestCase.php
```diff
@@ -193,4 +277,250 @@ public function go_to_new_post( string $post_status = 'publish' ): int {
...
+	/**
+	 * Assert availability of hooks.
+	 *
+	 * @param string[] $hooks WordPress hooks whose availability we have to verify.
+	 * @param bool     $availability_type TRUE if we want to check the presence of given hooks.
+	 *
+	 * @return void
+	 */
+	public function assert_wp_hooks_availablility( $hooks, $availability_type ) {
+		if ( ! $this->isPHPVersion7Dot2OrHigher() ) {
```
mehmoodak, line 286: Due to some reason `has_action` isn't working PHP 7.1 so using this function in codebase to avoid the checks.

I could have used just one approach i.e. calling the function directly instead of using `do_action` but chose this approach because `do_action` is more close to how WordPress works i.e. until we support PHP 7.1 we will use both approaches to write better tests and to make PHP 7.1 happy.

tests/Integration/TestCase.php
```diff
@@ -193,4 +277,250 @@ public function go_to_new_post( string $post_status = 'publish' ): int {
...
+			);
+		}
+	}
+
+	/**
+	 * Asserts that a passed style is not registered.
+	 *
+	 * @param string $handle Style handle to test.
+	 */
+	public function assert_is_style_not_registered( string $handle ): void {
```
mehmoodak, line 397: Same as above script code .... the only difference is it verifies the status of styles instead of script.

tests/Unit/Utils/UtilsTest.php
```diff
@@ -0,0 +1,244 @@
...
+				'msg'             => 'Should match with the format of the future date.',
+			),
+		);
+
+		foreach ( $tests_data as $t ) {
+			$args   = $t['args'];
+			$output = get_utc_date_format( $args['days'] );
+
+			self::assertEquals( $t['expected_output'], $output, $t['msg'] );
+		}
```
mehmoodak, line 95: Instead of using multiple asserts using loop to verify different arguments (FYI: This pattern is very common in GoLang)

mehmoodak (commented):

src/Utils/utils.php
mehmoodak: Doing something similar [here](https://github.com/Parsely/wp-parsely/blob/develop/src/RemoteAPI/class-remote-api-base.php#L97) so returning this error should be fine 🤔

wp-parsely.php
```diff
@@ -116,17 +118,11 @@ function parsely_initialize_plugin(): void {
...
-
-	$plugins_actions = new Plugins_Actions();
-	$plugins_actions->run();
-
-	$row_actions = new Row_Actions( $parsely );
-	$row_actions->run();
-
-	$site_health = new Site_Health( $parsely );
-	$site_health->run();
+	( new Admin_Warning( $parsely ) )->run();
```
acicovic, line 121: Generally, as the specific refactoring is not related to this PR's goal, it would be better to do this in another dedicated refactorings PR. This keeps context, separation of concerns, and can make PRs smaller and easier to review.

Having said that, let's keep these changes here.

src/UI/class-admin-columns-parsely-stats.php
```diff
@@ -0,0 +1,370 @@
...
+	public function __construct( Parsely $parsely ) {
+		$this->parsely = $parsely;
+	}
+
+	/**
+	 * Registers action and filter hook callbacks.
+	 *
+	 * @since 3.7.0
+	 *
+	 * @return void
```
acicovic: Let's remove `return void` comments, unless they are strictly needed by PHPStan.

https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#1-functions-class-methods

> return void should not be used outside of the default bundled themes.

src/UI/class-admin-columns-parsely-stats.php
```diff
@@ -0,0 +1,370 @@
...
+	 */
+	public function run(): void {
+		if ( ! $this->parsely->site_id_is_set() || ! $this->parsely->api_secret_is_set() ) {
+			return;
```
acicovic, line 87: If the Site ID or API Secret is not set, do we currently provide any kind of error message or warning about it?

src/UI/class-admin-columns-parsely-stats.php
```diff
@@ -0,0 +1,370 @@
...
+		}
+
+		$admin_settings_asset = require_once plugin_dir_path( PARSELY_FILE ) . 'build/admin-parsely-stats.asset.php';
+		$built_assets_url     = plugin_dir_url( PARSELY_FILE ) . 'build/';
+
+		wp_enqueue_style(
+			'admin-parsely-stats-styles',
+			$built_assets_url . 'admin-parsely-stats.css',
+			$admin_settings_asset['dependencies'] ?? null,
```
acicovic, line 124: Some times, this does not work well for enqueuing CSS files, so in these cases I add `array()` here. This comment is more of a FYI and no need to change this if it works.

src/UI/class-admin-columns-parsely-stats.php
```diff
@@ -0,0 +1,370 @@
...
+	/**
+	 * Add `Parse.ly Stats` column on admin columns.
+	 *
+	 * @param array<string, string> $columns Columns array which contain keys and labels.
+	 *
+	 * @return array<string, string>
+	 */
+	public function add_parsely_stats_column_on_list_view( array $columns ): array {
+		if ( $this->is_tracked_as_post_type() ) {
+			$columns['parsely-stats'] = 'Parse.ly Stats';
```
acicovic: Is this a title? in this case it should be translatable. Also, I would recommend checking with @danielabloch what's the final title we want to have for the column.

src/UI/class-admin-columns-parsely-stats.php
```diff
@@ -0,0 +1,370 @@
...
+
+		wp_enqueue_script(
+			'admin-parsely-stats-script',
+			$built_assets_url . 'admin-parsely-stats.js',
+			$admin_settings_asset['dependencies'] ?? null,
+			$admin_settings_asset['version'] ?? Parsely::VERSION,
+			true
+		);
+
+		wp_localize_script(
```
acicovic: https://developer.wordpress.org/reference/functions/wp_localize_script/

>`wp_localize_script()` is often used to pass generic data from PHP to JavaScript, because it was originally the only official way to do that.

> [wp_add_inline_script()](https://developer.wordpress.org/reference/functions/wp_add_inline_script/) was [introduced in WP v4.5](https://make.wordpress.org/core/2016/03/08/enhanced-script-loader-in-wordpress-4-5/), and is now the best practice for that use case. `wp_localize_script()` should only be used when you actually want to localize strings.

src/UI/class-admin-columns-parsely-stats.php
```diff
@@ -0,0 +1,370 @@
...
+			'admin-parsely-stats-script',
+			'wpParselyPostsStatsResponse',
+			$parsely_stats_response
+		);
+	}
+
+	/**
+	 * Call Parsely Analytics API and get stats data.
```
acicovic:
```suggestion
	 * Call Parse.ly Analytics API and get stats data.
```

src/UI/class-admin-columns-parsely-stats.php
```diff
@@ -0,0 +1,370 @@
...
+		}
+
+		$response = $analytics_api->get_posts_analytics(
+			array(
+				'period_start'   => get_utc_date_format( - Analytics_Posts_API::ANALYTICS_API_DAYS_LIMIT ),
+				'period_end'     => get_utc_date_format(),
+				'pub_date_start' => $date_params['pub_date_start'] ?? '',
+				'pub_date_end'   => $date_params['pub_date_end'] ?? '',
+				'limit'          => Analytics_Posts_API::MAX_RECORDS_LIMIT,
+				'sort'           => 'avg_engaged', // Note: API sends different stats on different sort options.
```
acicovic, line 236: FYI @abelsonlive, we have found that the Parse.ly API returns a different set of metrics depending on the value of the `sort` parameter that is being passed. This is currently not a blocker for this PR, but something noteworthy for us when we were experimenting with the API. We can exchange more information if needed.

src/UI/class-admin-columns-parsely-stats.php
```diff
@@ -0,0 +1,370 @@
...
+		add_action( 'admin_enqueue_scripts', array( $this, 'enqueue_parsely_stats_styles' ) );
+		add_filter( 'manage_posts_columns', array( $this, 'add_parsely_stats_column_on_list_view' ) );
+		add_action( 'manage_posts_custom_column', array( $this, 'update_published_times_and_show_placeholder' ) );
+		add_filter( 'manage_pages_columns', array( $this, 'add_parsely_stats_column_on_list_view' ) );
+		add_action( 'manage_pages_custom_column', array( $this, 'update_published_times_and_show_placeholder' ) );
+		add_action( 'admin_footer', array( $this, 'enqueue_parsely_stats_script_with_data' ) );
+	}
+
+	/**
+	 * Set current screen property.
```
acicovic: Function summaries should be in third person.

> A function, hook, class, or method is a third-person singular element, meaning that third-person singular verbs should be used to describe what each does.

Please review https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#language for examples/guidelines, and update any function summaries needed.

src/UI/class-admin-columns-parsely-stats.php
```diff
@@ -0,0 +1,370 @@
...
+		return array(
+			'pub_date_start' => ( new DateTime( min( $published_times ) ) )->format( DATE_UTC_FORMAT ),
+			'pub_date_end'   => ( new DateTime( max( $published_times ) ) )->format( DATE_UTC_FORMAT ),
+		);
+	}
+
+	/**
+	 * Get unique key which we can use for Parse.ly stats map.
+	 *
+	 * Mainly we need this because Parse.ly Analytics API doesn't have anything unique through which we can
```
acicovic:
```suggestion
	 * Needed because Parse.ly Analytics API doesn't have anything unique through which we can
```

src/UI/class-admin-columns-parsely-stats.php
```diff
@@ -0,0 +1,370 @@
...
+	 * @since 3.7.0
+	 *
+	 * @return bool
+	 */
+	private function is_tracked_as_post_type(): bool {
+		if ( is_null( $this->current_screen ) ) {
+			return false;
+		}
+
+		// Analytics API don't include Non-Post data so we can only show stats on tracked post types.
```
acicovic:
```suggestion
		// Analytics API doesn't include Non-Post data, so we can only show stats on tracked post types.
```

src/Utils/utils.php
```diff
@@ -12,7 +12,29 @@
 
 namespace Parsely\Utils;
 
+use NumberFormatter;
 use WP_Post;
+use WP_Error;
+
+const DATE_UTC_FORMAT     = 'Y-m-d';
+const WP_DATE_TIME_FORMAT = 'Y-m-d H:i:s';
```
acicovic, line 20: Should any WP dates be locale-aware?

src/Utils/utils.php
```diff
@@ -129,3 +151,73 @@ function get_time_format(): string {
...
+	}
+
+	return $formatted_number;
+}
+
+/**
+ * Get time in formatted form.
+ *
+ * Example:
+ *   - Input `1000` (seconds) and Output `16:40` which represent "16 minutes, 40 seconds”
```
```suggestion
 *   - Input `1000` (seconds) and Output `16:40` which represents "16 minutes, 40 seconds”
```

src/Utils/utils.php
acicovic: @mehmoodak, regarding my comment about the error code, the docs linked state:

> Error codes are slugs that are used to identify each error. They are mostly useful when a piece of code can produce several different errors, and you want to handle each of those errors differently.

> The error codes used in WordPress are not integers, but strings, with any spaces between words replaced with underscores (example: an_error_code). The error codes used in WordPress are usually based on the error message associated with that code.

We still have some places with number error codes in the codebase, but we should avoid this in newly added code. String error codes are what allow us to display error hints in the Content Helper for example, giving us much more specific context than a generic HTTP error number.

About @chriszarate's comment, I'll let him elaborate himself if he sees a need for it.

src/js/admin-parsely-stats.ts
```diff
@@ -0,0 +1,93 @@
...
+ * Replace Parse.ly Stats placeholder from default to differentiate while the API request
+ * is in progress or completed.
+ */
+function updateParselyStatsPlaceholder(): void {
+	getAllPostStatsElements()?.forEach( ( statsElement: Element ): void => {
+		statsElement.innerHTML = '—';
+	} );
+}
+
+function showParselyStats( parselyStatsMap: ParselyStatsMap ): void {
```
acicovic, line 58: Noting that some function comments are missing in this file.

src/js/admin-parsely-stats.ts
```diff
@@ -0,0 +1,93 @@
...
+		}
+
+		if ( stats.avg_time ) {
+			statsElement.innerHTML += `<span class="parsely-post-avg-time">${ stats.avg_time }</span><br/>`;
+		}
+	} );
+}
+
+function showParselyStatsError( parselyStatsError: ParselyAPIErrorInfo ): void {
+	const headerEndElement = document.querySelector( '.wp-header-end' ); // WP have this element before admin notices.
```
acicovic:
```suggestion
	const headerEndElement = document.querySelector( '.wp-header-end' ); // WP has this element before admin notices.
```

acicovic (commented):
I've not reviewed tests yet, but I'm posting this as a first set of comments/questions.

src/Utils/utils.php
mehmoodak: Okay thanks for clarification ...... replacing `400` with `parsely_encoding_failed`

src/UI/class-admin-columns-parsely-stats.php
mehmoodak: Not strictly needed so removing these.

src/UI/class-admin-columns-parsely-stats.php
mehmoodak: For `Site ID` there is already a default message as follows

```
The Parse.ly plugin is not active. You need to [provide your Parse.ly Dash Site ID](http://parsely.test/wp-admin/options-general.php?page=parsely) before things get cooking.
```

For `API Secret` there is nothing ...... as far as I understand not all users have API Secret set in the options so if we show error on this then it will show errors for those users everytime.

FYI: If I remove `api_secret` condition then our error parsely stats error message does show `Forbidden` error.

Let me now what approach we should take here. Thanks

src/UI/class-admin-columns-parsely-stats.php
mehmoodak: Thanks for letting me know, for now things are working so making no change.

src/UI/class-admin-columns-parsely-stats.php
mehmoodak: Making this translatable .... this has been taken from Figma designs so as far as I know it is final but let's wait for @danielabloch also to be sure.

src/Utils/utils.php
mehmoodak: We are only using this in Tests now so sees like we don't need it to be locale-aware.

mehmoodak:
Implemented the feedback, please have a look again. Thanks

src/UI/class-admin-columns-parsely-stats.php
danielabloch: `Parse.ly Stats` is good. We'd all discussed Context initially but Parse.ly Stats is better because it is clearer and includes the name of the data source (Parse.ly). 

Thanks to you both.

src/UI/class-admin-columns-parsely-stats.php
acicovic: Does it look broken when the forbidden error is shown?

If not, let's show the error. It is better to signal to the user that something is wrong, and we can also add a note about this in the documentation.

src/UI/class-admin-columns-parsely-stats.php
```diff
@@ -0,0 +1,361 @@
...
+	/**
+	 * Sets current screen property.
+	 *
+	 * @since 3.7.0
+	 */
+	public function set_current_screen(): void {
+		$this->current_screen = get_current_screen();
+	}
+
+	/**
```
acicovic, line 108: It seems we're missing `since` in some function comments.

src/Utils/utils.php
```diff
@@ -12,7 +12,29 @@
...
 
+use NumberFormatter;
 use WP_Post;
+use WP_Error;
+
+const DATE_UTC_FORMAT     = 'Y-m-d';
+const WP_DATE_TIME_FORMAT = 'Y-m-d H:i:s';
+
+/**
+ * Get UTC Date.
```
acicovic: We still have function summaries that need to be converted to third person.

tests/Integration/TestCase.php
```diff
@@ -168,6 +177,81 @@ public function create_test_post( string $post_status = 'publish' ): int {
 		return self::factory()->post->create( $post_data );
 	}
 
+	/**
+	 * Create test posts in sequence.
```
acicovic: We still have function summaries that need to be converted to third person.

tests/Integration/TestCase.php
```diff
@@ -187,10 +271,232 @@ public function go_to_new_post( string $post_status = 'publish' ): int {
...
+		foreach ( $assert_false as $status ) {
+			self::assertFalse(
+				wp_style_is( $handle, $status ),
+				"Unexpected style status: $handle status should NOT be '$status'"
+			);
+		}
+	}
+
+	/**
+	 * Return TRUE if minimum PHP version is 7.2 or higher. We uses this if something works
```
acicovic: We still have function summaries that need to be converted to third person. Please check other places as well, as I there seem to be occurrences in multiple places.

tests/Integration/TestCase.php
acicovic: Maybe we would like to add a PHP comment about this, in order to not lose the context in the future?

src/UI/class-admin-columns-parsely-stats.php
mehmoodak: No it doesn't looks broken. Is it okay to show this error to all users who are using Parse.ly and doesn't have API Secret because of additional pricing? .... skipped showing error because of this case.

image

tests/Integration/TestCase.php
mehmoodak: I think the needed context is already there in PHPDoc of `isPHPVersion7Dot2OrHigher` function.

mehmoodak (commented):
Feedback Implemented. Please have a look again. Thanks.

tests/Integration/TestCase.php
acicovic: Yes, the function name is self-explanatory. I was referring more to the "why we need to do this". But feel free to mark this as resolved without any changes, it is not an issue.

src/UI/class-admin-columns-parsely-stats.php
acicovic: Will the error we see in the screenshot be displayed even if the user hides the column?

mehmoodak: In current implementation .. Yes.

By default this column will appear for our customers so will it be good if they have to hide the column due to this error ..... or we can just keep the current implementation to silently hide it if key isn't set?

acicovic: If this is possible, here is how I would approach this:

- If the column is shown, show the error when no API Secret is set.
- If the column is not shown, do not show the error.

The point is that the error concerns the column, so if the user chooses to hide it, there is no point in displaying the error. However, if the user wants the column, they must be notified that something is wrong. Else they could perceive the lack of data as a bug.

I'd also make the error message more helpful if possible, as we do in the Content Helper sidebar.

acicovic:
Apart the open discussion we have regarding the error message, this looks ready to me.

src/UI/class-admin-columns-parsely-stats.php
mehmoodak: The recommended approach is possible but is it fine to force all users to disable the column if they are not using the API Secret because by default we are showing the Parse.ly Stats column.

acicovic: There was actually a discussion to hide by default all Content Helper features when an API Secret is not set (this would be a different PR of course). I do not know how much this discussion has progressed and if there was a decision about it.

@danielabloch, can you read this thread and let us know of your thoughts? We're happy to answer questions if something is unclear.

acicovic: If we activate the column by default, we probably do not want to show the error message, because it looks like we're activating a broken feature. @mehmoodak, I guess that this is why you had it hidden.

mehmoodak: Yes that is the reason.

If we go with `No Default` then it is fine to show the error but it doesn't seems like a good idea to make this feature hidden by default.

danielabloch: Reviewing the questions...

`Is it okay to show this error to all users who are using Parse.ly and doesn't have API Secret because of additional pricing?` 
Yes, it is okay to show this error. if we are unable to show Parse.ly data to a customer, we should display an error message that explains the issue and how to resolve it.

> **Proposed error message:**
> We are unable to retrieve data for Parse.ly Stats. Please contact support@parsely.com for help resolving this issue. 

> Previous error message: 
> Unable to retrieve data for Parse.ly Stats.
> Detail: (403) Forbidden

`Will the error we see in the screenshot be displayed even if the user hides the column?`
Do not show the error message if the user hid the column. I agree that 'if the user chooses to hide it [the column], there is no point in displaying the error. However, if the user wants the column, they must be notified that something is wrong. Else they could perceive the lack of data as a bug.'

`Is it fine to force all users to disable the column if they are not using the API Secret because by default we are showing the Parse.ly Stats column.`
We do not want to force users to disable the column. If they don't want to see the column, they have the option to hide it. Otherwise, we want them to know it is available. And, if no data displays in that column, we want to help them find out how to get data to display in the Parse.ly Stats column. 

`There was actually a discussion to hide by default all Content Helper features when an API Secret is not set (this would be a different PR of course). I do not know how much this discussion has progressed and if there was a decision about it.`
There is no decision on this, and if we need to make changes we can do so in a separate PR.

mehmoodak: > Else they could perceive the lack of data as a bug.

There will be no lack of data because as of now we are hiding the entire column so it should not perceive as bug.

Current Approach is as follows:
- Do not show the Parse.ly Stats column if Site ID and API Secret aren't set.
  - If Site ID isn't set then there is already a message in place which shows the error.
  - If there is no API Secret set then we shows no error and no Parse.ly Stats column i.e. silently hide the column (just in place to avoid any compulsory user action)

Recommended Approach is as follows:
- Do not show the Parse.ly Stats column if Site ID isn't set.
  - If Site ID isn't set then there is already a message in place which shows the error.
- For customers that doesn't have API Secret it will show the error and then they have to contact with support or either hide the column.

I think current approach is better because on recommended approach user still have to take some action. Thoughts?

acicovic: I think that the "better approach" boils down to whether we want to show the column by default, even if no Site ID or API Secret is set. So far this is what we have been doing. For example, we currently show the CHS (Content Helper Sidebar) even when no Site ID is set, and even though there is no data to offer (we display a friendly "Contact Us" message instead).

So by auto-hiding the column, we break the previous convention, but we do save the user from the manual step of hiding the column if they do not want it. However, a side-effect of auto-hiding is that many users might totally miss the existence of the column and never know about it (unless they consult the docs or get informed about it in some other way).

This is not really an engineering problem, but rather what's the strategy/approach we use for our CH features. So I'll leave the final decision to @danielabloch about this.

mehmoodak: As of now the feature doesn't seem broken so for now I will recommend to merge this PR and handle any additional feedback on a separate PR which will make the review simpler.

Apart from this we can also get additional feedback from the team about current implementation.

acicovic (approved):
Approving this. We will be addressing any last feedback as separate PR(s).

src/UI/class-admin-columns-parsely-stats.php
danielabloch: Makes sense to me to move additional feedback on a separate PR. That said, I'll weigh in one more time.

`So by auto-hiding the column, we break the previous convention, but we do save the user from the manual step of hiding the column if they do not want it. However, a side-effect of auto-hiding is that many users might totally miss the existence of the column and never know about it (unless they consult the docs or get informed about it in some other way).`

One of the main objectives with this work is showing people what is possible with Parse.ly data in WordPress. 

- If we auto-hide the column, then people don't know it exists. This goes against our goal to show people what is possible with Parse.ly. 
- We want people to see the column and then decide if they want to use it. In this instance, it is desirable to have people take an action.

mehmoodak: Column will be visible by default if API Secret is set and as a part of this project I think we will create API Secret for the customers.

With that being said @danielabloch please confirm the final flow. Is it same as you have described above or are there any changes?

danielabloch: If API secret is set
- column is visible

If API secret is not set
- column is visible, with messaging on how to activate parsely (contact us)
- this is an edge case for VIP customers
- this is a common case for non-VIP customers

mehmoodak: Ticket: https://github.com/Parsely/wp-parsely/issues/1330

PR #1272, opened 2023-01-02
Improve Remote APIs naming

mehmoodak:
## Description
In codebase we are calling Parse.ly API from two places:
1. From `RemoteAPI` folder which gets data by calling Parsely API endpoints.
2. From `Endpoints` folder which calls the WP REST API endpoints created by us which under the hood calls `RemoteAPIs`. These endpoints serve as proxy for us.

But the naming issue is that we have also used the word `Proxy` in all of the naming related to `RemoteAPI` which is technically not right so therefore this PR improves the naming which eventually makes it easy to understand the code by just reading.

## How Has This Been Tested?
- Ran existing tests to make sure that everything is working fine.


src/Endpoints/class-base-api-proxy.php
```diff
@@ -148,11 +147,7 @@ protected function get_data( WP_REST_Request $request, bool $require_api_secret
 		}
 
 		// A proxy with caching behavior is used here.
-		$response = $this->proxy->get_items( $params ); // @phpstan-ignore-line.
-
-		if ( false === $response ) {
-			return new stdClass();
-		}
```
mehmoodak, line 155: This was added while incorporating PHPStan but now after giving correct type we don't need this.

src/RemoteAPI/class-analytics-post-detail-api.php
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Class for Analytics Post Detail API (`/analytics/post/detail`).
+ *
+ * @since 3.6.0
+ */
+class Analytics_Post_Detail_API extends Remote_API_Base {
+	protected const ENDPOINT     = Parsely::PUBLIC_API_BASE_URL . '/analytics/post/detail';
+	protected const QUERY_FILTER = 'wp_parsely_analytics_post_detail_endpoint_args';
+}
```
mehmoodak, line 23: All of these new files just contain renaming and updated comments.

mehmoodak (commented):
Left comments for easier review.

src/RemoteAPI/class-remote-api-base.php
```diff
@@ -77,7 +77,7 @@ public function get_api_url( array $query ): string {
 	}
 
 	/**
-	 * Implements the fetcher for the proxy interface.
+	 * Get items from the specified endpoint.
```
acicovic:
```suggestion
	 * Gets items from the specified endpoint.
```

acicovic (approved):
Great work!

Left a minor comment.

PR #1288, opened 2023-01-04
Refactor Content Helper for better structure

acicovic:
## Description
The current code in the Content Helper needs some refactoring in order to be structured in a consistent way that promotes maintenability.

For example, the current file structure is not very well organized, and some renaming is needed. Also, we have multiple SCSS files that could be joined into one, or at least harmonized to follow the same coding conventions. There might be additional issues which do not affect functionality, but do not offer an optimal maintenability experience, and make it difficult to understand the codebase.

This PR attempts to address the aforementioned and any other related issues.

## Motivation and Context
Better structured code that leads to easier maintenance, and a convention that we can follow in the future when adding new panels to the Content Helper.

## How Has This Been Tested?
- Existing tests pass.
- Manually checked that the changes in CSS do not result in style breakage.


src/blocks/content-helper/content-helper.tsx
```diff
@@ -9,8 +9,8 @@ import { registerPlugin } from '@wordpress/plugins';
 /**
  * Internal dependencies
  */
-import CurrentPostDetails from './current-post-details/component';
-import RelatedTopPostList from './components/related-top-post-list';
+import PerformanceDetails from './performance-details/component';
+import RelatedTopPostList from './related-top-performing-posts/component-list';
```
mehmoodak: Here `related-top-posts` seems better than `related-top-performing-posts`.

mehmoodak (approved):
Just left one comment but approved the PR.

BTW I have a suggestion that maybe it is better to follow Angular like naming convention here like:
- index.component.tsx
- list.component.tsx
- list-item.component.tsx
- content-helper.provider.ts
- content-helper.model.ts

src/blocks/shared/variables.scss
```diff
@@ -1,10 +1,43 @@
...
-
-// Colors
-$black: #000;
-// gray
-$gray-300: #edeeef;
-$gray-600: #586069;
-$gray-700: #444d56;
-// blue
-$blue-550: #2596db;
+/**
```
acicovic, line 4: Moved all related CSS code to the shared SCSS file.

src/blocks/content-helper/content-helper.scss
```diff
@@ -1,313 +1,312 @@
...
 @import "../shared/variables";
 @import "../shared/functions";
 
+.wp-parsely-content-helper {
 
-.parsely-spinner-wrapper {
-	display: flex;
-	justify-content: center;
-	margin-top: to_rem(40px);
+	.parsely-spinner-wrapper {
```
acicovic, line 6: Most of the changes in this file are indentation differences (done to avoid too many indentation levels). I have moved a couple of things around and done some limited re-organizing.

acicovic (commented):
Left a couple of comments for easier review. Most of the changes are just file names, but there are some more significant changes to SCSS files.

src/blocks/content-helper/content-helper.tsx
acicovic: I've renamed `related-top-performing-posts` to `related-top-posts` as suggested. I did not rename with Angular naming conventions but we can do so in the future if needed.

I'll soon mark this PR as ready for review.

acicovic:
Opening this for review. There are a couple of additional enhancements that could be made, but I think that this covers the basics for now.

src/blocks/shared/variables.scss
```diff
@@ -1,10 +1,43 @@
...
+
+	/** Additional variables. **/
+	--font-size--large: 1rem;
+	--font-size--extra-large: 1.2rem;
+	--black: #000;
+	--sidebar-black: #1e1e1e;
+}
+
+/** SASS variables **/
+$html-font-size: 16px; // Used in functions.scss.
```
mehmoodak: It is better to put code without nesting on top.

src/blocks/shared/variables.scss
```diff
@@ -1,10 +1,43 @@
-$html-font-size: 16px;
-
-// Colors
-$black: #000;
-// gray
-$gray-300: #edeeef;
-$gray-600: #586069;
-$gray-700: #444d56;
-// blue
-$blue-550: #2596db;
```
mehmoodak, line 10: Personally I like using SCSS variables because CSS ones are verbose. But also fine if we are going with one convention.

mehmoodak (approved):
Looks Good.

src/blocks/shared/variables.scss
acicovic: I've done this, though it doesn't have any difference in the output files.

src/blocks/shared/variables.scss
acicovic: I like SCSS variables for the same reason and more so because you don't have to use `var()` everywhere. However, I've seen generally recommended to use CSS variables where possible, and most of the code was in CSS variables already.

We can look into that if it becomes important.

PR #1305, opened 2023-01-12
Content Helper: Add WordPress Dashboard Widget

### LINKED ISSUE #1221: Add Dashboard Widget
wp-parsely users would benefit from additional data and content performance context within their WordPress dashboard. 

We want to add the following data points to a new Parse.ly Content Helper screen element for the wp-admin dashboard. 

**Screen element** 
As a user, I'd like the option to select the `Parse.ly Content Helper` as a screen element on my dashboard, so I can look at Parse.ly data when I log in to wp-admin. 
screen-element

If the user has selected the `Parse.ly Content Helper` element, they see the following components of the screen element
mockup

**Factoids**
As a user, I would like a quick snapshot of how my content is performing today for the big, key metrics I could see in my Parse.ly dashboard:
- page views
- visitors
- ~minutes~
- avg. time
- ~soc. interactions~
- ~new posts~

**Top Posts Today (top 3)**
As a user, I would like a quick snapshot of how my top content is performing today, similar to how I could see this on the overview in the dashboard. Metrics I'd like to see:
- post title
- ~section~
- ~author~
- ~referrer~
- page views
- ~spark line~

acicovic:
## Description
This PR adds a WordPress Dashboard Widget which displays the site's 3 top posts in the last 7 days.

## Motivation and context
- Start showing some useful data in the WordPress Dashboard.
- Closes #1221.

## How has this been tested?
Some existing tests have been modified and pass, and some additional tests will be written as a separate PR.

###	Parse.ly flavored responsive design.
image

image

### Identical behavior with the Content Helper sidebar for errors and messages.
image


src/blocks/content-helper/performance-details/component.tsx
```diff
@@ -300,60 +301,4 @@ function ActionsSection( props: PerformanceSectionProps ) {
 	);
 }
 
-/**
```
acicovic, line 303: Moved the function below to another file for better re-use.

src/blocks/content-helper/performance-details/provider.ts
```diff
@@ -183,31 +187,6 @@ class PerformanceDetailsProvider {
 
 		return response.data;
 	}
-
-	/**
```
acicovic, line 187: Moved the functions below to another file for better re-use

src/Endpoints/class-analytics-posts-api-proxy.php
```diff
@@ -53,13 +53,15 @@ protected function generate_data( $response ): array {
...
-					'id'      => $item->url,
-					'dashUrl' => Parsely::get_dash_url( $site_id, $item->url ),
-					'title'   => $item->title,
-					'url'     => $item->url,
-					'views'   => $item->metrics->views,
+					'author'         => $item->author,
+					'dashUrl'        => Parsely::get_dash_url( $site_id, $item->url ),
+					'date'           => wp_date( $date_format, strtotime( $item->pub_date ) ),
+					'id'             => $item->url,
+					'postId'         => url_to_postid( $item->url ), // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.url_to_postid_url_to_postid
```
acicovic, line 62: Added Thumb URL and Post ID fields.

acicovic (commented):
Left a couple of comments for easier review.

src/blocks/shared/functions.ts
```diff
@@ -0,0 +1,55 @@
...
+ * Implements the "Imprecise Number" functionality of the Parse.ly dashboard.
+ *
+ * Note: This function is not made to process float numbers.
+ *
+ * @param {string} value          The number to process. It can be formatted.
+ * @param {number} fractionDigits The number of desired fraction digits.
+ * @param {string} glue           A string to put between the number and unit.
+ * @return {string} The number formatted as an imprecise number.
+ */
+export function impreciseNumber( value: string, fractionDigits = 1, glue = '' ): string {
```
mehmoodak: 1. `formatNumber` seems like a better name.
2. Can you please also add tests to make it resilient for future changes.

src/blocks/shared/functions.ts
```diff
@@ -0,0 +1,55 @@
...
+	const number = parseInt( value.replace( /\D/g, '' ), 10 );
+
+	if ( number < 1000 ) {
+		return value;
+	} else if ( number < 10000 ) {
+		fractionDigits = 1;
+	}
+
+	const unitNames = {
```
mehmoodak: Type here?

src/blocks/shared/utils/date.ts
```diff
@@ -1,8 +1,54 @@
...
+ * the passed date is within the current year), respecting the user's language.
+ *
+ * @param {Date} date The date to be formatted.
+ * @return {string} The resulting date in its final format.
+ */
+export function getSmartShortDate( date: Date ): string {
+	let dateFormat = SHORT_DATE_FORMAT;
+
+	if ( date.getUTCFullYear() === new Date().getUTCFullYear() ) {
+		dateFormat = SHORT_DATE_FORMAT_WITHOUT_YEAR;
```
mehmoodak, line 22: Intead of adding `SHORT_DATE_FORMAT_WITHOUT_YEAR` maybe its better to remove `year` key from `dateFormat`

src/Endpoints/class-analytics-posts-api-proxy.php
mehmoodak: 1. Here the keys doesn't makes much sense i.e. `id` should have the actual ID of the post instead of the `postId` but we are passing `url` in that even when there is a separate `url` field.

2. How much accurately we can get `ID` from the `url` using `url_to_postid` ..... does it have same assumption that permalink configuration of the local site should be same as per production?

src/Endpoints/class-analytics-posts-api-proxy.php
```diff
@@ -53,13 +53,15 @@ protected function generate_data( $response ): array {
...
+					'thumbUrlMedium' => $item->thumb_url_medium,
+					'title'          => $item->title,
+					'url'            => $item->url,
+					'views'          => $item->metrics->views,
```
mehmoodak, line 66: Should we add a check for `$item->metrics` .... for cases where API doesn't send any metrics?

src/blocks/shared/utils/date.ts
```diff
@@ -1,8 +1,54 @@
...
+
+/**
+ * Removes the given number of days from a "YYYY-MM-DD" string, and returns
+ * the result in the same format.
+ *
+ * @param {string} date The date in "YYYY-MM-DD" format.
+ * @param {number} days The number of days to remove from the date.
+ * @return {string} The resulting date in "YYYY-MM-DD" format.
+ */
+export function removeDaysFromDate( date: string, days: number ): string {
```
mehmoodak, line 39: `removeDaysFromDate` from date is giving the assumption that we are just taking year and months and removing days from the date. Maybe it is better to name the function `addDaysInDate` and make it generic so that we can handle both addition and subtraction of days.

src/content-helper/dashboard-widget/class-dashboard-widget.php
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	/**
+	 * Adds the Widget and its contents to the WordPress Dashboard.
+	 *
+	 * @since 3.7.0
+	 */
+	public function add_dashboard_widget(): void {
+		wp_add_dashboard_widget(
+			'wp-parsely-dashboard-widget',
```
mehmoodak, line 42: This id seems very generic so `'wp-parsely-top-posts-dashboard-widget'` seems like a good condidate for ID.

src/content-helper/dashboard-widget/class-dashboard-widget.php
```diff
@@ -0,0 +1,77 @@
...
+			__( 'Parse.ly Top Posts (Last 7 Days)', 'wp-parsely' ),
```
mehmoodak, line 43: Maybe it is better to have `id` and `title` of the widget as constants in a dedicated php file which represents top posts and then import it from there.

src/content-helper/dashboard-widget/class-dashboard-widget.php
```diff
@@ -0,0 +1,77 @@
...
+	 * @param string $hook_suffix The current admin page.
+	 */
+	public function enqueue_assets( $hook_suffix ): void {
+		if ( 'index.php' === $hook_suffix ) {
+			$asset_php        = require_once plugin_dir_path( PARSELY_FILE ) . 'build/content-helper/dashboard-widget.asset.php';
+			$built_assets_url = plugin_dir_url( PARSELY_FILE ) . 'build/content-helper/';
+
+			wp_enqueue_script(
+				'wp-parsely-dashboard-widget',
+				$built_assets_url . 'dashboard-widget.js',
```
mehmoodak, line 62: Better if we change `handle` and `filename` here also i.e. be more specific.

src/content-helper/dashboard-widget/class-dashboard-widget.php
```diff
@@ -0,0 +1,77 @@
...
+				$asset_php['dependencies'] ?? null,
+				$asset_php['version'] ?? Parsely::VERSION,
+				true
+			);
+
+			wp_enqueue_style(
+				'wp-parsely-dashboard-widget',
+				$built_assets_url . 'dashboard-widget.css',
```
mehmoodak, line 70: Same as above.

src/content-helper/dashboard-widget/dashboard-widget.scss
```diff
@@ -0,0 +1,125 @@
...
+	.parsely-contact-us {
+		margin-top: to_rem(15px) !important;
+	}
+
+	p.parsely-error-hint {
+		color: var(--gray-700);
+	}
+}
+
+#wp-parsely-dashboard-widget .parsely-top-posts-wrapper {
```
mehmoodak, line 26: We can avoid using `#wp-parsely-dashboard-widget` here and nest this block above.

src/content-helper/dashboard-widget/dashboard-widget.scss
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	// ol element.
+	.parsely-top-posts {
+		counter-reset: item; // Needed to increment post numbers.
+		list-style: none;
+		margin: 0;
+	}
+
+	// li elements.
```
mehmoodak: This and above comment seems extra i.e. names seems enough.

src/content-helper/dashboard-widget/dashboard-widget.scss
```diff
@@ -0,0 +1,125 @@
...
+
+	.parsely-top-post-views {
+		float: right;
+		font-family: var(--numeric-font);
+		font-size: to_rem(18px);
+		padding-left: to_rem(10px);
+	}
+}
+
+@media only screen and (max-width: 380px) {
```
mehmoodak: Instead of having a separate media query block like this I think it is much better to include these media queries in their specific selectors ...... seems much more maintainable and easy to read in this way.

tests/Integration/Endpoints/AnalyticsPostsProxyEndpointTest.php
```diff
@@ -136,22 +136,26 @@ function () use ( &$dispatched ) {
...
-						'views'   => 142,
+						'author'         => 'Aakash Shah',
+						'date'           => wp_date( $date_format, strtotime( '2020-04-06T13:30:58' ) ),
+						'id'             => 'https://blog.parse.ly/web-analytics-software-tools/?itm_source=parsely-api',
+						'dashUrl'        => PARSELY::DASHBOARD_BASE_URL . '/blog.parsely.com/find?url=https%3A%2F%2Fblog.parse.ly%2Fweb-analytics-software-tools%2F%3Fitm_source%3Dparsely-api',
+						'thumbUrlMedium' => 'https://images.parsely.com/XCmTXuOf8yVbUYTxj2abQ4RSDkM=/85x85/smart/https%3A//blog.parse.ly/wp-content/uploads/2021/06/Web-Analytics-Tool.png%3Fw%3D150%26h%3D150%26crop%3D1',
+						'title'          => '9 Types of Web Analytics Tools — And How to Know Which Ones You Really Need',
+						'url'            => 'https://blog.parse.ly/web-analytics-software-tools/?itm_source=parsely-api',
+						'views'          => 142,
+						'postId'         => 0,
```
mehmoodak, line 147: Can we pass valid ID and checks that `url_to_postId` is working as expected?

src/content-helper/dashboard-widget/top-posts/model.ts
```diff
@@ -0,0 +1,11 @@
+export interface TopPostData {
+	author: string;
+	dashUrl: string;
+	date: string;
+	id: number;
+	postId: number;
```
mehmoodak, line 6: `id` and `postId` seems similar ... can we use only one?

src/content-helper/dashboard-widget/dashboard-widget.tsx
```diff
@@ -0,0 +1,21 @@
...
+ * External dependencies
+ */
+import { render } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import TopPostList from './top-posts/component-list';
+
+// Insert Dashboard Widget content when the page loads.
```
mehmoodak: Comment seems extra

src/content-helper/dashboard-widget/provider.ts
```diff
@@ -0,0 +1,104 @@
...
+import {
+	convertDateToString,
+	removeDaysFromDate,
+} from '../../blocks/shared/utils/date';
+
+/**
+ * The form of the response returned by the /stats/posts WordPress REST API
+ * endpoint.
+ */
+interface TopPostsApiResponse {
```
mehmoodak, line 25: Instead of defining new interface I think it is better to extend the existing interface i.e. if we are using same API then interfaces should also be common.

src/content-helper/dashboard-widget/provider.ts
```diff
@@ -0,0 +1,104 @@
...
+	error?: Error;
+	data?: TopPostData[];
+}
+
+/**
+ * The form of the result returned by the getTopPosts() function.
+ */
+export interface GetTopPostsResult {
```
mehmoodak: Is it better to just return `TopPostData[]` and avoid this interface ...... for me it seems better because its simple.

src/content-helper/dashboard-widget/provider.ts
```diff
@@ -0,0 +1,104 @@
...
+	posts: TopPostData[];
+}
+
+export const TOP_POSTS_DEFAULT_LIMIT = 3;
+export const TOP_POSTS_DEFAULT_TIME_RANGE = 7; // In days.
```
mehmoodak, line 31: For related posts it is `3 days` here we are going with `7 days` should we make it consistent?

If we are making this consistent then should we define this in some common location and use from that or should we just enforce this by default on our API endpoint?

src/content-helper/dashboard-widget/provider.ts
```diff
@@ -0,0 +1,104 @@
...
+}
+
+/**
+ * The form of the result returned by the getTopPosts() function.
+ */
+export interface GetTopPostsResult {
+	posts: TopPostData[];
+}
+
+export const TOP_POSTS_DEFAULT_LIMIT = 3;
```
mehmoodak, line 30: I know the decision has been already taken to only show 3 posts but for me the numbers seems very very low. I think we should go with at least 5 but I am more inclined toward 7 or 10.

src/content-helper/dashboard-widget/provider.ts
```diff
@@ -0,0 +1,104 @@
...
+
+	/**
+	 * Constructor.
+	 */
+	constructor() {
+		this.dataPeriodEnd = convertDateToString( new Date() ) + 'T23:59';
+		this.dataPeriodStart = removeDaysFromDate(
+			this.dataPeriodEnd,
+			TOP_POSTS_DEFAULT_TIME_RANGE - 1
+		) + 'T00:00';
```
mehmoodak, line 45: Using same logic inside `PerformanceDetailsProvider` .... should we make a common function for it?

src/content-helper/dashboard-widget/provider.ts
```diff
@@ -0,0 +1,104 @@
...
+	public async getTopPosts(): Promise<GetTopPostsResult> {
+		let data;
+
+		try {
+			data = await this.fetchTopPostsFromWpEndpoint();
+		} catch ( contentHelperError ) {
+			return Promise.reject( contentHelperError );
+		}
+
+		return { posts: data };
```
mehmoodak: If we change the type as suggested above then we can write this code as follows which is more easy to read:

```
try {
  return await this.fetchTopPostsFromWpEndpoint();
} catch ( contentHelperError ) {
  return Promise.reject( contentHelperError );
}
```

src/content-helper/dashboard-widget/top-posts/component-list.tsx
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Internal dependencies
+ */
+import DashboardWidgetProvider from '../provider';
+import TopPostListItem from './component-list-item';
+import { TopPostData } from './model';
+import { ContentHelperError } from '../../../blocks/content-helper/content-helper-error';
+import { getDateInUserLang, SHORT_DATE_FORMAT } from '../../../blocks/shared/utils/date';
+
+const FETCH_RETRIES = 3;
```
mehmoodak, line 17: Using this in multiple places .... maybe we should define this in some common location 🤔

src/content-helper/dashboard-widget/provider.ts
```diff
@@ -0,0 +1,104 @@
...
+				response.error.message,
+				ContentHelperErrorCode.ParselyApiResponseContainsError
+			) );
+		}
+
+		return response?.data || [];
+	}
+}
+
+export default TopPostsProvider;
```
mehmoodak: Name doesn't matches with path ..... seems like directory`dashboard-widget` should be renamed to `top-posts`

src/content-helper/dashboard-widget/top-posts/component-list-item.tsx
```diff
@@ -0,0 +1,91 @@
...
+interface TopPostListItemProps {
+	post: TopPostData;
+}
+
+/**
+ * Returns a single list item depicting a post.
+ *
+ * @param {TopPostData} post The Post to be shown.
+ */
+function TopPostListItem( { post }: TopPostListItemProps ): JSX.Element {
```
mehmoodak, line 23: `top-post-list-item.tsx` seems like a better filename.

src/content-helper/dashboard-widget/top-posts/component-list-item.tsx
```diff
@@ -0,0 +1,91 @@
...
+	return (
+		<li className="parsely-top-post">
+			<div className="parsely-top-post-wrapper">
```
mehmoodak: Using `wrapper` makes more sense if we uses this `div` to wrap around `li` .... I think `parsely-top-post-content` is a better name.

src/content-helper/dashboard-widget/top-posts/component-list-item.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+				<div className="parsely-top-post-thumbnail">
+					<span className="screen-reader-text">Thumbnail</span>
+					<img src={ post.thumbUrlMedium } alt={ __( 'Post thumbnail', 'wp-parsely' ) } />
```
mehmoodak: What wil happen if there is no thumbnail?

src/content-helper/dashboard-widget/top-posts/component-list-item.tsx
```diff
@@ -0,0 +1,91 @@
...
+				</div>
+
+				<div className="parsely-top-post-data">
+
+					<span className="parsely-top-post-views">
+						<span className="screen-reader-text">Number of Views</span>
+						{ impreciseNumber( post.views.toString() ) }
+					</span>
+
+					{ getPostTitleElement( { post } ) }
```
mehmoodak, line 39: This takes the user to edit view of the post ..... shouldn't it be in normal view mode?

src/content-helper/dashboard-widget/top-posts/component-list-item.tsx
```diff
@@ -0,0 +1,91 @@
...
+	let titleLink =
+		<a className="parsely-top-post-title" href={ titleLinkUrl } target="_blank" rel="noreferrer">
+			<span className="screen-reader-text">
+				{ __( 'View in Parse.ly (opens in new tab)', 'wp-parsely' ) }
+			</span>
+			{ post.title }
+		</a>;
+
+	if ( 0 === post.postId ) {
+		titleLink = <div className="parsely-top-post-title">{ post.title }</div>;
```
mehmoodak: We can move this at start of the function and return directly.

src/content-helper/dashboard-widget/top-posts/component-list.tsx
```diff
@@ -0,0 +1,92 @@
...
+function TopPostList() {
+	const [ loading, setLoading ] = useState<boolean>( true );
+	const [ error, setError ] = useState<ContentHelperError>();
+	const [ posts, setPosts ] = useState<TopPostData[]>( [] );
+	const provider = new DashboardWidgetProvider();
+
+	useEffect( () => {
+		const fetchPosts = async ( retries: number ) => {
+			provider.getTopPosts()
+				.then( ( result ): void => {
```
mehmoodak, line 31: Can we add type on param `result`?

mehmoodak (changes requested):
Done with 1st review and left some comments. Apart from this

1. Can you please adds tests for the added functionality in same PR? I think it is a nice practice to follow this instead of separate PR just to enforce proper testing.

2. While testing locally I am getting `Error: The response is not a valid JSON response.` inside widget so please let me know if there is any other configuration needed on my side to test this. Apart from that the WP REST API returns `404` so maybe it should display the right error.

acicovic:
> While testing locally I am getting Error: The response is not a valid JSON response. inside widget so please let me know if there is any other configuration needed on my side to test this. Apart from that the WP REST API returns 404 so maybe it should display the right error.

This is strange. No additional configuration is needed and it works for me locally. Have you tried restarting your environment? This seems like a glitch.

src/content-helper/dashboard-widget/class-dashboard-widget.php
acicovic: We actually want this name to be generic, as we should be adding more data to it in next iterations. The reason we've only added top posts so far is because this is the only thing we had available from the API until now.

mehmoodak: Maybe we should change it in future then when its actually needed ...... just a suggestion

mehmoodak:
> This is strange. No additional configuration is needed and it works for me locally. Have you tried restarting your environment? This seems like a glitch.

I notice that this happens when I change default Permalink Settings .... can you please try it to see if it produces same error for you?

acicovic:
> I notice that this happens when I change default Permalink Settings .... can you please try it to see if it produces same error for you?

Cannot reproduce this, but will keep looking.

src/Endpoints/class-analytics-posts-api-proxy.php
acicovic: 1. The `id` key has been used for React at some places and it is not a new field. `postId` can return 0 if the post is not found, so it is not suitable for an ID, since we can have multiple 0s. For now, URLs are still better for uniqueness.
3. I have no issue with it switching permalink configurations, as it keeps working.

src/content-helper/dashboard-widget/top-posts/model.ts
acicovic: Related to [this](https://github.com/Parsely/wp-parsely/pull/1305#discussion_r1086894570), they cannot be used interchangeably.

src/blocks/shared/functions.ts
acicovic: Added

src/Endpoints/class-analytics-posts-api-proxy.php
acicovic: @abelsonlive, is it possible for the `analytics/posts` endpoint to return without metrics data?

src/blocks/shared/utils/date.ts
acicovic: This function was moved from a place to another and was left unchanged. I'd prefer not to proceed with any refactorings that do not give any direct value to this specific PR. We can of course update this function when the need arises.

src/content-helper/dashboard-widget/class-dashboard-widget.php
acicovic: At this point the plugin has only one Dashboard Widget, which should be populated with more data in the next iterations. As such, I doubt we get value in renaming it now and then having to rename it back.

src/content-helper/dashboard-widget/class-dashboard-widget.php
acicovic: See my comment above about renaming.

src/content-helper/dashboard-widget/class-dashboard-widget.php
acicovic: See my comment above about renaming.

src/content-helper/dashboard-widget/provider.ts
acicovic: 👍 I'm planning a refactoring PR after this one, which should harmonize Content Helper code a bit more. We have made some progress for re-use and will be continuing that.

src/content-helper/dashboard-widget/provider.ts
acicovic: Yes, but not in this PR. See [this](https://github.com/Parsely/wp-parsely/pull/1305#discussion_r1086950059).

src/content-helper/dashboard-widget/provider.ts
acicovic: With the latest changes, I'm not `return`ing immediately here, since I'm checking for an empty array below.

src/content-helper/dashboard-widget/top-posts/component-list.tsx
acicovic: Related to [this](https://github.com/Parsely/wp-parsely/pull/1305#discussion_r1086950059). I'll be noting this for that PR.

src/content-helper/dashboard-widget/provider.ts
acicovic: See my comment about renaming.

src/content-helper/dashboard-widget/top-posts/component-list-item.tsx
acicovic: See my comment about renaming.

src/content-helper/dashboard-widget/top-posts/component-list-item.tsx
acicovic: @abelsonlive, I guess is it possible for the `analytics/posts` endpoint to return an empty thumb URL?

src/content-helper/dashboard-widget/top-posts/component-list-item.tsx
acicovic: This is as per the requirements [here](https://github.com/Parsely/wp-parsely/issues/1221#issuecomment-1353319380). I've posted about this in internal communications, in case we decide to change it in the future.

src/content-helper/dashboard-widget/provider.ts
acicovic: This is as per the requirements [here](https://github.com/Parsely/wp-parsely/issues/1221#issuecomment-1353319380). I've asked about this myself in internal communications, and the decision was for it to remain 3. So for now this remains as-is.

src/content-helper/dashboard-widget/provider.ts
acicovic: I'd flag this in our internal communications, and we can get back to it if needed in another PR. It is a small change to make if we need to. This is as per the requirements [here](https://github.com/Parsely/wp-parsely/issues/1221#issuecomment-1353319380)

src/Endpoints/class-analytics-posts-api-proxy.php
abelsonlive: @acicovic no, there will never be missing metrics data, nor null values for metrics, though some metrics may have the value of `0`.

src/content-helper/dashboard-widget/top-posts/component-list-item.tsx
abelsonlive: @acicovic yes, it could be null for a number of reasons, most commonly because a post has only recently been published and has not yet been crawled. You might consider using the `image_url` as a fallback, though that will have different dimensions.

src/content-helper/dashboard-widget/dashboard-widget.scss
acicovic: I find it actually helps when skimming files to have comments like this (or comments that quickly describe blocks of code), but removed all comments that you refer as extra.

src/blocks/shared/functions.ts
acicovic: Renaming this to `formatToImpreciseNumber`, to make this function have a name that is function-like but also specific to what the function does. "Imprecise Number" is the Parse.ly term used for this type of formatting.

src/Endpoints/class-analytics-posts-api-proxy.php
acicovic: @mehmoodak, so we should be OK here without any additional checks.

src/blocks/shared/utils/date.ts
acicovic: I went to make this change, but I'm unsure this would be cleaner or more reliable. Would you use `delete` for this, or something else? Maybe there is something I'm missing.

src/content-helper/dashboard-widget/class-dashboard-widget.php
acicovic: What advantage does this offer us in the current implementation?

src/content-helper/dashboard-widget/dashboard-widget.scss
acicovic: Yes, I've done this on purpose to avoid too many nested levels that can result in huge CSS parent blocks that are harder to read. This approach of defining major CSS blocks separately has been taken in other files as well (see content-helper.scss). This should not have an impact on the generated CSS file, so I do not see a good reason to change this. However, if your opinion differs let me know.

src/content-helper/dashboard-widget/dashboard-widget.scss
acicovic: Good catch! done.

tests/Integration/Endpoints/AnalyticsPostsProxyEndpointTest.php
acicovic: Those are remote URLs we are testing, so these posts do not exist in the local WP instance. As such, `url_to_postid` will always return 0 here. We could probably create some local WP posts and test with that. This would need rework though, so we could do that as part of the tests PR.

src/content-helper/dashboard-widget/provider.ts
acicovic: Done, and I've also added some code to handle the case of an empty data array.

src/content-helper/dashboard-widget/provider.ts
acicovic: Renamed the function to `DashboardWidgetProvider` which reflects its correct purpose.

src/content-helper/dashboard-widget/top-posts/component-list-item.tsx
acicovic: Regarding this file, I've updated it to include the suggestions. We now take care of an empty thumbnail and we're using early returns.

src/content-helper/dashboard-widget/top-posts/component-list.tsx
acicovic: We're doing the same thing in the Related Top Posts list. Maybe this would go into the refactorings. What would you suggest doing here exactly?

src/blocks/shared/utils/date.ts
mehmoodak: If this is more cleaner then lets keep it.

src/content-helper/dashboard-widget/class-dashboard-widget.php
mehmoodak: Let's leave it as it is ..... I was thinking it differently before.

src/Endpoints/class-analytics-posts-api-proxy.php
mehmoodak: Can we please have point 1 as comment somewhere?

src/content-helper/dashboard-widget/provider.ts
mehmoodak: I think it is better to do in same PR .... but your call 🙂

src/content-helper/dashboard-widget/provider.ts
mehmoodak: Since our main Goal here is to show users what Parse.ly is capable so I think more than 3 is better.

@danielabloch can you please give it another thought. Thanks

src/content-helper/dashboard-widget/top-posts/component-list.tsx
mehmoodak: Only suggesting to add typing which can be helpful while understanding the code.

mehmoodak (approved):

src/Endpoints/class-analytics-posts-api-proxy.php
acicovic: I've added a couple of comments here in the hope that it will be clearer. This section might change as the API gets updated and as we start sending Post IDs in metadata.

src/content-helper/dashboard-widget/provider.ts
danielabloch: This is on my list of things for us to revisit as well @mehmoodak and @acicovic but for now, let's keep it at 3.

acicovic:
@mehmoodak, I've fiddled with permalink settings, but been unable to reproduce this. Let me know if you still have the issue. If yes, can you also try to give specific instructions for reproduction?

Thank you!

acicovic:
Note for posterity: The issue mentioned by @mehmoodak turned out to be something local, so it should not be a concern here.

PR #1331, opened 2023-01-26
Improve Parse.ly Stats on post list screen

### LINKED ISSUE #1330: Show Contact Us message on Parsely Stats if API Secret is not set
Currently we don't show the column if API Secret isn't set so we should changes this flow to following:

- If API secret is set
  - column is visible
- If API secret is not set
  - if column is visible then show contact us message
  - if column is hidden then don't show anything and also not make any API call

mehmoodak:
## Description
Closes: https://github.com/Parsely/wp-parsely/issues/1330

## How has this been tested?
- Added integration tests tests.

### DIFF
--- a/src/UI/class-admin-columns-parsely-stats.php
+++ b/src/UI/class-admin-columns-parsely-stats.php
@@ -83,7 +83,7 @@ class Admin_Columns_Parsely_Stats {
 	 * @since 3.7.0
 	 */
 	public function run(): void {
-		if ( ! $this->parsely->site_id_is_set() || ! $this->parsely->api_secret_is_set() ) {
+		if ( ! $this->parsely->site_id_is_set() ) {
 			return;
 		}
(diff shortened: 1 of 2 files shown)


src/UI/class-admin-columns-parsely-stats.php
```diff
@@ -223,6 +240,20 @@ public function get_parsely_stats_response( $analytics_api ) {
 			return null;
 		}
 
+		if ( ! $this->parsely->api_secret_is_set() ) {
+			return array(
+				'data'  => null,
+				'error' => array(
```
acicovic, line 246: Can/should we make the messages translatable?

mehmoodak: Oops ... Done

acicovic (approved):

PR #1339, opened 2023-01-31
Settings page: Replace screen options with tabs

### LINKED ISSUE #939: Settings Page: Add tabs for sections
Our current settings page implementation uses Screen Options to show and hide the "Requires Recrawl" and "Advanced Settings" sections.

The downsides of this approach are:
- Some users might not be familiar with Screen Options and might miss it (we've [included this in the docs](https://www.parse.ly/help/integration/troubleshooting#i-cant-see-all-options-in-the-parsely-settings-page) as a means to improve this).
- It needs some clicks to enable/disable the sections.

Implement tabs to make switching between sections more obvious and quicker, as well as to make the settings page more compact for users.

Notes:
- We should cater for tab accessibility.
- This could be part of a total Settings page revamp, taking into account #485 and #799.

mehmoodak:
## Description
Added tabs in Parse.ly settings and removed screen options.

Closes: #939 

## How has this been done?
- Removes screen options and help text.
- Added tabs.
- Refactors validate functions i.e. divided one big function which now reprsents fields of  each tab.

- Renamed `Requires Recrawl Settings` to `Recrawl Settings` because of better naming

## How has this been tested?
- Tested manually that each field are storing fine (better if someone else also does this just to be safe)
- Changed existing tests to reflect new changes

## Screenshots

image


src/UI/class-settings-page.php
```diff
@@ -316,7 +271,7 @@ public function initialize_settings(): void {
 	private function initialize_basic_section(): void {
 		add_settings_section(
 			'basic_settings',
-			__( 'Basic Settings', 'wp-parsely' ),
```
mehmoodak, line 319: Removed because now tabs already shows this and having title also seems redundant.

src/UI/class-settings-page.php
```diff
@@ -998,13 +995,16 @@ public function validate_options( $input ) {
 			}
 		}
 
+		if ( ! isset( $input['meta_type'] ) ) {
+			$input['meta_type'] = $options['meta_type'];
+		} else {
+			$input['meta_type'] = sanitize_text_field( $input['meta_type'] );
+		}
```
mehmoodak, line 958: Moved this to basic section function.

src/UI/class-settings-page.php
```diff
@@ -998,13 +995,16 @@ public function validate_options( $input ) {
...
+
 		if ( '' === $input['logo'] ) {
 			$input['logo'] = self::get_logo_default();
 		}
 
-		// Validate 'Track post type as'.
-		$this->validate_options_post_type_tracking( $input );
```
mehmoodak, line 1007: Moved this to recrawl settings function

src/UI/class-settings-page.php
```diff
@@ -1026,18 +1026,6 @@ public function validate_options( $input ) {
...
-		} elseif ( 'true' !== $input['disable_autotrack'] && 'false' !== $input['disable_autotrack'] ) {
-			add_settings_error(
-				Parsely::OPTIONS_KEY,
-				'disable_autotrack',
-				__( 'Value passed for disable_autotrack must be either "Yes" or "No".', 'wp-parsely' )
-			);
-		} else {
-			$input['disable_autotrack'] = 'true' === $input['disable_autotrack'];
-		}
-
```
mehmoodak, line 1041: Moved this to advanced settings function

src/UI/class-settings-page.php
```diff
@@ -1057,14 +1045,20 @@ public function validate_options( $input ) {
 			$input['disable_amp'] = 'true' === $input['disable_amp'];
 		}
 
-		$input['api_secret'] = sanitize_text_field( $input['api_secret'] );
```
mehmoodak, line 1061: Was a duplicate so removed.

views/parsely-settings.php
```diff
@@ -11,27 +11,45 @@
...
-<?php
-if ( is_multisite() && is_main_site() ) {
-	?>
-		<div class="notice notice-info">
-			<p><?php esc_html_e( 'Attention: this is the main site of your Multisite Network.', 'wp-parsely' ); ?></p>
-		</div>
+
+	<?php $wp_parsely_settings->show_setting_tabs(); ?>
+
+	<div class="tab-content">
```
mehmoodak: Added tabs and wrapped the function inside `.tab-content`

views/parsely-settings.php
```diff
@@ -11,27 +11,45 @@
...
+			<div class="notice notice-info">
+				<p><?php esc_html_e( 'Attention: this is the main site of your Multisite Network.', 'wp-parsely' ); ?></p>
+			</div>
+		<?php
+	}
+	?>
+		<form
+			name="parsely"
+			method="post"
+			action=<?php echo esc_url( 'options.php?tab=' . $wp_parsely_settings->get_active_tab() ); ?>
```
mehmoodak: Passed tab info so that validations will be run only for the fields in active tab.

wp-parsely.php
```diff
@@ -134,8 +134,8 @@ function parsely_admin_init_register(): void {
  * Network Admin Sites List table.
  */
 function parsely_wp_admin_early_register(): void {
-	$settings_page = new Settings_Page( $GLOBALS['parsely'] );
-	$settings_page->run();
+	$GLOBALS['parsely_settings_page'] = new Settings_Page( $GLOBALS['parsely'] );
```
mehmoodak, line 140: Using global variable because we are using this same instance for showing tabs on `views/parsely-settings.php`

acicovic:
I absolutely love the screenshot!

mehmoodak (commented):

acicovic (changes requested):
First of all, I'll push this back to 3.8 (as was originally the plan) so we have more time to reflect and discuss this change.

E2E tests constantly fail, so this needs fixing.

While I like the simplicity of the current implementation, the user experience has issues. As every tab is treated as a separate page, we have a page reload when moving between tabs. I've noticed that the current solution:

1. Requires users to click the save button on every tab they change settings in.
2. Does not warn that settings will be lost when changes are performed but are not saved and the user navigates to another tab. Users do not expect their changes to be lost as they navigate between tabs.

Most of the tab implementations I've seen do not treat separate tabs as different pages, and I think that this relates to the UX issues above.

So my take would be:
- Let's aim to do this without needing page reloads between tabs.
- If we conclude for any reason that we do not want to go that route, we should notify the user if their form is dirty and they try to navigate to another tab.

mehmoodak:
> E2E tests constantly fail, so this needs fixing.

It was intentional i.e. trying to get first review and then target fixing of e2e tests.

mehmoodak:
@acicovic  feedback implemented about changing tabs without reload. Please have a look again. Thanks

FYI: Will fix e2e tests soon just wanted to get first review before fixing.

mehmoodak:
e2e tests are also updated.

acicovic:
Hello, I'll start reviewing this. Can you resolve the conflict?

src/UI/class-settings-page.php
```diff
@@ -284,6 +203,45 @@ public function display_settings(): void {
...
+	 * Shows setting tabs.
+	 *
+	 * @since 3.8.0
+	 */
+	public function show_setting_tabs(): void {
+		?>
+		<nav class="nav-tab-wrapper">
+			<?php foreach ( $this->setting_tabs as $t ) { ?>
+			<a
+				class="nav-tab <?php echo esc_attr( $t['key'] . '-tab' ); ?>"
```
acicovic: At least for PHP code, my suggestion is to rename `$t` to `$tab` (for here and everywhere else this is done). Makes things more obvious/clearer.

src/UI/class-settings-page.php
```diff
@@ -284,6 +203,45 @@ public function display_settings(): void {
...
+		</nav>
+		<?php
+	}
+
+	/**
+	 * Shows content of setting tabs.
+	 *
+	 * @since 3.8.0
+	 */
+	public function show_content_of_setting_tabs(): void {
```
acicovic: Could be renamed to `show_setting_tabs_content()` (I think it's better but you can decide).

src/UI/class-settings-page.php
```diff
@@ -475,16 +431,18 @@ private function initialize_basic_section(): void {
...
 	 */
-	private function initialize_requires_recrawl_section(): void {
+	private function initialize_recrawl_section( string $section_key ): void {
 		add_settings_section(
-			'requires_recrawl_settings',
+			$section_key,
 			__( 'Requires Recrawl Settings', 'wp-parsely' ),
 			function (): void {
-				echo '<strong>' . wp_kses_post( __( '<span style="color:#d63638">Important:</span> Changing any of these values below on a site currently tracked with Parse.ly will require reprocessing of your Parse.ly data.', 'wp-parsely' ) ) . '</strong><br />';
+				echo '<br /><strong>' . wp_kses_post( __( '<span style="color:#d63638">Important:</span> Changing any of these values below on a site currently tracked with Parse.ly will require reprocessing of your Parse.ly data.', 'wp-parsely' ) ) . '</strong><br />';
```
acicovic, line 373: I cannot see this text in the UI, and I think it should be there. Also, why is this `<br>` needed?

src/UI/class-settings-page.php
```diff
@@ -475,16 +431,18 @@ private function initialize_basic_section(): void {
...
 	/**
-	 * Registers section and settings for Requires Recrawl section.
+	 * Registers section and settings for Recrawl section.
+	 *
+	 * @param string $section_key Key of the section.
 	 *
 	 * @since 3.2.0
 	 */
-	private function initialize_requires_recrawl_section(): void {
+	private function initialize_recrawl_section( string $section_key ): void {
```
acicovic: If we keep this renaming, maybe it's worth adding a since comment that the method has been renamed int 3.8.0.

src/UI/class-settings-page.php
```diff
@@ -1058,14 +1023,20 @@ public function validate_options( $input ) {
...
+	}
 
-		// Custom taxonomy as section.
-		if ( ! isset( $input['meta_type'] ) ) {
-			$input['meta_type'] = $options['meta_type'];
-		} else {
-			$input['meta_type'] = sanitize_text_field( $input['meta_type'] );
-		}
+	/**
+	 * Validate fields of Recrawl Section.
```
acicovic:
```suggestion
	 * Validates fields of Recrawl Section.
```

src/UI/class-settings-page.php
```diff
@@ -1163,6 +1134,31 @@ public function validate_options( $input ) {
 		return $input;
 	}
 
+	/**
+	 * Validate fields of Advanced Section.
```
acicovic:
```suggestion
	 * Validates fields of Advanced Section.
```

wp-parsely.php
acicovic: What does the transition to the global variable offer us? Was it required to make things work?

acicovic (commented):
@mehmoodak, this is great! The user experience is much smoother in my opinion!

Did a first review, leaving some some comments and questions.

@danielabloch, this PR renames the `Requires Recrawl Settings` section to `Recrawl Settings`. Is this OK with you, provided we have the text that explains that a recrawl is required after setting changes?

Also, I wonder if we want to remove the `Settings` string on every tab (see screenshot below - may be rendundant?). If we're unsure, we can leave as is though.

image

danielabloch:
> this PR renames the Requires Recrawl Settings section to Recrawl Settings. Is this OK with you, provided we have the text that explains that a recrawl is required after setting changes?

> Also, I wonder if we want to remove the Settings string on every tab (see screenshot below - may be rendundant?). If we're unsure, we can leave as is though.


1. Yes, Recrawl Settings is a good change, still makes sense. 
2. Agreed, we should provide text explaining that recrawl is required after a setting changes.
3. I also think Settings string on every tab is redundant.

src/UI/class-settings-page.php
acicovic: Marking this as resolved since we will be keeping the new naming.

wp-parsely.php
mehmoodak: I have to use this same instance in `views/parsely-settings.php` so therefore put it in globals.

src/UI/class-settings-page.php
mehmoodak: 1. Added the text in UI
2. Added `<br>` to improve spacing else text appears right below the tabs which looks a bit odd

mehmoodak:
Feedback incorporated and screenshots for plugins directory are also updated.

Please have a look again. Thanks

acicovic (commented):
I'd remove the version strings from the screenshots, because they currently convey inaccurate information (the feature is not available in 3.6.2) and they will make our screenshots look outdated very soon.

Apart from that, everything looks good to me 👍

Let's not merge this yet though, as the decision to add this to 3.7 or 3.8 is pending.

acicovic (approved):
Looks good. Thanks for bearing with me 🙂

Approving this, but let's not merge it yet please.

acicovic:
Scheduling this for 3.8. I've changed the base branch, so it can be merged.

PR #1366, opened 2023-02-03
 Increase PHPStan coverage by including tests in analysing

### LINKED ISSUE #1365: Increase PHPStan coverage by including tests in analysing
As of now we don't run PHPStan on our `tests` directory so as a part of this ticket we will increase our PHPStan coverage by including `tests` directory in PHPStan analysation.

mehmoodak:
## Description
As of now we don't run PHPStan on our tests directory so as a part of this ticket we will increase our PHPStan coverage by including tests directory in PHPStan analysation.

Closes: #1365

## How has this been tested?
- Make sure that existing tests are passing fine after incorporating PHPStan changes.


src/Metadata/class-post-builder.php
```diff
@@ -22,7 +22,7 @@
  *
  * @since 3.4.0
  *
- * @phpstan-import-type ParselyOptions from Parsely
+ * @phpstan-import-type Parsely_Options from Parsely
```
mehmoodak, line 25: Using `_` in naming which is similar to how we name other things. This has been done to some other types also.

tests/Integration/DashboardLinkTest.php
```diff
@@ -39,7 +39,7 @@ public function set_up(): void {
 	 */
 	public function test_generate_parsely_post_url(): void {
 		$post_id = self::factory()->post->create();
-		$post    = get_post( $post_id );
+		$post    = $this->get_post( $post_id );
```
mehmoodak, line 42: `get_post` returns `WP_Post|null` so we will need a lot of checks in the code where first we will check the `null` and then code the rest. This will not look good in the codebase so to avoid this we implemented `get_post` method in our `TestCase` class which will only return the `WP_Post` to ensures strong typing and cleaner code.

Note: This has been done to some other methods also like
  - get_permalink
  - get_term
  - get_term_link
  
and more.

mehmoodak (commented):

tests/Integration/Metadata/DateArchiveTest.php
```diff
@@ -187,11 +187,4 @@ public function tear_down(): void {
 		update_option( 'date_format', 'F j, Y' ); // reset to default.
 		update_option( 'time_format', 'g:i a' ); // reset to default.
 	}
-
-	/**
-	 * Runs once after all tests.
-	 */
-	public static function tear_down_after_class(): void {
```
acicovic, line 194: Can you explain why this was removed? It wasn't useful?

acicovic (approved):

acicovic:
I've added some labels and also changed the milestone to 3.7.0. We can merge this safely as there are no user-facing changes.

tests/Integration/Metadata/DateArchiveTest.php
mehmoodak: As per typing on line 25 we can't assign `null` ..... removing it was easier then changing the types and then add checks in rest of the code.

PR #1362, opened 2023-02-03
Fix PHP 8 incompatibilities

acicovic:
## Description
We have found parts of our code which are incompatible with PHP 8.x.

## Motivation and context
- Closes #1361.
- More context is available in that issue.

## How has this been tested?
1. We first introduced integration tests that would expose the incompatibilities and let GitHub run them.
2. As expected, the tests failed.
3. We committed the fixes that solve the issue.
4. The tests then passed.

PR #1368, opened 2023-02-04
Fix Fatal Error: switch `self` to `$this` since method is not static

mjangda:
I spotted this on our test site:

```
Uncaught Error: Non-static method Parsely\Endpoints\Analytics_Post_Detail_API_Proxy::get_duration() cannot be called statically in /var/www/wp-content/plugins/wp-parsely/src/Endpoints/class-analytics-post-detail-api-proxy.php:53
```

Looks like we are incorrectly calling the `get_duration` method using `self::` rather than `$this->`.

This PR fixes that.

### DIFF
--- a/src/Endpoints/class-analytics-post-detail-api-proxy.php
+++ b/src/Endpoints/class-analytics-post-detail-api-proxy.php
@@ -50,7 +50,7 @@ final class Analytics_Post_Detail_API_Proxy extends Base_API_Proxy {
 		return array_map(
 			static function( stdClass $item ) use ( $site_id ) {
 				return (object) array(
-					'avgEngaged' => self::get_duration( (float) $item->avg_engaged ),
+					'avgEngaged' => $this->get_duration( (float) $item->avg_engaged ),
 					'dashUrl'    => Parsely::get_dash_url( $site_id, $item->url ),
 					'url'        => $item->url,
 					'views'      => number_format_i18n( $item->metrics->views ),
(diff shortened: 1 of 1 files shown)


mjangda:
Oh, I totally missed that this is inside an `array_map` 🤦

acicovic:
Thank you for taking the initiative to work on this!

Actually, we've gone the other way round and [converted the function to static](https://github.com/Parsely/wp-parsely/pull/1362/files#diff-f20b47ffa91acb38714c4047d8793b8ed575125959d642266947ae58ccbea965), since it is not a method that needs to be bound to a specific instance (and may be moved elsewhere in the future).

However, let us know if you think that there is a good reason to use the approach proposed here instead of what we've done.

mehmoodak (approved):

mehmoodak:
Based on Alex comment seems like issue is already resolved.

mjangda:
IMO the `foreach` (as in this PR) simplifies the code (e.g. no need for `use` to pass in additional variables or changing signatures of other methods to work here) which then makes it easier to understand and also change in the future. I'm also not sure if the `array_map` gives us any benefits (happy to learn if there are some).

But, I don't have a strong opinion on this so I'm okay with closing this in favour of the fix in the PHP 8 PR.

acicovic:
Adding this to the 3.8 milestone since we have already a fix and can talk about this later.

mehmoodak:
> IMO the foreach (as in this PR) simplifies the code

Agreed.

> But, I don't have a strong opinion on this so I'm okay with closing this in favour of the fix in the PHP 8 PR.

Closing the PR since there is no strong opinion.

PR #1374, opened 2023-02-08
Content Helper: Fix top referrers percentage displaying as "NaN"

acicovic:
## Description
In certain cases, the top referrers percentage in the Content Helper would display as `NaN`. This was not a problem in the JS side, but rather an erroneous calculation in PHP which resulted in this.

## Motivation and context
Fix bugs!

## How has this been tested?
1. We first introduced integration tests that expose the issue.
2. We committed the fix that solves the issue.
3. The tests passed.

### DIFF
--- a/src/Endpoints/class-referrers-post-detail-api-proxy.php
+++ b/src/Endpoints/class-referrers-post-detail-api-proxy.php
@@ -14,6 +14,8 @@ use stdClass;
 use WP_REST_Request;
 use WP_Error;
 
+use function Parsely\Utils\convert_to_positive_integer;
+
 /**
  * Configures the `/referrers/post/detail` REST API endpoint.
  *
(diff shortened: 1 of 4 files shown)


acicovic:
Marking this as ready for review, though SonarCloud complains about duplication in a test file.

src/Endpoints/class-referrers-post-detail-api-proxy.php
```diff
@@ -212,6 +216,17 @@ private function generate_referrers_data(
...
+	/**
+	 * Converts a string to a positive integer, removing any non-numeric
+	 * characters.
+	 *
+	 * @param string $string The string to be converted to an integer.
+	 * @return int The integer resulting from the conversion.
+	 */
+	private function convert_to_positive_integer( string $string ): int {
+		return (int) preg_replace( '/\D/', '', $string );
+	}
```
mehmoodak: Should we move this to `utils.php`?

mehmoodak (approved):

src/Endpoints/class-referrers-post-detail-api-proxy.php
acicovic: Good catch!

Issue #1377, opened 2023-02-09
Fix PHP warnings on Performance Details panel


mehmoodak:
## Describe the bug
In some cases Parsely API sends incomplete data and since on plugin side we are not expecting this so warnings occurred while accessing the `Undefined` properties. On UI side `Referrers Section` also don't get populated.

## Expected behavior
- No Warnings or Errors
- Referrers sections gets populated on Performance Details panel.

PR #1379, opened 2023-02-09
 Fix Undefined warnings on Performance Details panel

mehmoodak:
## Description
Warnings were appearing on Performance Details panel if API sends incomplete data so just added `isset` checks to avoid the warnings.

Related: #1377 

## How has this been tested?
- Ran existing tests
- Manual test
- Tested on a live VIP site


acicovic (approved):
I've spotted a small difference in numbers compared to the previous state of the code, but it seems to me that this was the API updating its data.

PR #1381, opened 2023-02-10
Fix referral distribution in Performance Details panel

### LINKED ISSUE #1367: [bug] PCH referral distribution only shows direct traffic
## Describe the bug
The referrer section of Performance Details only shows direct traffic, even when there are other referrers. 

## To Reproduce
1. Go to (https://docs.parse.ly/wp-admin/post.php?post=5378&action=edit)
2. Click on PCH sidebar
3. Scroll down to Performance Details
4. See error --> The PCH says it's all direct
5. View in Parse.ly
6. View in Parse.ly - change filter to last 7 days (Jan 29 - Feb 2)
7. If you look at the past 7 days in parse.ly, there are actually a few search referrals

## Expected behavior
The referrals performance details shows you all referrer channels. 

## Screenshots
PCH: Distribution for last 7 days (Jan 29 - today): 41 page views from Direct
Screenshot 2023-02-03 at 1 27 48 PM

View in Parse.ly takes me to Pulse view for today, not last 7 days (new bug)

Change time period to Jan 29 - today) in Parse.ly: 33 page views from Direct
The distribution we'd want to see on the post
Screenshot 2023-02-03 at 1 31 40 PM

## Environment
Chrome browser
Active Parse.ly plugin

mehmoodak:
## Description
In Performance Details panel we are calling two API's which under the hood manages same cache. The root cause of the issue was the cache collision means when we request 1st API then it populates the cache and on 2nd API call instead of requesting data from API it serves the results from cache which are wrong.

Fixes: #1367

## How this PR addresses the issue?
- Updated the cache key logic to avoid cache collisions.

## How has this been tested?
- Ran existing tests.

## Screenshots 

Before
image

After
image

### DIFF
--- a/src/RemoteAPI/class-remote-api-cache.php
+++ b/src/RemoteAPI/class-remote-api-cache.php
@@ -54,7 +54,11 @@ class Remote_API_Cache implements Remote_API_Interface {
 	 *                                             response is empty.
 	 */
 	public function get_items( $query, $associative = false ) {
-		$cache_key = 'parsely_api_' . wp_hash( (string) wp_json_encode( $this->remote_api ) ) . '_' . wp_hash( (string) wp_json_encode( $query ) );
+		$cache_key = (
+			'parsely_api_' .
+			wp_hash( $this->remote_api->get_endpoint() ) . '_' .
+			wp_hash( (string) wp_json_encode( $query ) )
+		);
 
 		/**
 		 * Variable.
--- a/src/RemoteAPI/class-remote-api-base.php
+++ b/src/RemoteAPI/class-remote-api-base.php
@@ -52,6 +52,17 @@ abstract class Remote_API_Base implements Remote_API_Interface {
 	}
 
 	/**
+	 * Gets Parse.ly API endpoint.
+	 *
+	 * @since 3.6.2
+	 *
+	 * @return string
+	 */
+	public function get_endpoint(): string {
+		return static::ENDPOINT;
+	}
+
+	/**
 	 * Gets the URL for a particular Parse.ly API endpoint.
 	 *
 	 * @since 3.2.0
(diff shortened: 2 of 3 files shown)


src/RemoteAPI/class-remote-api-cache.php
```diff
@@ -54,7 +54,11 @@ public function __construct( Remote_API_Base $remote_api, Cache $cache ) {
 	 *                                             response is empty.
 	 */
 	public function get_items( $query, $associative = false ) {
-		$cache_key = 'parsely_api_' . wp_hash( (string) wp_json_encode( $this->remote_api ) ) . '_' . wp_hash( (string) wp_json_encode( $query ) );
```
mehmoodak, line 57: This `wp_json_encode( $this->remote_api )` returns `{}` for all remote API's which means if the query params all same then collisions can occurs.

Updated this with API endpoints which are unique.

mehmoodak (commented):
Added comment for easier review.

acicovic:
Thank you so much for your work!

This looks good to me on first inspection. It would be worth adding tests for proving that `get_endpoint()` is unique for every endpoint.

src/RemoteAPI/class-remote-api-cache.php
```diff
@@ -54,7 +54,11 @@ public function __construct( Remote_API_Base $remote_api, Cache $cache ) {
...
+		$cache_key = (
```
acicovic, line 57: Out of curiosity, what's the reason for using parentheses for this assignment?

acicovic (approved):

src/RemoteAPI/class-remote-api-cache.php
mehmoodak: No reason. It could have been avoided.

acicovic: OK, we can leave as is for now.

PR #1401, opened 2023-02-17
Upgrade minimum PHP version to 7.2

acicovic:
## Description
After internal discussion (@mjangda, @mehmoodak, @acicovic), we have decided to upgrade the plugin's minimum PHP version to 7.2.

## Impact and compatibility

As this PR does not update any code to PHP 7.2 syntax, this change is backwards-compatible (thus we include it in a minor version upgrade). However, sites using PHP 7.1 will not be able to update to wp-parsely 3.8 automatically. The recommended solution is to upgrade to PHP 7.2 or newer.

## Motivation and context
Merge PRs	#1269 and #1270 into our codebase.

## How has this been tested?
Existing tests should pass.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -23,7 +23,6 @@ jobs:
       WP_VERSION: ${{ matrix.wp }}
 
     strategy:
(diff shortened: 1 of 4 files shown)


mjangda (approved):
✅

PR #1407, opened 2023-02-20
Show/hide Parse.ly widget and stats column based on user capabilities

### LINKED ISSUE #1406: Hide Parsely Widget and Stats if user isn't allowed to make REST API calls
## Describe the bug
Currently if user isn't allowed to make API calls then still we shows widget and Parsely Stats column.

- Widget does get shown but with an error
- Parsely Stats column also get shown to the user with actual data

## Expected behavior
Widget and Stats column shouldn't appear on the UI if there are not enough permissions

mehmoodak:
## Description
Hides UI of Parsely Widget and Stats Column based on user capability.

Closes #1406

## What was causing the issue?
For _Widget_ the root cause of the issue is that we are performing checks on proxy endpoints which does this verification when the API is actually called not while including assets and due to this UI appears even when API responds with error.

For _Stats Column_ we don't use proxy endpoint which means no checks and due to this UI appears with data.

## How this issue is being fixed?
- Moved permission checks from proxy endpoints to actual remote endpoints.
  - Uses `$is_public_endpoint` attribute and based on this we will either make the endpoint public or use the default `publish_posts` capability check (can extend this more in future)
- Proxy endpoints can define permissions on REST endpoints by utilizing the checks from base remote endpoints.

## How has this been tested?
- Ran existing tests with some changes.

### DIFF
--- a/src/Endpoints/class-analytics-post-detail-api-proxy.php
+++ b/src/Endpoints/class-analytics-post-detail-api-proxy.php
@@ -24,7 +24,7 @@ final class Analytics_Post_Detail_API_Proxy extends Base_API_Proxy {
 	 * Registers the endpoint's WP REST route.
 	 */
 	public function run(): void {
-		$this->register_endpoint( '/stats/post/detail', 'publish_posts' );
+		$this->register_endpoint( '/stats/post/detail' );
 	}
 
 	/**
--- a/src/Endpoints/class-base-api-proxy.php
+++ b/src/Endpoints/class-base-api-proxy.php
@@ -29,15 +29,6 @@ abstract class Base_API_Proxy {
 	protected $parsely;
 
 	/**
-	 * Capability of the user based on which we should allow access to endpoint.
-	 *
-	 * `null` should be used for all public endpoints.
-	 *
-	 * @var string|null
-	 */
-	protected $user_capability;
-
-	/**
 	 * Proxy object which does the actual calls to the Parse.ly API.
 	 *
 	 * @var Remote_API_Interface
@@ -72,17 +63,7 @@ abstract class Base_API_Proxy {
 	 * @return bool
 	 */
 	public function permission_callback(): bool {
-		// This endpoint does not require any capability checks.
-		if ( is_null( $this->user_capability ) ) {
-			return true;
-		}
-
-		// The user has the required capability to access this endpoint.
-		if ( current_user_can( $this->user_capability ) ) {
-			return true;
-		}
-
-		return false;
+		return $this->api->is_user_allowed_to_make_api_call();
 	}
 
 	/**
@@ -99,13 +80,9 @@ abstract class Base_API_Proxy {
 	/**
 	 * Registers the endpoint's WP REST route.
 	 *
-	 * @param string      $endpoint The endpoint's route (e.g. /stats/posts).
-	 * @param string|null $user_capability Capability of the user based on which we should allow access to endpoint.
-	 * @param bool        $show_in_index Show endpoint in /wp-json view if TRUE.
+	 * @param string $endpoint The endpoint's route (e.g. /stats/posts).
 	 */
-	protected function register_endpoint( string $endpoint, ?string $user_capability, $show_in_index = false ): void {
-		$this->user_capability = $user_capability;
-
+	protected function register_endpoint( string $endpoint ): void {
 		$filter_key = trim( str_replace( '/', '_', $endpoint ), '_' );
 		if ( ! apply_filters( 'wp_parsely_enable_' . $filter_key . '_api_proxy', true ) ) {
 			return;
@@ -131,7 +108,7 @@ abstract class Base_API_Proxy {
 				'callback'            => array( $this, 'get_items' ),
 				'permission_callback' => array( $this, 'permission_callback' ),
 				'args'                => $get_items_args,
-				'show_in_index'       => $show_in_index,
+				'show_in_index'       => $this->permission_callback(),
 			),
 		);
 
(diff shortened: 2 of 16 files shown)


src/Endpoints/class-analytics-post-detail-api-proxy.php
```diff
@@ -24,7 +24,7 @@ final class Analytics_Post_Detail_API_Proxy extends Base_API_Proxy {
 	 * Registers the endpoint's WP REST route.
 	 */
 	public function run(): void {
-		$this->register_endpoint( '/stats/post/detail', 'publish_posts' );
```
mehmoodak, line 27: Removing permission changes from proxy endpoints and we will add it in actual remote endpoints.

src/RemoteAPI/class-analytics-post-detail-api.php
```diff
@@ -20,4 +20,13 @@
...
 	protected const QUERY_FILTER = 'wp_parsely_analytics_post_detail_endpoint_args';
+
+	/**
+	 * Indicates whether the endpoint is public or protected behind permissions.
+	 *
+	 * @since 3.7.0
+	 *
+	 * @var bool
+	 */
+	protected $is_public_endpoint = false;
```
mehmoodak, line 31: In remote endpoints we are using this attribute to define the access level on each endpoint.

If `false` then it means the endpoint is private and we will use the default `publish_posts` capability to define the access ..... else the endpoint is public and will not need any checks.

tests/Integration/TestCase.php
```diff
@@ -208,15 +209,28 @@ private function create_posts_and_get_ids( int $num_of_posts = 1, $post_type = '
...
+		 */
+		$one_day_interval = date_interval_create_from_date_string( '1 days' );
 
 		for ( $i = 1; $i <= $num_of_posts; $i++ ) {
 			/**
 			 * Variable.
 			 *
 			 * @var DateTime
 			 */
-			$post_date = date_add( $date, date_interval_create_from_date_string( '1 days' ) ); // Increment by 1 day like sequence.
```
mehmoodak, line 219: PHPStan was giving error so makes the types more clear by comments.

mehmoodak (commented):

src/RemoteAPI/class-remote-api-base.php
```diff
@@ -134,4 +155,25 @@ public function get_items( $query, $associative = false ) {
 
 		return $associative ? convert_to_associative_array( $response ) : $response;
 	}
+
+	/**
+	 * Check if user is capable for making API call.
```
acicovic: I think this occurrence (and additional occurrences) can be improved to:

```suggestion
	 * Checks if the current user is allowed to make the API call.
```

src/Endpoints/class-base-api-proxy.php
```diff
@@ -72,17 +63,7 @@ abstract public function get_items( WP_REST_Request $request );
 	 * @return bool
 	 */
 	public function permission_callback(): bool {
```
acicovic, line 65: Question: Now that this is done, do we need the `permission_callback()` function, or can we just call `is_user_allowed_to_make_api_call()` directly?

Note: We can leave this as-is if we think it is clearer that way.

src/UI/class-admin-columns-parsely-stats.php
```diff
@@ -87,6 +94,13 @@ public function run(): void {
 			return;
 		}
 
+		$this->analytics_api = new Analytics_Posts_API( $this->parsely );
+
+		// Avoid adding column if user is not allowed to make API call.
```
acicovic:
```suggestion
		// Don't add the column if the user is not allowed to make the API call.
```

src/content-helper/dashboard-widget/class-dashboard-widget.php
```diff
@@ -28,6 +29,13 @@ class Dashboard_Widget {
 	 * @since 3.7.0
 	 */
 	public function run(): void {
+		$posts_api = new Analytics_Posts_API( $GLOBALS['parsely'] );
+
+		// Avoid adding widget if user is not allowed to make API call.
```
acicovic:
```suggestion
		// Don't add the widget if the user is not allowed to make the API call.
```

src/RemoteAPI/class-remote-api-base.php
```diff
@@ -34,21 +34,42 @@ abstract class Remote_API_Base implements Remote_API_Interface {
...
+
 	/**
 	 * Parsely Instance.
 	 *
 	 * @var Parsely
 	 */
 	private $parsely;
 
+	/**
+	 * Capability of the user based on which we should allow access to endpoint.
```
acicovic:
```suggestion
	 * User capability based on which we should allow access to the endpoint.
```

acicovic (commented):
Looks good, left some minor comments.

acicovic:
Note: the title of this PR should be adjusted to reflect what it really does as the PR title is being used in our changelogs, and we also need it to be correct for organizational/searching purposes.

src/Endpoints/class-base-api-proxy.php
mehmoodak: As far as I know we cannot call `is_user_allowed_to_make_api_call()` directly because it is inside `$api` property ..... so seems like keeping the `permission_callback` is fine and clear.

mehmoodak:
> Note: the title of this PR should be adjusted to reflect what it really does as the PR title is being used in our changelogs, and we also need it to be correct for organizational/searching purposes.

Done please have a look again. Thanks

acicovic (approved):

PR #1416, opened 2023-02-22
Add shell script for updating the wp-parsely version number

acicovic:
## Description
With this PR we're adding a shell script that updates wp-parsely's version number strings in all our codebase easily.

The script will create a new branch and commit the changes there. The only thing after that is inspect the correctness of the changes, and push the branch to GitHub.

## How to run
While being in the plugin directory, run `bin/uvn.sh <version>` (e.g. `bin/uvn.sh 3.7.0`).

## Motivation and context
Update version numbers in our code faster.

## How has this been tested?
Local testing, only tested on macOS sed which requires the `''` after `sed -i`.

### DIFF
--- a/bin/uvn.sh
+++ b/bin/uvn.sh
@@ -1,8 +1,11 @@
 #!/usr/bin/env bash
 
-# Script to update version numbers (uvn) in the code.
+# Script to update version numbers (uvn) in the code. It will create a new
+# branch and commit the changes. After inspecting the results, it can be pushed
+# to GitHub as a PR.
+#
 # Usage: Specify the new version as an argument. (e.g. bin/uvn.sh 3.7.0)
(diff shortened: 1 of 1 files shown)


bin/uvn.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+# Script to update version numbers (uvn) in the code. It will create a new
```
mehmoodak, line 3: I think we should update file name from `uvn.sh` to `update_plugin_version_number.sh` which is more descriptive.

mehmoodak (approved):
Left a tiny comment but approving the PR.

bin/uvn.sh
acicovic: Merging this as-is, but noting that the name will probably be revised.

PR #1417, opened 2023-02-22
Add filters to configure user capability on remote APIs

### LINKED ISSUE #1415: Use WP filter hooks to configure user capability on remote apis
As of now we are using `publish_posts` as default capability on private remote APIs but we should allow users to configure it as follows:

- We should provide a filter by using which users can change default `publish_posts` capability on all remote APIs.
- We should provide unique filter against each remote API by using which user can configure the access on each remote API

mehmoodak:
## Description
Added following filters hooks on remote endpoints so that users can configure it as per their needs:

1. `wp_parsely_user_capability_for_all_private_apis` which changes the default user capability for all remote apis
2. `wp_parsely_user_capability_for_[ENDPOINT_FILTER_KEY]_api` which changes the user capability of each API

Closes: #1415 

## How has this been tested?
- Added integration tests.

### DIFF
--- a/src/RemoteAPI/class-analytics-post-detail-api.php
+++ b/src/RemoteAPI/class-analytics-post-detail-api.php
@@ -18,7 +18,7 @@ use Parsely\Parsely;
  * @since 3.6.0
  */
 class Analytics_Post_Detail_API extends Remote_API_Base {
-	protected const ENDPOINT     = Parsely::PUBLIC_API_BASE_URL . '/analytics/post/detail';
+	protected const ENDPOINT     = '/analytics/post/detail';
 	protected const QUERY_FILTER = 'wp_parsely_analytics_post_detail_endpoint_args';
 
 	/**
--- a/src/Utils/utils.php
+++ b/src/Utils/utils.php
@@ -238,6 +238,8 @@ function convert_to_positive_integer( string $string ): int {
  *
  * @param string $endpoint Route of the endpoint.
  *
+ * @since 3.7.0
+ *
  * @return string
  */
 function convert_endpoint_to_filter_key( string $endpoint ): string {
--- a/src/Utils/utils.php
+++ b/src/Utils/utils.php
@@ -234,7 +234,7 @@ function convert_to_positive_integer( string $string ): int {
 }
 
 /**
- * Convert endpoint to filter key by replacing `/` with `_`.
+ * Converts endpoint to filter key by replacing `/` with `_`.
  *
  * @param string $endpoint Route of the endpoint.
  *
--- a/src/Utils/utils.php
+++ b/src/Utils/utils.php
@@ -232,3 +232,14 @@ function convert_to_associative_array( $obj ) {
 function convert_to_positive_integer( string $string ): int {
 	return (int) preg_replace( '/\D/', '', $string );
 }
+
+/**
+ * Convert endpoint to filter key by replacing `/` with `_`.
+ *
+ * @param string $endpoint Route of the endpoint.
+ *
+ * @return string
+ */
+function convert_endpoint_to_filter_key( string $endpoint ): string {
+	return trim( str_replace( '/', '_', $endpoint ), '_' );
+}
--- a/tests/Integration/Endpoints/AnalyticsPostsProxyEndpointTest.php
+++ b/tests/Integration/Endpoints/AnalyticsPostsProxyEndpointTest.php
@@ -141,6 +141,56 @@ final class AnalyticsPostsProxyEndpointTest extends ProxyEndpointTest {
 	}
 
 	/**
+	 * Verifies default user capability filter.
+	 *
+	 * @covers \Parsely\Endpoints\Analytics_Posts_API_Proxy::permission_callback
+	 *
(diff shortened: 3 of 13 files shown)


src/RemoteAPI/class-analytics-post-detail-api.php
```diff
@@ -18,7 +18,7 @@
  * @since 3.6.0
  */
 class Analytics_Post_Detail_API extends Remote_API_Base {
-	protected const ENDPOINT     = Parsely::PUBLIC_API_BASE_URL . '/analytics/post/detail';
```
mehmoodak, line 21: - Removed this base URL from here so that we can change the endpoint to filter key in more cleaner way.
- Added this base URL inside `get_api_url` function of Remote_Base_API (only needed once because its a common location)

tests/Integration/Endpoints/AnalyticsPostsProxyEndpointTest.php
```diff
@@ -140,6 +140,56 @@ public function test_get_items_fails_when_api_secret_is_not_set(): void {
...
+	 *
+	 * @covers \Parsely\Endpoints\Analytics_Posts_API_Proxy::permission_callback
+	 *
+	 * @uses \Parsely\RemoteAPI\Remote_API_Base::__construct
+	 * @uses \Parsely\RemoteAPI\Remote_API_Base::is_user_allowed_to_make_api_call
+	 */
+	public function test_user_is_allowed_to_make_proxy_api_call_if_endpoint_specific_user_capability_is_changed(): void {
+		$this->login_as_contributor();
+		add_filter(
+			'wp_parsely_user_capability_for_analytics_posts_api',
```
mehmoodak, line 179: Notice that this filter is changed for each remote api.

mehmoodak (commented):

src/Utils/utils.php
```diff
@@ -232,3 +232,16 @@ function convert_to_associative_array( $obj ) {
 function convert_to_positive_integer( string $string ): int {
 	return (int) preg_replace( '/\D/', '', $string );
 }
+
+/**
+ * Convert endpoint to filter key by replacing `/` with `_`.
```
```suggestion
 * Converts endpoint to filter key by replacing `/` with `_`.
```

tests/Integration/RemoteAPI/RelatedRemoteAPITest.php
```diff
@@ -69,4 +69,24 @@ public function data_api_url() {
 			Parsely::PUBLIC_API_BASE_URL . '/related?apikey=my-key&limit=5&sort=score',
 		);
 	}
+
+	/**
+	 * Verifies that public apis does not have filter which checks user capability.
```
acicovic:
```suggestion
	 * Verifies that the endpoint does not have filters that check user capability.
```

tests/Integration/TestCase.php
```diff
@@ -445,6 +451,14 @@ public function set_admin_user( $admin_user_id = 1 ): void {
 		wp_set_current_user( $admin_user_id );
 	}
 
+	/**
+	 * Create a user with role `contributor` and login.
```
acicovic:
```suggestion
	 * Creates a user with role `contributor` and login.
```

tests/Integration/RemoteAPI/RelatedRemoteAPITest.php
```diff
@@ -69,4 +69,24 @@ public function data_api_url() {
...
+	 *
+	 * @covers \Parsely\RemoteAPI\Remote_API_Base::is_user_allowed_to_make_api_call
+	 *
+	 * @uses \Parsely\RemoteAPI\Remote_API_Base::__construct
+	 */
+	public function test_public_endpoint_does_not_have_user_capability_checks(): void {
```
acicovic:
```suggestion
	public function test_related_endpoint_does_not_have_user_capability_filters(): void {
```

acicovic (commented):
Very nice work!

I think we should have at least one test that checks that when both the default and endpoint-specific filters are set, that the endpoint-specific filter will have the priority.

Or do we have this and I missed it?

mehmoodak:
> I think we should have at least one test that checks that when both the default and endpoint-specific filters are set, that the endpoint-specific filter will have the priority.

Added this test and incorporated other feedback. Please have a look again. Thanks

tests/Integration/RemoteAPI/AnalyticsPostsRemoteAPITest.php
```diff
@@ -39,4 +39,75 @@ public function data_api_url() {
...
+			}
+		);
+
+		$api = new Analytics_Posts_API( new Parsely() );
+
+		self::assertTrue( $api->is_user_allowed_to_make_api_call() );
+	}
+
+	/**
+	 * Verifies that endpoint specific user capability filter have more priority than default.
```
acicovic:
```suggestion
	 * Verifies that the endpoint specific user capability filter has more priority than the default capability filter.
```

acicovic (approved):
I think we're good, but left a comment improvement that I think is clearer.

PR #1419, opened 2023-02-27
Simplify the wp-parsely release process

### LINKED ISSUE #1156: Simplify the wp-parsely release process
A few months ago we fully automated our release process using GitHub Workflows. However, this has proven to present some issues, with the most notable ones being:

- We are not currently taking advantage of the full automation, since it would require exceedingly relaxed branch restrictions or other workarounds that we do not wish to allow or implement.
- Some times and for unknown reasons, not all PRs get listed in the automated release draft. As such, manual inspection is still needed. This is especially painful and time consuming for the dozens of dependency updates that we have in every release, especially when there is a large time gap between releases. Currently, the majority of time for releases is spent on checking dependency PRs.
- Our release-related materials reflect the fully automated release workflow, presenting inaccurate data.

All this makes the current release process clunky and more complicated than the original process, as well as incomprehensible for any new contributors.

We should aim for the maximum simplicity and clarity possible for the release process. Some related steps would be:

- [x] Remove the unused parts of the automation from our GitHub workflows.
- [x] Remove the dependency updates section from CHANGELOG.md, or if possible add an automated link pointing to this repo with the dependency updates of each release.
- [x] Possibly create a script that will update version numbers, so we don't have to do this manually on every release (although easy/quick to do). 
- [x] Update all release-related materials so they reflect the new release workflow.

This should result in a more manual but also clearer and more streamlined process.

acicovic:
## Description
This PR simplifies the steps needed to create a new wp-parsely release. It replaces our current release workflow that did not work so well for us for various reasons and never got used in its full potential.

This is a first iteration, and it could possibly be improved in the future. Even as it is, it should eliminate some pain-points we had with the previous process, the main one being incomplete changelog generation.

## Motivation and context
Closes #1156.

## How has this been tested?
This workflow has been successfully used to publish the 3.7.0 release.


bin/update-version-numbers.sh
```diff
@@ -1,10 +1,10 @@
...
 
-# Script to update version numbers (uvn) in the code. It will create a new
-# branch and commit the changes. After inspecting the results, it can be pushed
-# to GitHub as a PR.
+# Script to update version numbers in the code. It will create a new branch and
+# commit the changes.
 #
-# Usage: Specify the new version as an argument. (e.g. bin/uvn.sh 3.7.0)
+# Usage: Specify the version to update to. For example, to update to 3.7.0:
+#   `bin/update-version-numbers.sh 3.7.0)`
```
mehmoodak: I think we should change `numbers` to `number` .... also in filename.

.github/ISSUE_TEMPLATE/release-template.md
```diff
@@ -6,49 +6,45 @@ labels: 'Type: Maintenance'
...
+- [ ] [Create a PR](https://github.com/Parsely/wp-parsely/compare) that merges `develop` into `trunk`, named _Release wp-parsely x.y.z_.
+- [ ] Merge the PR into `trunk`.
+
+**3. Create the GitHub release**
+- [ ] [Create a GitHub release](https://github.com/Parsely/wp-parsely/releases/new) that targets `trunk`.
+- [ ] For the release's text, use the changelog text that was posted by the `bin/release.php` script into the PR in step 1.
+- [ ] Verify that you're using the appropriate name and tag (should be the version that is being released, without a leading 'v' character), and publish the release.
+
+**4. Deploy the release to wordpress.org**
+- [ ] Run the [Deploy to WordPress.org](https://github.com/Parsely/wp-parsely/actions/workflows/deploy.yml) GitHub workflow, selecting the new version tag.
```
mehmoodak, line 34: 👍 for adding this link .... much clear now.

bin/release.php
```diff
@@ -0,0 +1,158 @@
+<?php
+// phpcs:ignoreFile
```
mehmoodak, line 2: I have few questions regarding this file

1. Any particular reason of doing this in PHP? I think writing a bash script will be more appropriate.
2. Instead of using `phpcs:ignoreFile` I think we should ignore lines in which we are facing false error (if we are using bash script then this issue shouldn't occur)
3. Can we also add this file in PHPStan by updating configs from `phpstan.neon` (if we are not upgrading to bash script)

bin/release.php
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ * This script creates a new branch, updates version numbers and the changelog,
```
mehmoodak: Nit: `number` instead of `numbers`

bin/release.php
```diff
@@ -0,0 +1,158 @@
...
+
+// Examine and assign arguments passed to the script.
+if ( count( $argv ) < 3 ) {
+	exit(
+		'Wrong number of arguments. Should be: bin/' .
+		basename( __FILE__ ) . ' $milestone_from $milestone_to'
+	);
+}
+$milestone_from = $argv[1];
+$milestone_to   = $argv[2];
```
mehmoodak, line 34: Suggestion: Variable names `$old_version` and `$new_version` seems a bit better to me personally.

bin/release.php
```diff
@@ -0,0 +1,158 @@
...
+
+// Update and commit version numbers.
+echo "Updating and committing version numbers...\n";
+shell_exec( 'bin/update-version-numbers.sh ' . $milestone_to );
```
mehmoodak: This and above line will change if we are changing `numbers` text.

bin/release.php
```diff
@@ -0,0 +1,158 @@
...
+// phpcs:ignoreFile
+
+/**
+ * This script creates a new branch, updates version numbers and the changelog,
+ * and commits the changes into 2 separate commits. Once this is done, it also
+ * asks whether it should push those changes as a PR on GitHub. It will also
+ * include the changelog into the PR's body, so it can be easily copy/pasted
+ * into the related GitHub release.
+ *
+ * Requirements: GitHub CLI (`brew install gh`).
```
mehmoodak, line 11: It seems like we are not checking the availability of this package ..... so should we check this package at the start of the script to make sure that things go uninterrupted once script is triggered?

bin/release.php
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ * Writes the passed string to the changelog file.
+ *
+ * @param string $string The string to be written to the changelog file.
+ * @param string $milestone_from The milestone we're upgrading from.
+ * @param string $milestone_to The milestone we're upgrading to.
+ */
+function write_to_changelog_file(
+	string $string,
```
mehmoodak: Nit: `$release_log` seems more easy to understand than `$string`

bin/release.php
```diff
@@ -0,0 +1,158 @@
...
+ * @param string $milestone The milestone of the PRs that we want to fetch.
+ * @param string $label The label of the PRs that we want to fetch.
+ * @param int    $limit The maximum number of results to be returned.
+ * @return string|false|null The result of the `shell_exec` function.
+ */
+function get_pull_request_data( string $milestone, string $label, $limit = 100 ) {
+	return shell_exec(
+		'gh pr list --repo Parsely/wp-parsely --search "milestone:' . $milestone .
+		'" --label "' . $label . '"  --state merged --limit ' . $limit .
+		' --json title,number,url --jq \'.[] | "- \(.title) ([#\(.number)](\(.url)))"\''
```
mehmoodak, line 109: Although this is fine but I think having each argument on new line is more easy to read (same feedback below where we are creating PR)

mehmoodak (changes requested):
Done with 1st review. Added some feedback but since I am not much experienced with bash scripts so more 👀 here will be nicer (on second thought since we have tested the script and it is not user facing so we can also ignore more reviews .... yours call 🙂 )

bin/update-version-numbers.sh
acicovic: What do you think about renaming to `update-version` instead? Maybe even better?

bin/release.php
acicovic: > Any particular reason of doing this in PHP? I think writing a bash script will be more appropriate.

I have no experience with bash, and wanted to get this done quickly. So writing this in PHP makes this faster to code, maintain and amend, at least for me. Developers who join the project might not know bash, but they will surely know PHP. So PHP makes sense to me. It could definitely be converted to bash in the future if there's a good reason for that, but it's not something that I'm planning in this first iteration.

> Instead of using phpcs:ignoreFile I think we should ignore lines in which we are facing false error (if we are using bash script then this issue shouldn't occur)

There were _lots_ of lines with errors, so picked this approach. I don't think there is a major reason to ignore specific lines, since this is not really a part of the plugin code.

> Can we also add this file in PHPStan by updating configs from phpstan.neon (if we are not upgrading to bash script)

Good idea. Will do 🙂

bin/release.php
acicovic: The reason for picking variable names with `milestone` is that I wanted to make it as obvious as possible in the code that we're using milestones for our comparisons and data fetching. It makes it clearer IMO, as everything we do is based on milestones. A version can be a tag or a branch for example, but a milestone is something specific.

bin/update-version-numbers.sh
mehmoodak: Sounds Good.

bin/release.php
acicovic: I agree, but do not want to touch this currently 🙂

phpstan.neon
```diff
@@ -7,6 +7,7 @@ parameters:
     - wp-parsely.php
     - uninstall.php
   scanFiles:
+    - bin/release.php
```
mehmoodak: As far as I know this location is for adding PHPStan stubs so we should add it above inside `paths` .... apart from this since we could have more php scripts inside bin folder so it is better to add `bin/` in the paths instead of specific file.

mehmoodak (approved):

phpstan.neon
acicovic: Good catch. This actually revealed some PHPStan errors.

acicovic:
@mehmoodak, can you take a look at https://github.com/Parsely/wp-parsely/pull/1419/commits/6e95e3457f70742b82478a185a98a970f5598a23?

mehmoodak (approved):
Changes looks good. Thanks

PR #1423, opened 2023-02-27
Hide Parse.ly Stats column if API Secret is not available

mehmoodak:
## Description
Currently we are showing error message on Parse.ly Stats column if the API Secret is not present in settings but instead we should hide the column.

## How has this been tested?
- Updated tests to reflect this use case.

image

### DIFF
--- a/src/UI/class-admin-columns-parsely-stats.php
+++ b/src/UI/class-admin-columns-parsely-stats.php
@@ -90,7 +90,7 @@ class Admin_Columns_Parsely_Stats {
 	 * @since 3.7.0
 	 */
 	public function run(): void {
-		if ( ! $this->parsely->site_id_is_set() ) {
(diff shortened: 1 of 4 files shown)


src/UI/class-admin-columns-parsely-stats.php
```diff
@@ -90,14 +90,7 @@ public function __construct( Parsely $parsely ) {
...
 	 */
 	public function run(): void {
-		if ( ! $this->parsely->site_id_is_set() ) {
-			return;
-		}
-
-		$this->analytics_api = new Analytics_Posts_API( $this->parsely );
-
-		// Don't add the column if the user is not allowed to make the API call.
-		if ( ! $this->analytics_api->is_user_allowed_to_make_api_call() ) {
```
mehmoodak, line 100: Added API Secret check and move to a function so that we can use the same logic in tests easily.

acicovic (approved):

PR #1436, opened 2023-03-01
Update the "Track Logged-in Users" option to off by default

### LINKED ISSUE #1394: Turn the "Track Logged-in Users" option to off by default

mehmoodak:
## Description
We should not track logged in users by default because this is what most customers expect.

Closes: #1394 

## How has this been tested?
- Added tests

### DIFF
--- a/tests/Integration/OtherTest.php
+++ b/tests/Integration/OtherTest.php
@@ -226,7 +226,7 @@ final class OtherTest extends TestCase {
 		update_option( Parsely::OPTIONS_KEY, 'someinvalidvalue' );
 
 		$options = self::$parsely->get_options();
-		self::assertSame( self::EMPTY_DEFAULT_OPTIONS, $options );
+		self::assertSame( self::$parsely->get_default_options(), $options );
 	}
 
 	/**
--- a/tests/Integration/ScriptsTest.php
+++ b/tests/Integration/ScriptsTest.php
@@ -316,7 +316,7 @@ final class ScriptsTest extends TestCase {
 	}
 
 	/**
-	 * Verifies that disabling authenticated user tracking works.
+	 * Verifies that enabling authenticated user tracking works.
 	 *
 	 * @covers \Parsely\Scripts::enqueue_js_tracker
 	 * @covers \Parsely\Scripts::register_scripts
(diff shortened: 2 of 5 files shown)


tests/Integration/OtherTest.php
```diff
@@ -226,7 +226,7 @@ public function test_corrupted_options(): void {
 		update_option( Parsely::OPTIONS_KEY, 'someinvalidvalue' );
 
 		$options = self::$parsely->get_options();
-		self::assertSame( self::EMPTY_DEFAULT_OPTIONS, $options );
```
mehmoodak, line 229: We don't need the duplicated default settings in tests because we can use the one from `Parsely`

tests/Integration/ScriptsTest.php
```diff
@@ -316,7 +316,7 @@ public function test_enqueue_js_api_with_secret(): void {
 	}
 
 	/**
-	 * Verifies that disabling authenticated user tracking works.
+	 * Verifies that enabling authenticated user tracking works.
```
mehmoodak, line 319: Since we have `false` by default so performing the test with `true` value. Apart from this there is also a test with default value below.

tests/Integration/ScriptsTest.php
```diff
@@ -331,27 +331,27 @@ public function test_enqueue_js_api_with_secret(): void {
...
 			array(
 				'api_secret'                => 'hunter2',
-				'track_authenticated_users' => false,
+				'track_authenticated_users' => true,
 			)
 		);
 		$new_user_id = $this->create_test_user( 'bill_brasky' );
 		wp_set_current_user( $new_user_id );
-
+		$this->go_to_new_post();
```
mehmoodak, line 343: Test was passing for all cases so added this line to make test work as expected.

mehmoodak (commented):

tests/Integration/ScriptsTest.php
```diff
@@ -436,6 +436,45 @@ public function test_do_not_track_logged_in_users_multisite(): void {
...
+			Parsely::OPTIONS_KEY,
+			( new Parsely() )->get_default_options()
+		);
+
+		$this->set_admin_user();
+		$this->go_to_new_post();
+		self::$scripts->register_scripts();
+		self::$scripts->enqueue_js_tracker();
+
+		// As track_authenticated_users options is false by default, enqueuing should fail.
```
```suggestion
		// As track_authenticated_users option is false by default, enqueuing should fail.
```

acicovic (approved):
Looks good to me. Left a minor comment.

PR #1443, opened 2023-03-02
Fix flaky test that verifies loader script

### LINKED ISSUE #1442: Fix flaky test that verifies loader script
## Describe the bug
Our integration tests suite have become a lot flaky recently and the root cause is only one test with name `test_tracker_markup_has_attribute_when_cfasync_filter_is_used` so we should fix it for smooth CI and CD.

## To Reproduce
Run `composer testwp` multiple times.

mehmoodak:
## Description
Recently we have fixed some SonarCloud issues and in that PR we have replaced loading of files via `require` to `require_once` as suggested by SonarCloud but those changes were causing the loader script test to fail under some race conditions. The root cause of the problem is that `require_once` loads the data first time and for subsequent times it just returns `true` due to which value doesn't match and test fails.

Closes: #1442 

## How does this PR fixes the issue?
- This PR replaces `require_once` with `require` so that we load same data each time
- To avoid multiple SonarCloud issues a util function `get_asset_info` is used which we have to only ignore once in SonarCloud.

## How has this been tested?
Ran `composer testwp` couple of hundred times and verfies that no test is failing.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -15,3 +15,5 @@ parameters:
       param_type: 59
       property_type: 0 # We can't use property types until we support PHP 7.4.
       print_suggestions: false
+  typeAliases:
+    Asset_Info: 'array{ dependencies: array<string>, version: string }'
--- a/src/UI/class-admin-columns-parsely-stats.php
+++ b/src/UI/class-admin-columns-parsely-stats.php
@@ -16,6 +16,7 @@ use Parsely\RemoteAPI\Remote_API_Base;
 use Parsely\RemoteAPI\Analytics_Posts_API;
 use WP_Screen;
 
+use function Parsely\Utils\get_asset_info;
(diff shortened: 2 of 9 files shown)


phpstan.neon
```diff
@@ -15,3 +15,5 @@ parameters:
       param_type: 59
       property_type: 0 # We can't use property types until we support PHP 7.4.
       print_suggestions: false
+  typeAliases:
+    Asset_Info: 'array{ dependencies: array<string>, version: string }'
```
mehmoodak, line 19: In `PHPStan` we can either define the type on classes or globally on config files. Since we are using this type on util fuction which is not wrapped by class so therefore we define the type globally.

acicovic (approved):
Thank you for your work. Looks good to me!

PR #1454, opened 2023-03-03
Content Helper: Update Related Top Posts range to 7 days

### LINKED ISSUE #1452: Update Related Top Posts range from 3 days to 7 days
Currently on most places we pull data for 7 days from Parse.ly but in `Related Top Posts` panel we pull data for 3 days which is inconsistent.

We should pull 7 days data on `Related Top Posts` panel to make it consistent with other parts of the plugin.

mehmoodak:
## Description
Currently on most places we pull data for 7 days from Parse.ly but in Related Top Posts panel we pull data for 3 days which is inconsistent so we should make Related Top Posts panel range consitent with others.

Closes: #1452

While making changes I have noticed following things:

- We are duplicating the logic for calculating period range for API so a util function has been created to avoid this duplication.
- We are calculating period ranges in custom manner due to which we have discrepancy in `Parsely Stats Column` and `Dasboard Widget` so instead of calculating the time in custom manner we are now using `Relative Time` approach provided by Parse.ly API

## How has this been tested?
- Updated tests.


src/blocks/content-helper/performance-details/provider.ts
```diff
@@ -40,22 +37,19 @@ interface ReferrersApiResponse {
 	data: PerformanceReferrerData;
 }
 
+const PERFORMANCE_DETAILS_DEFAULT_TIME_RANGE = 7; // In days.
```
mehmoodak, line 40: Moved the limit here from constructor which is consistent with other places.

mehmoodak (commented):

src/UI/class-admin-columns-parsely-stats.php
```diff
@@ -290,8 +289,7 @@ public function get_parsely_stats_response( $analytics_api ) {
 
 		$response = $analytics_api->get_posts_analytics(
 			array(
-				'period_start'   => get_utc_date_format( - Analytics_Posts_API::ANALYTICS_API_DAYS_LIMIT ),
-				'period_end'     => get_utc_date_format(),
+				'period_start'   => Analytics_Posts_API::ANALYTICS_API_DAYS_LIMIT . 'd',
```
mehmoodak, line 292: `period_end` is by default current time so therefore just passing `period_start`

acicovic:
After some thinking, I remembered that the reason I introduced `period_end` and `period_start` fields was to allow possible custom time ranges in the future. Relative times are good, but not useful in custom ranges.

So maybe we can have those 2 coexist?

mehmoodak:
As of now all of our current cases can be tackled by relative time which is more simpler or there are lot of chances that for future cases also we only going to need relative time. So for me it is fine to keep the simple code for now and for future we can get this custom logic back if needed.

danielabloch:
> 

custom time ranges would be a very cool capability to include in PCH.

mehmoodak:
So for now we are using relative time but I did keep custom date functions in case we have to use those in future.

src/blocks/shared/utils/api.ts
```diff
@@ -0,0 +1,34 @@
+export interface AnalyticsApiQueryParams extends AnalyticsApiOptionalQueryParams {
```
acicovic, line 1: I don't see us using this somewhere. Do we plan on using this in the future?

src/blocks/shared/utils/api.ts
```diff
@@ -0,0 +1,34 @@
...
+	type: string,
+}
+
+export interface AnalyticsApiOptionalQueryParams extends ApiPeriodRange {
```
acicovic, line 5: I'm unsure we need `Optional` in the name?

src/blocks/shared/utils/api.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export interface ApiPeriodRange {
+	period_start?: string; // Defaults to 3 days ago.
+	period_end?: string; // Defaults to current date and time.
+}
+
+/**
+ * Gets `period_start` and `period_end` params for API.
+ *
+ * @param {number} days No. of days by which we have to calculate period range.
```
acicovic:
```suggestion
 * @param {number} days Number of days for which to calculate the period range.
```

acicovic (commented):
I've not tested this yet, but leaving some minor comments.

src/blocks/shared/utils/api.ts
mehmoodak: In future maybe ..... I think we should have a common definition of API interfaces in the codebase so in other places where we are defining local interfaces we should use these common interfaces.

It is not used but I have kept it here just to match it with the API specs i.e. just the interface is enough to get full idea of API params.

src/blocks/shared/utils/api.ts
mehmoodak: In Parsely API docs we have separate section for Optional params so just follows that and it is also making things clear just from the name.

acicovic (approved):

PR #1455, opened 2023-03-03
Fix @wordpress/eslint-plugin v14 warnings

### LINKED ISSUE #1444: Fix #1441 warnings
#1441 is showing some warnings, and we would like to fix them. 

Fix the warnings and merge #1441. This can also be done in the inverse order if needed.

Just ignoring the warnings, but they introduce noise (so this is not an optimal solution).

mehmoodak:
## Description
Upgrading eslint to v14 is generating some warning so this PR fixes those.

Closes: #1444 

## How has this been tested?
- Tested manually and make sure that existing test suite is passing.


src/blocks/content-helper/performance-details/component.tsx
```diff
@@ -30,9 +30,10 @@ function PerformanceDetails() {
 	const [ loading, setLoading ] = useState<boolean>( true );
 	const [ error, setError ] = useState<ContentHelperError>();
 	const [ postDetailsData, setPostDetails ] = useState<PerformanceData>();
-	const provider = new PerformanceDetailsProvider();
```
mehmoodak, line 35: Instead of passing this as dependency moved it inside `useEffect`

src/blocks/recommendations/components/parsely-recommendations-inspector-controls.tsx
```diff
@@ -40,23 +39,23 @@ const ParselyRecommendationsInspectorControls = ( {
 				<TextControl
 					label={ __( 'Title', 'wp-parsely' ) }
 					value={ title }
-					onChange={ useCallback( ( value: string ): void => setAttributes( { title: value } ), [ title ] ) }
```
mehmoodak, line 43: We added these `useCallback` while implementing the feedback of SonarCloud but after the upgrade we have to remove it because eslint show warnings that these are extra and SonarCloud also don't show any issue on these now.

mehmoodak (commented):

acicovic (approved):

Issue #1461, opened 2023-03-06
Fix Warnings


mehmoodak:
## Describe the bug
In some cases plugin is generating following warnings which we should fix:

```
Undefined property: WP_Post_Type::$post_type
```

```
Undefined property: stdClass::$post_type
```

```
Undefined array key "display_author"
```

```
Undefined array key "personalize_results"
```

acicovic:
Related: #1445.

Issue #1460, opened 2023-03-06
Fix fatal errors


mehmoodak:
## Describe the bug
In some cases plugin is generating following fatal errors:

```
Uncaught TypeError: Parsely\Utils\get_int_query_var(): Return value must be of type int, string returned
```

```
Uncaught TypeError: Parsely\UI\Recommended_Widget::get_api_url(): Argument #2 ($published_within) must be of type ?int, string given, called
```

PR #1477, opened 2023-03-08
Fix some fatal errors and warnings

### LINKED ISSUE #1445: Fix admin bar warnings in Custom Post Type Archives
## Describe the bug
The admin bar produces some warnings on Custom Post Type Archives.

## To Reproduce
1. Create a CPT.
2. Visit the CPT's Archive page.
3. PHP errors should appear in the log.

## Expected behavior
No warnings should occur.

## Screenshots
image

acicovic:
## Description
This PR fixes all fatal errors we had in the last 15 days, as well as some warnings.

For some other warnings, I'll be keeping an eye to monitor the situation.

One of the issues that prompted this PR mentions a couple of warnings in the Recommended Widget. These have not occurred in the last 15 days and have been ignored since the widget is to be removed in the future and it has very low usage.

This PR closes #1445, #1460 and #1461.

## Motivation and context
We don't want errors!

## How has this been tested?
Existing tests pass.

### DIFF
--- a/src/Endpoints/class-analytics-posts-api-proxy.php
+++ b/src/Endpoints/class-analytics-posts-api-proxy.php
@@ -55,7 +55,7 @@ final class Analytics_Posts_API_Proxy extends Base_API_Proxy {
 				return (object) array(
 					'author'         => $item->author,
 					'dashUrl'        => Parsely::get_dash_url( $site_id, $item->url ),
-					'date'           => wp_date( $date_format, strtotime( $item->pub_date ) ),
+					'date'           => $item->pub_date ? wp_date( $date_format, strtotime( $item->pub_date ) ) : '',
 					// Unique ID (can be replaced by Parse.ly API ID if it becomes available).
 					'id'             => $item->url,
 					// WordPress Post ID (0 if the post cannot be found, might not be unique).
--- a/src/UI/class-admin-bar.php
+++ b/src/UI/class-admin-bar.php
@@ -73,7 +73,7 @@ final class Admin_Bar {
 		 */
 		$current_object = $GLOBALS['wp_the_query']->get_queried_object();
 
-		if ( null === $current_object || '' === $current_object->post_type ) {
+		if ( null === $current_object || null === $current_object->post_type || '' === $current_object->post_type ) {
 			return;
 		}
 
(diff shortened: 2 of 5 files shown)


src/Endpoints/class-analytics-posts-api-proxy.php
```diff
@@ -55,7 +55,7 @@ static function( stdClass $item ) use ( $date_format, $site_id ) {
 				return (object) array(
 					'author'         => $item->author,
 					'dashUrl'        => Parsely::get_dash_url( $site_id, $item->url ),
-					'date'           => wp_date( $date_format, strtotime( $item->pub_date ) ),
+					'date'           => $item->pub_date ? wp_date( $date_format, strtotime( $item->pub_date ) ) : '',
```
acicovic, line 58: `$item->pub_date` could be null, resulting in fatals because the `strtotime()` function would receive a null argument.

src/UI/class-admin-bar.php
```diff
@@ -73,7 +73,7 @@ public function admin_bar_parsely_stats_button( WP_Admin_Bar $admin_bar ): void
 		 */
 		$current_object = $GLOBALS['wp_the_query']->get_queried_object();
 
-		if ( null === $current_object || '' === $current_object->post_type ) {
+		if ( null === $current_object || null === $current_object->post_type || '' === $current_object->post_type ) {
```
acicovic, line 76: We were continuing even if `$current_object->post_type` was null, which was generating warnings down the line.

src/UI/class-recommended-widget.php
```diff
@@ -143,7 +143,7 @@ public function widget( $args, $widget_settings ): void /* @phpstan-ignore-line
 		// Set up the variables.
 		$api_url = $this->get_api_url(
 			$this->parsely->get_site_id(),
-			$instance['published_within'],
+			(int) $instance['published_within'], // @phpstan-ignore-line
```
acicovic, line 146: Quick fix for a fatal in the Recommended Widget, which is to be dropped in a future release.

src/Utils/utils.php
```diff
@@ -123,7 +123,7 @@ function get_int_query_var( $var ): int {
 	 *
 	 * @var int
 	 */
-	return get_query_var( $var );
+	return get_query_var( $var, 0 );
```
acicovic, line 126: The function was supposed to return an int, but the `get_query_var()` function returns an empty string by default, which would result in a fatal.

mehmoodak (approved):
Thanks for quickly fixing these warnings and errors.

PR #1484, opened 2023-03-13
Fix Content Helper breakages due to null pub_date in API

acicovic:
## Description
We recently discovered that the Parse.ly API's `pub_date` field can be `null`. This was breaking parts of the PCH. Instead of breaking, the PCH will now display a `Date N/A` message in these cases.

## Motivation and context
Prevent PCH from breaking.

## How has this been tested?
- Existing tests pass
- Added some related unit tests


src/Endpoints/class-analytics-posts-api-proxy.php
```diff
@@ -55,7 +55,7 @@ static function( stdClass $item ) use ( $date_format, $site_id ) {
 				return (object) array(
 					'author'         => $item->author,
 					'dashUrl'        => Parsely::get_dash_url( $site_id, $item->url ),
-					'date'           => $item->pub_date ? wp_date( $date_format, strtotime( $item->pub_date ) ) : '',
+					'date'           => $item->pub_date ? wp_date( $date_format, strtotime( $item->pub_date ) ) : null,
```
acicovic, line 58: API can return the value as `null`, so using this instead of an empty string.

src/blocks/content-helper/performance-details/component.tsx
```diff
@@ -96,9 +96,8 @@ function PerformanceDetailsSections( props: PerformanceSectionProps ) {
  * shown.
  */
 function DataPeriodSection(): JSX.Element {
-	const periodStart = getDateInUserLang(
+	const periodStart = getSmartShortDate(
```
acicovic, line 99: Using this for consistency across the PCH, so we show the year when the date isn't within the current year.

src/blocks/content-helper/related-top-posts/component-list.tsx
```diff
@@ -30,10 +29,7 @@ function RelatedTopPostList() {
 				.then( ( result ): void => {
 					const mappedPosts: RelatedTopPostData[] = result.posts.map(
 						( post: RelatedTopPostData ): RelatedTopPostData => (
-							{
-								...post,
-								date: getDateInUserLang( new Date( post.date ), SHORT_DATE_FORMAT ),
-							}
+							{ ...post }
```
acicovic: Let the final component decide how the passed date should be displayed.

src/blocks/shared/utils/date.ts
```diff
@@ -1,7 +1,33 @@
...
+	if ( 'string' === typeof date ) {
+		date = new Date( date );
+	}
+
+	// Return false if the object is not a valid Date object, or if its value is
+	// equal to the Unix Epoch.
+	return date instanceof Date && ! isNaN( +date ) && 0 !== date.getTime();
+}
 
 export function getDateInUserLang( date: Date, options: Intl.DateTimeFormatOptions ): string {
```
acicovic, line 27: With the newest changes, we aren't using this function but I have not removed it, in case we want to display dates differently in the future.

tests/js/blocks/shared/utils/date.test.tsx
```diff
@@ -0,0 +1,21 @@
+import { canProcessDate } from '../../../../../src/blocks/shared/utils/date';
+
+const basePath = 'src/blocks/shared/utils/date.ts';
+
+describe( `${ basePath } canProcessDate()`, (): void => {
+	it( 'should return true for valid dates', (): void => {
+		expect( canProcessDate( new Date() ) ).toBe( true );
+		expect( canProcessDate( new Date( '2022-11-10T04:42:03' ) ) ).toBe( true );
+		expect( canProcessDate( new Date( '2022-11-10' ) ) ).toBe( true );
+	} );
```
mehmoodak, line 12: In this case it is generally recommended to have 1 assertion per test so seems like it is better to use test.each from Jest ([https://jestjs.io/docs/26.x/api#testeachtablename-fn-timeout](https://jestjs.io/docs/26.x/api#testeachtablename-fn-timeout))

src/content-helper/dashboard-widget/top-posts/component-list.tsx
```diff
@@ -31,12 +30,7 @@ function TopPostList() {
 			provider.getTopPosts()
 				.then( ( result ): void => {
 					const mappedPosts: TopPostData[] = result.map(
-						( post: TopPostData ): TopPostData => (
-							{
-								...post,
-								date: getDateInUserLang( new Date( post.date ), SHORT_DATE_FORMAT ),
-							}
-						)
+						( post: TopPostData ): TopPostData => ( { ...post } )
```
mehmoodak: Is this mapping needed now? seems like `result` already have the required structure.

src/blocks/content-helper/related-top-posts/component-list.tsx
mehmoodak: Is this mapping needed now? seems like `result` already have the required structure.

src/blocks/shared/utils/date.ts
```diff
@@ -1,7 +1,33 @@
...
+ * @return {boolean} Whether the date can be processed further.
+ */
+export function canProcessDate( date: Date|string ): boolean {
+	if ( 'string' === typeof date ) {
+		date = new Date( date );
+	}
+
+	// Return false if the object is not a valid Date object, or if its value is
+	// equal to the Unix Epoch.
+	return date instanceof Date && ! isNaN( +date ) && 0 !== date.getTime();
```
mehmoodak: What is the purpose of this expression `0 !== date.getTime()`? .... maybe we should add the comment to make it clear.

mehmoodak (approved):
Added few comments but approving the PR.

src/blocks/shared/utils/date.ts
acicovic: `0 === date.getTime()` would mean that the passed date is the Unix Epoch date, as mentioned in the second part of the comment above. We could break the expression into 2 vars (`isValidDateObject` and `isUnixEpoch`) if you think it is needed, but I find that the comment is adequately explanatory.

src/content-helper/dashboard-widget/top-posts/component-list.tsx
acicovic: Good catch!

tests/js/blocks/shared/utils/date.test.tsx
acicovic: Thanks, I wasn't aware of this. The link to the docs BTW was very helpful.

src/blocks/shared/utils/date.ts
mehmoodak: Recommendation of breaking the expression seems like a more easy to understand approach but I will leave the decision of adding it or not to you :)

acicovic: Ah, just saw your comment. Let me do this before you re-review.

acicovic: Done.

mehmoodak (approved):

PR #1492, opened 2023-03-15
Fix NumberFormatter constructor failures

acicovic:
## Description
We had a report of the following fatal error occurring when using the PCH Stats Column:

```
Fatal error: Uncaught Exception: Constructor failed
in /mysitepath.com/wp-content/plugins/wp-parsely/src/Utils/utils.php on line 170

Call stack:

NumberFormatter::__construct()
wp-content/plugins/wp-parsely/src/Utils/utils.php:170
...
```

The point of interest is that we were passing the locale `en_US` in the constructor. After some research, it seems that some installations could be considering this value invalid, and preferring `en-US` instead. However, we had no guarantee that replacing `en_US` to `en-US` wouldn't be problematic for other installations.

As such, we replaced the offending value to `en` which should be globally acceptable, and yields the same result.

## Motivation and context
Fix bugs.

## How has this been tested?
Existing related tests pass.

### DIFF
--- a/src/Utils/utils.php
+++ b/src/Utils/utils.php
@@ -149,7 +149,7 @@ function get_time_format(): string {
  * @return string
  */
 function get_formatted_number( $number ): string {
-	$number_formatter = new NumberFormatter( 'en_US', NumberFormatter::PADDING_POSITION );
+	$number_formatter = new NumberFormatter( 'en', NumberFormatter::PADDING_POSITION );
 	$formatted_number = $number_formatter->format( $number );
 
 	if ( false === $formatted_number ) {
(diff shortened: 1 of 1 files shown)


mehmoodak (approved):
Change looks good but I think we should either verify it on our own or confirm it from customer that things are working fine after this fix and then go ahead with the merge.

acicovic:
I agree, but don't know it is possible. @arhine, are you able to test this before we merge it? You would need to make the 2 modifications to the installed plugin.

acicovic:
We've got no response so far about this, and we're about to make a new release. I assume that this change cannot be worse than what we already have, so I'll be merging this soon so we can include it in 3.8.4.

Issue #1338, opened 2023-01-31
Add hooks to cater for platform incompatibilities and UX tailoring


acicovic:
## Is your feature request related to a problem?
Our plugin is being used in platforms where:
- Some of our features may be incompatible, causing issues or breakages.
- The user experience (UX) may need to be tailored.

## Describe the solution you'd like
- Document which features can cause issues/breakage for the platforms in question.
- Introduce hooks so the specific features can be disabled.
- Document how user experience must be tailored for every platform.
- Introduce hooks that will allow this custom experience to happen.

acicovic:
For the platforms and users that need to disable the Content Helper in whole or in part, here are the hooks I propose:

- `enable_content_helper`: When `false`, disables all PCH features.
- `enable_content_helper_dashboard_widget`: When `false`, disables the PCH Dashboard Widget.
- `enable_content_helper_post_list_column`: When `false`, disables the PCH column in Post/CPT listings.
- `enable_content_helper_editor_sidebar`: When `false`, disables the PCH Sidebar within Gutenberg.

All the above will be `true` by default.

@mehmoodak, @mjangda: ideas, comments, objections?

mjangda:
Maybe we should change `enable_content_helper_sidebar` to `enable_content_helper_editor_sidebar` instead? Specifically, including `editor` provides some context around where the sidebar is, similar to the other two features (dashboard, post_list).

acicovic:
Yes, good idea. I've updated my first post with this suggestion.

mehmoodak:
> Some of our features may be incompatible, causing issues or breakages.

Although the hooks looks good to me but can you please explain a bit why we needed these hooks ..... is it just a safe guard if some feature is causing breakage or there is some other reason for it?

acicovic:
> please explain a bit why we needed these hooks

The main reasons would be:
- I've heard of at least one Parse.ly customer who would want to disable the PCH (I'm unsure partly or totally).
- When the plugin is used in platforms that are not self-hosted installations, there may be incompatibilities that make the plugin crash, the plaform may want to tailor the experience of the plugin, or it may want to disable the features in certain sites.

GaryJones:
[...]

PR #1539, opened 2023-04-06
Add filters for disabling Content Helper features

acicovic:
## Description
This PR introduces the Content Helper filters that were first discussed in #1338. By default, all features are enabled. Filters evaluate to `false` if their value is different than `true` and `null` (default value which means that a filter has not been explicitly set).

Here are the filters which were added:

- `wp_parsely_enable_content_helper`
- `wp_parsely_enable_content_helper_dashboard_widget`
- `wp_parsely_enable_content_helper_stats_column`
- `wp_parsely_enable_content_helper_editor_sidebar`

In the code, `wp_parsely_enable_content_helper` is referred to as the `global filter`, while others are referred to as `feature filters`. To make things work, we're introducing an abstract class called `Content_Helper_Feature`, which all Content Helper features extend from.

## Usage examples
```php
// Disable the Content Helper (global filter).
add_filter( 'wp_parsely_enable_content_helper', '__return_false' );

// Disable all Content Helper features except for the Dashboard Widget.
add_filter( 'wp_parsely_enable_content_helper', '__return_false' );
add_filter( 'wp_parsely_enable_content_helper_dashboard_widget', '__return_true' );

// Disable the Content Helper Editor Sidebar only.
add_filter( 'wp_parsely_enable_content_helper_editor_sidebar', '__return_false' );
```

## Motivation and context
Sometimes, disabling the Content Helper in part or in whole might be needed.

## How has this been tested?
- Manually tested (usage examples were part of the testing).
- Existing tests pass.
- New integration tests have been written to ensure that asset loading is working correctly.


src/content-helper/class-content-helper-feature.php
```diff
@@ -0,0 +1,106 @@
...
+	 * @return bool Whether the feature is enabled by filters.
+	 */
+	protected function is_enabled_by_filters(): bool {
+		// Get filter values.
+		$global  = apply_filters( self::get_global_filter_name(), null ); // phpcs:ignore
+		$feature = apply_filters( static::get_feature_filter_name(), null ); // phpcs:ignore
+
+		// If not set, the feature filter will get its value from the global
+		// filter.
+		$global_filter_is_false = null !== $global && true !== $global;
```
mehmoodak, line 86: Instead of dealing with null, true and any other value I think we should just deal with true/false to keep logic simple.

src/content-helper/class-content-helper-feature.php
```diff
@@ -0,0 +1,106 @@
...
+	abstract public function run(): void;
+
+	/**
+	 * Returns whether the feature should be enabled.
+	 *
+	 * @since 3.9.0
+	 *
+	 * @return bool Whether the feature should be enabled.
+	 */
+	abstract public function should_be_enabled(): bool;
```
mehmoodak: For me naming of `should_be_enabled` and `is_enabled_by_filters` seems confusing. Maybe we should improve the naming here for better understanding my suggestions are:

- Rename `should_be_enabled` to `is_feature_enabling_allowed`
- Rename `is_enabled_by_filters` to `is_feature_enabled`

mehmoodak: Apart from this I think we should call `is_feature_enabling_allowed` ourselves inside `is_feature_enabled` ..... so that there will be only one function to deliver the info whether the feature is enabled or not.

Example: [here](https://github.com/Parsely/wp-parsely/pull/1539/files#diff-64a195c30e4b4274fcb559a839a6ce099b66c914c809d565fd709a01b17bd686R92) we are dealing with two function while only using one function will make things a bit simple.

tests/Integration/ContentHelperFeatureTest.php
```diff
@@ -0,0 +1,55 @@
...
+	/**
+	 * Sets the global and feature filters according to the passed values.
+	 *
+	 * Note: Passing null for a filter will not set it.
+	 *
+	 * @since 3.9.0
+	 *
+	 * @param string $feature_filter The feature filter to be set.
+	 * @param mixed  $global_filter_value The global filter value.
+	 * @param mixed  $feature_filter_value The feature filter value.
```
mehmoodak, line 49: I think it is better to improve typing and only allow `true` or `false`.

tests/Integration/UI/AdminColumnsParselyStatsTest.php
```diff
@@ -64,6 +65,382 @@ public function tear_down(): void {
...
+	 * @covers \Parsely\UI\Admin_Columns_Parsely_Stats::set_current_screen
+	 * @covers \Parsely\UI\Admin_Columns_Parsely_Stats::is_tracked_as_post_type
+	 * @covers \Parsely\UI\Admin_Columns_Parsely_Stats::enqueue_parsely_stats_script_with_data
+	 * @covers \Parsely\UI\Admin_Columns_Parsely_Stats::get_script_id
+	 *
+	 * @group content-helper
+	 * @group content-helper-admin-stats-column
+	 */
+	public function test_assets_get_enqueued_by_default(): void {
+		self::assert_enqueued_status( null, null, true );
```
mehmoodak: I think we shouldn't add tests for `true` cases to reduce or simplify tests because existing tests already cover those scenarios i.e. if script is there or api is getting called then it means feature is enabled.

tests/Integration/UI/AdminColumnsParselyStatsTest.php
```diff
@@ -64,6 +65,382 @@ public function tear_down(): void {
...
+	 * @covers \Parsely\UI\Admin_Columns_Parsely_Stats::set_current_screen
+	 * @covers \Parsely\UI\Admin_Columns_Parsely_Stats::is_tracked_as_post_type
+	 * @covers \Parsely\UI\Admin_Columns_Parsely_Stats::enqueue_parsely_stats_script_with_data
+	 * @covers \Parsely\UI\Admin_Columns_Parsely_Stats::get_script_id
+	 *
+	 * @group content-helper
+	 * @group content-helper-admin-stats-column
+	 */
+	public function test_assets_get_enqueued_when_global_filter_is_true(): void {
+		self::assert_enqueued_status( true, null, true );
```
mehmoodak: Same as above

tests/Integration/UI/AdminColumnsParselyStatsTest.php
```diff
@@ -64,6 +65,382 @@ public function tear_down(): void {
...
+	 * @covers \Parsely\UI\Admin_Columns_Parsely_Stats::is_tracked_as_post_type
+	 * @covers \Parsely\UI\Admin_Columns_Parsely_Stats::enqueue_parsely_stats_script_with_data
+	 * @covers \Parsely\UI\Admin_Columns_Parsely_Stats::get_script_id
+	 *
+	 * @group content-helper
+	 * @group content-helper-admin-stats-column
+	 */
+	public function test_assets_get_enqueued_when_global_filter_is_invalid_and_feature_filter_is_true(): void {
+		self::assert_enqueued_status( 'string', true, true );
+	}
```
mehmoodak, line 158: Currently these tests are almost identical so maybe we should only add it once. My suggestion is to add these tests only once inside `ContentHelperFeatureTest.php` file and make sure that these tests are running for each feature.

tests/Integration/content-helper/ContentHelperEditorSidebarTest.php
```diff
@@ -0,0 +1,332 @@
...
+ * Integration Tests for the PCH Editor Sidebar.
+ */
+final class ContentHelperEditorSidebarTest extends ContentHelperFeatureTest {
+	/**
+	 * Teardown method called after each test.
+	 */
+	public function tear_down(): void {
+		parent::tear_down();
+		wp_dequeue_script( Content_Helper::get_script_id() );
+		wp_dequeue_style( Content_Helper::get_style_id() );
```
mehmoodak: Any particular reason why we haven't done this for `AdminColumnsParselyStatsTest` and `ContentHelperDashboardWidgetTest`?

src/content-helper/dashboard-widget/class-dashboard-widget.php
```diff
@@ -59,29 +122,28 @@ public function add_dashboard_widget(): void {
...
-			);
-
-			wp_enqueue_style(
-				'wp-parsely-dashboard-widget',
-				$built_assets_url . 'dashboard-widget.css',
-				array(),
-				$asset_php['version']
-			);
+	public function enqueue_assets(): void {
+		if ( ! $this->should_be_enabled() ) {
```
mehmoodak: We are using this function in two place i.e. inside each callback of the action ..... so can we just use it once inside `run` function?

tests/Integration/ContentHelperFeatureTest.php
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Integration Tests: Base class for all Content Helper feature tests
+ *
+ * @package Parsely\Tests
+ * @since   3.9.0
+ */
+
+declare(strict_types=1);
+
+namespace Parsely\Tests\ContentHelper;
```
mehmoodak, line 11: Should we also use Snake Case here?

mehmoodak (changes requested):
Done with first review and left some comments.

tests/Integration/ContentHelperFeatureTest.php
acicovic: I thought about this as well, but currently all our PHP test names do not use snake case. So be convention, I put it like that. We could address this in a refactoring PR.

src/content-helper/dashboard-widget/class-dashboard-widget.php
acicovic: The reason for this is explained in the [run() function](https://github.com/Parsely/wp-parsely/blob/f5db4851f6958aac2fccf044ba1d8c9aa654be0e/src/content-helper/dashboard-widget/class-dashboard-widget.php#L66-L67), but I see that the function name in the comment is wrong. I'll update it.

tests/Integration/content-helper/ContentHelperEditorSidebarTest.php
acicovic: - This is not needed in `AdminColumnsParselyStatsTest` since some functions already address it.
- In `ContentHelperDashboardWidgetTest`, this is taken care of within the `assert_enqueued_status()` function [here](https://github.com/Parsely/wp-parsely/blob/f5db4851f6958aac2fccf044ba1d8c9aa654be0e/tests/Integration/content-helper/ContentHelperDashboardWidgetTest.php#L51-L55).

I'll remove the block you refer to and implement it into the `assert_enqueued_status()` function of `ContentHelperEditorSidebarTest` in order to follow a common approach with `ContentHelperDashboardWidgetTest`.

tests/Integration/UI/AdminColumnsParselyStatsTest.php
acicovic: This is the initial approach I took but I concluded that it does not make things simpler or more concise, as I ended up with more lines of code, tighter coupling and slightly more complexity.

The reason is that we still need to create a function for every test in the child class to add the `covers` DocBlocks there (because they are different for every child class), as well as call the parent testing function from there.

So unless you had a specific methodology in mind to address this, for me the current approach is fine.

tests/Integration/UI/AdminColumnsParselyStatsTest.php
acicovic: The `test_assets_*` functions specifically test asset loading in relation to the value of global/feature filters. I agree that there is some overlap with existing Parse.ly Stats tests, but this is not exactly what the existing tests are testing (though I'm sure some of them would fail as a side-effect in case of an issue).

We do introduce some extra code to have these new tests, but having more tests and aiming for testing a broader set of values (even if today they are being defaults or assumed to be a certain value) is a good thing in my opinion, especially as it will be clear from the tests where the issue is exactly.

src/content-helper/class-content-helper-feature.php
acicovic: I think that there are advantages in the current implementation. For example, see the second usage example in the PR's description. That example turns off existing and future Content Helper features. A true/false only solution would need more filters set and would require manual filtering of future additions.

Also, though we would want the user to pass a boolean value, the user can pass invalid values when using the filter, and we would still need to convert that to a boolean somehow. I do not think we should throw exceptions if the value is not a boolean.

So I'm unsure jumping to a pure true/false solution has real advantages, and the current function code (2 `if`s) is not complicated in my opinion. @mjangda, happy to hear your opinion about this in case you would also prefer a stricter true/false approach.

tests/Integration/ContentHelperFeatureTest.php
acicovic: Loose typing here is to allow testing invalid values that could be passed by the user to the filters. If we revise our approach, we can see if this needs any change.

src/content-helper/class-content-helper-feature.php
acicovic: I'm happy to look into whether we can reduce this into a single function. As long as we have 2, it is more probable to be confusing.

acicovic: I've pushed commit https://github.com/Parsely/wp-parsely/pull/1539/commits/bbc4912ea1442e32d35280d533f307c310ff5ccc which uses a [single function](https://github.com/Parsely/wp-parsely/blob/857f1c684f5d04a94cfdf908899b9dde64ce3d3c/src/content-helper/class-content-helper-feature.php#L79-L103) everywhere (except in the widget for convenience purposes). This is simpler and flexible as it will allow for advanced enablement logic that we might have in the future for newer features.

src/content-helper/class-content-helper-feature.php
acicovic: My recommendation: Leave this as is.

tests/Integration/ContentHelperFeatureTest.php
acicovic: My recommendation: Leave this as is.

tests/Integration/UI/AdminColumnsParselyStatsTest.php
acicovic: My recommendation: Leave this as is.

tests/Integration/UI/AdminColumnsParselyStatsTest.php
acicovic: My recommendation: Leave this as is.

tests/Integration/content-helper/ContentHelperEditorSidebarTest.php
acicovic: This has been implemented.

src/content-helper/dashboard-widget/class-dashboard-widget.php
acicovic: This is done.

tests/Integration/ContentHelperFeatureTest.php
acicovic: My recommendation: Leave this as is for the moment.

tests/Integration/UI/AdminColumnsParselyStatsTest.php
mehmoodak: Currently we have `RemoteAPITest` which we use as extend class in our other tests, all tests inside `RemoteAPITest` class runs for each API seprately so what I am thinking here is that if we have move all of these duplicated tests inside `ContentHelperFeatureTest` then we can achieve same thing with less code and in future also it will save our time.

mehmoodak (approved):
Left one comment but approving the PR.

tests/Integration/UI/AdminColumnsParselyStatsTest.php
acicovic: Hey @mehmoodak, I had not thought of adding all `covers` and `uses` in the base class. I [implemented your feedback](https://github.com/Parsely/wp-parsely/pull/1539/commits/c40af6c01fed4c63f992a04c4d8a829d7d206880). Please take a look again 🙂

mehmoodak (approved):

Issue #1341, opened 2023-01-31
Content Helper: Consider removing or improving fetch retries


acicovic:
## Is your feature request related to a problem?
Within parts of the Content Helper's code, we have this where the Parse.ly API gets called:
```ts
const FETCH_RETRIES = 3 // Number of attempts to fetch the data before displaying an error.

function MyComponent() {
	useEffect( () => {
		const fetchData = async ( retries: number ) => {
			provider.getData()
				.then( ( result ) => {
					// Populate component with data.
				} )
				.catch( async ( err ) => {
					if ( retries > 0 ) {
						await new Promise( ( r ) => setTimeout( r, 500 ) );
						await fetchPosts( retries - 1 );
					} else {
						// Show the error
					}
				} );
		};

		setLoading( true );
		fetchPosts( FETCH_RETRIES );
	}, [] );
}
```
This means that if the initial fetch of the data fails, we attempt to fetch it up to two times more, and display the error if all attempts failed. This seems like a good idea but it actually results in extra API calls, that could be a significant amount at scale and with many users.

This retry mechanism has been firing a lot when it shouldn't, for example when we have wrong credentials. There are many valid scenarios where we should get an error and stop there, but this code just keeps on firing, and another side effect of this is that errors appear with an increased latency, since this code has resulted in 3 calls instead of 1.

I can propose two solutions:
1. Get rid of this logic. Having it so far, I do not think we got any real advantages out of it.
2. Flag the errors that should not be retried, as to only retry for errors that are perceived as "unknown/unexpected".

## Describe the solution you'd like
I'm opening this for discussion to see if an action should be taken, and which of the 2 approaches is preferred. Tagging @mehmoodak, @mjangda and @chriszarate for thoughts and discussion.

I'd just remove it if there are no strong advantages in keeping it (which I think is the case).

mehmoodak:
As of now I think going with Solution 1 is fine because as you already mentioned this is not giving any real advantage.

If needed we can implement it back in future.

mjangda:
On retries, I think there's benefit to having a single retry to account for network issues or unexpected issues (i.e. go from 3 => 1). In majority of the cases this shouldn't kick in anyway.

However, we should be smarter and more defensive about the API failure response. For example, if the Parsely API is down, we should catch that and stop making calls, even if it's for a short time period. To do this, we can have a counter via transients that tracks failures and after a certain threshold, we set an option which blocks API calls for 5 minutes.

[...]

PR #1541, opened 2023-04-07
Content Helper: Limit unnecessary requests to the Parse.ly API

acicovic:
## Description
Most Content Helper features employ a retrying mechanism which resends (up to 3 times) any requests to the Parse.ly API that failed to return any data.

However, in many instances this unnecessarily increases the amount of API requests. For example, if the user has a wrong API secret and gets an access forbidden response, there is no reason in repeating the request 3 more times. In addition, the user experiences the latency of 4 API calls until they see the error.

This PR adjusts the retries from 3 to 1 and introduces a boolean `retryFetch` field in our Content Helper's error object which determines whether the request should be retried. This will drastically reduce the number of requests, especially if we account for hundreds or thousands of users. It will also speed-up the appearance of any errors that do not need to be retried. All unknown errors are retried by default.

In future iterations of this improvement, we can consider taking additional measures such as those suggested [here](https://github.com/Parsely/wp-parsely/issues/1341#issuecomment-1452163887). 

## Motivation and context
- Limit unnecessary requests to the Parse.ly API.
- Bring down API usage cost per user.
- Make user experience faster.
- Work on a simple fix that will have a big impact on limiting unneeded API requests.

## How has this been tested?
- Manual testing for verifying that the correct amount of requests are fired.
- The existing structure test has been modified and passes.
- The existing an E2E tests pass, an one has been modified to work with the reduced response delays.


src/blocks/content-helper/performance-details/component.tsx
```diff
@@ -43,7 +43,7 @@ function PerformanceDetails() {
 					setLoading( false );
 				} )
 				.catch( async ( err ) => {
-					if ( retries > 0 ) {
+					if ( retries > 0 && err.retryFetch ) {
```
mehmoodak, line 46: Here the expectation is to retry or not if the error is not in `noRetryList`?

As of now it seems like we cannot retry in this case.

tests/e2e/specs/blocks/content-helper/top-bar-icon.spec.ts
```diff
@@ -21,7 +21,7 @@ const pluginButton = 'button[aria-label="Parse.ly Editor Sidebar"]';
  * Tests for the PCH Editor Sidebar top bar icon.
  */
 describe( 'PCH Editor Sidebar top bar icon in the WordPress Post Editor', () => {
-	const contentHelperTitle = 'Performance DetailsRelated Top Posts';
+	const contentHelperTitle = 'Performance DetailsThis post is not published, so its details are unavailable.Related Top Posts';
```
mehmoodak, line 24: We didn't made any changes around this so any reason why we are changing this?

tests/js/blocks/content-helper/structure.test.tsx
mehmoodak, line 1: We should also add a test somewhere to test:

- We are retrying if the error is not present in `noRetryList`
- We are not retrying if the error is present in `noRetryList`

mehmoodak (changes requested):
Done with first review and left some comments.

src/blocks/content-helper/performance-details/component.tsx
acicovic: This works as expected in my testing. Note that we're doing `this.retryFetch = ! noRetryFetchErrors.includes( this.code );`. So if the error is not in the (exclusion) array, we are retrying it.

tests/e2e/specs/blocks/content-helper/top-bar-icon.spec.ts
acicovic: When the icon is clicked, the side-panel opens and it performs API requests. Because this is done with an unpublished post, this produces an error, which would previously result in 4 API requests, and thus more delay. So under these conditions, the test would pass.

However, as latency is much smaller now (no retry), we're getting the error message faster, and thus the test fails if we do not include it.

src/blocks/content-helper/performance-details/component.tsx
mehmoodak: Okay. Got it ... we are retrying only once now.

tests/js/blocks/content-helper/structure.test.tsx
acicovic: I introduced [some unit tests](https://github.com/Parsely/wp-parsely/pull/1541/files#diff-0ef66260ff5ecced6e6dd525764987663af4b7cc49d8f55556bf48eff5a143ef) to address this.

mehmoodak (approved):

PR #1574, opened 2023-04-27
Improve settings page

acicovic:
## Description
This PR brings a number of small improvements to the settings page like:
- Some labels were slightly renamed.
- Some error messages have been updated.
- Basic validation has been added for the API Secret field.
- The Metadata Secret field was moved to the Advanced section.
- When the Site ID or Secrets fail validation they don't get saved anymore.
- The "Wipe Parse.ly Metadata Info" setting was removed.

This PR also introduces a managed credentials mode, reserved for web host usage.

## Motivation and context
Related to #1338.

## How has this been tested?
- Existing tests pass.
- Fixed some E2E tests that were failing.
- Some new integration tests were added.


src/class-parsely.php
```diff
@@ -586,4 +589,54 @@ public function get_all_track_types(): array {
...
+	 * This allows hosting providers to provide a more customized experience for
+	 * the plugin by handling credentials automatically.
+	 *
+	 * @since 3.9.0
+	 * @access private
+	 *
+	 * @return bool Whether credentials are being managed at the platform level.
+	 */
+	public static function are_credentials_managed(): bool {
+		return has_filter( 'wp_parsely_credentials' );
```
mjangda: I don't think `has_filter` is safe for use here.

It might be better for `inject_managed_credentials` to add a value that we can check instead?

```suggestion
				$options = $this->get_options();
				if ( isset( $options['managed_credentials'] ) && true === $options['managed_credentials'] ) {
					return true;
				}

				return false;
```

return isset( $this->options{
```

File: src/UI/class-settings-page.php [hunk 2]
```diff
@@ -218,62 +219,45 @@ private function initialize_basic_section(): void {
 		$field_id   = 'apikey';
 		$field_args = array(
 			'option_key'    => $field_id,
-			'help_text'     => __( 'Your Site ID is typically your own site domain without <code>http(s)://</code> prefixes or trailing <code>/</code> (e.g. <code>mydomain.com</code>).', 'wp-parsely' ),
+			'help_text'     => $are_credentials_managed ? '' : __( 'Your Site ID is typically your own site domain without <code>http(s)://</code> prefixes or trailing <code>/</code> (e.g. <code>mydomain.com</code>).', 'wp-parsely' ),
```
mehmoodak: I think it is fine to not hide this text.

src/UI/class-settings-page.php
```diff
@@ -218,62 +219,45 @@ private function initialize_basic_section(): void {
...
 			'label_for'     => $field_id,
 			'optional_args' => array(
 				'required'    => 'required',
 				'placeholder' => 'mydomain.com',
+				'disabled'    => $are_credentials_managed ? true : false,
```
mehmoodak: Since `$are_credentials_managed` is already a boolean so we can directly use it without ternary

src/UI/class-settings-page.php
```diff
@@ -930,44 +941,42 @@ public function validate_options( $input ) {
...
-		}
-
-		$input['api_secret'] = $this->get_unobfuscated_value( $input['api_secret'], $this->parsely->get_api_secret() );
 
-		$input['metadata_secret'] = $this->get_unobfuscated_value( $input['metadata_secret'], $this->parsely->get_options()['metadata_secret'] );
-		if ( '' !== $input['metadata_secret'] ) {
-			if ( strlen( $input['metadata_secret'] ) !== 10 ) {
+			$input['api_secret'] = $this->get_unobfuscated_value( $input['api_secret'], $this->parsely->get_api_secret() );
+			$api_secret_length   = strlen( $input['api_secret'] );
+			if ( $api_secret_length > 0 && $api_secret_length <= 30 ) {
```
mehmoodak, line 958: Did we discussed this validation with Parse.ly team? I think we shouldn't add this because

1. It is not a strong validation and customers can still enter invalid values.
2. If Parse.ly team changes this on their side then a change in the plugin will be needed and if not synced then it can cause issue.

src/class-parsely.php
```diff
@@ -416,10 +416,13 @@ public function get_options() {
...
 
 		if ( ! is_array( $options ) ) {
-			return $this->option_defaults;
+			$options = $this->option_defaults;
 		}
 
-		return array_merge( $this->option_defaults, $options );
+		return array_merge(
+			$this->option_defaults,
+			$this->inject_managed_credentials( $options )
```
mehmoodak: To me having `array_merge` like this seems more preferrable

```php
array_merge(
			$this->option_defaults,
			$options,
			$this->get_managed_credentials()
);
```

In above example we are not passing `$options` to the function and if we introduce dedicated types for the function then it will make it easier to understand the code specially because there is no possibility of mutation.

Same code after changes.
```php
	private function get_managed_credentials() {
		$credentials = apply_filters( 'wp_parsely_credentials', false );

		if ( ! is_array( $credentials ) ) {
			return [];
		}

		$managed_options = array();

		if ( isset( $credentials['site_id'] ) ) {
			$$managed_options['apikey'] = $credentials['site_id'];
		}

		if ( isset( $credentials['api_secret'] ) ) {
			$$managed_options['api_secret'] = $credentials['api_secret'];
		}

		if ( isset( $credentials['metadata_secret'] ) ) {
			$$managed_options['metadata_secret'] = $credentials['metadata_secret'];
		}

		return $managed_options;
	}
```

src/class-parsely.php
mehmoodak: 1. @mjangda how it is not safe 🤔 
2. Agreed that the suggested approach is better and just to be clear that we should set the boolean ourselves in the plugin and don't rely on hosting providers to set it.

tests/Integration/OtherTest.php
```diff
@@ -213,6 +213,96 @@ public function test_checking_site_id_is_set_or_not(): void {
...
+	 * @since 3.9.0
+	 *
+	 * @covers \Parsely\Parsely::are_credentials_managed
+	 * @covers \Parsely\Parsely::inject_managed_credentials
+	 * @uses \Parsely\Parsely::__construct
+	 * @uses \Parsely\Parsely::get_default_options
+	 */
+	public function test_managed_credentials_get_set_as_expected(): void {
+		// Valid and invalid credential sets to be tested.
+		$credential_sets = array(
```
mehmoodak, line 234: Here I think we should provide array as following:

```php
array(
    array(
        'input' => array(), // having credentials
        'expected' => array(), // having expected credentials
     ),
      array(
        'input' => array(), // having credentials
        'expected' => array(), // having expected credentials
     )
)
```

More changes will be that we don't need to invoke the credentials function ourselves since we can use the existing `get_options` function.

The main benefits of this approach are

1. Easier because we just have to compare the returned values with `expected` array and we can avoid this function

```php
if ( is_array( $credentials ) && 0 < count( $credentials ) ) {
				foreach ( $credentials as $key => $value ) {
					if ( 'site_id' === $key ) {
						self::assertSame( $actual_options['apikey'], $value );
					} elseif ( 'api_secret' === $key ) {
						self::assertSame( $actual_options['api_secret'], $value );
					} elseif ( 'metadata_secret' === $key ) {
						self::assertSame( $actual_options['metadata_secret'], $value );
					} else {
						// Invalid keys should not be added to the options array.
						self::assertFalse( isset( $actual_options[ $key ] ) );
					}
				}
			} else {
				// Credentials are of wrong data type or an empty array.
				self::assertSame( $default_options, $actual_options );
			}
```

2. Easier to add new cases because we don't have that function

3. We can use credentials type (if defined above) to make it easy to understand the code

tests/Integration/UI/SettingsPageTest.php
```diff
@@ -43,6 +43,274 @@ public function set_up(): void {
...
+	 * @uses \Parsely\UI\Settings_Page::validate_advanced_section
+	 * @uses \Parsely\UI\Settings_Page::validate_options
+	 * @uses \Parsely\UI\Settings_Page::validate_options_post_type_tracking
+	 * @uses \Parsely\UI\Settings_Page::validate_recrawl_section
+	 * @uses \Parsely\UI\Settings_Page::validate_site_id
+	 *
+	 * @group settings-page
+	 * @group settings-page-validation
+	 */
+	public function test_valid_site_id_values_are_retained_when_validated(): void {
```
mehmoodak, line 72: Thanks for adding these tests 👍👏

tests/Integration/UI/SettingsPageTest.php
```diff
@@ -43,6 +43,274 @@ public function set_up(): void {
...
+	 * @uses \Parsely\UI\Settings_Page::validate_site_id
+	 *
+	 * @group settings-page
+	 * @group settings-page-validation
+	 */
+	public function test_valid_api_secret_values_are_retained_when_validated(): void {
+		$options = self::$parsely->get_options();
+
+		// More than 30 characters.
+		$options['api_secret'] = 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa';
```
mehmoodak: If we are keeping the validation then a string like this seems more preferable `test_secret_for_parsely_api_access` .... maybe we should keep this in a constant and use that

tests/Integration/UI/SettingsPageTest.php
```diff
@@ -43,6 +43,274 @@ public function set_up(): void {
...
+	 * @group settings-page-validation
+	 */
+	public function test_valid_site_id_values_are_retained_when_validated(): void {
+		$options = self::$parsely->get_options();
+
+		$options['apikey'] = 'mydomain.com';
+		$expected          = $options;
+
+		$actual = self::$settings_page->validate_options( $options );
+		self::assertSame( $expected, $actual );
```
mehmoodak, line 79: Instead of comparing whole arrays I think it is better to use it like this

```php
self::assertSame( 'mydomain.com', $actual['apikey'] );
```

tests/Integration/UI/SettingsPageTest.php
```diff
@@ -43,6 +43,274 @@ public function set_up(): void {
...
+	 * @group settings-page-validation
+	 */
+	public function test_invalid_site_id_values_are_emptied_when_validated(): void {
+		$expected = self::$parsely->get_options();
+		$options  = self::$parsely->get_options();
+
+		$options['apikey'] = 'invalid';
+
+		$actual = self::$settings_page->validate_options( $options );
+		self::assertSame( $expected, $actual );
```
mehmoodak, line 115: To me the function name sounds confusing because there is a word `emptied` here but we are not asserting `''` in test ..... if we follow above suggestion then it will also make things clear here.

Same feedback for other asserts.

mehmoodak (changes requested):
Reviewed the PR and left some feedback.

src/UI/class-settings-page.php
mjangda: This was briefly discussed as a simple way to prevent mixing up the api and metadata secret. It's not a comprehensive check but would address some portion of tickets.

acicovic: Yes, this has been mentioned several times in different occasions. It is indeed quite simplistic, but it will prevent a customer from injecting shorter keys within the API Secret field (thing which has happened lately according to Parse.ly support).

Ideally in the future we can have real validation with an API call or such.

src/class-parsely.php
mjangda: has_filter tells us if a callback was attached not if the values were changed. If the callback returns early, managed mode should not be activated.

acicovic: Although I'm not sure there's an issue here, I agree that an option will be more resilient and I'm happy to go that route. I've started an internal discussion about that.

src/UI/class-settings-page.php
acicovic: Do you think it's better to have the help text? I removed it since users won't be interacting with the field, thus I deemed it has no practical value.

src/UI/class-settings-page.php
acicovic: Good catch. I'm wondering why I even did this.

acicovic: Updated.

src/class-parsely.php
acicovic: Updated.

tests/Integration/OtherTest.php
acicovic: Updated but left some asserts for options, since the functions that return the options and the managed credentials are different (getting the correct managed credentials does not mean that they get correctly merged into the options array).

tests/Integration/UI/SettingsPageTest.php
acicovic: The readability of the keys has been improved in the integration test. For E2E tests, I created a constant.

tests/Integration/UI/SettingsPageTest.php
acicovic: I agree that zooming to the specific value can be nice for readability.

However, comparing the whole array can catch other things that might have gone wrong. For example, if the function was correctly setting the value we're looking for but was mangling another value, this would not be caught when comparing the specific value.

So I prefer having a broader testing scope here, even if it's a bit more abstract. I do not think it makes things much more complicated either.

The above also applies for the review comment below.

acicovic:
@mehmoodak, I've left a question and I consider your feedback generally implemented (with the exception of the two last comments). I'll address the issue brought up by @mjangda in another commit. I'm informing you of this in case we can start closing some of the comments already.

Thanks!

src/UI/class-settings-page.php
mehmoodak: Personally to me it seems like an additional info that help me understand the field which is fine even in managed mode.

tests/Integration/UI/SettingsPageTest.php
mehmoodak: Agreed, comparing whole arrays seems like a better approach for testing.

mehmoodak:
@acicovic gone through the PR and closed most of the comments, let me know when the PR is fully ready and needs a final review. Thanks

src/UI/class-settings-page.php
acicovic: OK, done.

src/class-parsely.php
acicovic: @mehmoodak, @mjangda, I've implemented something based on our discussion.

src/class-parsely.php
```diff
@@ -586,4 +544,62 @@ public function get_all_track_types(): array {
...
+	 */
+	private function are_credentials_managed(): bool {
+		$credentials = apply_filters( 'wp_parsely_credentials', false );
+
+		if ( ! is_array( $credentials ) ) {
+			return false;
+		}
+
+		$required_credentials = array( 'site_id', 'api_secret' );
+		$is_valid             = array();
```
mehmoodak: The current implementation is fine but we can also remove this `is_valid` array and makes the code a bit simple

```php
	$credentials = apply_filters( 'wp_parsely_credentials', false );

	if ( ! is_array( $credentials ) ) {
		return false;
	}

	$required_credentials = array( 'site_id', 'api_secret' );

	foreach ( $required_credentials as $key ) {
		if (
			! isset( $credentials[ $key ] ) ||
			! is_string( $credentials[ $key ] ) ||
			'' === $credentials[ $key ]
		) {
			return false;
		}
	}

	return true;
}
```

mehmoodak (approved):

src/class-parsely.php
acicovic: OK, I've updated this in https://github.com/Parsely/wp-parsely/pull/1574/commits/a7dbe1e172e203d474931de98e75beb00e61b331

src/UI/class-settings-page.php
```diff
@@ -930,44 +926,42 @@ public function validate_options( $input ) {
...
-		}
 
-		$input['api_secret'] = $this->get_unobfuscated_value( $input['api_secret'], $this->parsely->get_api_secret() );
-
-		$input['metadata_secret'] = $this->get_unobfuscated_value( $input['metadata_secret'], $this->parsely->get_options()['metadata_secret'] );
-		if ( '' !== $input['metadata_secret'] ) {
-			if ( strlen( $input['metadata_secret'] ) !== 10 ) {
+			$input['api_secret'] = $this->get_unobfuscated_value( $input['api_secret'], $this->parsely->get_api_secret() );
+			$api_secret_length   = strlen( $input['api_secret'] );
+			if ( $api_secret_length > 0 && $api_secret_length <= 30 ) {
```
mjangda, line 958: Just to confirm, based on the logic here, the API Secret will always be 31 or more characters?

src/class-parsely.php
```diff
@@ -608,4 +566,65 @@ public function get_all_track_types(): array {
...
+
+	/**
+	 * Returns whether credentials are being managed at the platform level.
+	 *
+	 * @since 3.9.0
+	 * @access private
+	 *
+	 * @return bool Whether credentials are being managed at the platform level.
+	 */
+	private function are_credentials_managed(): bool {
```
mjangda, line 609: Minor: it might be better to:

- make the `are_credentials_managed` property `private`
- rename this method to `check_for_managed_credentials()` (or similar)
- add a new `are_credentials_managed()` method that returns the value of the property and use that outside the class

Not a blocker. More of a code cleanliness / hardening thing.

mjangda: > rename this method to check_for_managed_credentials() (or similar)

Alternate idea would be to call this `load_managed_credentials` and combine it with `get_managed_credentials`.

Specifically, after validating that the credentials are passed through, it can set `are_credentials_managed = true` and also a new property called `credentials`.

Main benefit is that we simplify things into 1-step and avoid having to maintain the filter in two places.

Again, not a blocker. Can be done later if we want to.

mjangda (approved):
LGTM.

Had 1 question and suggestion; neither are blockers.

src/UI/class-settings-page.php
acicovic: They are usually longer than the limit here. The 30 limit was added here to cater for obvious mistakes (for example pasting the metadata secret).

mjangda: wfm 👍

src/class-parsely.php
acicovic: Thanks for the feedback. I'll merge this as it is for now but also note the improvements stated here so we can work on them in another PR.

PR #1588, opened 2023-05-04
Content Helper: Improve error handling

acicovic:
## Description
This PR improves the Content Helper's error handling.
- Local endpoints are no longer being called when credentials are not set.
- The message that is displayed when credentials are not set is now customizable through the `wp_parsely_message_credentials_not_set` filter.

Some changes that were done while implementing this:
- Moved the `$parsely` internal variable to `the Content_Helper_Feature` class. Constructors were left in the child classes to allow future flexibility. All child classes now need to set `$parsely` in their constructor.
- A new JS variable `wpParselyCredentialsNotSetMessage` was introduced. This variable is being passed by WordPress to the PCH and it contains the message that is displayed when no credentials are set.
- A new `ContentHelperErrorMessage` component was introduced.
- Additional smaller changes to glue everything together and make it work.

## Motivation and context
- We had an internal request for both of those features.
- Being able to customize the error message relates to #1338.

## How has this been tested?
Fixed/updated the tests that started failing due to the changes described above.


src/@types/assets/window.d.ts
```diff
@@ -26,5 +26,6 @@ declare global {
 		wpParselyDisableAutotrack?: boolean;
 		wpParselyHooks?: _Hooks;
 		wpParselyPostsStatsResponse: string;
+		wpParselyCredentialsNotSetMessage: string;
```
mehmoodak: `wpParselyEmptyCredentialsMessage` ?

src/blocks/content-helper/content-helper-error.tsx
```diff
@@ -52,14 +66,58 @@ export class ContentHelperError extends Error {
 		Object.setPrototypeOf( this, ContentHelperError.prototype );
 	}
 
-	public ProcessedMessage( className = '' ): JSX.Element {
-		// Errors that need to display the "Contact Us" message.
-		const contactUsErrorCodes = [
+	/**
+	 * Returns an error message element that can contain HTML.
+	 *
+	 * Warning: Any HTML passed to this function should be sanitized.
```
mehmoodak: Should we sanitize the HTML in our function? ..... this will make it easy to use the function and helps us to avoid mistakes where we forget to pass the sanitized html?

src/blocks/content-helper/content-helper-error.tsx
```diff
@@ -52,14 +66,58 @@ export class ContentHelperError extends Error {
...
+	/**
+	 * Renders the error's message.
+	 *
+	 * @param {ContentHelperErrorMessageProps|null} props The props needed for the function.
+	 *
+	 * @return {JSX.Element} The resulting JSX Element.
+	 */
+	public renderMessage( props: ContentHelperErrorMessageProps|null = null ): JSX.Element {
+		// Handle cases where credentials are not set.
+		const CredentialsNotSetErrorCodes = [
```
mehmoodak, line 74: camelCase?

src/content-helper/class-content-helper-feature.php
```diff
@@ -101,4 +112,66 @@ protected function can_enable_feature( ...$conditions ): bool {
...
 	}
+
+	/**
+	 * Injects any required inline scripts.
+	 *
+	 * @since 3.9.0
+	 */
+	protected function inject_inline_scripts(): void {
+		$are_credentials_set = $this->parsely->site_id_is_set() &&
+		$this->parsely->api_secret_is_set();
```
mehmoodak: indentation.

src/content-helper/class-content-helper-feature.php
```diff
@@ -101,4 +112,66 @@ protected function can_enable_feature( ...$conditions ): bool {
...
+		// Override default message if the respective filter is set.
+		$message = apply_filters(
+			'wp_parsely_message_credentials_not_set',
+			$default_message
+		);
+
+		// Remove unnecessary whitespace to avoid broken output.
+		$message = str_replace( array( "\r", "\n", "\t" ), '', $message );
+
+		$sanitized_message = wp_kses_post( $message );
```
mehmoodak: We can directly return ... SonarCloud also complains about this.

src/blocks/content-helper/class-content-helper.php
```diff
@@ -74,6 +84,8 @@ public function run(): void {
 			true
 		);
 
+		$this->inject_inline_scripts();
```
mehmoodak, line 87: Instead of calling this function seprately should we moved it inside the constructor of `Content_Helper_Feature`?

src/content-helper/dashboard-widget/class-dashboard-widget.php
```diff
@@ -124,6 +135,8 @@ public function enqueue_assets(): void {
 			true
 		);
 
+		$this->inject_inline_scripts();
```
mehmoodak, line 138: Same as above

src/content-helper/dashboard-widget/dashboard-widget.tsx
```diff
@@ -7,12 +7,24 @@ import { render } from '@wordpress/element';
...
 	function() {
+		let element = <TopPostList />;
+
+		if ( window.wpParselyCredentialsNotSetMessage ) {
+			const error = new ContentHelperError(
+				'',
+				ContentHelperErrorCode.PluginCredentialsNotSetMessageDetected
+			);
+
+			element = error.renderMessage();
```
mehmoodak: Can we create `ContentHelperEmptyCredentialseError` component instead of calling `renderMessage` function separately ..... using that would be easier 🤔

tests/e2e/utils.ts
```diff
@@ -97,13 +97,14 @@ export const insertRecordIntoTaxonomy = async ( recordName: string, taxonomyType
...
+ * @param {string} selector The selector from which to extract the message.
  * @return {Promise<string>} The message returned.
  */
-export const getTopRelatedPostsMessage = async ( category = '', tag = '', timeout = 500 ): Promise<string> => {
+export const getTopRelatedPostsMessage = async ( category = '', tag = '', timeout = 500, selector = '.content-helper-error-message' ): Promise<string> => {
 	// Selectors
 	const addCategoryButton = 'button.components-button.editor-post-taxonomies__hierarchical-terms-add.is-link';
 	const pluginButton = 'button[aria-label="Parse.ly Editor Sidebar"]';
-	const contentHelperMessage = '.wp-parsely-content-helper div.components-panel__body.is-opened .parsely-top-posts-descr';
+	const contentHelperMessage = '.wp-parsely-content-helper div.components-panel__body.is-opened ' + selector;
```
mehmoodak: Better to include word selector in variable.

tests/e2e/specs/content-helper/top-bar-icon.spec.ts
```diff
@@ -22,6 +22,7 @@ const pluginButton = 'button[aria-label="Parse.ly Editor Sidebar"]';
  */
 describe( 'PCH Editor Sidebar top bar icon in the WordPress Post Editor', () => {
 	const contentHelperTitle = 'Performance DetailsThis post is not published, so its details are unavailable.Related Top Posts';
+	const contentHelperTitleNoCredentials = 'Performance DetailsContact us about advanced plugin features and the Parse.ly dashboard.Existing Parse.ly customers can enable this feature by setting their Site ID and API Secret in wp-parsely options.Related Top Posts';
```
mehmoodak: Same as above using `EmptyCredentials` terminology seems better

mehmoodak (changes requested):
Reviewed the PR and added some feedback

src/@types/assets/window.d.ts
acicovic: Yes, good idea. Or maybe even `wpParselyNoCredentialsMessage`? Which do you prefer?

src/content-helper/class-content-helper-feature.php
acicovic: Can't find an issue here, indentation seems correct to me?

src/blocks/content-helper/content-helper-error.tsx
acicovic: Since we force `wp_kses_post` in `get_credentials_not_set_message()`, I do not see a problem with the current approach, unless we get careless in the future.

When researching, the only method I found to sanitize in the JS function is using third-party libraries (for examlpe DOMPurify), and I wanted to avoid it if possible. Is this the approach you're suggesting?

src/blocks/content-helper/content-helper-error.tsx
acicovic: We're currently using this like a function rather than a component, thus the camelCase and the `render` in the name. The `error` object (from which this class extends) has a `message` property itself, so this name was picked to distinguish it from that.

Maybe there's a better way to restructure everything and give better names, but I'm unsure it would be part of this PR.

src/blocks/content-helper/class-content-helper.php
acicovic: When writing this, I saw that this function needs to be below the `wp_enqueue_script` function to work and I wanted to implement it with minimal changes. It seems good to me as it is, and I'm unsure I understand how we would move it into the constructor.

src/@types/assets/window.d.ts
acicovic: Went with `wpParselyEmptyCredentialsMessage`.

src/blocks/content-helper/content-helper-error.tsx
acicovic: Please see https://github.com/Parsely/wp-parsely/pull/1588/commits/45dbe709532851ac8cac53113bd1a6b258f0ad29

src/content-helper/dashboard-widget/dashboard-widget.tsx
acicovic: Please see https://github.com/Parsely/wp-parsely/pull/1588/commits/45dbe709532851ac8cac53113bd1a6b258f0ad29

acicovic:
@mehmoodak, I've answered to all your feedback as code updates or comments. Please take a look again 🙂

src/blocks/content-helper/content-helper.tsx
```diff
@@ -12,19 +12,20 @@ import { registerPlugin } from '@wordpress/plugins';
...
+import { ElementOrEmptyCredentialsMessage } from './content-helper-error';
 
 const BLOCK_PLUGIN_ID = 'wp-parsely-block-editor-sidebar';
 
 const renderSidebar = () => (
 	<PluginSidebar icon={ <LeafIcon /> } name="wp-parsely-content-helper" className="wp-parsely-content-helper" title={ __( 'Parse.ly Editor Sidebar', 'wp-parsely' ) }>
 		<Panel>
 			<PanelBody title={ __( 'Performance Details', 'wp-parsely' ) } initialOpen={ true }>
-				<PerformanceDetails />
+				{ ElementOrEmptyCredentialsMessage( <PerformanceDetails /> ) }
```
mehmoodak: Is it better if we implement it this way?

```jsx
<VerifyCredentials>
    <PerformanceDetails />
</VerifyCredentials>
```

Here `VerifyCredentials` will show the component passed to it via `props.children` if credentials are good else show the empty credentials message.

src/content-helper/class-content-helper-feature.php
mehmoodak: As far as I know it is better to have next lines indented if the statement is continuing .... but if the linter is not giving any warning/error then it is fine.

```
are_credentials_set = $this->parsely->site_id_is_set() &&
		$this->parsely->api_secret_is_set();
```

mehmoodak (approved):

src/content-helper/class-content-helper-feature.php
acicovic: You're right, my eye didn't catch this. Fixing this right now.

src/blocks/content-helper/content-helper.tsx
acicovic: Yes! Implemented in https://github.com/Parsely/wp-parsely/pull/1588/commits/a8901418d6859995bc4df3c84ce1b77bee1bd72b.

PR #1601, opened 2023-05-12
Fix some SonarCloud warnings

acicovic:
## Description
Currently, SonarCloud is showing 9 bugs and 31 code smells in our develop branch. This PR should solve some of them.

## Motivation and context
Quickly eliminate some SonarCloud warnings that are easy to solve. 

## How has this been tested?
Existing tests pass.


tests/e2e/utils.ts
```diff
@@ -148,7 +148,7 @@ export const getTopRelatedPostsMessage = async ( category = '', tag = '', timeou
 		'document.querySelector("' + contentHelperMessageSelector + '").innerText.length > 0',
 		{ polling: 'mutation', timeout: 5000 }
 	);
-	const text = await page.$eval( contentHelperMessageSelector, ( element: Element ): string => element.textContent || '' );
+	const text = await page.$eval( contentHelperMessageSelector, ( element: Element ): string => element.textContent ?? '' );
```
mehmoodak, line 151: Instead of relying on SonarCloud can we enforce this rule via eslint?

mehmoodak (approved):

tests/e2e/utils.ts
acicovic: I've added this. Maybe we should also use the [SonarLint VSCode extension ](https://marketplace.visualstudio.com/items?itemName=SonarSource.sonarlint-vscode) as it should behave exactly as SonarCloud.

Issue #1473, opened 2023-03-07
Content Helper/Recommendations Block: Add an appropriate itm_source to URLs


acicovic:
## Is your feature request related to a problem?
The URLs we're getting from the Parse.ly API have an `itm_source` value of `parsely-api`. However, when those and other URLs are used from within the Content Helper or the Recommendations Block, we want to have an appropriate `itm_source` value for them.

## Describe the solution you'd like
1. Decide what will be the `itm_source` values used. They could all be `wp-parsely` for example, or they could be different between the Content Helper and the Recommendations Block. We could even target specific Content Helper sections with different values, though I have no context about whether this is a good idea or not (@arhine, @danielabloch).
2. Implement the solution.

danielabloch:
My first thought is to create two new `itm-source`: `content-helper` and `recs-block`. It would be helpful to start collecting data on these two separately.

GaryJones:
Maybe also consider a prefix - `plugin-content-helper`, `plugin-recommendations-block` or similar?

acicovic:
I think that prefixing is a good idea. I'd probably prefix with `wp-parsely-`. @danielabloch, what do you think? Also, should we do `recommendations-block` or `recs-block`? Although longer, I do prefer the first variant.

danielabloch:
@acicovic agree on both
prefix: `wp-parsely-`
naming: `recommendations-block`

acicovic:
This was implemented, so closing this.

PR #1625, opened 2023-05-29
Improve itm_source handling

acicovic:
## Description
We had `itm_source` parameters with the value of `parsely-api` in URLs in the PCH and the Recommendations Block. Besides not really reflecting the correct source of visits, the value was also being added to other URL strings where it was unneeded.

This PR adds correct `itm_source`s of `wp-parsely-content-helper` and `wp-parsely-recommendations-block` and improves the overall handling of `itm_source` so it does not get added to places where it isn't of use.

## Motivation and context
See #1473.

## How has this been tested?
- Tested manually.
- An integration test has been updated.


src/Endpoints/class-analytics-post-detail-api-proxy.php
```diff
@@ -45,14 +45,15 @@ public function get_items( WP_REST_Request $request ) {
...
+		$itm_source = $this->itm_source;
 
 		return array_map(
-			static function( stdClass $item ) use ( $site_id ) {
+			static function( stdClass $item ) use ( $site_id, $itm_source ) {
 				return (object) array(
 					'avgEngaged' => self::get_duration( (float) $item->avg_engaged ),
 					'dashUrl'    => Parsely::get_dash_url( $site_id, $item->url ),
-					'url'        => $item->url,
+					'url'        => Parsely::set_url_itm_source( $item->url, $itm_source ),
```
mehmoodak:
```suggestion
					'url'        => Parsely::get_url_with_itm_source( $item->url, $itm_source ),
```

mehmoodak (approved):

Issue #1602, opened 2023-05-12
Fix SonarCloud warnings in develop branch


acicovic:
## Is your feature request related to a problem?
SonarCloud is flagging our `develop` branch as having bugs and code smells. Some are simple but some others might need a bit more research and effort to fix.

## Describe the solution you'd like
1. Examine the issues SonarCloud is reporting.
2. Mark as `wontfix` those who don't make sense.
3. Solve the rest of the issues 🙂

acicovic:
With the latest merge, SonarCloud passes with 0 or 1 "bugs" (the number fluctuates) and 20 "code smells". Previously the test was not passing and SonarCloud was flagging 5 "bugs".

PR #1637, opened 2023-05-31
Content Helper: Update file and class structures

acicovic:
## Prologue
I apologize in advance for the large PR (sorry!). Having said that:
- It contains many renames without changes or minor changes.
- The majority of code changes are tests in comments.

So overall it might not be as bad as it looks and I've tried to keep changes to ones that are required or very simple.

## Description
This PR moves all files related to the PCH in a central `src/content-helper` folder. It also improves PCH class structure, which means that:
- The `Content_Helper` class has been renamed to `Editor_Sidebar` which reflects its real purpose.
- The `Admin_Columns_Parsely_Stats` class has been renamed to `Post_List_Stats`.
- Both of these classes have been moved to the `Parsely\Content_Helper` namespace.

Although the refactoring could have possibly been more extensive, it has been left to the basics for time-saving and simplicity reasons.

## Motivation and context
As we're adding more features to the Content Helper, we should have a folder and class structure that makes sense and eases our development work.

## How has this been tested?
- Manually verified that the PCH works as previously.
- Existing tests pass.


acicovic:
The failing test is SonarCloud complaining about [old code](https://sonarcloud.io/project/issues?id=Parsely_wp-parsely&pullRequest=1637&resolved=false&types=BUG) (not code introduced in this PR), so I think we should not worry about it here.

We already have an open issue for addressing SonarCloud objections [here](https://github.com/Parsely/wp-parsely/issues/1602).

src/blocks/recommendations/edit.tsx
```diff
@@ -11,7 +11,7 @@ import { useBlockProps } from '@wordpress/block-editor';
 import ParselyRecommendations from './components/parsely-recommendations';
 import ParselyRecommendationsInspectorControls from './components/parsely-recommendations-inspector-controls';
 import RecommendationsStore from './recommendations-store';
-import LeafIcon from '../shared/components/leaf-icon';
+import LeafIcon from '../../content-helper/common/icons/leaf-icon';
```
mehmoodak, line 14: Instead of `icons` I think we should go with generic folder i.e. `images`

src/content-helper/common/css/functions.scss
mehmoodak, line 1: Should we go ahead `shared` name instead of `common`?

src/content-helper/editor-sidebar/class-editor-sidebar.php
```diff
@@ -8,18 +8,22 @@
...
 
 use function Parsely\Utils\get_asset_info;
 
+use const Parsely\PARSELY_FILE;
+
 /**
  * Class that generates and manages the PCH Editor Sidebar.
  *
  * @since 3.5.0
+ * @since 3.9.0 Renamed FQCN from `Parsely\Content_Helper` to `Parsely\Content_Helper\Editor_Sidebar`
```
mehmoodak, line 24: What does it mean by `FQCN`?

mehmoodak (approved):

src/content-helper/editor-sidebar/class-editor-sidebar.php
acicovic: FQCN = Fully Qualified Class Name, which can be translated to "the full class name, including its full namespace path". In this specific instance, we haven't just changed the class name but also changed the namespace it belongs to, thus the reference to FQCN instead of saying "renamed class".

mehmoodak: Thanks for explaining but still I likes simple `class name` better here ..... feel free to keep this change if you think its better

GaryJones: FQCN is standard terminology in multiple languages. The use is correct here.

PR #1663, opened 2023-06-16
utils/date.ts: Implement additional unit tests

acicovic:
## Description
This PR introduces some additional unit tests for the `src/content-helper/common/utils/date.ts`, and introduces a couple of additional checks in its functions. 

## Motivation and context
Improving testing and completing the work (I had written a couple of those tests some time ago, but had not completed them).

## How has this been tested?
Ran the tests 🙂


src/content-helper/common/utils/date.ts
```diff
@@ -68,6 +73,10 @@ export function getSmartShortDate( date: Date ): string {
  * @return {string} The resulting date in "YYYY-MM-DD" format.
  */
 export function removeDaysFromDate( date: string, days: number ): string {
+	if ( false === Number.isInteger( days ) ) {
+		return __( 'days parameter must be an integer', 'wp-parsely' );
```
mehmoodak, line 77: Instead of returning `string` should we throw `UnexpectedValueException` here?

src/content-helper/common/utils/date.ts
```diff
@@ -15,6 +15,11 @@ const DATE_NOT_AVAILABLE_MESSAGE = __( 'Date N/A', 'wp-parsely' );
  */
 export function canProcessDate( date: Date|string ): boolean {
 	if ( 'string' === typeof date ) {
+		// Verify that date format is "YYYY-MM-DD".
+		if ( ! /^\d{4}-\d{2}-\d{2}$/.test( date ) ) {
```
mehmoodak, line 19: What will happen if the passed date is in `YYYY-DD-MM` format i.e. it will pass this condition but will there be any unexpected behavior?

tests/js/content-helper/common/utils/date.test.tsx
```diff
@@ -23,3 +32,76 @@ describe( `${ basePath } canProcessDate()`, (): void => {
...
+ */
+describe( `${ basePath } getDateInUserLang()`, (): void => {
+	it( 'should format dates according to passed options', (): void => {
+		expect( getDateInUserLang(
+			new Date( '2022-11-11' ),
+			{ month: 'short', day: 'numeric', year: 'numeric' },
+		) ).toBe( 'Nov 11, 2022' );
+
+		expect( getDateInUserLang(
+			new Date( '2022-12-12' ),
```
mehmoodak: Should we also add a test with date greater than `12` ... just to verify the proper format?

mehmoodak (approved):

src/content-helper/common/utils/date.ts
acicovic: We could, though I don't necessarily think we should since we're able to show the error 🤔

src/content-helper/common/utils/date.ts
acicovic: You've got a point.

There are regexes that are much more thorough (and very complicated) but I think that the real fix here would be to remove the capability of passing the date as a string.

The only function that accepts a date as a string is `removeDaysFromDate()`, so the fix might not be that complicated. This would be outside of the scope of this PR though, so noting this to look into it as a separate item.

tests/js/content-helper/common/utils/date.test.tsx
acicovic: You're right. I updates the existing test.

src/content-helper/common/utils/date.ts
acicovic: `canProcessDate()` no longer accepts a string as an argument and doesn't do any type of regex, so closing this comment as resolved.

PR #1670, opened 2023-06-20
Fix PHPStan errors

acicovic:
## Description
After several PHPStan updates, running PHPStan against our develop branch would result in 110 errors in my environment. This PR fixes them all.

## Motivation and context
- Improve code quality.
- Improve developer workflow (no pre-commit hooks failing every time we want to commit).
- To work around the previous point, we needed to skip pre-commit checks which was not a good practice since some things could slip through if the checks weren't done manually.

## How has this been tested?
- All existing tests pass.
- Running `composer static-analysis` returns no errors in my environment.


phpstan.neon
```diff
@@ -12,8 +13,8 @@ parameters:
     - vendor/php-stubs/wordpress-stubs/wordpress-stubs.php
     - vendor/php-stubs/wordpress-tests-stubs/wordpress-tests-stubs.php
   type_coverage:
-      return_type: 88
-      param_type: 59
+      return_type: 92
+      param_type: 65
```
mehmoodak, line 17: Glad to see it increased 👍

.phpcs.xml.dist
```diff
@@ -79,4 +79,9 @@
 		</properties>
 	</rule>
 
+	<!-- Allow short @var assignments in tests. -->
+ 	<rule ref="Generic.Commenting.DocComment.MissingShort">
+    	<exclude-pattern>/tests/</exclude-pattern>
+ 	</rule>
```
mehmoodak, line 85: 1. What is the benefit of adding this rule instead of going what phpcs and wpcs provides by default?
2. Why we are excluding `/tests/` and not following the consistencny?

.vscode/wp-parsely.code-workspace
```diff
@@ -39,6 +39,13 @@
 		"[markdown]": {
 			"files.trimTrailingWhitespace": false
 		},
+		// Error Lens
+		"errorLens.exclude": [
+			// Exclude type coverage errors to reduce noise. The errors are
+			// still visible in all other parts of VSCode and will cause
+			// pre-commit hooks to fail.
+			" actually have it. Add more (return|param) types to get over "
```
mehmoodak, line 47: By noise you mean a prominent messages shows by `Error Lens` or something else? (if we are talking about that prominent message then I think we should keep it as this noise is beneficial and forces us to add more type coverage.

tests/Integration/Metadata/CustomTaxonomyTermArchiveTest.php
```diff
@@ -76,7 +78,8 @@ public function test_metadata_is_correctly_constructed_for_custom_taxonomy_term_
 		// Create the structured data for that term archive. The term archive
 		// metadata doesn't use the post data, but the construction method
 		// requires it for now.
-		$metadata        = new Metadata( $parsely );
+		$metadata = new Metadata( $parsely );
+		/** @var array<string, mixed> $structured_data */
```
mehmoodak, line 82: Instead of giving type here can we return this type in `construct_metadata` function?

mehmoodak (changes requested):
Done with review and left some comments.

.phpcs.xml.dist
acicovic: As stated by its comment, with this rule we allow short `@var` assignments  (which isn't allowed by default) only in tests. This allows us to use single-line `/** @var string */` docblocks instead of the following which takes multiple lines and is often not that useful:

```
/**
 * Variable.
 *
 * @var string
 */
```

We can of course still use it this form if we want to provide meaningful documentation in the first line. This was done so tests are not full of comment blocks such as the above. We still require the longer form in all other code.

.vscode/wp-parsely.code-workspace
acicovic: While writing code, Error Lens shows messages like "Out of X possible param types, only X % actually have it. Add more param types to get over X %".

The problem with this is that:
- Those errors are shown all over the place next to method signatures that cannot be fixed, or would need refactorings of questionable real value.
- Personally for me, they inhibit my eyes when I'm writing code and they get in the way because they move.
- If someone runs the `static-analysis` command, PHPStan will pass. So actually those red warnings represent an error that doesn't really exist when we run `composer static-analysis` for our whole project.
 
Due to all the above, IMO there is no sense in keeping those errors in the editor, since in-editor errors should be useful and those aren't (they are even misleading). Here's a sample screenshot:

image

The reason I put it this exclusion the global workspace is that when I put it in my personal settings it doesn't work. Error Lens gives some commands to disable it temporarily so I could use that instead, but then all errors are gone and this is worse.

acicovic: Noting that even when excluded, the function has a red squiggle that shows the error when hovered.

tests/Integration/Metadata/CustomTaxonomyTermArchiveTest.php
acicovic: The `construct_metadata` function already returns a PHPStan type which is more specific, and thus would be considered better at least theoretically.

```
	/**
	 * Creates Parse.ly metadata object from post metadata.
	 *
	 * @param WP_Post $post object.
	 *
	 * @return Metadata_Attributes
	 */
	public function construct_metadata( WP_Post $post ) {
```

The above type seems to be incompatible with the array type that is expected in another function a bit below. So this is why we are casting it to array here. Maybe it would be possible to fix this by modifying `construct_metadata` or the receiving function, but this was out of scope for the PR.

Maybe this could be the object of a future iteration on PHPStan-related improvements.

acicovic: Sidenote: In general, PHPStan types seem to have some side-effects in my experience, and they can be the reason for the `Out of X possible param types, only X % actually have it. Add more param types to get over X %` errors mentioned previously.

So in the future, we can look at all the data and see how to proceed regarding that.

.phpcs.xml.dist
mehmoodak: This short form syntax is good maybe we should use it in whole codebase (ofcourse not in this PR)

.vscode/wp-parsely.code-workspace
mehmoodak: This sounds good. I wasn't seeing these errors on my end therefore was confused .... thanks for the explanation and screenshot.

mehmoodak (approved):

PR #1677, opened 2023-06-26
Implement Custom Post Types tracking by default

### LINKED ISSUE #1112: Feature: track Custom Post Types by default
wp-parsely tracks Posts and Pages, but does not track any Custom Post Types by default. This is to avoid tracking content that should not be tracked.
However, customers use Custom Post Types for their content and for non-post pages that are still important for them to update, track performance on, and enhance over time. This means that the plugin will not work as intended out of the box in those cases, and that it will need further configuration by the customer or us.

If a user has no settings, CPTs will be selected by default in the "Track as" option.  aka Track custom page types by default.
Customers can then adjust how they are tracked, or turn off tracking for page types depending on their needs. 
This prevents issues where customers think tracking is broken because they don't see data for some of their pages, which we see quite frequently.
Whether we handle a normal Post or Custom Post Type, we go though [this function](https://github.com/Parsely/wp-parsely/blob/747c7be772fbeac60e603af70aeb5584f2da84ad/src/Metadata/class-post-builder.php#L278-L309) which is called [here](https://github.com/Parsely/wp-parsely/blob/747c7be772fbeac60e603af70aeb5584f2da84ad/src/Metadata/class-post-builder.php#L168-L170).

acicovic:
## Description
This PR implements CPT tracking by default. The general logic for it is (as suggested by @mjangda):

- if ( public && supports editor && is hierarchical ) { track as page }
- else if ( public && supports editor && is not hierarchical ) { track as post }
- else { do no track }

## Motivation and context
- Discussed internally and in #1112.
- Closes #1112.

## How has this been tested?
- Manual testing.
- Modified an integration test and added a couple more.


acicovic:
Noting that the lint/CS warning should be taken care of by the work in #1670.

src/class-parsely.php
```diff
@@ -362,9 +362,10 @@ public function get_options() {
 		 *
 		 * @var Parsely_Options|null
 		 */
-		$options = get_option( self::OPTIONS_KEY, $this->option_defaults );
+		$options = get_option( self::OPTIONS_KEY, null );
 
 		if ( ! is_array( $options ) ) {
+			$this->set_default_track_as_values();
```
mehmoodak, line 368: Just to confirm: This will only effect new plugin istalls and doesn't change anything for existing installs ... right?

tests/Integration/OptionsTest.php
```diff
@@ -0,0 +1,175 @@
+<?php
+/**
+ * Integration Tests: Plugin options
```
mehmoodak, line 3: I think this is not an integration test.

tests/Integration/OptionsTest.php
```diff
@@ -0,0 +1,175 @@
...
+	 *
+	 * @covers \Parsely\Parsely::get_options
+	 */
+	public function test_default_options_are_returned_when_options_are_corrupted_or_not_set(): void {
+		update_option( Parsely::OPTIONS_KEY, 'someinvalidvalue' );
+
+		$options = self::$parsely->get_options();
+		self::assertSame( self::$parsely->get_default_options(), $options );
+
+		delete_option( Parsely::OPTIONS_KEY );
```
mehmoodak, line 59: This `delete_option` and the above `update_option` seems unnecessary because `get_options` function isn't setting any WP Option.

tests/Integration/OptionsTest.php
```diff
@@ -0,0 +1,175 @@
...
+	 * @since 3.9.0
+	 *
+	 * @covers Parsely\Parsely::get_options
+	 */
+	public function test_get_options_returns_correct_track_as_defaults(): void {
+		$options = self::$parsely->get_options();
+		self::assertSame( array( 'post' ), $options['track_post_types'] );
+		self::assertSame( array( 'page' ), $options['track_page_types'] );
+
+		delete_option( Parsely::OPTIONS_KEY );
```
mehmoodak: Same as above

tests/Integration/OptionsTest.php
```diff
@@ -0,0 +1,175 @@
...
+		register_post_type(
+			'excluded_page_type_3',
+			array(
+				'public'       => false,
+				'supports'     => array( 'editor' => false ),
+				'hierarchical' => true,
+			)
+		);
+
+		delete_option( Parsely::OPTIONS_KEY );
```
mehmoodak: Same as above

tests/Integration/OptionsTest.php
```diff
@@ -0,0 +1,175 @@
...
+
+		// Private page without editor support.
+		register_post_type(
+			'excluded_page_type_3',
+			array(
+				'public'       => false,
+				'supports'     => array( 'editor' => false ),
+				'hierarchical' => true,
+			)
+		);
```
mehmoodak, line 159: Can we use loop while registering multiple post types .... I think that will be easier to read and maintain.

tests/Integration/OptionsTest.php
```diff
@@ -0,0 +1,175 @@
...
+
+		// Clean up.
+		unregister_post_type( 'custom_post_type' );
+		unregister_post_type( 'excluded_post_type_1' );
+		unregister_post_type( 'excluded_post_type_2' );
+		unregister_post_type( 'excluded_post_type_3' );
+		unregister_post_type( 'custom_page_type' );
+		unregister_post_type( 'excluded_page_type_1' );
+		unregister_post_type( 'excluded_post_type_2' );
+		unregister_post_type( 'excluded_post_type_3' );
```
mehmoodak: Can we use loop while unregistering multiple post types .... I think that will be easier to read and maintain.

mehmoodak (changes requested):
Done with review and left some comments.

src/class-parsely.php
acicovic: Yes. The idea here is to set the default values only if the options are unset or invalid. This won't affect existing installs, as the settings are already an array there.

tests/Integration/OptionsTest.php
acicovic: In this test, we are testing how `get_options()` behaves when the options are damaged or not present in the database. The `update_option()` function makes the option invalid, and the `delete_option()` removes the option from the database to test these scenarios.

tests/Integration/OptionsTest.php
acicovic: Right. So what you're saying if I understand correctly is that since we have never set the option, it will be already absent from the database, thus deletion makes no sense.

tests/Integration/OptionsTest.php
acicovic: How so? We're using the `register_post_type` and `unregister_post_type` WordPress functions in one of the tests. I don't feel that there's a need to break this file into unit and integration tests, but let me know if you do.

src/class-parsely.php
acicovic: I've pushed a test that verifies that the `set_default_track_as_values()` function doesn't fire when options are saved.

tests/Integration/OptionsTest.php
acicovic: I haven't performed any changes for the specific item, since I think that what is done here is correct. For everything else, I've applied changes that should address your feedback.

src/class-parsely.php
```diff
@@ -380,6 +381,31 @@ public function get_options() {
...
+				continue;
+			}
+
+			if ( is_post_type_hierarchical( $post_type ) ) {
+				$this->option_defaults['track_page_types'][] = $post_type;
+			} else {
+				$this->option_defaults['track_post_types'][] = $post_type;
+			}
+		}
+	}
```
mjangda, line 407: Rather than changing the state of the object, I wonder if it's better to just calculate and return the configuration and then pass into the `array_merge`. This feels a bit cleaner but it's a bit subjective.

```suggestion
	public function get_default_post_type_tracking_config(): void {
		$defaults = array(
			'track_page_types' => array(),
			'track_post_types'  => array(),
		);

		$post_types = get_post_types( array( 'public' => true ) );

		foreach ( $post_types as $post_type ) {
			if ( ! post_type_supports( $post_type, 'editor' ) ) {
				continue;
			}

			if ( is_post_type_hierarchical( $post_type ) ) {
				$defaults['track_page_types'][] = $post_type;
			} else {
				$defaults['track_post_types'][] = $post_type;
			}
		}

		return $defaults;
	}
```

tests/Integration/OptionsTest.php
```diff
@@ -0,0 +1,169 @@
...
+
+		$options = self::$parsely->get_options();
+
+		self::assertSame( array( 'post', 'custom_post_type' ), $options['track_post_types'] );
+		self::assertSame( array( 'page', 'custom_page_type' ), $options['track_page_types'] );
+
+		// Clean up.
+		foreach ( $custom_post_types as $key => $value ) {
+			unregister_post_type( $key );
+		}
```
mjangda: Alternative option that's cleaner: add a tear_down method that calls `$this->reset_post_types()`

https://core.trac.wordpress.org/browser/trunk/tests/phpunit/includes/abstract-testcase.php#L294

tests/Integration/OptionsTest.php
```diff
@@ -0,0 +1,169 @@
...
+		$options['track_post_types'] = array( 'new_post_type' );
+		$options['track_page_types'] = array( 'new_page_type' );
+
+		add_option( Parsely::OPTIONS_KEY, $options );
+		$saved_options = self::$parsely->get_options();
+
+		self::assertSame( array( 'new_post_type' ), $saved_options['track_post_types'] );
+		self::assertSame( array( 'new_page_type' ), $saved_options['track_page_types'] );
+
+		delete_option( Parsely::OPTIONS_KEY );
```
mjangda: This cleanup shouldn't be necessary since the testcase class attempts to reset the state after each test.

src/class-parsely.php
```diff
@@ -362,9 +362,10 @@ public function get_options() {
 		 *
 		 * @var Parsely_Options|null
 		 */
-		$options = get_option( self::OPTIONS_KEY, $this->option_defaults );
+		$options = get_option( self::OPTIONS_KEY, null );
 
 		if ( ! is_array( $options ) ) {
```
mjangda, line 367: One edge case where this can be a problem is if the DB query to lookup options fails. We've seen this happen on VIP a few times which can end up resetting the state.

But, it's an edge case and a lower-level issue so I'm not sure it's worth solving for here. If we did want to solve for it, we'd need to check for something like this:

```suggestion
		if ( ! is_array( $options ) && ! $wpdb->last_error ) {
```

mjangda (commented):
Looks good 👍 

Added some minor notes and suggestions.

src/class-parsely.php
acicovic: My research indicates that `$wpdb->last_error` is flushed before every query, so we don't risk getting an error that comes from a previous query. If we were to do this check, I'd probably go with something like this:

```php
global $wpdb

// If options aren't an array or a database error occurred, return the default options.
if ( ! is_array( $options ) || '' !== $wpdb_last_error ) {
```

Unsure this adds value though, as in case of a database reading failure, `$options` should be `false` or `null`, which will still return the default options.

tests/Integration/OptionsTest.php
acicovic: Done.

tests/Integration/OptionsTest.php
acicovic: Removed and verified that tests work without issues.

src/class-parsely.php
acicovic: I think I agree that this is cleaner.

However, this change has proven to be more complicated than I expected when I started working on it. We could implement it as part of a refactoring including some other things, but this is out of scope for this PR.

src/class-parsely.php
mjangda: Sorry, what I meant was that we want to prevent the default options from being used when the db query fails. Otherwise, a temporary issue with the database would reset all Parsely options. 

But like I said, it's an edge case and also not a new problem, so i'm okay with skipping this for now.

mjangda (approved):

PR #1680, opened 2023-06-27
install-wp-tests.sh: Allow directories overwriting

acicovic:
## Description
Once `install-wp-tests.sh` would download the required files and create its directories, it would then quit on subsequent runs without doing anything because the directories in question were already present.

This however would get in the way when trying switch between different WordPress versions when testing. With this PR, we're giving the user the choice to choose a directory overwrite, re-running the whole script.

## Motivation and context
- Make it easier to juggle between different WordPress versions when testing.
- This was written as part of facilitating testing for https://github.com/Parsely/wp-parsely/pull/1681.

## How has this been tested?
- Tested manually a few times, jumping between `latest` and `trunk` parameters.


bin/install-wp-tests.sh
```diff
@@ -58,12 +58,22 @@ set -ex
 install_wp() {
 
 	if [ -d $WP_CORE_DIR ]; then
-		return;
+		while true; do
+			read -p "The $WP_CORE_DIR directory already exists. Overwrite it? (y/n) " yn
+			case $yn in
+				y|Y ) rm -rf $WP_CORE_DIR; break;;
+				n|N ) exit;;
+				* ) echo invalid response;
```
mehmoodak: Should we make `invalid response` a string?

mehmoodak (approved):
Left a comment but approved the PR.

bin/install-wp-tests.sh
acicovic: I added quotes and updated the message to be a bit more descriptive.

PR #1681, opened 2023-06-27
Fix failing integration test for WordPress 6.3 versions and newer

acicovic:
## Description
Upon the launch of WordPress 6.3 Alpha versions, one of our integration tests started failing:

```
Content Helper Post List Stats (Parsely\Tests\ContentHelper\ContentHelperPostListStats)
 ✘ Script of parsely stats admin column on valid posts and valid response
   │
   │ Unexpected deprecation notice for WP_Scripts::print_inline_script.
   │ Function WP_Scripts::print_inline_script is deprecated since version 6.3.0! Use WP_Scripts::get_inline_script_data() or WP_Scripts::get_inline_script_tag() instead.
   │ Failed asserting that an array is empty.
```

This PR fixes the failure by calling the correct function when the test is running on WordPress 6.3 Alpha or newer.

## Motivation and context
No more red warnings everywhere in our PRs.

## How has this been tested?
The test that was failing now passes when tested with the latest WordPress 6.3 Alpha.

### DIFF
--- a/tests/Integration/content-helper/ContentHelperPostListStatsTest.php
+++ b/tests/Integration/content-helper/ContentHelperPostListStatsTest.php
@@ -719,7 +719,15 @@ final class ContentHelperPostListStatsTest extends ContentHelperFeatureTest {
 		global $wp_scripts;
 
 		ob_start();
-		var_dump( $wp_scripts->print_inline_script ( 'post-list-stats-script', 'before' ) ); // phpcs:ignore
+
+		global $wp_version;
+		if ( true === version_compare( substr( $wp_version, 0, 3 ), '6.3', '>=' ) ) {
+			// @phpstan-ignore-next-line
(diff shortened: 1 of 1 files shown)


tests/Integration/content-helper/ContentHelperPostListStatsTest.php
```diff
@@ -719,7 +719,15 @@ public function test_script_of_parsely_stats_admin_column_on_valid_posts_and_val
 		global $wp_scripts;
 
 		ob_start();
-		var_dump( $wp_scripts->print_inline_script ( 'post-list-stats-script', 'before' ) ); // phpcs:ignore
+
+		global $wp_version;
+		if ( true === version_compare( substr( $wp_version, 0, 3 ), '6.3', '>=' ) ) {
+			// @phpstan-ignore-next-line
```
acicovic, line 725: The PHPStan comment is because PHPStan complains that we're calling an undefined function. We will probably be able to remove it in the future.

mehmoodak (approved):

PR #1696, opened 2023-07-04
Add PHP 8.3 to integration and unit tests

acicovic:
## Description
This PR is adds PHP 8.3 testing to integration and unit tests.

## Motivation and context
PHP 8.3 is to be released in November.

## How has this been tested?
Re-running existing tests.


.github/workflows/integration-tests.yml
```diff
@@ -47,6 +47,10 @@ jobs:
             wp: 'trunk'
             coverage: none
             experimental: false
+          - php: '8.3'
+            wp: 'trunk'
+            coverage: none
+            experimental: false
```
mehmoodak, line 53: - Any changes for line 31, 36 of this file?
- Any changes for `unit-tests.yml` file?

mehmoodak (changes requested):
Done with review and left a comment.

.github/workflows/integration-tests.yml
acicovic: - The comments on lines 26-31 are actually outdated, and I plan to correct them in another PR when everything related is merged.
- No changes are needed in line 36 as far as I'm aware, unless I'm missing something.

acicovic: Thanks for spotting that I should also make changes to `unit-tests.yml`, this was an omission from my part.

.github/workflows/unit-tests.yml
```diff
@@ -34,22 +34,16 @@ jobs:
       # - coverage: Whether to run the tests with code coverage.
       # - experimental: Whether the build is "allowed to fail".
       matrix:
-        php: ['7.2', '7.3', '7.4']
+        php: ['7.2', '7.3', '7.4', '8.0', '8.2', '8.3']
```
mehmoodak, line 37: Any reason why we have php versions different here and in above file?

mehmoodak (approved):
Left a comment but approving the PR.

.github/workflows/unit-tests.yml
acicovic: Yes! 😃

In `integration-tests.yml`, PHP versions `8.1`, `8.2` and `8.3` are defined in the `include` section in order to run the tests with specific configurations (for example code coverage or to run the tests against WP `trunk`).

In `unit-tests.yml` we have only put the `8.1` version in the `include` section to run it with code coverage (we cannot run unit tests against WP `trunk`). This is why there are differences between the lines.

We can confirm that things work correctly by going to the PR's page and verifying that integration and unit tests run for PHP 7.4 - 8.3, which they do.

Edit: In the previous paragraph, I mistakenly wrote 7.4 when I meant 7.2

mehmoodak: I think if we are adding new versions then we should also remove some old versions.

acicovic: Our plugin supports PHP version 7.2 and up, so we still need to test against 7.2.

But we'll remove versions as our PHP requirements go up.

PR #1724, opened 2023-07-18
Add managed options feature

### LINKED ISSUE #1338: Add hooks to cater for platform incompatibilities and UX tailoring

acicovic:
## Description
This PR adds a managed options feature, to be used by hosts for managing plugin options.

- Managed options can be set using the [wp_parsely_managed_options](https://github.com/Parsely/wp-parsely/blob/8d31931e0ab51cadf8f9f1a55f81a74fdf1ead06/tests/Integration/OptionsTest.php#L269-L282) filter.
- Badges are shown next to every managed option. Badge text and URL can be customized using the [wp_parsely_managed_options_badge](https://github.com/Parsely/wp-parsely/blob/8d31931e0ab51cadf8f9f1a55f81a74fdf1ead06/src/UI/class-settings-page.php#L101-L107) filter.

## Motivation and context
Closes https://github.com/Parsely/wp-parsely/issues/1338.

## How has this been tested?
Added some related tests.


src/class-parsely.php
```diff
@@ -653,4 +670,93 @@ private function are_credentials_managed(): bool {
...
+
+		if ( ! is_array( $managed_options ) ) {
+			return;
+		}
+
+		// Don't allow certain options to be set as managed.
+		unset(
+			$managed_options['apikey'],
+			$managed_options['api_secret'],
+			$managed_options['metadata_secret'],
```
mjangda, line 693: Just to confirm: we're separating "managed options" from "managed credentials"? i.e. for overriding apikey / api_secret / metadata_secret, we'd use the `wp_parsely_credentials` filter.

acicovic: Correct.

src/UI/class-settings-page.php
```diff
@@ -1294,4 +1328,63 @@ private function get_unobfuscated_value( $current_value, $previous_value ): stri
...
+
+	/**
+	 * Returns the taxonomies allowed to be used as custom_taxonomy_section
+	 * option values.
+	 *
+	 * @since 3.9.0
+	 *
+	 * @return array<string> Array of taxonomies.
+	 */
+	public static function get_section_taxonomies(): array {
```
acicovic, line 1375: This was extracted as a function because it is being called in 2 places now.

src/css/admin-settings.scss
```diff
@@ -1,49 +1,85 @@
...
+
+.settings_page_parsely {
 	--padding-default: 15px;
-}
 
-#track-post-types {
-	max-width: 550px;
-	width: 100%;
-}
+	#track-post-types {
```
acicovic, line 7: `#track-post-types` styles have been re-arranged under a single element, with no real code changes. This is also true for `@media` CSS.

src/css/admin-settings.scss
```diff
@@ -1,49 +1,85 @@
...
-#track-post-types thead th:nth-child(3) {
-	word-break: break-word;
-}
+		thead th:nth-child(3) {
+			word-break: break-word;
+		}
+	}
 
-@media only screen and (max-width: 380px) {
+	fieldset:disabled,
```
acicovic, line 38: `fieldset` and badge styles have been added in this section, until `@media` styles.

views/parsely-settings.php
```diff
@@ -35,7 +35,7 @@
 ?>
 
 <div class="wrap">
-	<h1 class="wp-heading-inline"><?php echo esc_html( get_admin_page_title() ); ?></h1>
+	<h1 class="wp-heading-inline"><?php echo esc_html_e( 'Parse.ly Settings', 'wp-parsely' ); ?></h1>
```
acicovic, line 38: This was updated because it was creating some problems for tests.

src/class-parsely.php
```diff
@@ -653,4 +670,93 @@ private function are_credentials_managed(): bool {
...
+	 *
+	 * @since 3.9.0
+	 * @access private
+	 */
+	private function set_managed_options(): void {
+		$managed_options = apply_filters( 'wp_parsely_managed_options', false );
+
+		if ( ! is_array( $managed_options ) ) {
+			return;
+		}
```
mjangda, line 687: Minor: Should we move this below the `unset`? That way, if we unset a bunch of keys and end up with an empty we correctly return early.

src/class-parsely.php
```diff
@@ -653,4 +670,93 @@ private function are_credentials_managed(): bool {
...
+	 * @param string      $option_id The option's ID.
+	 * @param bool|string $value The option's value.
+	 *
+	 * @return bool|string The sanitized option value.
+	 */
+	private function sanitize_managed_option( string $option_id, $value ) {
+		$option_value_type = gettype( $this->option_defaults[ $option_id ] );
+
+		if ( 'boolean' === $option_value_type && ! is_bool( $value ) ) {
+			return false;
```
mjangda, line 753: We should trigger a `_doing_it_wrong` here.

src/class-parsely.php
```diff
@@ -653,4 +670,93 @@ private function are_credentials_managed(): bool {
...
+		}
+
+		if ( 'string' === $option_value_type ) {
+			if ( ! is_string( $value ) ) {
+				$value = strval( $value );
```
mjangda, line 769: We should trigger a `_doing_it_wrong` here.

src/class-parsely.php
```diff
@@ -653,4 +670,93 @@ private function are_credentials_managed(): bool {
...
+			}
+
+			// String options that are restricted to specific values.
+			$restricted_value_options = array(
+				'custom_taxonomy_section' => Settings_Page::get_section_taxonomies(),
```
mjangda, line 774: Minor: This feels a bit risky. Even though it's calling code within our own plugin, I worry that it may cause problems at some point due to load order or something else.

I'm not sure if we need to enforce that an incorrect taxonomy was passed in, but I think it's also okay to leave as-is.

src/class-parsely.php
```diff
@@ -653,4 +670,93 @@ private function are_credentials_managed(): bool {
...
+				'meta_type'               => array( 'json_ld', 'repeated_metas' ),
+			);
+
+			// Verify that the above values are respected.
+			foreach ( $restricted_value_options as $option_key => $valid_values ) {
+				if ( $option_id === $option_key ) {
+					if ( ! in_array( $value, $valid_values, true ) ) {
+						$value = $this->option_defaults[ $option_id ];
```
mjangda, line 794: We should trigger a `_doing_it_wrong` here.

src/UI/class-settings-page.php
```diff
@@ -1294,4 +1328,63 @@ private function get_unobfuscated_value( $current_value, $previous_value ): stri
...
+	public function set_field_label_contents( string $title, string $option_id ): string {
+		$is_managed = key_exists( $option_id, $this->parsely->managed_options );
+
+		if ( $is_managed ) {
+			$text = $this->managed_options_badge['text'] ?? '';
+			if ( ! is_string( $text ) || '' === $text ) {
+				return $title;
+			}
+
+			$text = esc_attr( $text );
```
mjangda: We should escape as late as possible. Additionally `esc_attr` is not ideal for non-attribute content.

```suggestion
```

src/UI/class-settings-page.php
```diff
@@ -1294,4 +1328,63 @@ private function get_unobfuscated_value( $current_value, $previous_value ): stri
...
+			$url  = filter_var( $this->managed_options_badge['url'] ?? '', FILTER_VALIDATE_URL );
+
+			if ( false === $url ) {
+				$badge = '<span class="managed-option-badge">' . $text . '</span>';
+			} else {
+				$badge = '<a class="managed-option-badge" href="' . esc_url( $url ) .
+					'" target="_blank" rel="noopener">' . $text . '</a>';
```
mjangda:
```suggestion
					'" target="_blank" rel="noopener">' . esc_html( $text ) . '</a>';
```

src/UI/class-settings-page.php
```diff
@@ -1294,4 +1328,63 @@ private function get_unobfuscated_value( $current_value, $previous_value ): stri
...
+			$text = $this->managed_options_badge['text'] ?? '';
+			if ( ! is_string( $text ) || '' === $text ) {
+				return $title;
+			}
+
+			$text = esc_attr( $text );
+			$url  = filter_var( $this->managed_options_badge['url'] ?? '', FILTER_VALIDATE_URL );
+
+			if ( false === $url ) {
+				$badge = '<span class="managed-option-badge">' . $text . '</span>';
```
mjangda:
```suggestion
				$badge = '<span class="managed-option-badge">' . esc_html( $text ) . '</span>';
```

src/UI/class-settings-page.php
```diff
@@ -81,13 +81,34 @@ final class Settings_Page {
...
 	 * @param Parsely $parsely Instance of Parsely class.
 	 */
 	public function __construct( Parsely $parsely ) {
 		$this->parsely = $parsely;
+
+		$managed_options_badge = apply_filters(
+			'wp_parsely_managed_options_badge',
+			array(
+				'text' => __( 'Upgrade', 'wp-parsely' ),
+				'url'  => 'https://www.parse.ly/getdemo/',
```
mjangda, line 105: Note that when we start to use the integrations framework for joint customers as well, this will not make much sense. We'll need to remember to use the filter to handle P100 vs Joint users.

mjangda (approved):
Just had some small suggested, but overall this looks good.

src/class-parsely.php
```diff
@@ -653,4 +670,93 @@ private function are_credentials_managed(): bool {
...
 	}
+
+	/**
+	 * Sets the values of managed options.
+	 *
+	 * @since 3.9.0
+	 * @access private
+	 */
+	private function set_managed_options(): void {
+		$managed_options = apply_filters( 'wp_parsely_managed_options', false );
```
mehmoodak, line 683: Seems like it will be helpful to have a comment here that explains this filter and if needed then some explanation how its different from `wp_parsely_credentials` filter.

src/class-parsely.php
```diff
@@ -653,4 +670,93 @@ private function are_credentials_managed(): bool {
...
+
+		// Set managed options values.
+		foreach ( $managed_options as $key => $value ) {
+			$is_option_valid = isset( $this->option_defaults[ $key ] );
+
+			if ( $is_option_valid ) {
+				if ( null === $value ) {
+					// When null, the option gets its value from the database.
+					$this->managed_options[ $key ] =
+						$current_options[ $key ] ?? $this->option_defaults[ $key ];
```
mehmoodak, line 718: Can we just `unset` here?

src/class-parsely.php
```diff
@@ -653,4 +670,93 @@ private function are_credentials_managed(): bool {
...
+				} else {
+					$this->managed_options[ $key ] =
+						$this->sanitize_managed_option( $key, $value );
+				}
+			}
```
mehmoodak, line 723: Can we refactor this `foreach` loop like following?

```php
// Set managed options values.
foreach ( $managed_options as $key => $value ) {
	$is_option_valid = isset( $this->option_defaults[ $key ] );

	if ( ! $is_option_valid || null === $value ) {
		continue;
	}

	$this->managed_options[ $key ] = $this->sanitize_managed_option( $key, $value );
}
```

src/class-parsely.php
```diff
@@ -131,17 +132,32 @@ class Parsely {
...
+	/**
+	 * Holds the managed options and their values.
+	 *
+	 * This allows hosting providers to provide a more customized experience for
+	 * the plugin by handling options automatically.
+	 *
+	 * @since 3.9.0
+	 * @access private
+	 *
+	 * @var array<empty>|array<string, bool|string|null>
```
mehmoodak, line 149: I think we can remove `array<empty>`

src/class-parsely.php
```diff
@@ -653,4 +670,93 @@ private function are_credentials_managed(): bool {
...
+
+		// Don't allow certain options to be set as managed.
+		unset(
+			$managed_options['apikey'],
+			$managed_options['api_secret'],
+			$managed_options['metadata_secret'],
+			$managed_options['track_post_types'],
+			$managed_options['track_page_types'],
+			$managed_options['plugin_version']
+		);
```
mehmoodak, line 697: Just a thought .... Instead of blacklisting I think we should follow whitelisting approach here i.e. only define the values that we want to be managed.

src/class-parsely.php
acicovic: PHPStan complains if we do:

 ------ ------------------------------------------------------------------------------
  Line   src/class-parsely.php
 ------ ------------------------------------------------------------------------------
  712    Property Parsely\Parsely::$managed_options (array<string, bool|string|null>)
         does not accept array<array<string>|bool|string|null>.
 ------ ------------------------------------------------------------------------------

src/class-parsely.php
acicovic: I'm unsure I understand how `unset()` would help here.

mehmoodak: Ignore this as I posted a code snippet below.

src/class-parsely.php
acicovic: I like the idea of simplifying code where we can. However, are you certain that those two are functionally equivalent? I'm getting integration test errors when using this.

mehmoodak: To me it seems similar, if you are getting errors then I think error message can tell what's wrong.

src/class-parsely.php
acicovic: I get your point.

- We could move this to the `Validator` class, but as long as we're moving this to an external file, we've got the same load order dangers.
- We could implement the function within this file, but I think this function is a bit out of scope for this class.

I think that validating values makes sense, and I think I'd leave this as-is for the moment, until we can come up with something better.

src/UI/class-settings-page.php
acicovic: The context here is that we don't want HTML passed into the badge, and this is why `esc_attr()` was picked. Can switch this to `esc_html` if this is what you think we should do.

I prefer to escape the variable before since it is only for display purposes, but since this is a WP convention I'll adjust to that.

src/class-parsely.php
acicovic: Both approaches would work, I think this is largely opinion-based. I prefer the blocklist because:
- There are less items to add to the list.
- An allowlist would need to be updated every time a new option is added. Admittedly, not something that happens often, but would introduce a friction point and a "oh I forgot" potential.

src/class-parsely.php
acicovic: This won't help with empty arrays, since it checks whether the user actually passed a non-array. Will add an early return after the `unsets` though if the array is empty.

acicovic: Done in https://github.com/Parsely/wp-parsely/pull/1724/commits/8fefc99de4d93c9443ae79943cc6ee541d18ed48.

src/class-parsely.php
acicovic: Added a function comment that should help with this in https://github.com/Parsely/wp-parsely/pull/1724/commits/65f10b89f456e352ae84f89fb3fe6dc34f3d4d88.

src/class-parsely.php
acicovic: Done in https://github.com/Parsely/wp-parsely/pull/1724/commits/6d7681ce2b94ede00534d7505220f97f8944290c with some fixes in https://github.com/Parsely/wp-parsely/pull/1724/commits/9789500925076c9f3138661db208f01a89a44021. Will mark similar requests as resolved.

src/UI/class-settings-page.php
acicovic: Applied your suggestions below using `esc_html()` in https://github.com/Parsely/wp-parsely/pull/1724/commits/d5a58d5378e80a8be9980819d53c51797ceb59ae. Marking the related suggestions as resolved, and keeping this open for context.

src/UI/class-settings-page.php
acicovic: Started a private communication thread about this.

src/class-parsely.php
acicovic: This code doesn't do anything if the value of the option is `null`, which is not what we want. So it isn't really an alternative we can use.

PR #1727, opened 2023-07-20
Add Validator class

acicovic:
## Description
This PR introduces a `Validator` class. For now, only secrets are being validated but this could be expanded in the future to include all current validation needs (with an option to move to dedicated `*_Validator` classes if needed).

## Motivation and context
With some of the latest changes in #1724, some unit tests began failing. I could have changed the implementation and I also tried mocking the `__()` function to make unit tests pass, but moving validation code to a `Validator` class actually makes more sense. So here it is.

## How has this been tested?
Existing tests pass, and some additional unit tests were added.


src/class-validator.php
```diff
@@ -0,0 +1,71 @@
...
+declare(strict_types=1);
+
+namespace Parsely;
+
+/**
+ * Contains a variety of validation functions.
+ *
+ * @since 3.9.0
+ */
+class Validator {
```
mjangda, line 18: Rather than just true / false, I think we should return a `WP_Error` object when the value is invalid. That way, for validations where we have multiple checks, we can provide more specific guidance on what the problem is (e.g. length vs allowed characters), both for end users and also in automated tests.

Not a blocker since we're just extracting existing code but something to consider.

src/UI/class-settings-page.php
```diff
@@ -1141,15 +1143,15 @@ private function validate_advanced_section( $input ) {
...
-					add_settings_error(
-						Parsely::OPTIONS_KEY,
-						'metadata_secret',
-						__( 'The Metadata Secret was not saved because it is incorrect. Please contact Parse.ly support!', 'wp-parsely' )
-					);
-					$input['metadata_secret'] = $options['metadata_secret'];
-				}
+			$metadata_secret_length   = strlen( $input['metadata_secret'] );
+			if ( $metadata_secret_length > 0 &&
+					false === Validator::validate_metadata_secret( $input['metadata_secret'] ) ) {
```
mjangda, line 1148: `$metadata_secret_length > 0` is unnecessary here. That's the job of the validator now.

```suggestion
			if ( false === Validator::validate_metadata_secret( $input['metadata_secret'] ) ) {
```

src/UI/class-settings-page.php
```diff
@@ -955,7 +956,8 @@ private function validate_basic_section( $input ) {
 
 			$input['api_secret'] = $this->get_unobfuscated_value( $input['api_secret'], $this->parsely->get_api_secret() );
 			$api_secret_length   = strlen( $input['api_secret'] );
-			if ( $api_secret_length > 0 && $api_secret_length <= 30 ) {
+			if ( $api_secret_length > 0 &&
```
mjangda, line 959: The length check is unnecessary here since the validator should be doing it.

tests/Unit/ValidatorTest.php
```diff
@@ -0,0 +1,130 @@
...
+	/**
+	 * Verifies that invalid Site IDs fail validation.
+	 *
+	 * @since 3.3.0
+	 * @since 3.9.0 Extracted from test_validate_site_ids() as a separate test.
+	 *
+	 * @covers \Parsely\Validator::validate_site_id
+	 */
+	public function test_validate_invalid_site_ids(): void {
+		$invalid_values = array(
```
mjangda, line 54: Data Providers are a nice way to loop through test cases: https://phpunit.de/manual/3.7/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers.examples.DataTest.php

mjangda (approved):
Nice improvement! Left some suggestions but not blockers.

src/class-validator.php
acicovic: I agree. We can enhance this when we improve validation.

tests/Unit/ValidatorTest.php
acicovic: Thanks for the mention. I discovered data providers very recently and I'm using them in one of the new PRs. In the specific case, I like the separation because the function names refer to what is exactly being tested. Obviously, this is subjective.

src/UI/class-settings-page.php
acicovic: We could make the validation function accept an empty value as valid through an `$allow_empty` argument, though I would not do that as a default behavior (since an empty value cannot really be a valid secret value). This is why this piece was left there, but I understand the reasoning.

src/UI/class-settings-page.php
acicovic: Same as above.

PR #1733, opened 2023-07-21
Improve PHP tests

acicovic:
## Description
This PR improves our PHP tests by:
- Replacing almost all instances of `assertEquals()` to `assertSame()`, improving the checks with strict equality comparison.
- Improving test function comments (mainly whitespace inconsistencies).
- Improving test function declarations by adding type hints.

As a bonus, PHPStan type coverage has now increased from 92% to 94% for return types, and from 65% to 72% for param types.

## Motivation and context
Fix inconsistencies and bring some quick and easy improvements.

## How has this been tested?
PHP tests pass.


tests/Integration/Metadata/AuthorArchiveTest.php
```diff
@@ -68,7 +68,7 @@ public function test_author_archive(): void {
 		$this->assert_data_has_required_properties( $structured_data );
 
 		// The headline should be the category name.
-		self::assertEquals( 'Author - parsely', $structured_data['headline'] ?? null );
-		self::assertEquals( $author_posts_url, $structured_data['url'] ?? null );
+		self::assertSame( 'Author - parsely', $structured_data['headline'] ?? null );
+		self::assertSame( $author_posts_url, $structured_data['url'] ?? null );
```
mjangda, line 72: Aside: `?? null` is a bug waiting to happen :)

The value would either match `$structured_data['url']` or null and there should be separate tests if we're expecting both cases.

Not a huge issue, but something we should watch out for since this pattern is used several times in our tests.

phpstan.neon
```diff
@@ -13,8 +13,8 @@ parameters:
     - vendor/php-stubs/wordpress-stubs/wordpress-stubs.php
     - vendor/php-stubs/wordpress-tests-stubs/wordpress-tests-stubs.php
   type_coverage:
-      return_type: 92
-      param_type: 65
+      return_type: 94
+      param_type: 72
       property_type: 0 # We can't use property types until we support PHP 7.4.
```
mjangda, line 18: Aside: we support 7.4; should this be changed?

mjangda (approved):
This is great!

phpstan.neon
acicovic: Good question!

Property types were introduced in PHP 7.4. We have PHP 7.2 set as the minimum for the plugin, and internally we're still using PHP 7.1 syntax.

If I try to introduce this somewhere (e.g. `protected array $metadata = array();`), I get complaints from IntelliSense (`PHP 7.4+ syntax found. Targeting PHP 7.1.0.`) and PHPCS (`Typed properties are not supported in PHP 7.3 or earlier`).

So I think the comment is valid, it just needs an improved phrasing. I'll add this into my todo list.

tests/Integration/Metadata/AuthorArchiveTest.php
acicovic: Noted. I'll look into this further once I get the chance.

phpstan.neon
acicovic: Addressed in https://github.com/Parsely/wp-parsely/pull/1752

PR #1763, opened 2023-08-10
PCH Performance Details: Fix "Date N/A"

acicovic:
## Description
We had "Date N/A" being displayed in the Performance Details pane of the Content Helper. This PR fixes the issue and simplifies the related code.

## Motivation and context
Fix the bug and simplify the code, making it less prone to fail in the future.

## How has this been tested?
Unit tests were adjusted and pass.


src/content-helper/common/utils/date.ts
```diff
@@ -13,20 +13,14 @@ const DATE_NOT_AVAILABLE_MESSAGE = __( 'Date N/A', 'wp-parsely' );
...
-	if ( 'string' === typeof date ) {
-		// Verify that date format is "YYYY-MM-DD".
-		if ( ! /^\d{4}-\d{2}-\d{2}$/.test( date ) ) {
-			return false;
-		}
-
-		date = new Date( date );
+export function canProcessDate( date: Date ): boolean {
+	// Return false if the date is not a valid Date object.
+	if ( isNaN( +date ) ) {
```
mehmoodak, line 18: It seems like we can remove this `canProcessDate` function and use `isNaN( date )` directly to check if the data is valid or not.

src/content-helper/common/utils/date.ts
```diff
@@ -13,20 +13,14 @@ const DATE_NOT_AVAILABLE_MESSAGE = __( 'Date N/A', 'wp-parsely' );
...
+		return false;
 	}
 
-	const isValidDateObject = date instanceof Date && ! isNaN( +date );
-	const isNotEpochDate = 0 !== date.getTime();
-
-	return isValidDateObject && isNotEpochDate;
+	// Return false if the date is the Unix Epoch time.
+	return 0 !== date.getTime();
```
mehmoodak, line 23: The comment is not entirely true because `new Date( 86400000 )` will pass this check.

I think we should remove this condition also because not sure what value it is adding.

mehmoodak (approved):

src/content-helper/common/utils/date.ts
acicovic: Having a named function makes code more semantic, and also allows us to add to or change our logic in the future more easily if needed.

src/content-helper/common/utils/date.ts
acicovic: > The comment is not entirely true because new Date( 86400000 ) will pass this check.

Unless I misunderstood your comment, `new Date( 86400000 )` doesn't resolve to the epoch time.

> I think we should remove this condition also because not sure what value it is adding.

Tested a bit, and actually removing this will also allow invalid dates like `new Date( '2023-03-32' )`. I think that the current implementation is a good enough compromise. JS is just weird with dates.

PR #1766, opened 2023-08-10
Fix risky Validator unit tests

acicovic:
## Description
Some of our `Validator` unit tests would be flagged as risky due to lack of a `@covers` docblock. This PR fixes this.

## Motivation and context
Don't flag valid tests as risky.

## How has this been tested?
Ran the tests locally and they were flagged as risky. Fixed and ran tests again, and now they pass without being flagged as risky.


acicovic:
Merging this as the same code had been approved by @mehmoodak [here](https://github.com/Parsely/wp-parsely/pull/1761).

Issue #1161, opened 2022-11-07
Implement tests where needed


acicovic:
## Is your feature request related to a problem?
We strive to write tests for all new code that is added to the plugin. Due to workload, sometimes tests come after the integration of the code.

This thread is:
- A catch-all to track where tests need to be implemented for recent code.
- A call to action to examine what tests should be written for older code that is of high impact or prone to breakages.

## Describe the solution you'd like
1. Write tests (unit, integration, E2E or combination) where applicable for recent code.
2. If resources allow it, expand this effort to older code that is of high impact or prone to breakages.

## Recent code that has no or needs more tests
- [ ] https://github.com/Parsely/wp-parsely/issues/986.
- [ ] https://github.com/Parsely/wp-parsely/pull/1159.
- [ ] PARTIALLY DONE - After a bug which came from a [missing require](https://github.com/Parsely/wp-parsely/pull/959) (which also resulted in our [implementation of autoloading](https://github.com/Parsely/wp-parsely/pull/985)), we noticed that our metadata `Tag_Builder` and `Date_Builder` classes have no tests.
- [ ] The Dashboard Widget needs to have tests implemented (We've added integration tests [here](https://github.com/Parsely/wp-parsely/blob/6e930acd1b44e59aa92af6a3a8e394fa28506200/tests/Integration/content-helper/ContentHelperDashboardWidgetTest.php) as part of #1539.
- [ ] https://github.com/Parsely/wp-parsely/pull/1784#discussion_r1294436552
- [ ] https://github.com/Parsely/wp-parsely/pull/1772#discussion_r1294434065
- [ ] https://github.com/Parsely/wp-parsely/pull/1789#discussion_r1298124726

## Other related issues
- [x] DONE - https://github.com/Parsely/wp-parsely/issues/1121

acicovic:
I'll be closing this issue. Generally, tests are being added to new code when needed, or when bugs get fixed. The issue as it stands is broad and is a bit of a catch-all, making it hard to resolve in a short amount of time.

PR #1772, opened 2023-08-11
Disable plugin setup message in managed credentials mode

acicovic:
## Description
With this PR, we're disabling the setup warning message when the plugin is in managed mode.

## Motivation and context
Improve user experience (don't provide users with confusing messages).

## How has this been tested?
Existing tests pass and a test new case has been added.


src/UI/class-admin-warning.php
```diff
@@ -73,6 +73,10 @@ private function should_display_admin_warning(): bool {
 			return false;
 		}
 
+		if ( defined( 'is_managed' ) && true === constant( 'is_managed' ) ) {
```
mehmoodak: I thought we were going to return `is_managed` value in `wp_parsely_credentials` filter as an array key/value pair .... or is it not the case?

mehmoodak (changes requested):

src/UI/class-admin-warning.php
mehmoodak: If possible kindly add tests for this also.

acicovic: About the test, please see the "How has this been tested?" section. Noting this in #1161 though, so we can consider doing this in the future.

About the Parse.ly `wp_parsely_credentials` filter returning an `is_managed` value, there were no such plans or discussions to my knowledge.

Isn't `is_managed` a PHP constant that gets set in the platform side? If not, we'll need to revise our approach.

mehmoodak: As far as I know `is_managed => true` will be return via `wp_parsely_credentials` filter if the plugin is managed by Platform. cc: @mjangda for double check if my understanding is wrong.

acicovic: Noting that there are 2 "managed" aspects in the plugin: `credentials` and `options` which are governed by their respective filters. Credentials and filters behave differently, and thus we've got 2 different filters to manage them at the moment (they could possibly be managed by a single filter, but this would be a refactoring).

If we use the `wp_parsely_credentials` filter, the passed credentials will be in managed mode (read-only or hidden). Same for options with `wp_parsely_managed_options`. This allows any combination of settings, and allows different behaviors if the platform sets different filters for different user types or situations.

mehmoodak: `is_managed => true` will only be return by `wp_parsely_credentials` and plugin should have been able to use this for any purpose.

mjangda: This should the address the issue in this PR:

```suggestion
		if ( $this->parsely->are_credentials_managed ) {
```

To Mehmood's point though, we have a separate issue described here: https://github.com/Automattic/vip-parsely/issues/53 (which can be fixed in a separate PR).

Basically, there will be cases where we have a site that we expect to have managed credentials but the credentials will not be generated yet. And we want to restrict that site's ability to manage the credentials until they are generated. Currently, `are_credentials_managed` only works if we pass in valid values. So, the idea is to be able to set another flag that tells the plugin that this site has its credentials managed, so don't expose things related to that (form controls, warnings, etc.) that customers can't action.

acicovic: OK, this makes sense. I'd examine this flag as above in the function that sets the `are_credentials_managed` value, so we have a single check for everything. Would this work for everybody?

mehmoodak: Works for me (but can sure after viewing the changes)

acicovic: Available in https://github.com/Parsely/wp-parsely/pull/1790.

acicovic:
This should be merged after we finalize https://github.com/Parsely/wp-parsely/pull/1790.

mehmoodak (approved):

PR #1785, opened 2023-08-15
Try to fix failing E2E test

acicovic:
## Description
This PR was used as a testbed to address a GitHub E2E test failure that wasn't reproducible locally for me.


tests/e2e/specs/settings-track-post-types-as.spec.ts
mehmoodak, line 1: I think we shouldn't do this and find the root cause. Please feel free to take another opinion.

acicovic: For full context:

My opinion is that it is OK to take temporary workarounds sometimes when our workflow is hindered, and we can't fix the issue immediately because there are more significant priorities to tackle.

FWIW, I've already dedicated some time to fix this with no result so far. This only fails on GitHub and makes E2E tests fail constantly, which makes all our PRs red even if everything is fine with them. Rerunning them hasn't helped.

@mjangda, I guess since we have different opinions, we can do whatever you think best.

acicovic:
Closing in favor of #1851 because this PR contains lots of junk commits.

PR #1789, opened 2023-08-17
Analytics_Posts_API: Increase request timeout

### LINKED ISSUE #1741: Post List Stats: Increase data fetching timeout
From time to time, the PCH Post List Stats will show an error, resembling the following:

```
Error while getting data for Parse.ly Stats.
Detail: (http_request_failed) cURL error 28: Operation timed out after 10000 milliseconds with 53248 bytes received
```

Our discussion with the Parse.ly back-end team revealed that little can be done from the API's side.

Try to increase the timeout from our side, and see whether this creates performance problems or not.

Leave things as they are.

acicovic:
## Description
This PR:
- Adds the ability to all classes extending `Remote_API_Base` to be able to set their own particular request options like timeout, headers, and [all other WP_Http::request accepted arguments](https://developer.wordpress.org/reference/classes/wp_http/request/#parameters).
- Sets a custom timeout to the `Analytics_Posts_API` because of [timeout issues](https://github.com/Parsely/wp-parsely/issues/1741). The new timeout is currently set to `30 seconds`, but we're awaiting feedback and this might be subject to change.

## Motivation and context
- Closes #1741.
- Gives us the flexibility to set custom request options per endpoint.

## How has this been tested?
- Checked that the code runs and that the timeout is passed into the request.
- This work is based on a private branch that communicates with an API with large timeouts, and it has worked.
- Due to the hard-to-reproduce nature of the issue, we'll need to see what customers report after this gets released.

### DIFF
--- a/src/RemoteAPI/class-analytics-posts-api.php
+++ b/src/RemoteAPI/class-analytics-posts-api.php
@@ -86,7 +86,7 @@ class Analytics_Posts_API extends Remote_API_Base {
 	 *
 	 * @return array<string, mixed> The array of options.
 	 */
-	protected function set_request_options(): array {
+	protected function get_request_options(): array {
 		return array(
 			'timeout' => 30, // phpcs:ignore WordPressVIPMinimum.Performance.RemoteRequestTimeout.timeout_timeout
 		);
--- a/src/RemoteAPI/class-analytics-posts-api.php
+++ b/src/RemoteAPI/class-analytics-posts-api.php
@@ -80,7 +80,7 @@ class Analytics_Posts_API extends Remote_API_Base {
 	}
 
 	/**
-	 * Sets the request's options for the remote API call.
+	 * Returns the request's options for the remote API call.
 	 *
 	 * @since 3.9.0
 	 *
--- a/src/RemoteAPI/class-analytics-posts-api.php
+++ b/src/RemoteAPI/class-analytics-posts-api.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Class for Analytics Posts API (`/analytics/post`).
+ * Class for Analytics Posts API (`/analytics/posts`).
  *
  * @package Parsely
  * @since   3.4.0
@@ -10,11 +10,10 @@ declare(strict_types=1);
 
 namespace Parsely\RemoteAPI;
 
-use Parsely\Parsely;
 use WP_Error;
 
 /**
- * Class for Analytics Posts API (`/analytics/post`).
+ * Class for Analytics Posts API (`/analytics/posts`).
  *
  * @since 3.4.0
  *
@@ -78,5 +77,18 @@ class Analytics_Posts_API extends Remote_API_Base {
 	 */
 	public function get_posts_analytics( $api_params ) {
 		return $this->get_items( $api_params, true ); // @phpstan-ignore-line
+	}
+
+	/**
+	 * Sets the request's options for the remote API call.
+	 *
+	 * @since 3.9.0
+	 *
+	 * @return array<string, mixed> The array of options.
+	 */
+	protected function set_request_options(): array {
+		return array(
+			'timeout' => 30, // phpcs:ignore WordPressVIPMinimum.Performance.RemoteRequestTimeout.timeout_timeout
+		);
 	}
 }
--- a/src/RemoteAPI/class-remote-api-base.php
+++ b/src/RemoteAPI/class-remote-api-base.php
@@ -158,16 +158,6 @@ abstract class Remote_API_Base implements Remote_API_Interface {
 			return $response;
 		}
 
-		$response_code = wp_remote_retrieve_response_code( $response );
-		if ( 200 !== $response_code ) {
-			$error_message = wp_remote_retrieve_response_message( $response );
-			if ( '' === $error_message ) {
-				$error_message = __( 'Unknown error', 'wp-parsely' );
-			}
-
-			return new WP_Error( $response_code, $error_message );
-		}
-
 		$body    = wp_remote_retrieve_body( $response );
 		$decoded = json_decode( $body );
 
--- a/src/RemoteAPI/class-remote-api-base.php
+++ b/src/RemoteAPI/class-remote-api-base.php
@@ -151,14 +151,24 @@ abstract class Remote_API_Base implements Remote_API_Interface {
 	 */
 	public function get_items( $query, $associative = false ) {
 		$full_api_url = $this->get_api_url( $query );
+		$options      = $this->set_request_options();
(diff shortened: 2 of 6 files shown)


src/RemoteAPI/class-analytics-posts-api.php
```diff
@@ -79,4 +78,17 @@ class Analytics_Posts_API extends Remote_API_Base {
...
 	}
+
+	/**
+	 * Sets the request's options for the remote API call.
+	 *
+	 * @since 3.9.0
+	 *
+	 * @return array<string, mixed> The array of options.
+	 */
+	protected function set_request_options(): array {
```
mehmoodak: The name here should be `get_request_options`

src/RemoteAPI/class-analytics-posts-api.php
```diff
@@ -79,4 +78,17 @@ class Analytics_Posts_API extends Remote_API_Base {
...
+		return array(
+			'timeout' => 30, // phpcs:ignore WordPressVIPMinimum.Performance.RemoteRequestTimeout.timeout_timeout
```
mehmoodak, line 91: As far as I know Parsely API have a timout of `10s` so how about we use it directly instead of making it configurable via base class?

src/RemoteAPI/class-remote-api-base.php
```diff
@@ -151,14 +151,24 @@ public function get_api_url( array $query ): string {
...
+		$options      = $this->set_request_options();
+		$response     = wp_safe_remote_get( $full_api_url, $options );
 
-		$result = wp_safe_remote_get( $full_api_url, array() );
+		if ( is_wp_error( $response ) ) {
+			return $response;
+		}
+
+		$response_code = wp_remote_retrieve_response_code( $response );
+		if ( 200 !== $response_code ) {
```
mehmoodak: Any particular reason of adding this code? Does API returns anything else other than 200 which is not a 4xx or 5xx?

src/RemoteAPI/class-analytics-posts-api.php
acicovic: As Parse.ly API endpoints evolve or get added, they may have different timeouts and we may want to be explicit about the values we set. Although we could hardcode a max value for everything, I think that configurability is useful, since it allows us to also pass other things if the need ever comes up.

mehmoodak: This implementation is fine but if they are hard coding the value then possible we can also do the same and keep our code simple.

acicovic: I like the customizability and flexibility of the current solution. Having said that, no strong feelings if you just prefer us to hardcode a value.

mehmoodak: I like less code 😅

acicovic: In this case, since we'll be possibly querying content generation APIs that may take some time to respond, I'll add a large timeout.

mehmoodak: If the timeout is going to be different then let's keep it and better to have tests to make sure these are working as expected.

acicovic: OK, keeping this part of the code as is then.

acicovic: I'm considering the particular discussion resolved, but feel free to reopen if I missed anything.

src/RemoteAPI/class-analytics-posts-api.php
acicovic: Done in https://github.com/Parsely/wp-parsely/pull/1789/commits/9be6751372f8c6e282dd1d19ee5942d313845c8c

src/RemoteAPI/class-remote-api-base.php
acicovic: This is an API we do not control, so it might return various HTTP codes or an empty error message. We're performing a `wp_safe_remote_get()` request and handling the result if its an error, regardless of its HTTP code.

Do you see any particular issue with the code? I'm unsure I understand the reason for this question.

mehmoodak: > Do you see any particular issue with the code? I'm unsure I understand the reason for this question.

As far as I know `is_wp_error` already handles everything which is not a 2xx.

Just for an example say API returns 204/206 then we will throw the error as per this code which is technically not right.

tests/Integration/Endpoints/AnalyticsPostsProxyEndpointTest.php
```diff
@@ -228,7 +229,11 @@ public function test_get_items(): void {
 			function () use ( &$dispatched ) {
 				$dispatched++;
 				return array(
-					'body' => '{"data":[
+					'response' => array(
+						'code'    => 200,
+						'message' => 'OK',
+					),
+					'body'     => '{"data":[
```
mehmoodak: These changes can be removed based on the decision of above comment.

mehmoodak (changes requested):

src/RemoteAPI/class-remote-api-base.php
acicovic: Trying to remember why this code was introduced, it was probably because an experimental API endpoint had a particular behavior. But this is not applicable right now, so removed it.

tests/Integration/Endpoints/AnalyticsPostsProxyEndpointTest.php
acicovic: Removed.

acicovic:
30 seconds seems like a reasonable timeout according to @flavius-popan, so we can keep it as is.

src/RemoteAPI/class-remote-api-base.php
```diff
@@ -198,4 +198,15 @@ public function is_user_allowed_to_make_api_call(): bool {
...
 	}
+
+	/**
+	 * Returns the request's options for the remote API call.
+	 *
+	 * @since 3.9.0
+	 *
+	 * @return array<string, mixed> The array of options.
+	 */
+	protected function get_request_options(): array {
```
mehmoodak, line 209: Kindly add tests to make sure that this is getting called while requesting data from API.

mehmoodak (approved):
Left a comment but approving the PR.

src/RemoteAPI/class-remote-api-base.php
acicovic: Unsure about how we can test this.

mehmoodak: Add a test spy on it and make sure that it is getting called and correct values are returning.

PR #1790, opened 2023-08-17
Improve managed credentials check

acicovic:
## Description
This PR improves the managed credentials check by returning the `is_managed` field value if it is set and it is a boolean.

## Motivation and context
This PR is a continuation of an internal discussion.

## How has this been tested?
Added a couple of test cases.

### DIFF
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -640,6 +640,10 @@ class Parsely {
 			return false;
 		}
 
+		if ( isset( $credentials['is_managed'] ) && true === $credentials['is_managed'] ) {
+			return true;
+		}
+
 		$required_credentials = array( 'site_id', 'api_secret' );
 
 		foreach ( $required_credentials as $key ) {
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -640,8 +640,9 @@ class Parsely {
 			return false;
 		}
 
-		if ( isset( $credentials['is_managed'] ) && true === $credentials['is_managed'] ) {
-			return true;
+		if ( isset( $credentials['is_managed'] ) &&
+				is_bool( $credentials['is_managed'] ) ) {
+			return $credentials['is_managed'];
 		}
 
 		$required_credentials = array( 'site_id', 'api_secret' );
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -640,23 +640,10 @@ class Parsely {
 			return false;
 		}
 
-		if ( isset( $credentials['is_managed'] ) &&
-				is_bool( $credentials['is_managed'] ) ) {
+		if ( isset( $credentials['is_managed'] ) ) {
 			return $credentials['is_managed'];
 		}
 
-		$required_credentials = array( 'site_id', 'api_secret' );
-
-		foreach ( $required_credentials as $key ) {
-			if (
-				! isset( $credentials[ $key ] ) ||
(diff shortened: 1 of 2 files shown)


src/class-parsely.php
```diff
@@ -634,6 +634,10 @@ private function get_managed_credentials() {
 	 * @return bool Whether credentials are being managed at the platform level.
 	 */
 	private function are_credentials_managed(): bool {
+		if ( defined( 'is_managed' ) && true === constant( 'is_managed' ) ) {
+			return true;
+		}
```
mehmoodak: Who will define this constant?

As per my understanding this `is_managed` value will be return by platform as an array key inside the `wp_parsely_credentials` filter data.

acicovic: So when the platform sets that, it will do it like so in its code?
```
add_filter(
	'wp_parsely_credentials',
	function() {
		return array(
			'site_id'    => '',
			'api_secret' => '',
			'is_managed' => true
		);
	}
);
```
Is this an understanding, or do we have documentation that confirms the decision?

acicovic: If we don't know what the platform will do exactly, we need to sort this out first.

mehmoodak: > So when the platform sets that, it will do it like so in its code?

Yes

acicovic: OK, thanks. I'll work on a solution for that.

acicovic: Updated, I hope it's what we want.

src/class-parsely.php
```diff
@@ -640,6 +640,11 @@ private function are_credentials_managed(): bool {
 			return false;
 		}
 
+		if ( isset( $credentials['is_managed'] ) &&
+				is_bool( $credentials['is_managed'] ) ) {
```
mehmoodak: `is_managed` will always be boolean so we can remove this condition.

```
is_bool( $credentials['is_managed'] )
```

src/class-parsely.php
```diff
@@ -640,6 +640,11 @@ private function are_credentials_managed(): bool {
...
+			return $credentials['is_managed'];
+		}
+
```
mehmoodak, line 646: We can remove all of the code below and just return `false`.

If `site_id` and `api_secret` is set then `is_managed` will also always be set so the code below is unreachable.

mehmoodak (changes requested):

src/class-parsely.php
acicovic: This is deliberate. Can we be that certain that no user error or bug could occur that makes this a non-boolean? I generally prefer to be cautious, even if the values are platform-provided.

src/class-parsely.php
acicovic: So to conclude, our only criterion is whether `is_managed` is set. We don't care about anything else. Correct?

mehmoodak: Correct

src/class-parsely.php
mehmoodak: It's hard coded to `true` but if you think keeping this is find then keep it.

src/class-parsely.php
acicovic: Updated. Let me know if it suits the needs.

acicovic: Tests are failing after the update, looking into it.

mehmoodak (approved):

PR #1804, opened 2023-08-22
Add early returns to managed credentials functions

acicovic:
## Description
This PR adds a couple of early returns for better code readability and for not processing unneeded code.

## How has this been tested?
Existing tests pass.


src/class-parsely.php
```diff
@@ -608,7 +608,12 @@ private function get_managed_credentials() {
 		}
 
 		$credentials = apply_filters( 'wp_parsely_credentials', array() );
-		$result      = array();
+
+		if ( ! is_array( $credentials ) || 0 === count( $credentials ) ) {
```
mehmoodak, line 612: Can we use empty instead of `0 === count`?

src/class-parsely.php
```diff
@@ -636,7 +641,7 @@ private function get_managed_credentials() {
 	private function are_credentials_managed(): bool {
 		$credentials = apply_filters( 'wp_parsely_credentials', array() );
 
-		if ( ! is_array( $credentials ) ) {
+		if ( ! is_array( $credentials ) || 0 === count( $credentials ) ) {
```
mehmoodak, line 644: Same as above.

mehmoodak (approved):

src/class-parsely.php
acicovic: This was done in the past, but at some point all `empty()` occurrences were removed. From then on, checks were updated to be as specific as possible and this has become a convention for this project. If you search the code, you'll see that `empty()` is not being used anywhere. Generally, many people consider `empty()` problematic and there are articles about that on the web.

PR #1805, opened 2023-08-22
Content Helper: Show hint when HTTP errors are encountered

acicovic:
## Description
While I was offline due to a power outage (no internet), I noticed that PCH errors due to HTTP failures were a bit cryptic and did not provide any context. In my case, the error was simply "A valid URL was not provided."

This PR introduces a hint so users can understand that there's a connectivity issue, while also keeping the original error message for possible troubleshooting.

## Motivation and context
Give a clue to users that the API is unreachable and that they might be offline.

## How has this been tested?
- Existing tests pass.
- Added a test to verify that this type of error gets retried.

image


mehmoodak:
I tried to replicate this case and turn my WiFi off but got this which seems completely fine for the case. Are you doing anything different?

image

acicovic:
This is interesting.

Turning off my router or disabling my LAN network through `System Settings > Network` resulted in the screenshot I posted.

For whatever reason, you're getting a different error and thus a different message.

mehmoodak (approved):

PR #1816, opened 2023-08-24
Fix PHPStan errors

acicovic:
## Description
The recent `composer update` I ran revealed some PHPStan errors that are getting fixed in this PR.

## Motivation and context
Eliminate PHPStan errors and hindered workflow.

## How has this been tested?
Existing tests pass.


src/RemoteAPI/class-remote-api-base.php
```diff
@@ -151,8 +153,13 @@ public function get_api_url( array $query ): string {
 	 */
 	public function get_items( $query, $associative = false ) {
 		$full_api_url = $this->get_api_url( $query );
-		$options      = $this->get_request_options();
-		$response     = wp_safe_remote_get( $full_api_url, $options );
+		/**
+		 * GET request options.
+		 *
+		 * @var WP_HTTP_Request_Args $options
+		 */
```
mehmoodak, line 160: Instead of adding this as comment I think we can improve type of `get_request_options` function.

src/class-parsely.php
```diff
@@ -41,6 +41,26 @@
  *   plugin_version: string,
  * }
  *
+ * @phpstan-type WP_HTTP_Request_Args array{
```
mehmoodak, line 45: 1. I think we shouldn't have `WP_` in interface because this indicates that the type is defined via WordPress but in actual it is not.

2. Maybe we should only define the keys that we are using in our codebase.

mehmoodak (changes requested):

src/RemoteAPI/class-remote-api-base.php
acicovic: Unless I'm doing something wrong, adding PHPStan types as `@return` comments have significant negative side-effects in my experience. Let's take the following function for example:

```php
	/**
	 * Returns the request's options for the remote API call.
	 *
	 * @since 3.9.0
	 *
	 * @return array<string, mixed> The array of options.
	 */
	protected function get_request_options(): array {
		return array();
	}
```

Now let's add the PHPStan type to the `@return` comment:

```php
	/**
	 * Returns the request's options for the remote API call.
	 *
	 * @since 3.9.0
	 *
	 * @return WP_HTTP_Request_Args The array of options.
	 */
	protected function get_request_options(): array {
		return array();
	}
```

We now got a PHPStan error:
```
Method Parsely\RemoteAPI\Remote_API_Base::get_request_options() should return array{method: string, timeout: float, redirection: int, httpversion: string, user-agent: string, reject_unsafe_urls: bool, blocking: bool, headers: array<string, string>, ...} but returns array{}.
```

OK, let's remove the function's return type then:
```php
	/**
	 * Returns the request's options for the remote API call.
	 *
	 * @since 3.9.0
	 *
	 * @return WP_HTTP_Request_Args The array of options.
	 */
	protected function get_request_options() {
		return array();
	}
```

We now get 2 PHPStan errors:
```
Out of 5 possible return types, only 60 % actually have it. Add more return types to get over 94 %.
Method Parsely\RemoteAPI\Remote_API_Base::get_request_options() should return array{method: string, timeout: float, redirection: int, httpversion: string, user-agent: string, reject_unsafe_urls: bool, blocking: bool, headers: array<string, string>, ...} but returns array{}.
```

OK, let's add a `@var` comment then:
```php
	/**
	 * Returns the request's options for the remote API call.
	 *
	 * @since 3.9.0
	 *
	 * @return WP_HTTP_Request_Args The array of options.
	 */
	protected function get_request_options() {
		/**
		 * GET request options.
		 *
		 * @var WP_HTTP_Request_Args
		 */
		return array();
	}
```

We're still getting the `Out of 5 possible return types, only 60 % actually have it. Add more return types to get over 94 %.` error though. And we need to perform this manipulation to other functions that are now incompatible with this one. We can try adding `: array` back to the function, but this only leads to other errors.

The conclusion is that not only our code has become bulkier, but also our global return type coverage has lowered, and we would need to lower our global return type coverage in `phpstan.neon`. This is why:
- I'm totally against using PHPStan types as parameter or return types.
- I'm totally in favor of using normal PHP types because of the many benefits they bring, including type recognition and parameter/return type coverage increase.

Please let me know if you think I'm missing something.

src/class-parsely.php
acicovic: > I think we shouldn't have WP_ in interface because this indicates that the type is defined via WordPress but in actual it is not.

This is an exact copy of [WP_Http::request args](https://developer.wordpress.org/reference/classes/wp_http/request/#parameters), so this is why `WP` was included. But I can remove it if you feel strongly about it.

acicovic: > Maybe we should only define the keys that we are using in our codebase.

Removing the keys works as well, but did as you recommended.

src/RemoteAPI/class-remote-api-base.php
mehmoodak: > Method Parsely\RemoteAPI\Remote_API_Base::get_request_options() should return array{method: string, timeout: float, redirection: int, httpversion: string, user-agent: string, reject_unsafe_urls: bool, blocking: bool, headers: array<string, string>, ...} but returns array{}.

It seems like we can marked the fields as optional in PHPStan type to avoid this issue.

src/class-parsely.php
mehmoodak: > But I can remove it if you feel strongly about it.

No strong preferences just mentioning it because it seems like we are breaking a convention.

mehmoodak (approved):
Left some comments but approving the PR because these are just PHPStan changes that have no effect on functionality.

src/class-parsely.php
acicovic: OK, got it. Will leave as is for now because IMO it makes sense to have it, but if we see that this confuses us at any point we'll remove the `WP_` prefix.

src/RemoteAPI/class-remote-api-base.php
acicovic: Experimented with this a bit more, but couldn't get to a state without errors. So keeping as is, but happy to re-discuss about it in the future if we think it's important.

Issue #1796, opened 2023-08-18
Upgrade dependencies that fail node builds


acicovic:
## Is your feature request related to a problem?
We've got an increasing number of dependabot PRs that fail node builds for various reasons. As we're focusing on other things, we've been postponing taking care of this, but it will definitely become an issue at some point.

For reference:
- #1794
- #1793
- #1783
- #1779

## Describe the solution you'd like
Find why the builds fail and fix the PRs so we can merge them.

## Additional context
Related/similar to #1265.

acicovic:
This should be fixed in #1847.

PR #1847, opened 2023-09-06
Upgrade to React 18 and fix depedency issues

### LINKED ISSUE #1265: Update React to version 18
Many of the WordPress packages that we use as dependencies are switching to React 18 and we already cannot merge the respective dependabot PRs due to version conflicts. I assume that the situation will only get intensified as time goes by. Yesterday we had 2 PRs like that, today we have 3.

Update to React 18.

## Additional context
Related/similar to #1796.

acicovic:
## Description
We had lots of dependabot PRs that could not pass node builds, creating a general maintenance nightmare like not being able to upgrade our Gutenberg packages. This PR:
- Updates React to version 18.
- Fixes all dependency build failures.
- Brings all dependencies up to date, including all Gutenberg packages.
- Removes the (now uneeded) `@types/wordpress__core-data` package and adds the (now needed) `@types/jest` package.
- Fixes some small issues that resulted from the above updates.

## Motivation and context
- Eliminate tech debt and conflicting dependency loops.
- Closes #1265 and #1796.

## How has this been tested?
- Linting passes and builds succeed without errors.
- Existing tests pass.
- Did a quick manual PCH smoke test (Dashboard Widget + Sidebar).


jest-unit.config.js
```diff
@@ -0,0 +1,9 @@
+// eslint-disable-next-line
```
acicovic, line 1: A particular unit test would fail without this code.

chriszarate: It can be useful to disable the specific rule, so that others know what is being worked around and additional violations aren't introduced.
```suggestion
// eslint-disable-next-line rule-name
```

chriszarate (approved):
Small suggestion

jest-unit.config.js
acicovic: Thanks for the suggestion, I'll look into adding the specific rule here and in a similar file for E2E tests. Just to be clear, the comment about unit tests failing are not referring to the particular line, but to the addition of`transformIgnorePatterns`.

PR #1860, opened 2023-09-15
Remove Intl dependency from utils.php

### LINKED ISSUE #1558: Fix errors in utils.php
## Describe the bug
Presently we have 2 reports that something is wrong in utils.php

```
Fatal error: Uncaught Exception: Constructor failed
in /wp-content/plugins/wp-parsely/src/Utils/utils.php on line 170

    NumberFormatter::__construct()
    wp-content/plugins/wp-parsely/src/Utils/utils.php:170
    Parsely\Utils\get_formatted_number()
    wp-content/plugins/wp-parsely/src/UI/class-admin-columns-parsely-stats.php:348
    Parsely\UI\Admin_Columns_Parsely_Stats::get_parsely_stats_response()
    wp-content/plugins/wp-parsely/src/UI/class-admin-columns-parsely-stats.php:218
    Parsely\UI\Admin_Columns_Parsely_Stats::enqueue_parsely_stats_script_with_data()
    wp-includes/class-wp-hook.php:308
    WP_Hook::apply_filters()
    wp-includes/class-wp-hook.php:332
    WP_Hook::do_action()
    wp-includes/plugin.php:517
    do_action()
    wp-admin/admin-footer.php:78
    require_once()
    wp-admin/edit.php:502
```

```
An error of type E_ERROR was caused in line 152 of the file /wp-content/plugins/wp-parsely/src/Utils/utils.php.
Error message: Uncaught IntlException: Constructor failed in /wp-content/plugins/wp-parsely/src/Utils/utils.php:152

    #0 wp-content/plugins/wp-parsely/src/Utils/utils.php(152): NumberFormatter->__construct('en', 14)
    #1 wp-content/plugins/wp-parsely/src/UI/class-admin-columns-parsely-stats.php(348): Parsely\Utils\get_formatted_number(256)
    #2 wp-content/plugins/wp-parsely/src/UI/class-admin-columns-parsely-stats.php(218): Parsely\UI\Admin_Columns_Parsely_Stats->get_parsely_stats_response(Object(Parsely\RemoteAPI\Analytics_Posts_API))
    #3 wp-includes/class-wp-hook.php(308): Parsely\UI\Admin_Columns_Parsely_Stats->enqueue_parsely_stats_script_with_data('')
    #4 wp-includes/class-wp-hook.php(332): WP_Hook->apply_filters(NULL, Array)
    #5 wp-includes/plugin.php(517): WP_Hook->do_action(Array)
    #6 wp-admin/admin-footer.php(78): do_action('admin_footer', '')
```

The first occurrence was reported [here](https://wordpress.org/support/topic/error-in-quick-edit-2/). When we first investigated the issue, we concluded that for whatever reason, the host's environment did not support the `NumberFormatter::PADDING_POSITION` constant. So the following test code was always failing:
```php
<?php
$f = new NumberFormatter("en_US", NumberFormatter::PADDING_POSITION);
echo $f->format(10000);
?>
```

## Desired solution
Investigate if the reports boil down to the same issue, and if yes consider an alternative implementation without using `intl`, `NumberFormatter` or `PADDING_POSITION`. 

## To Reproduce
The first customer reported they had the issue when visiting wp-admin's edit.php page. More generally, this error should occur on wp-admin pages that integrate the PCH Stats Column, as it is the component that calls the related functions.

## Expected behavior
The functions in question should work without generating errors or crashing wp-admin.

acicovic:
## Description
The `get_formatted_time()` and `get_formatted_number()` functions located in utils.php were relying on the `Intl` PHP extension by using `NumberFormatter`. However, it seems that the said extension is not installed in all hosting environments, which was resulting in fatal errors for certain installations. This PR replaces the functions  that were using Intl with functions that don't.

## Motivation and context
Fixes #1558.

## How has this been tested?
Existing tests pass, and some of them were modified to work with the new functions.

### DIFF
--- a/src/Utils/utils.php
+++ b/src/Utils/utils.php
@@ -172,14 +172,31 @@ function get_formatted_number( $number ): string {
  * @return string
  */
 function get_formatted_time( $seconds ): string {
-	$time_formatter = new NumberFormatter( 'en', NumberFormatter::DURATION );
-	$formatted_time = $time_formatter->format( $seconds );
+	$seconds = round( $seconds );
+	$hours   = floor( $seconds / 3600 );
 
-	if ( false === $formatted_time ) {
-		return '';
+	if ( $hours >= 1 ) {
+		$seconds = $seconds - ( $hours * 3600 );
+		$minutes = floor( $seconds / 60 );
+		$seconds = round( $seconds % 60 );
+
+		return esc_html( /* translators: 1: Number of hours 2: Number of minutes 3: Number of seconds */
+			sprintf( __( '%1$d:%2$02d:%3$02d', 'wp-parsely' ), $hours, $minutes, $seconds )
+		);
 	}
 
-	return $formatted_time;
+	$minutes = floor( $seconds / 60 );
+	$seconds = round( $seconds % 60 );
+
+	if ( $minutes >= 1 ) {
+		return esc_html( /* translators: 1: Number of minutes 2: Number of seconds */
+			sprintf( __( '%1$d:%2$02d', 'wp-parsely' ), $minutes, $seconds )
+		);
+	}
+
+	return esc_html( /* translators: 1: Number of seconds */
+		sprintf( __( '%1$d sec.', 'wp-parsely' ), round( $seconds ) )
+	);
 }
 
 /**
--- a/src/Utils/utils.php
+++ b/src/Utils/utils.php
@@ -12,7 +12,6 @@ declare(strict_types=1);
 
 namespace Parsely\Utils;
 
-use NumberFormatter;
 use WP_Post;
 use WP_Error;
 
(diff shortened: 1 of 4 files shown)


acicovic:
Unsure why we're getting the warnings in CS & Lint. Will be taken care of in a separate PR if they persist.

acicovic:
> Unsure why we're getting the warnings in CS & Lint. Will be taken care of in a separate PR if they persist.

It's because the branch name ends with .php. Renaming the branch will most probably close the PR, so leaving it as is.

src/Utils/utils.php
```diff
@@ -137,26 +136,69 @@ function get_time_format(): string {
...
+ * @param int    $fraction_digits The number of desired fraction digits.
+ * @param string $glue            A string to put between the number and unit.
  *
- * @return string
+ * @return string The number formatted as an imprecise number.
  */
-function get_formatted_number( $number ): string {
-	$number_formatter = new NumberFormatter( 'en', NumberFormatter::PADDING_POSITION );
-	$formatted_number = $number_formatter->format( $number );
+function get_formatted_number( string $value, int $fraction_digits = 1, string $glue = '' ): string {
```
mehmoodak, line 156: The implementation of this function seems complex can we make it something like following which is an easier implementation?

```php
function formatNumber($number) {
    $suffixes = ["", "K", "M", "B", "T", "Q"];
    $index = 0;

    while ($number >= 1000 && $index < count($suffixes) - 1) {
        $number = $number / 1000;
        $index++;
    }

    return round($number, 1) . $suffixes[$index];
}

// Example usage:
echo formatNumber(10000); // Output: 10K
echo formatNumber(1500000); // Output: 1.5M
echo formatNumber(2500000000); // Output: 2.5B
echo formatNumber(3500000000000); // Output: 3.5T
echo formatNumber(4500000000000000); // Output: 4.5Q
```

acicovic: I agree this is complicated. This is the way Parse.ly does imprecise numbers on the dashboard, so the same approach got used for consistency and for not increasing support questions like "why do I see number `x` on the Post List Stats, but number `y` in the Parse.ly dash?".

I'm more than happy to use the simplified version if it proves to have the same output as the Parse.ly dash, which would need writing related tests on our end. The focus of this PR is to just drop the dependency so I'd leave things as they are, but I'm happy to open an issue for going towards a simplification and a verification that our numbers match the Parse.ly dash.

mehmoodak: If Parse.ly code is in PHP then its fine but if its not then I think we should take on the refactor now as this is not a huge effort.

acicovic: The original function is in JS and was ported to typeScript in our code. Then I ported the TS code to PHP. Replacing the function isn't a lot of work, but testing that our numbers match those of the Parse.ly dash and writing tests for that is more work.

mehmoodak: Okay approving the PR then but feel free to take another opinion.

mehmoodak (approved):

PR #1869, opened 2023-09-19
Post Endpoints: Unify data extraction functions

acicovic:
## Description
We were extracting data separately in each of our Post endpoints (`/posts` and `/post/detail`), duplicating some code and opening up the possibility for result inconsistencies in the future. Given that post endpoints return similar data (the exact same in some cases), this PR refactors and centralizes the related code into one single place.

It also:
- Introduces a bugfix to the `get_formatted_duration()` function (single-digit minutes were returned with a leading space).
- Takes into account that the `avg_engaged` metric can be in different places of the Parse.ly API response.

This update was done as part of working on some PCH enhancements, but we're posting it as a separate PR to keep things simpler in the final PCH PR.

## How has this been tested?
- Manually did a quick test to see that PCH still works as expected.
- Existing tests pass (some were tinkered).


src/Endpoints/class-analytics-post-detail-api-proxy.php
```diff
@@ -42,48 +41,10 @@ public function get_items( WP_REST_Request $request ) {
...
-	 *   - $time of 1.005 yields '1:00'.
-	 *   - $time of 1.5 yields '1:30'.
-	 *   - $time of 1.999 yields '2:00'.
-	 *
-	 * @since 3.6.0
-	 *
-	 * @param float $time The time as a float number.
-	 * @return string The resulting formatted time duration.
-	 */
-	private static function get_duration( float $time ): string {
```
acicovic, line 78: Was renamed to `get_formatted_duration` and moved to `utils.php`.

src/Endpoints/class-base-api-proxy.php
```diff
@@ -172,4 +174,89 @@ protected function get_data( WP_REST_Request $request, bool $require_api_secret
...
+	 *
+	 * Should only be used with endpoints that return post data.
+	 *
+	 * @since 3.10.0
+	 *
+	 * @param stdClass $item The object to extract the data from.
+	 *
+	 * @return array<string, mixed> The extracted data.
+	 */
+	protected function extract_post_data( stdClass $item ): array {
```
mehmoodak, line 189: Can we please confirm with Parse.ly Backend team that for both endpoints they are using same code for generating post data object?

- If YES then we can keep this change (although I do prefer to map API responses separately to be cautious and it's also helpful to only have the needed fields after mapping)

- If NO then I think we should also avoid doing this on our side

src/Endpoints/class-base-api-proxy.php
```diff
@@ -172,4 +174,89 @@ protected function get_data( WP_REST_Request $request, bool $require_api_secret
...
+	/**
+	 * Extracts the post data from the passed object.
+	 *
+	 * Should only be used with endpoints that return post data.
+	 *
+	 * @since 3.10.0
+	 *
+	 * @param stdClass $item The object to extract the data from.
+	 *
+	 * @return array<string, mixed> The extracted data.
```
mehmoodak, line 187: Adding proper typing here will be helpful

src/Endpoints/class-base-api-proxy.php
```diff
@@ -172,4 +174,89 @@ protected function get_data( WP_REST_Request $request, bool $require_api_secret
...
+			// Set thumbnail URL, falling back to the Parse.ly thumbnail if needed.
+			$thumbnail_url = get_the_post_thumbnail_url( $post_id, 'thumbnail' );
+			if ( false !== $thumbnail_url ) {
+				$data['thumbnailUrl'] = $thumbnail_url;
+			} elseif ( isset( $item->thumb_url_medium ) ) {
+				$data['thumbnailUrl'] = $item->thumb_url_medium;
+			}
+		}
+
+		ksort( $data );
```
mehmoodak: Why we need to use `ksort` here? If it's a must then I think comment can help.

mehmoodak (changes requested):

src/Endpoints/class-base-api-proxy.php
acicovic: Getting an answer to that could take some time, which is of the essence right now. Regardless, implementations that are out of our control can change without notice. So today it could be the same code, tomorrow it could differ. This shouldn't make any difference for us in our function(s).

If an (unlikely) code change on the Parse.ly back-end results in breaking our implementation (for example if a field was moved or renamed), it would have broken the old implementation as well. So we would end up with the exact same problem, whether we use one or multiple functions.

I prefer a single function that can be flexible rather than code duplication, especially if more `/post*` endpoints get added in the future.

However, if you strongly disagree with the approach here, we can ignore this PR for the time being. However I'll need to know that to update the code to work without this approach.

src/Endpoints/class-base-api-proxy.php
acicovic: This is to have a predictable array, sorted alphabetically by key. We can live without it if you think we should remove it, it doesn't affect functionality.

src/Endpoints/class-base-api-proxy.php
acicovic: You mean something like `@return array<string, string|bool|int|null>`, or you mean using PHPStan types? The reasons I don't use PHPStan types have been discussed in another PR. We use `@return array<string, mixed>` regularly in our codebase, but we can see if it would be good to change that with more specific types.

src/Endpoints/class-base-api-proxy.php
mehmoodak: In that case I think we should remove it because it's an extra operation.

src/Endpoints/class-base-api-proxy.php
mehmoodak: I mean define whole structure of the return object [Parsely_Options](https://github.com/Parsely/wp-parsely/blob/c3fda7876e2f43a382cdd230f6115cd2b32e151e/src/class-parsely.php#L23)

src/Endpoints/class-base-api-proxy.php
mehmoodak: Let's take another opinion cc: @mjangda

src/Endpoints/class-base-api-proxy.php
acicovic: Done

src/Endpoints/class-base-api-proxy.php
acicovic: This would need a new PHPStan type.

I've expressed why I consider them problematic [here](https://github.com/Parsely/wp-parsely/pull/1816#discussion_r1304094701). My opinion remains the same currently and I don't wish to invest further time in that, unless there's a suggestion that fixes the shortcomings mentioned in that thread.

Issue #1661, opened 2023-06-14
Content Helper: Add engaged time filter


danielabloch:
**Customer request: show me what content keeps people on our site**
This customer said they need help identifying what keeps people on their site. They want to know which posts people are reading for the longest time, and which posts keep people on site longest / get them to read other content. 

The two metrics that do this:
1 ) engage time
2) recirculation rate (out of scope)

**Acceptance criteria**
Customer can filter the sidebar data by engaged time (new) instead of only page views (old).

acicovic:
The API provides a host of engaged-related metrics (related page [here](https://docs.parse.ly/api-available-metrics/)). No recirculation data at the moment.

Questions that come to mind:
1. Which of the engaged metrics would we like to use?
2. While the sidebar is mentioned in this issue, the Dashboard Widget is probably a good candidate for that feature?

danielabloch:
Yep, recirculation would need to be added to the API. 
I suggest we use `avg_engaged` (Engaged minutes spent by total visitors)
and agreed that it would be good for both the sidebar and the widget

acicovic:
Recirculation is out of scope.  Adding `avg_time` has been done for:
- [x] Dashboard Widget
- [x] Performance Details pane (PCH Sidebar)
- [x] Top Related Posts pane (PCH Sidebar)

acicovic:
Closing this issue as implemented per the work done in #1876 and #1898.

Issue #1352, opened 2023-02-02
Content Helper: Allow user to select time period filter(s)


danielabloch:
### Is your feature request related to a problem?
Different customers care about different time periods of data (1 day, 7 days, 30 days, etc.). By providing customers with filtering options, we let them see the time frame that is most relevant to them while editing or authoring content in their CMS.

### Describe the solution you'd like
Ability for customers to select 1 of several time period options with which to evaluate their content performance.

Time period filtering option ideas

- Last 24 hours
- Right now (real-time)
- Last 7 days
- Last 30 days

Available on

- Performance Details
- Top Related Posts
- Dashboard Widget

### Describe alternatives you've considered
Creating a default time period that customers cannot edit. We launched this in 3.6. 

### Notes

danielabloch:
We currently offer last 7 days by default.

I recommend we prioritize additional time filters in the following order:
1. Real time: "Real-time data includes the last 24 hours and updates every five seconds. It can be seen throughout the dashboard." ([source](https://docs.parse.ly/does-parse-ly-show-realtime-data/#:~:text=Parse.ly%20provides%20both%20real,to%20the%20Parse.ly%20dashboard.))
2. Last 30 days -- stretch goal only, not a requirement

acicovic:
Waiting for feedback on real-time due to caching issues, this is done for:
- [x] Dashboard Widget
- [x] Performance Details pane (PCH Sidebar)
- [x] Top Related Posts pane (PCH Sidebar)

acicovic:
Closing this issue as implemented per the work done in #1876 and #1898. Real-time is still blocked due to caching issues, but it's under our radar and in internal discussion. If the caching issue gets fixed, we already have the code to implement it.

PR #1876, opened 2023-09-21
Dashboard Widget: Add period and metric filters

acicovic:
## Description
This PR adds period (`1`, `7` and `30 Days`) and metric (`Page Views` and `Avg. Time`) filters to the Dashboard Widget, so users can have more choices regarding the data they're viewing.

This wasn't my initial intention, but this PR turned out to be a small refactoring as well. Sorry for that!

## Motivation and context
- Add more options to the Dashboard Widget.
- Related to https://github.com/Parsely/wp-parsely/issues/1352 and https://github.com/Parsely/wp-parsely/issues/1661.

## How has this been tested?
- Manual testing.
- Existing tests pass.

## Screenshots
image
image


acicovic:
@danielabloch, can you check the screenshots and let me know if the wording in the filter dropdowns looks good to you?

Thanks!

danielabloch:
this is fantastic! looks good to me

src/content-helper/common/filter-select.tsx
```diff
@@ -0,0 +1,94 @@
+/**
+ * Defines the props structure for FilterSelect.
```
mehmoodak: How about we call it just `Select` or `WidgetSelect` (if its only specific to widgets)?

src/content-helper/common/filter-select.tsx
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Returns a prefilled FilterSelect containing time options.
+ *
+ * @since 3.10.0
+ *
+ * @param {PrefilledFilterSelectProps} props The component's props.
+ *
+ * @return {JSX.Element} The JSX Element.
+ */
+export const FilterSelectTime = (
```
mehmoodak: I think we should only have one component in this file i.e. the generic one.

We are only using this `TopPostsFilters` so maybe it's better to also move this component there.

src/content-helper/common/filter-select.tsx
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Returns a prefilled FilterSelect containing metric options.
+ *
+ * @since 3.10.0
+ *
+ * @param {PrefilledFilterSelectProps} props The component's props.
+ *
+ * @return {JSX.Element} The JSX Element.
+ */
+export const FilterSelectMetric = (
```
mehmoodak: Same as above.

src/content-helper/dashboard-widget/provider.ts
```diff
@@ -25,19 +25,21 @@ interface TopPostsApiResponse {
...
 	/**
 	 * Returns the site's top posts.
 	 *
+	 * @param {string} period The period to fetch data for.
+	 * @param {string} metric The metric to sort by.
+	 *
 	 * @return {Promise<Array<TopPostData>>} Object containing message and posts.
 	 */
-	public async getTopPosts(): Promise<TopPostData[]> {
+	public async getTopPosts( period: string, metric: string ): Promise<TopPostData[]> {
```
mehmoodak: 1. `timePeriod` instead of just period which is more descriptive.
1. If backend only supports few values for both of these params then how about using enums here instead of string?

src/content-helper/dashboard-widget/components/top-posts-list-item.tsx
```diff
@@ -104,13 +108,54 @@ function getPostThumbnailElement( { post }: TopPostListItemProps ): JSX.Element
...
+					{ __( 'Average Time', 'wp-parsely' ) }
+				</span>
+				{ post.avgEngaged }
+			</span>
+		);
+	}
+
+	return (
+		<span className="parsely-top-post-metric-data">
+			{ __( 'Unsupported metric specified.', 'wp-parsely' ) }
```
mehmoodak: Isn't it too long .... how about a simple `-`?

src/content-helper/dashboard-widget/components/top-posts-list-item.tsx
```diff
@@ -85,7 +89,7 @@ export function TopPostListItem( { post }: TopPostListItemProps ): JSX.Element {
  *
  * @param {TopPostData} post The Post from which to get the data.
  */
-function getPostThumbnailElement( { post }: TopPostListItemProps ): JSX.Element {
+function ListItemThumbnail( { post }: TopPostDataProps ): JSX.Element {
```
mehmoodak, line 92: Good to see these changes 👍

src/content-helper/dashboard-widget/components/top-posts-list.tsx
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Returns an ordered list element containing Top Posts.
+ *
+ * @since 3.10.0
+ *
+ * @param {TopPostsListProps} props The component's props.
+ *
+ * @return {JSX.Element} The JSX Element.
+ */
+export function TopPostsList( { posts, metric }: TopPostsListProps ): JSX.Element {
```
mehmoodak: It will be more simpler and maintanable if we don't make this as a separate component. It does makes sense if we have some logic inside the component but this is just a simple loop.

src/content-helper/dashboard-widget/components/top-posts.tsx
```diff
@@ -3,31 +3,33 @@
...
 /**
  * List of the top posts.
  */
-export function TopPostList() {
+export function TopPosts() {
 	const [ loading, setLoading ] = useState<boolean>( true );
 	const [ error, setError ] = useState<ContentHelperError>();
 	const [ posts, setPosts ] = useState<TopPostData[]>( [] );
+	const [ period, setPeriodFilter ] = useState<string>( '7' );
+	const [ metric, setMetricFilter ] = useState<string>( 'views' );
```
mehmoodak: If we use enums then we can implement strong typing here and above.

src/content-helper/dashboard-widget/components/top-posts-filters.tsx
```diff
@@ -0,0 +1,38 @@
...
+}
+
+/**
+ * Returns a div element containing filter controls for the Top Posts widget.
+ *
+ * @since 3.10.0
+ *
+ * @param {TopPostsFiltersProps} props The component's props.
+ */
+export function TopPostsFilters(
```
mehmoodak: It will be more simpler and maintanable if we don't make this as a separate component. It does makes sense if we have some logic inside the component but these is just a simple wrapper.

mehmoodak (changes requested):

src/content-helper/common/filter-select.tsx
acicovic: Will be renaming this to Select.

src/content-helper/common/filter-select.tsx
acicovic: I'll integrate it into the `TopPostsFilters()` function itself. If we need to reuse it in the future, we extract it.

src/content-helper/common/filter-select.tsx
acicovic: I'll integrate it into the `TopPostsFilters()` function itself. If we need to reuse it in the future, we extract it.

src/content-helper/dashboard-widget/components/top-posts-list-item.tsx
acicovic: You're right, this isn't the whole row we're populating.

acicovic: Noting that when we've got a case like that, the API shows an error anyway and nothing is displayed.

src/content-helper/dashboard-widget/provider.ts
acicovic: 1. I think `period` is fine, and it's in line with Parse.ly API params. Let's leave it for brevity.
2. Will work to introduce Enums.

src/content-helper/dashboard-widget/components/top-posts-filters.tsx
acicovic: Will remove.

src/content-helper/dashboard-widget/components/top-posts-list.tsx
acicovic: OK, will remove.

src/content-helper/dashboard-widget/components/top-posts.tsx
acicovic: Will do.

acicovic:
Feedback addressed as described. Can you please take a look again? 🙂

src/content-helper/dashboard-widget/components/top-posts.tsx
```diff
@@ -0,0 +1,117 @@
...
+			<div className="parsely-top-posts-filters">
+				<Select
+					defaultValue={ period }
+					items={ [
+						[ Period.Day, 'Last 24 hours' ],
+						[ Period.Week, 'Last 7 days' ],
+						[ Period.Month, 'Last 30 days' ],
+					] }
+					onChange={ ( event ) => {
+						if ( Object.values( Period ).includes( event.target.value as Period ) ) {
```
mehmoodak, line 89: We can also do

```
if ( Period[ event.target.value ] )
```

src/content-helper/dashboard-widget/components/top-posts.tsx
```diff
@@ -0,0 +1,117 @@
...
+						}
+					} }
+				/>
+				<Select
+					defaultValue={ metric }
+					items={ [
+						[ Metric.Views, 'Page views' ],
+						[ Metric.AvgEngaged, 'Avg. Time' ] ] }
+					onChange={ ( event ) => {
+						if ( Object.values( Metric ).includes( event.target.value as Metric ) ) {
```
mehmoodak, line 100: Same as above

mehmoodak (approved):
Left few comments but approving the PR.

PR #1897, opened 2023-10-03
Validate API credentials using the Parse.ly API validation endpoint

### LINKED ISSUE #1853: Settings page: Enhance credentials validation
Currently, we validate credentials on the settings page with basic checks.

Enhance the validation logic so actual Site ID and API Secret validation takes place.

vaurdan:
## Description

Validate the API credentials by doing a API request to the validation endpoint. This assures that only valid credentials can be saved on the Settings page.

## Motivation and context
Fixes #1853 

## How has this been tested?
Tested on my local environment and with three new integration tests.

image


vaurdan:
Adding a "Do Not Merge" label as there is a problem with the e2e tests that need to be addressed. Looking into it.

vaurdan:
Removing the `Do Not Merge` label as the E2E issues were addressed. 

The issue was with testing with API credentials that aren't valid, since those are checked against the validate endpoint. I added a GET parameter called `e2e_parsely_skip_api_validate` that when is present in the settings page and set to `y`, it should skip the HTTP request to the `validate` endpoint and assume they are correct.

acicovic:
Hey there! Thank you for your work!

I haven't looked into the code just yet, but using the UI I noticed that it seems the settings page won't allow saving an empty Site ID or API Secret, which should be possible (especially for the API Secret since not all customers may have API access). You can easily see this if you've got a filled API Secret and then try to empty it.

I'd suggest looking how `develop` behaves with saving empty credentials and replicate the same behavior in this branch, unless we've got a reason to change it. In this case we can discuss it! 🙂

vaurdan:
@acicovic ah nice catch. I had it working like that previously but I think I accidentally reverted it when I amended a commit. Pushed the changes that should allow empty Site ID and Secret :)

src/class-validator.php
```diff
@@ -68,4 +79,51 @@ public static function validate_api_secret( string $api_secret ): bool {
...
+			$wp_error->add( self::INVALID_SITE_ID, __( 'Invalid Site ID', 'wp-parsely' ) );
+		}
+
+		if ( ! self::validate_api_secret( $api_secret ) ) {
+			$wp_error->add( self::INVALID_API_SECRET, __( 'Invalid API Secret', 'wp-parsely' ) );
+		}
+
+		if ( $wp_error->has_errors() ) {
+			return $wp_error;
+		}
```
acicovic: Any specific reason we've still using the `validate_*` functions? If I'm reading this correctly, we can remove them since we're now querying the API for valid credentials.

src/RemoteAPI/class-validate-api.php
```diff
@@ -0,0 +1,102 @@
+<?php
+/**
+ * Class for Validate API (`/validate`).
```
acicovic:
```suggestion
 * Class for Validate API (`/validate`)
```
Really super minor detail, but [file headers don't finish with a period](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#6-file-headers). Strange but true! 🙂

src/RemoteAPI/class-validate-api.php
```diff
@@ -0,0 +1,102 @@
...
+	 */
+	protected $is_public_endpoint = false;
+
+	/**
+	 * Gets the URL for the Parse.ly API credentials validation endpoint.
+	 *
+	 * @since 3.11.0
+	 *
+	 * @param array<string, mixed> $query The query arguments to send to the remote API.
+	 * @return string
```
acicovic, line 43: In any new code we're introducing, let's mind adding an empty line before `@return`

src/RemoteAPI/class-validate-api.php
```diff
@@ -0,0 +1,102 @@
...
+
+		$body    = wp_remote_retrieve_body( $response );
+		$decoded = json_decode( $body );
+
+		if ( ! is_object( $decoded ) ) {
+			return new WP_Error( 400, __( 'Unable to decode upstream API response', 'wp-parsely' ) );
+		}
+
+		if ( ! property_exists( $decoded, 'success' ) || false === $decoded->success ) {
+			return new WP_Error( $decoded->code ?? 400, $decoded->message ?? __( 'Unable to read data from upstream API', 'wp-parsely' ) );
```
acicovic: In any new code we're introducing, let's try to conform to [line wrapping guidelines](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#line-wrapping) for comments and code. Not always possible, but good to keep in mind.

tests/Integration/UI/SettingsPageTest.php
```diff
@@ -193,6 +207,136 @@ public function test_invalid_api_secret_values_are_emptied_when_validated(): voi
...
+	 * @uses   \Parsely\Validator::validate_api_credentials
+	 *
+	 * @group settings-page
+	 * @group settings-page-validation
+	 */
+	public function test_valid_api_credentials_are_retained_when_validated(): void {
+		$options = self::$parsely->get_options();
+
+		$options['apikey']     = 'mydomain.com';
+		$options['api_secret'] = 'valid_api_secret_key_based_on_length';
```
acicovic: Question: With the new changes, does it matter what value we put in `$options['api_secret']`? It would succeed regardless of the value due to `mock_request_api_credentials_validation_success`, right?

views/parsely-settings.php
```diff
@@ -44,6 +44,14 @@
 		<?php
 		settings_fields( Parsely::OPTIONS_KEY );
 		$wp_parsely_settings->show_setting_tabs_content();
+
+		// phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		if ( isset( $_GET['e2e_parsely_skip_api_validate'] ) && 'y' === $_GET['e2e_parsely_skip_api_validate'] ) {
+			?>
+			<input type="hidden" name="e2e_parsely_skip_api_validate" value="y" />
```
acicovic, line 51: Question: Does this field risk of being saved into the `parsely` options array?

src/RemoteAPI/class-remote-api-base.php
```diff
@@ -149,9 +149,9 @@ public function get_api_url( array $query ): string {
 	 * @param array<string, mixed> $query The query arguments to send to the remote API.
 	 * @param bool                 $associative When TRUE, returned objects will be converted into associative arrays.
 	 *
-	 * @return WP_Error|array<string, mixed>
+	 * @return WP_Error|array<string, mixed>|object|false
```
acicovic: Non-necessary cosmetic change: `@return` types could be in the same order for all `get_items()` functions.

acicovic (commented):
Excellent work! The effort that you've made to stick to project conventions is apparent and greatly appreciated!

Left some questions and a couple of minor adjustment proposals. This review is extremely super picky by the way and mostly focuses on details that are trivial.

Thank you so much!

src/class-validator.php
vaurdan: It is just validating the form (length and format) of the Site ID and API secret. It might save a few unnecessary API requests with invalid values. But I do agree it might be redundant and not offer any tactile gain in having both, apart from an API request here and there. 

What do you think? Is it worth to keep it, or rely exclusively on the API validation?

src/RemoteAPI/class-validate-api.php
vaurdan: Interesting! I adapted from the structure of the [ `class-related-api.php` ](https://github.com/Parsely/wp-parsely/blob/develop/src/RemoteAPI/class-related-api.php#L3) class and it also had the period.  :smile: Fixing it.

tests/Integration/UI/SettingsPageTest.php
vaurdan: If we keep the existing `validate_*` functions, then yes, it matters, since if you enter an invalid API secret (less than 30 chars), it will not run the API request to validate the credentials and instead will fail the validation immediately. The same is true for the API Key (If it's not a valid domain - www prefix and up to three dots)

views/parsely-settings.php
vaurdan: Should be safe since it isn't using the Settings API. But good question, I'm going to test that and make sure it isn't leaking to the options array.

vaurdan: Confirmed: it's not leaking the field into the options array.

src/class-validator.php
acicovic: I'd go for API exclusively, since the validations in place aren't that good anyway and were just a compromise until we had proper validation. I totally agree that having them could save us some API trips, but the Settings page is a set-and-forget thing that most customers don't come back to. So I think removing them is fine in this case.

src/RemoteAPI/class-validate-api.php
acicovic: Yes, some files that have a period haven't been updated. You happened to stumble on one of those it seems 🙂

views/parsely-settings.php
acicovic: Great, thank for checking! 👍

tests/Integration/UI/SettingsPageTest.php
acicovic: Makes sense, thanks! Asking because the `valid-api_secret_key_based_on_length` value won't make any sense if we stop relying on validating length.

acicovic:
@vaurdan, I had a second look. Thanks for addressing the feedback and replying. If you agree to wipe out the non-API validations, please go ahead and I can take a final look.

Thanks! 🙂

vaurdan:
@acicovic the older methods have been removed in favor of the new `validate_api_credentials`. Let me know your thoughts and feedback!

src/RemoteAPI/class-validate-api.php
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * Class for credentials validation API (`/validate`).
+ *
+ * @since 3.11.0
+ *
+ * @phpstan-import-type WP_HTTP_Request_Args from Parsely
+ */
+class Validate_API extends Remote_API_Base {
+	protected const ENDPOINT     = '/validate/secret';
+	protected const QUERY_FILTER = 'wp_parsely_related_endpoint_args';
```
acicovic:
```suggestion
	protected const QUERY_FILTER = 'wp_parsely_validate_secret_endpoint_args';
```
This should be based on the name of the endpoint.

src/UI/class-settings-page.php
```diff
@@ -968,37 +968,33 @@ private function validate_basic_section( $input ) {
...
-						'apikey',
-						__( 'The Site ID was not saved because it is incorrect. It should look like "example.com".', 'wp-parsely' )
-					);
-					$input['apikey'] = $options['apikey'];
-				} else {
-					$input['apikey'] = $site_id;
-				}
+			$site_id             = $this->sanitize_site_id( $input['apikey'] );
+			$api_secret          = $this->get_unobfuscated_value( $input['api_secret'], $this->parsely->get_api_secret() );
+			$input['api_secret'] = $api_secret;
```
acicovic: Is `$input['api_secret'] = $api_secret;` needed, or do we need to set the Site ID as well here?

src/class-validator.php
```diff
@@ -68,4 +35,42 @@ public static function validate_api_secret( string $api_secret ): bool {
...
+	 * @param Parsely $parsely The Parsely instance.
+	 * @param string  $site_id The Site ID to be validated.
+	 * @param string  $api_secret The API Secret to be validated.
+	 * @return WP_Error|true True if the API Credentials are valid, false otherwise.
+	 */
+	public static function validate_api_credentials( Parsely $parsely, string $site_id, string $api_secret ) {
+		$wp_error = new WP_Error();
+
+		// If the API secret is empty, the validation endpoint will always fail. Since it's possible to
+		// use only the API key without the secret, we'll skip the validation and assume it's valid.
```
acicovic: > Since it's possible to use only the API key without the secret

The wording confuses me a bit. Is there a way to simplify, clarify, or correct this if it's wrong?

src/class-validator.php
```diff
@@ -68,4 +35,42 @@ public static function validate_api_secret( string $api_secret ): bool {
...
+
+	/**
+	 * Validates the passed API Credentials.
+	 *
+	 * @since 3.11.0
+	 *
+	 * @param Parsely $parsely The Parsely instance.
+	 * @param string  $site_id The Site ID to be validated.
+	 * @param string  $api_secret The API Secret to be validated.
+	 * @return WP_Error|true True if the API Credentials are valid, false otherwise.
```
acicovic:
```suggestion
	 * @return true|WP_Error True if the API Credentials are valid, WP_Error otherwise.
```

src/class-validator.php
```diff
@@ -68,4 +35,42 @@ public static function validate_api_secret( string $api_secret ): bool {
...
+	 */
+	public static function validate_api_credentials( Parsely $parsely, string $site_id, string $api_secret ) {
+		$wp_error = new WP_Error();
```
acicovic: Not an issue, but I think we could probably instantiate this in line 67? We don't seem to be using this outside the `if`.

tests/Integration/UI/SettingsPageTest.php
```diff
@@ -44,153 +44,185 @@ public function set_up(): void {
...
+	 * @uses   \Parsely\UI\Settings_Page::validate_options
+	 * @uses   \Parsely\UI\Settings_Page::validate_options_post_type_tracking
+	 * @uses   \Parsely\UI\Settings_Page::validate_recrawl_section
+	 * @uses   \Parsely\Validator::validate_api_credentials
 	 *
 	 * @group settings-page
 	 * @group settings-page-validation
 	 */
-	public function test_valid_site_id_values_are_retained_when_validated(): void {
+	public function test_empty_site_id_api_secret_are_retained_when_validated(): void {
```
acicovic:
```suggestion
	public function test_empty_api_credentials_are_retained_when_validated(): void {
```
I think this is better? If not, feel free to keep as-is.

tests/Integration/UI/SettingsPageTest.php
```diff
@@ -44,153 +44,185 @@ public function set_up(): void {
...
+		$options['api_secret'] = '';
 
 		$actual = self::$settings_page->validate_options( $options );
-		self::assertSame( $expected, $actual );
+		self::assertSame( $options, $actual );
 	}
 
 	/**
-	 * Verifies that invalid Site ID values are emptied when validated.
+	 * Verifies that valid API credentials are retained when validated with the Validation API
```
acicovic:
```suggestion
	 * Verifies that valid API credentials are retained when validated with the Validation API.
```

tests/Integration/UI/SettingsPageTest.php
```diff
@@ -44,153 +44,185 @@ public function set_up(): void {
...
+
+		$expected = $options;
 
 		$actual = self::$settings_page->validate_options( $options );
 		self::assertSame( $expected, $actual );
 	}
 
 	/**
-	 * Verifies that valid API Secret values are retained when validated.
+	 * Verifies that invalid API credentials are emptied when validated with the Validation API
```
acicovic:
```suggestion
	 * Verifies that invalid API credentials are emptied when validated with the Validation API.
```

tests/Integration/UI/SettingsPageTest.php
```diff
@@ -44,153 +44,185 @@ public function set_up(): void {
...
 	 */
-	public function test_invalid_site_id_values_are_emptied_when_validated(): void {
-		$expected = self::$parsely->get_options();
-		$options  = self::$parsely->get_options();
+	public function test_valid_api_credentials_are_retained_when_validated(): void {
+		$options = self::$parsely->get_options();
 
-		$options['apikey'] = 'invalid';
+		$options['apikey']     = 'mydomain.com';
+		$options['api_secret'] = 'valid_api_secret_key_based_on_length';
```
acicovic:
```suggestion
		$options['api_secret'] = 'valid_api_secret';
```
Let's remove the "based on length" part since validation isn't being done by length anymore.

tests/Integration/UI/SettingsPageTest.php
```diff
@@ -44,153 +44,185 @@ public function set_up(): void {
...
+		remove_filter( 'pre_http_request', array( $this, 'mock_request_api_credentials_validation_success' ), 10 );
+		// Mock HTTP request to simulate a failed credentials validation.
+		add_filter( 'pre_http_request', array( $this, 'mock_request_api_credentials_validation_failure' ), 10, 3 );
+
+		$expected = self::$parsely->get_options();
+		$options  = self::$parsely->get_options();
 
-		// More than 30 characters.
+		$options['apikey']     = 'mydomain.com';
 		$options['api_secret'] = 'valid_api_secret_key_based_on_length';
```
acicovic:
```suggestion
		$options['api_secret'] = 'invalid_api_secret';
```
Although the value doesn't have any real effect in the test, let's change it so it's coherent with the test's purpose.

tests/Integration/UI/SettingsPageTest.php
```diff
@@ -215,7 +247,7 @@ public function test_invalid_api_secret_values_are_emptied_when_validated(): voi
 	 * @uses \Parsely\UI\Settings_Page::validate_options_post_type_tracking
 	 * @uses \Parsely\UI\Settings_Page::validate_recrawl_section
 	 * @uses \Parsely\Validator::validate_metadata_secret
-	 * @uses \Parsely\Validator::validate_site_id
+	 * @uses \Parsely\Validator::validate_api_credentials()
```
acicovic:
```suggestion
	 * @uses \Parsely\Validator::validate_api_credentials
```

tests/Integration/UI/SettingsPageTest.php
```diff
@@ -44,153 +44,185 @@ public function set_up(): void {
...
+	 * @uses   \Parsely\UI\Settings_Page::validate_options_post_type_tracking
+	 * @uses   \Parsely\UI\Settings_Page::validate_recrawl_section
+	 * @uses   \Parsely\Validator::validate_api_credentials
 	 *
 	 * @group settings-page
 	 * @group settings-page-validation
 	 */
-	public function test_valid_api_secret_values_are_retained_when_validated(): void {
-		$options = self::$parsely->get_options();
+	public function test_invalid_api_credentials_are_emptied_when_validated(): void {
```
acicovic: Regarding the test's name, I think that the credentials are retained to their previous values, not emptied to `''` per se.

tests/Integration/UI/SettingsPageTest.php
```diff
@@ -44,153 +44,185 @@ public function set_up(): void {
...
+	 * Tear down method called after each test.
+	 */
+	public function tear_down(): void {
+		parent::tear_down();
+		remove_filter( 'pre_http_request', array( $this, 'mock_request_api_credentials_validation_success' ), 10 );
 	}
 
 	/**
-	 * Verifies that valid Site ID values are retained when validated.
+	 * Verifies that empty Site ID and API Secret values are retained when validated.
```
acicovic:
```suggestion
	 * Verifies that empty API credentials are retained when validated.
```

acicovic (commented):
Great work! Left some comments and questions, mostly details.

tests/Integration/UI/SettingsPageTest.php
acicovic: If you think it's more explicative as it is, feel free leave it like that.

acicovic (approved):
Thanks for working on this and bearing with me 🙂

PR #1916, opened 2023-10-12
PCH Related Top Posts: Add filter type drop-down

acicovic:
## Description

The PCH's Related Top Posts pane used to automatically filter top related posts by `author`, `section` (category) or `tag`, depending on what was available to the post without giving the user any choice.

With this PR, we introduce the choice to select one of the above filter types. However, as previously, the feature still doesn't allow selecting the filter value and will return the first result it gets from WordPress. This is expected as this PR is an intermediate step towards offering fuller functionality.

## Motivation and context
Offer more choices regarding the data we're displaying in the Related Top Posts pane.

## How has this been tested?
Existing tests pass, and some were modified to accommodate the new changes.

### Before
image

### After
image
image


tests/e2e/specs/content-helper/related-top-posts-panel-filters.spec.ts
```diff
@@ -38,44 +37,10 @@ describe( 'PCH Editor Sidebar Related Top Post panel filters', () => {
...
-
-		// Author + tag.
-		expect( await getTopRelatedPostsMessage( '', tagName, 500, messageSelector ) )
-			.toMatch( `Top posts with tag "${ tagName }" in last 7 days.` );
-
-		// Author + category + tag.
-		expect( await getTopRelatedPostsMessage( categoryName, tagName, 500, messageSelector ) )
-			.toMatch( `Top posts with tag "${ tagName }" in last 7 days.` );
-	} );
-
```
acicovic, line 78: This test is now irrelevant since we don't pick our filter by prioritization but by user choice.

tests/e2e/specs/content-helper/related-top-posts-panel-filters.spec.ts
```diff
@@ -11,7 +11,6 @@ import {
 import {
 	VALID_API_SECRET,
 	getTopRelatedPostsMessage,
-	insertRecordIntoTaxonomy,
```
acicovic, line 14: Although `insertRecordIntoTaxonomy()` was being used only here, we haven't removed it since it could be of use in the future.

vaurdan (approved):
Looks great to me! 👍 

I do only have a more "conceptual" question. We are using "Section" to filter by category the related posts. Although I understand this is the term that's used by Parse.ly, I wonder if we should use it in WordPress context, or if we should keep using "Category". Take an editor, for example, that has no contact with Parse.ly whatsoever, they might be confused with the meaning of "Section" in a WordPress context. What are your thoughts on that?

acicovic:
> I do only have a more "conceptual" question. We are using "Section" to filter by category the related posts. Although I understand this is the term that's used by Parse.ly, I wonder if we should use it in WordPress context, or if we should keep using "Category". Take an editor, for example, that has no contact with Parse.ly whatsoever, they might be confused with the meaning of "Section" in a WordPress context. What are your thoughts on that?

@vaurdan, that's an excellent question!

Something we can say regarding WP categories vs Parse.ly sections is that they aren't 100% identical, since a post can be assigned to many WP categories, but Parse.ly allows only one section. So even if a post belongs to many WP categories, our metadata should [define only one of them as the Parse.ly section](https://github.com/Parsely/wp-parsely/blob/290d5ec4e56992d04d183107017fabbf230bd2fc/src/Metadata/class-post-builder.php#L165-L172).

I think that if we name this `Categories`, we should put there all the post's categories. If we name it `Section`, we should only add there the Parse.ly section the post belongs to. It's great you made that question, because our current approach regarding categories in Related Top Posts is flawed, and this clarification will help us fix it in the upcoming PRs.

@danielabloch, thoughts about this?

danielabloch:
> I think that if we name this Categories, we should put there all the post's categories. If we name it Section, we should only add there the Parse.ly section the post belongs to. It's great you made that question, because our current approach regarding categories in Related Top Posts is flawed, and this clarification will help us fix it in the upcoming PRs.

Agreed. The discrepancy is unfortunate, but I prefer that we use the naming conventions of Parse.ly (Section), since that is more accurate for the data being shown, and important for customers to know if/when they look for performance data in the Parse.ly Dashboard.

acicovic:
OK, we'll proceed with sections. Thank you both!

PR #1929, opened 2023-10-16
Dashboard Widget: Change number of visible top posts to 5 and add pagination

vaurdan:
## Description
Changes the Top Posts widget on the Dashboard to show 5 posts at the time, and includes a navigation on the bottom of the widget to allow changing pages.

## Motivation and context
Only three posts in the Top Posts widgets is clearly insufficient to have any meaningful insight. Having the possibility to see more posts will help understanding how the site posts are performing, directly on the dashboard.

## How has this been tested?
Tested locally on my development environment. Did test with different scenarios:

 * Many posts returned 
*  Less than 5 posts returned
 * When the total of posts is not a multiple of 5 (eg 7).
 * Changing period and metrics

Oct-16-2023 14-42-58


vaurdan:
@acicovic this widget does not have any E2E test setup. Do you think it makes sense to write them and include here, with this PR, or should be good to go, and work on that later down the line?

src/content-helper/dashboard-widget/components/top-posts.tsx
```diff
@@ -87,12 +90,38 @@ export function TopPosts() {
...
 	);
 
+	const navigation: JSX.Element = (
+		<div className="parsely-top-posts-navigation">
+			<button
+				className="parsely-top-posts-navigation-prev"
+				disabled={ page <= 1 }
+				onClick={ () => setPage( page - 1 ) }
+			>
+				{ '<< ' + __( 'Previous', 'wp-parsely' ) }
```
acicovic: Wondering whether we should include the `<<` and `>>` within the localized string. I'm thinking of RTL languages that might place them in a different position within the string.

You can also try the HTML entities `&laquo;`/`&raquo;` instead of `<<`/`>>` just to see if they look better, and keep whatever looks better to you.

src/content-helper/dashboard-widget/components/top-posts.tsx
```diff
@@ -112,13 +141,14 @@ export function TopPosts() {
...
 				loading ? ( spinner ) : (
-					<ol className="parsely-top-posts">
+					<ol className="parsely-top-posts" style={ { counterReset: 'item ' + ( ( page - 1 ) * TOP_POSTS_DEFAULT_LIMIT ) } }>
 						{ posts.map( ( post: PostData ): JSX.Element =>
 							<TopPostListItem key={ post.id } metric={ metric } post={ post } />
 						) }
 					</ol>
 				)
 			}
+			{ ( posts.length >= 5 || page > 1 ) && navigation }
```
acicovic: Can we replace `5` with `TOP_POSTS_DEFAULT_LIMIT` here? If not (or you think this could introduce a bug with an update), fine to leave as is.

src/content-helper/dashboard-widget/dashboard-widget.scss
```diff
@@ -61,7 +85,7 @@
 
 	.parsely-top-post-thumbnail {
 		height: 46px;
-		width: 46px;
+		min-width: 46px;
```
acicovic, line 88: Out of curiosity: Did you run into an issue and changed that? Wondering if this could result in different width images, which might make the widget look disarranged.

acicovic (commented):
Excellent work! I like the simplicity of this PR.

@vaurdan, regarding your question, it is totally fine to add tests as part of another PR.

Some things I've noticed with my manual testing:

- If the API call returns exactly 5 items, we still show a `Next` link, which when clicked displays a `No Top Posts data is available.` message, and the paging links disappear. In this case, users find themselves "locked" and have to change the period or metric to get a refresh. Not the end of the world, but something that could be looked into (into this PR or another).
- The current implementation allows for an unlimited number of pages, going deeper than top 10 results. @danielabloch, should we leave as is, or limit the feature to top 10 only?

src/content-helper/dashboard-widget/dashboard-widget.scss
vaurdan: I did change it to min-width to make all the thumbnails have the same width, which was not happening when mixing vertical and horizontal images,  for example:

image

With this change, they become aligned, and square, as intended (I think?):

image

acicovic: Ah, excellent!

vaurdan:
> If the API call returns exactly 5 items, we still show a Next link, which when clicked displays a No Top Posts data is available. message, and the paging links disappear. In this case, users find themselves "locked" and have to change the period or metric to get a refresh. Not the end of the world, but something that could be looked into (into this PR or another).

I believe this message is returned when the API returns no data. From my testing, it might happen from time to time. Maybe we should introduce a "refresh" button, in that case?

https://github.com/Parsely/wp-parsely/blob/1b15300e665d70f19f6eb966d3e8f0e0947dc04c/src/content-helper/dashboard-widget/provider.ts#L49-L55

EDIT: It might be a little more than than. Try to work on a solution :)

acicovic:
@vaurdan, regarding your question above, I guess there could be different solutions. For example, if all we need is a top 10, we could limit to 10 results, and at this point, paging would be just to show the first or second 5 results.

Or in the current scenario, in each call we could fetch 6 results to see if there will be a next page.

acicovic:
In the end of the day, if paging makes things more complicated than they should, we can consider alternatives.

vaurdan:
@acicovic should be fixed: 

image

This specific UI was being treated differently as it was  handling the error, so I included the navigation if the current page is greater than 1. 

Although, I like your idea of fetching `n+1` and displaying the `Next` button only if more posts are available.

acicovic:
Having a solution that doesn't make the user feel blocked like the one you've added should be enough. We can iterate if we want.

src/content-helper/dashboard-widget/components/top-posts.tsx
vaurdan: Moved the `<<` and `>>` to the translatable string. However I couldn't use the HTML entities, as they were being printed verbatim. Maybe the i18n library doesn't support those?

vaurdan:
@acicovic I have implemented your suggestions, feel free to take a look!

src/content-helper/dashboard-widget/components/top-posts.tsx
acicovic: Ah, you're right about the entities. React makes this complicated. Let's leave as is.

acicovic:
The code good looks to me. Let's wait on @danielabloch's confirmation on whether we are good as is, or if we need to limit the results to 10.

danielabloch:
Looks good to me! Excellent work Henrique!

acicovic (approved):

PR #1965, opened 2023-10-26
Add types/wordpress__editor devDependency

acicovic:
## Description
With this PR, we're adding the `@types/wordpress__editor` package to our `devDependencies`. This package helps with code completion when working with WordPress Editor functions, and has also revealed a couple of issues in our code.

## Motivation and context
- The package was added as an intermediate step to go towards improving the PCH's Related Top Posts pane with more options.
- Related to #1916.

## How has this been tested?
Existing tests pass.


src/content-helper/editor-sidebar/related-top-posts/provider.ts
```diff
@@ -183,6 +183,7 @@ export class RelatedTopPostsProvider {
 
 		if ( PostFilterType.Tag === filterType ) {
 			// Get post's first tag.
+			// @ts-ignore
```
acicovic, line 186: Ignoring this isn't an ideal solution, but the intention is to refactor/move the code soon. At that point, we'll see if we can find a better alternative.

vaurdan (approved):

PR #1976, opened 2023-11-03
Implement a base endpoint class

acicovic:
## Description
As we're preparing to introduce local-only API endpoints into our codebase, this PR implements a `Base_Endpoint` class that will be used as a common base class for local and remote endpoint classes.

The code in `Base_Endpoint` is largely borrowed from the `Remote_API_Base` class which has been renamed to `Base_Endpoint_Remote` and now inherits from `Base_Endpoint`.

## Motivation and context
Reduce code duplication, start organizing our endpoint code that is scattered and needs a refactoring.

## How has this been tested?
Manual test that the PCH works and existing tests pass.


mjangda:
I'd urge some caution here. OOP and inheritance is great until it's not (https://github.com/zakirullin/cognitive-load#inheritance-nightmare). It makes debugging hard and can more easily introduce bugs.

I've definitely had trouble with our REST API code when debugging something; it required jumping back-and-forth across child and parent classes and trying to understand what action was happening where. And also getting caught off-guard when a property within a child class drives the behaviour of an endpoint but that logic lives in the parent.

I also don't have any alternatives to offer immediately (which limits how useful this comment is) but I do wanted us to think deeply about this before we make our own and future team members' lives harder.

acicovic:
@mjangda, I can totally hear you on that and this has been hard to follow for me as well.

The API classes structure we currently have was introduced when the first endpoint was created, and we've been basing on it ever since. Personally, Besides the class jumping, the current naming scheme and the different locations of classes doesn't help either. Essentially, we've got something that's more complex than it should be IMO.

Having said that, we should think about how we need to restructure this to make it simpler to understand and maintain. IMO, the real problem isn't inheritance, traits or interfaces, but rather how we use them. We absolutely want to refactor this to simplify it in the future, and it would be good to do this before we add too many new endpoints.

Until we do that and regarding the specific PR, I don't think it's making things worse. But if this isn't the case, we can surface that during review.

vaurdan (approved):
Looks good! 

And I subscribe to what @acicovic is saying. I believe there's room for improvement and there is some unnecessary over complication on the codebase, making it more complex than it should.  Definitely something to work and improve down the line.

PR #1986, opened 2023-11-08
PCH Related Top Posts: Add filter values drop-down

### LINKED ISSUE #940: PCH Related Top Posts: Allow user to choose meta filter(s)
Currently, PCH Related Top Posts picks the filter for the results automatically (`tag` or `category` or `author`). We could enhance this by giving the user the ability to choose the filter(s) themselves.

In the sidebar:
- Show a drop-down containing all the categories assigned to the post
- Show a drop-down containing all the tags assigned to the post

Using the above, permit to the user to specify themselves the filter they are interested in, and if possible allow for a combination (posts that are in `category` and have `tag`) search.

## Additional context
The assumption here is that by giving users the ability to select, we would give them more flexibility in their research.

acicovic:
## Description
This PR introduces the ability to select a specific `Tag`, `Section` (Category) or `Author` value as a way to fine-grain results within the PCH's Related Top Posts panel. Once one of the filter types (let's say `Tag`) has been selected and the post has more than 1 tags, a second drop-down will be displayed to select the specific tag to filter by.

- Newly selected tags, categories and author in the Post Editor will be shown immediately as they're getting added/modified, without the need to save the post. 
- Filter types that aren't available (for example when a post doesn't have tags or categories) are not shown as options in the UI.
- When only a single filter type or value is available, it automatically gets selected without displaying any drop-downs.
- WordPress Pages don't have tags or categories by design, so they are limited to the author.
- The filter value drop-down only contains the values that are pertinent to the current post. For example, it won't show all the website's tags, but only the tags that are assigned to the specific post. It is also searchable to accommodate for many entries.
- This implementation doesn't current support the Co-Authors Plus plugin. This possibility will be investigated in the future.

## Motivation and context
- Completes the work that was started in #1916 and closes #940.
- This work is based on #1976 which should be reviewed and merged first.

## How has this been tested?
Some existing tests were slightly modified to pass. The testing was manual and there are no automated tests specifically targeting filter type and value selection.

### The filter type drop-down

| Shown | Clicked |
|--------|--------|
| image | image | 

| No category (section) assigned to the post | No tag or section, auto-fallback to author |
|--------|--------|
| image | image | 

### The filter value drop-down

| Post with many tags (collapsed) | Post with many tags (expanded) |
|--------|--------|
| image | image | 

The same view/functionality applies in the case of many categories or authors.


src/content-helper/editor-sidebar/editor-sidebar.tsx
```diff
@@ -24,10 +30,106 @@ import { RelatedTopPostList } from './related-top-posts/component-list';
...
+		apiFetch<SidebarPostData>( {
+			path: addQueryArgs( '/wp-parsely/v1/post-data/', { postId } ),
+		} ).then( ( response ) => {
+			setPostData( {
+				authors: response.authors,
+				tags: tagNames,
+				categories: categoryNames,
+			} );
+		} );
+	}, [ postId, tagNames, categoryNames ] );
```
vaurdan: Do we need to have these dependencies? If I'm understanding correctly, it will do a API fetch each time there's a change in tags or categories, and I'm not sure if it's necessary, since we already have those editor store.

vaurdan (commented):
Looks good to me! :) I have added an inline comment with a question, but otherwise, no blockers from me.

Regarding the need to refresh to get co-authors, the problem is related with the fact that the CAP terms are only added to the post on save, and therefore the new local API endpoint returns no co-authors, making the filter either empty or outdated, right?

src/content-helper/editor-sidebar/editor-sidebar.tsx
acicovic: Here's what I've come to understand as this is new to me: When `useEffect` runs, any of these dependencies can be initially null as they are being resolved in the `useSelect` block. Then, they will resolve and they might do so in different order. So the array here is needed to get the final postData correctly as different parts of the request resolve.

acicovic:
> Regarding the need to refresh to get co-authors, the problem is related with the fact that the CAP terms are only added to the post on save, and therefore the new local API endpoint returns no co-authors, making the filter either empty or outdated, right?

I haven't dug in detail about why this is exactly happening so I don't know, but it does feel quirky in regular use to the point I'm thinking of temporarily dropping Co-Authors support and directly getting the Author from Gutenberg. The final solution would be to get co-authors data from Gutenberg as well, as we can do with the other data.

acicovic:
I'm dropping Co-Authors support for now, to make the user experience better. We can then sort out in time how to get the data from the plugin.

acicovic:
I've removed Co-Authors Plus support, and modified the PR's description to reflect that.

src/content-helper/editor-sidebar/editor-sidebar.tsx
vaurdan: Ah makes sense! Thanks for the explanation!

vaurdan (approved):
Looks great! 👍

PR #1997, opened 2023-11-13
Update plugin screenshots

acicovic:
## Description
This PR updates the plugin's screenshots.

## Motivation and context
Keep our screenshots current!

## How has this been tested?
- No functionality changes.
- Current tests pass.
- Verified that the markdown renders correctly in VSCode preview.


mehmoodak (approved):
It will be better if screenshot-8 will be in same aspect ratio as screenshot 6 and 7 because the sidebar is more prominent in those.

acicovic:
@mehmoodak, thanks for spotting this! I actually had an updated screenshot but forgot to include it. This is now done. 🙂

PR #2001, opened 2023-11-13
Initialize Parse.ly option before registering the setting

vaurdan:
## Description

This update addresses an issue with the `register_setting` function in the Parse.ly plugin, where the `validate_options` callback was being called twice when the settings are saved for the first time. This behavior, related to a known WordPress core issue ([#21989](https://core.trac.wordpress.org/ticket/21989)), was causing validation issues on save. The fix involves adding the option using `add_option` before calling `register_setting`, thereby preventing the double sanitization of the uninitialized option.

I have also changed the third argument of `register_setting` to comply to the format specified [in the documentation](https://developer.wordpress.org/reference/functions/register_setting/).

## Motivation and context

Fixed the misleading validation errors when saving the settings for the first time.

## How has this been tested?

The changes have been tested in a local development environment by removing the `parsely` option. 

image


acicovic (approved):
Excellent work!

PR #2042, opened 2023-11-30
Fix reference error when telemetry is disabled

vaurdan:
## Description

This PR addresses a critical issue where a reference error was thrown in scenarios where telemetry was disabled. The error was due to the `wpParselyTracksTelemetry` being undefined. The proposed changes refactor the initialization of the telemetry system. Now, the telemetry JS code is always loaded, but it gets enabled only when the `wpParselyTracksTelemetry` object is present in the global context. This approach prevents the reference error and ensures that telemetry functionality is conditionally activated based on configuration.

## Motivation and context

This fix is necessary to resolve the JavaScript errors encountered when the telemetry feature is disabled by the user or in certain environments. By ensuring that the telemetry script is always loaded but only activated when enabled, we maintain the functionality's flexibility without introducing errors for users who opt out of telemetry.

## How has this been tested?

This fix was tested in both environments where telemetry is enabled and disabled. In both scenarios, the system behaved as expected: initializing and running telemetry when enabled, and remaining dormant without errors when disabled. 

## Screenshots \(if appropriate\)

N/A


acicovic (approved):
Thanks for catching this!

PR #2086, opened 2023-12-18
PCH Title Suggestions: Add tone and persona options

### LINKED ISSUE #2072: PCH: Add tone and style to title generator
The underlying title generation API supports setting both a tone and a style for the titles. The title generator functionality should support setting these values for a more personalized and adjusted experience. 

Add a expandable settings component to the title generator panel, that would allow setting the tone and the style using a dropdown.

vaurdan:
## Description
This pull request introduces enhancements to the WP Parse.ly AI Title Suggestions feature. The PR includes refinements to the title generation process, such as incorporating tone and persona parameters for more targeted, relevant title suggestions. Additionally, we've improved the codebase for better maintainability by updating the common components structure, currently used for the `BetaBadge`, and the newly introducted `ToneSelector` and `PersonaSelector`.

## Motivation and context
The enhancement of the AI Title Suggestions feature is driven by the need for more personalized and contextually appropriate title recommendations. By introducing tone and persona parameters, we aim to align the suggestions more closely with the content's intended audience and style. 

## How has this been tested?
This feature was tested in a local WordPress environment, where I generated titles with different tones and personas. Each set of suggested titles returned with the correct tone and persona, as set in the TItle Suggestion settings.

https://github.com/Parsely/wp-parsely/assets/5804923/f89e66cf-63c8-4ae3-908c-a6e64e9632cf


vaurdan:
@acicovic I was wondering if, for the [available tones](https://github.com/Parsely/wp-parsely/blob/d364cea71551dc2e8cdd5d9e72102cd188ccdb86/src/content-helper/common/components/tone-selector/component.tsx#L13-L58) and [personas](https://github.com/Parsely/wp-parsely/blob/d364cea71551dc2e8cdd5d9e72102cd188ccdb86/src/content-helper/common/components/persona-selector/component.tsx#L18-L63) lists, we should allow them to be overriden by a WordPress filter.

 This would allow more customization by the developers using the plugin, however it would require a bit of rework on how it's implemented, since we would need to make the lists with PHP to allow them to be filtered, and output them using `wp_add_inline_script` to be consumed by the Tone and Persona selectors components.

What are your thoughts? Should we implement that before merging, or could be addressed in the future with a new PR? The custom tone and persona should allow for fine tuning the results in the meantime.

src/content-helper/editor-sidebar/title-suggestions/store.ts
```diff
@@ -18,6 +20,8 @@ interface TitlesData {
 }
 interface TitlesState {
 	isLoading: boolean;
+	tone: ToneProp | string | undefined,
+	persona: PersonaProp | undefined,
```
acicovic: Should `string` be added to `persona` here, or removed from `tone`?

src/content-helper/editor-sidebar/title-suggestions/provider.ts
```diff
@@ -33,19 +35,29 @@ export class WriteTitleProvider {
 	/**
 	 * Returns a list of suggested titles for the given content.
 	 *
-	 * @param {string } content The content to generate titles for.
-	 * @param {number}  limit   The number of titles to return. Defaults to 3.
+	 * @param {string }     content The content to generate titles for.
+	 * @param {number}      limit   The number of titles to return. Defaults to 3.
+	 * @param {ToneProp}    tone    The tone to generate titles for.
+	 * @param {PersonaProp} persona The persona to generate titles for.
```
acicovic: Better update this to something like:

```suggestion
	 * @param {ToneProp}    tone    The tone to use when generating the titles.
	 * @param {PersonaProp} persona The persona to use when generating the titles.
```

src/content-helper/editor-sidebar/title-suggestions/provider.ts
```diff
@@ -11,6 +11,8 @@ import {
 	ContentHelperError,
 	ContentHelperErrorCode,
 } from '../../common/content-helper-error';
+import { getLabel as getLabelForTone, ToneProp } from '../../common/components/tone-selector';
+import { getLabel as getLabelForPersona, PersonaProp } from '../../common/components/persona-selector';
```
acicovic: My personal suggestion would be to rename these to `getToneLabel` and `getPersonaLabel` and directly put them into the `addQueryArgs` block, removing lines 48-53. I think that code clarity is also maintained as we would mostly assume we don't want to include smilies in an API request.

Let me know if any of this is unclear, or keep as-is if you think the current solution is better. In this case, I'd suggest to update the wording `humanReadable` to something else.

acicovic (commented):
Hey there! this is excellent work all-around and I appreciate the file structure introduced! We can certainly see about making this a convention.

I haven't looked into all the code yet, but leaving a few first comments.

In addition to that, it seems that most newly added `tsx` files lack `@since` and `@return` DockBlocks. Please take care of these, and I can take care of possible details that might remain.

Regarding your question, It's certain that some customers will want to limit/adjust these options or even remove the custom tone/persona capability. I'd recommend us to see if we can work on that on January 2024 as a separate PR.

vaurdan:
@acicovic Thanks for your feedback! I just pushed a commit addressing it.

> Regarding your question, It's certain that some customers will want to limit/adjust these options or even remove the custom tone/persona capability. I'd recommend us to see if we can work on that on January 2024 as a separate PR.

Sounds great to me! Thanks for your input :)

src/content-helper/editor-sidebar/title-suggestions/component.tsx
```diff
@@ -63,13 +74,30 @@ export const TitleSuggestionsPanel = (): JSX.Element => {
 		return getEditedPostAttribute( 'title' );
 	}, [] );
 
-	const generateTitles = async ( titleType: TitleType, content: string ): Promise<void> => {
+	const toneLabel = tone ? getToneLabel( tone ) : __( 'Neural', 'wp-parsely' );
```
acicovic:
```suggestion
	const toneLabel = tone ? getToneLabel( tone ) : __( 'Neutral', 'wp-parsely' );
```

src/content-helper/editor-sidebar/title-suggestions/component.tsx
```diff
@@ -84,8 +112,23 @@ export const TitleSuggestionsPanel = (): JSX.Element => {
 				request_more: titles.length > 0,
 				total_titles: titles.length,
 				total_pinned: titles.filter( ( title ) => title.isPinned ).length,
+				tone: tone ? tone : 'neutral',
+				persona: persona ? persona : 'journalist',
```
acicovic: The answer to this might be no, but do we need to localize these strings?
Also, maybe there would be a way to avoid possible spelling issues and repetition by accessing some default tone/persona values?

vaurdan: @acicovic we don't, since those basically work as the key for the `{ label, emoji }` object.  We might be able to expose the default by exporting it from the component, I'm going to take a look and see if I can implement that :)

acicovic (approved):
I pushed a small commit with mild improvements to whitespace and some comments. You also might notice that some imports have been moved around. I did some manual adjustment there, but also some of it comes from VSCode auto-sorting our imports.

This looks good to me! Thanks for the great work, and feel free to merge! 🙂

PR #2090, opened 2023-12-20
Increase PHPStan param_type coverage to 80

acicovic:
## Description
After a `composer update`, PHPStan gave me an error. Investigating, I found that the error was located in unneeded code, which I removed. While at it, I also increased our `param_type` coverage to 80.

## Motivation and context
Increasing our code coverage and removing unneeded code is always good.

## How has this been tested?
Tests continue to pass.

### DIFF
--- a/tests/Integration/ProxyEndpointTest.php
+++ b/tests/Integration/ProxyEndpointTest.php
@@ -141,7 +141,9 @@ abstract class ProxyEndpointTest extends TestCase {
 	 *
 	 * @param WP_REST_Request|null $request The request object to be used.
 	 */
-	public function run_test_get_items_fails_without_site_id_set( $request = null ): void {
+	public function run_test_get_items_fails_without_site_id_set(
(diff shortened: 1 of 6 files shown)


tests/Integration/ProxyEndpointTest.php
```diff
@@ -181,22 +185,13 @@ private function run_test_get_items_fails(
...
 		}
 
-		$dispatched = 0;
-		add_filter(
-			'pre_http_request',
-			function () use ( &$dispatched ) {
-				$dispatched++;
-				return null;
-			}
-		);
```
acicovic, line 198: This is the code that was removed. As far as I can understand, it was never fired as in the specific context an HTTP request was never made. We were also not testing the `dispatched` value, thing which we're doing in other tests which actually perform an HTTP request.

vaurdan (approved):
Looks good to me! 👍

Issue #1954, opened 2023-10-23
Content Helper: Remember user settings


acicovic:
## Is your feature request related to a problem?
As settings like `Period`, `Metric` and others are being introduced into the Content Helper, we might be running into usability issues since we're only relying on hard-coded default setting values.

For example, the Dashboard Widget has a default `Period` of 7 days. For some users though, the ideal setting might be 30 days. These users are currently obliged to switch to 30 days manually every time they load the Dashboard Widget.

## Describe the solution you'd like
For certain settings, we should remember the user's last selections and restore them.

mjangda:
This is the user settings api I had mentioned that we could hook into: https://developer.wordpress.org/reference/functions/wp_user_settings/

It stores the values per-user, per-blog in a cookie (so it's available client-side) and in user meta (so it persists across sessions).

We do need to be minimalistic with how much data we store with it though.

acicovic:
Pushing this to the 3.13 milestone so we can make a 3.12 release faster.

PR #2091, opened 2023-12-21
Content Helper Dashboard Widget: Remember settings

acicovic:
## Description
This PR enables the Content Helper's Dashboard Widget to remember the last selected settings. Instead of reverting to its default settings, the Widget will now load the latest selected settings whenever rendered. The settings are stored in the WordPress database and are per user.

To implement this, we've introduced a WordPress REST API endpoint that can save and retrieve WordPress user meta. Before the Dashboard Widget renders, we inject the settings into it. And every time a setting is changed, we automatically send the updated settings to the endpoint so they can be saved into the database.

## Motivation and context
- Having the Content Helper remember its settings is essential UX functionality.
- The next step of this work is to port this functionality to the Content Helper Sidebar.
- Related to #1954.

## How has this been tested?
Manual testing, and a few integration tests are coming as a separate PR. Some integration tests randomly started failing with this PR and a workaround for these failures has been implemented.


src/content-helper/dashboard-widget/class-dashboard-widget.php
```diff
@@ -144,4 +146,33 @@ public function enqueue_assets(): void {
...
+		$settings = '';
+
+		if ( ! defined( 'INTEGRATION_TESTS_RUNNING' ) ) {
+			$settings = rest_do_request(
+				new WP_REST_Request(
+					'GET',
+					'/wp-parsely/v1' . Dashboard_Widget_Settings_Endpoint::get_route()
+				)
+			)->get_data();
+		}
```
acicovic, line 166: This is the code that was making integration tests randomly fail. I tried some things but couldn't find a way around it, so I just created a variable to check whether we're running this in the context of integration tests.

I recognize that this is more of a workaround rather than a true solution, but I still prefer it to moving the logic in the React part of the Widget.

Having said that, I'm happy to hear any feedback, reconsider, etc. We may need to use this approach in other parts of the Content Helper as we're implementing settings retention.

src/Endpoints/user-meta/class-base-endpoint-user-meta.php
```diff
@@ -0,0 +1,208 @@
...
+	 * Constructor.
+	 *
+	 * @param Parsely $parsely The Parsely object instance.
+	 *
+	 * @since 3.13.0
+	 */
+	public function __construct( Parsely $parsely ) {
+		parent::__construct( $parsely );
+		// phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedConstantFound
+		define( 'ENDPOINT', static::get_route() );
```
vaurdan, line 71: What is the purpose of this `define` and this constant? Since it will be defined in the global namespace, if we need to use it, we should add some namespace prefix, eg `define( __NAMESPACE__ . '\ENDPOINT', static::get_route() );`

vaurdan (commented):
I reviewed the PR and overall looks good to me! I also tested locally and the saving seems to be working nicely on the dashboard widget. Good work!!

I added a comment regarding the `ENDPOINT` constant - take a look and let me know your thoughts on that.

src/Endpoints/user-meta/class-base-endpoint-user-meta.php
vaurdan: I just noticed that this was removed on #2092 so changing my review to Approved :)

vaurdan (approved):

src/content-helper/dashboard-widget/components/top-posts.tsx
```diff
@@ -20,21 +22,95 @@ import {
...
+ * defaults.
+ *
+ * @since 3.13.0
+ *
+ * @param {string} settingsJson The JSON containing the settings.
+ *
+ * @return {TopPostsSettings} The resulting settings object.
+ */
+const getSettingsFromJson = ( settingsJson: string ): TopPostsSettings => {
+	const parsedSettings: TopPostsSettings = JSON.parse( settingsJson );
```
vaurdan, line 60: I have a couple suggestions here:
 * Make `window.wpParselyContentHelperSettings` the default value of `settingsJson` (in the `TopPostsProps` component properties), so we don't need to explicitly set it with `<TopPosts settingsJson={ window.wpParselyContentHelperSettings } />` 
 *  in case `settingsJson` (aka `window.wpParselyContentHelperSettings`) is undefined for some reason, we could perhaps fallback to retrieving the user settings with a HTTP request to the API endpoint. This could add a little bit of complexity as it would require the function to be async, but might be worth to prevent any issues with the loading order of the scripts. 
 
If you think it makes sense, we can definitely tackle this down the line on a different PR. Let me know your thoughts.

acicovic: Thanks for the feedback!

- What I've done in the Sidebar (currently working on it) is accessing the settings directly in the component. So I think we could actually skip the whole prop situation here. Thoughts?
- My take is that I'd rather skip the complexity if I can. However, the function itself should definitely be resistant to `undefined` or malformed data. Technically, the endpoint returns defaults if something is broken, but this function is an extra protection if the endpoint isn't called or it happens to return junk data. So we should shield it as much as possible, and also see what happens if settings change form. So I think that there might be room for improvements here.

Whatever we do, let's do it in another PR. I'll leave this discussion open for further commenting.

PR #2106, opened 2024-01-09
Content Helper Editor Sidebar: Remember settings

### LINKED ISSUE #1954: Content Helper: Remember user settings

acicovic:
## Description
This PR enables the Content Helper's Editor Sidebar to remember the last selected settings. Instead of reverting to its default settings, the Sidebar will now load the latest selected settings whenever rendered. The settings are stored in the WordPress database and are per user.

This work is based on the initial effort done in #2091, and the code was refactored in the hopes to be usable across the whole PCH with minimal code duplication. For this reason, some code related to the Dashboard Widget has also been updated.

Noting that initially, I tried to work with settings that used nesting. However, this was actually introducing several complexities that could become pain points, so I reverted to a flat settings structure.

## Motivation and context
- Having the Content Helper remember its settings is essential UX functionality.
- Closes #1954. Yay!

## How has this been tested?
- Manual testing, and a few integration tests are coming as a separate PR. As part of the integration tests, I also expect to test things more.
- With the changes, some existing tests started failing and other had wrong coverage docblocks. These were all updated.

## Overview of how to use this in your own code
1. To create yous settings endpoint, create a class based on `Base_Endpoint_User_Meta`. You can use the `Editor_Sidebar_Settings_Endpoint` class as a template. Update it to reflect your settings.
2. Extend your PCH feature from the `Content_Helper_Feature` class. You can use the `Editor_Sidebar` class as a template. In the `$this->inject_inline_scripts` call, pass the `get_route()` function of the Endpoint class you created in the first step.
3. In your React component, use the new `useSaveSettings` hook to save any settings whenever they are updated. As a convention, try to put the call before any `useEffect` calls. Also, it is recommended to trigger the save as early as possible (for example, before any telemetry calls etc.).
4. To further protect your component against possibly corrupt or invalid options, create a function that will validate the options so anything bad can't pass through. You can take the `getSettingsFromJson` function of the `ContentHelperEditorSidebar` component as an example.


src/content-helper/editor-sidebar/title-suggestions/component.tsx
```diff
@@ -9,24 +9,41 @@ import { __ } from '@wordpress/i18n';
 /**
  * Internal dependencies
  */
+import { GutenbergFunction } from '../../../@types/gutenberg/types';
+import { Telemetry } from '../../../js/telemetry/telemetry';
+import { PersonaProp, getPersonaLabel } from '../../common/components/persona-selector';
+import { ToneProp, getToneLabel } from '../../common/components/tone-selector';
 import { ContentHelperError } from '../../common/content-helper-error';
+import { SidebarSettings } from '../editor-sidebar';
+import { TitleSuggestionsSettings } from './component-settings';
```
acicovic: FYI: This is mainly VSCode auto-sorting imports.

src/content-helper/editor-sidebar/title-suggestions/component.tsx
```diff
@@ -153,11 +170,19 @@ export const TitleSuggestionsPanel = (): JSX.Element => {
...
+			onSettingChange( 'TitleSuggestionsPersona', selectedPersona );
+			setPersona( selectedPersona );
+		} }
+		onSettingChange={ onSettingChange }
+		onToneChange={ ( selectedTone ) => {
+			onSettingChange( 'TitleSuggestionsTone', selectedTone );
+			setTone( selectedTone );
+		} }
+		persona={ initialPersona }
+		tone={ initialTone }
```
acicovic, line 185: This change of order is my own doing. Especially when there are many properties/entries, I like sorting things alphabetically. This does not apply to function declarations of course 🙂

src/content-helper/editor-sidebar/title-suggestions/store.ts
```diff
@@ -290,15 +290,15 @@ export const TitleStore = createReduxStore( 'wp-parsely/write-titles', {
 			}
 			return undefined;
 		},
-		getTone( state: TitlesState ): ToneProp {
+		getTone( state: TitlesState ): ToneProp|null {
```
acicovic, line 293: This was the easiest way to support settings without changing a lot of things, but please let me know if you think there's a better alternative.

src/content-helper/editor-sidebar/title-suggestions/component-settings.tsx
```diff
@@ -35,17 +38,18 @@ type TitleSuggestionsSettingsProps = {
...
+	isOpen,
+	onPersonaChange,
+	onSettingChange,
+	onToneChange,
+	persona,
+	tone,
 }: TitleSuggestionsSettingsProps ): JSX.Element => {
-	const [ isSettingActive, setIsSettingActive ] = useState<boolean>( false );
-	const [ isToneSelected, setIsToneSelected ] = useState<boolean>( false );
-	const [ isPersonaSelected, setIsPersonaSelected ] = useState<boolean>( false );
```
acicovic, line 46: I've changed the component to always display the tone/persona. When there's no user selection, the defaults will be used as before.

src/Endpoints/user-meta/class-editor-sidebar-settings-endpoint.php
```diff
@@ -0,0 +1,90 @@
...
+				'values'  => array( '10m', '1h', '2h', '4h', '24h', '7d', '30d' ),
+				'default' => '7d',
+			),
+			'TitleSuggestionsOpen'         => array(
+				'values'  => array( true, false ),
+				'default' => false,
+			),
+			'TitleSuggestionsPersona'      => array(
+				'values'  => array(),
+				'default' => __( 'journalist', 'wp-parsely' ),
```
acicovic: For `persona` and `tone`, I'm translating the values. If for example the website language is in Portuguese and the article is in Portuguese, we also want to send the persona and tone in Portuguese. Just citing this here in case my thinking isn't correct. As far as I can tell, that's what we're already doing in our persona/tone selector components.

acicovic (commented):
I'm opening this PR for review and leaving some initial comments.

src/content-helper/editor-sidebar/title-suggestions/store.ts
acicovic: In fact, I think that now we don't even need the store anymore. Thoughts?

src/Endpoints/user-meta/class-editor-sidebar-settings-endpoint.php
vaurdan: I believe that, since `journalist` is the key for the [persona list ](https://github.com/Parsely/wp-parsely/blob/develop/src/content-helper/common/components/persona-selector/component.tsx#L27-L30) it shouldn't be translatable. The label, however, should be, and it's a [translatable string in the persona-selector component](https://github.com/Parsely/wp-parsely/blob/develop/src/content-helper/common/components/persona-selector/component.tsx#L28). 

This key is what is sent to the API, and then it is used to write the prompt - which is in english. I'm not sure if mixing up languages in the prompt will have any side effects, but I think that having just a single language will make it more reliable. 

This comment is also valid for the tone.

src/content-helper/common/hooks/useSaveSettings.ts
```diff
@@ -0,0 +1,49 @@
...
+ * @since 3.13.0
+ *
+ * @param {string}     endpoint The settings endpoint to send the data to.
+ * @param {Settings}   data     The data to send.
+ * @param {ReactDeps } deps     The deps array that triggers saving.
+ */
+export const useSaveSettings = (
+	endpoint: string,
+	data: Settings,
+	deps: ReactDeps
```
vaurdan: From what I understand, it seems that the `deps` array elements are, most of the time, the same as the `data`  elements. Do you think it makes sense to make `deps` default to `undefined`, and in the function body do a check like the following?

```typescript
if ( undefined === deps ) { 
    deps = Object.values( data );
}
```

This would allow us to, when using the hook, omit the dependencies.

```typescript
useSaveSettings(
    'dashboard-widget-settings', { period, metric }
);
```
Let me know if it makes sense, and feel free to ignore the suggestion if you don't think so!

src/content-helper/editor-sidebar/title-suggestions/component.tsx
```diff
@@ -14,19 +14,36 @@ import { Telemetry } from '../../../js/telemetry/telemetry';
...
  * @since 3.12.0
  *
+ * @param {TitleSuggestionsPanelProps} props The component's props.
+ *
  * @return {JSX.Element} The Title Suggestions Panel.
  */
-export const TitleSuggestionsPanel = (): JSX.Element => {
+export const TitleSuggestionsPanel = ( {
+	initialPersona, initialSettingsOpen, initialTone, onSettingChange,
+}: TitleSuggestionsPanelProps ): JSX.Element => {
```
vaurdan, line 46: We could make the TitleSuggestionsPanelProps read-only. I originally forgot to add it when implemented it :) 

```suggestion
}: Readonly<TitleSuggestionsPanelProps> ): JSX.Element => {
```

src/content-helper/editor-sidebar/title-suggestions/store.ts
vaurdan: That's a good point. My intuition says that it shouldn't be necessary now, since we're storing this in the settings. If you want to, feel free to remove it from the store and give it a try, otherwise I can remove it later down the line when I open a PR for the cross linker settings.

vaurdan (commented):
Thanks for working on this, looks pretty good!

I have added a few comments, let me know your thoughts.

src/content-helper/editor-sidebar/title-suggestions/store.ts
acicovic: OK, I'll try remove the store as part of another PR to see what happens 🙂

src/Endpoints/user-meta/class-editor-sidebar-settings-endpoint.php
acicovic: OK, so my misunderstanding here is that I thought this was the actual string, not the key. So I'll remove the translation.

src/content-helper/common/hooks/useSaveSettings.ts
acicovic: Yes, this makes total sense, I'll try it 🙂

src/content-helper/editor-sidebar/title-suggestions/component.tsx
acicovic: From a first quick look, it seems that we have forgotten this in a few places. I'm noting it to make a PR specifically about it.

src/content-helper/common/hooks/useSaveSettings.ts
acicovic: Seems to work well.

acicovic:
Hey @vaurdan, thank you for your feedback. I've just pushed a couple of commits that take care of your feedback, except for the discussions that are still open here and will be taken care of by other PRs.

acicovic:
Don't re-review yet. One of my changes has a side-effect.

acicovic:
We should be good now. The builds take forever, but this might be a GitHub issue. I'm requesting a re-review and we can see about the builds tomorrow.

acicovic:
Actually, I have an indication of where the issue is, but need to dig into it further. Will look into it tomorrow.

acicovic:
We should be good now!

vaurdan (approved):
Approved! I tested it in a local environment, and the saving and loading of settings are  working just fine.

Thank you for working on this!

src/content-helper/editor-sidebar/title-suggestions/store.ts
acicovic: Done in #2122.

PR #2111, opened 2024-01-10
Allow remote requests to Parse.ly in controlled environments

vaurdan:
## Description
Added `allow_parsely_remote_requests` method to enable remote requests to Parse.ly in restrictive environments.

## Motivation and Context
This change allows the plugin to interact with Parse.ly APIs in environments, such as wp-now, that normally block remote requests.

## How has this been tested?
Tested in a wp-now environment to ensure that remote requests to Parse.ly APIs are successfully allowed.


src/class-parsely.php
```diff
@@ -812,4 +814,35 @@ private function sanitize_managed_option( string $option_id, $value ) {
...
+			self::PUBLIC_API_BASE_URL,
+			self::PUBLIC_SUGGESTIONS_API_BASE_URL,
+		);
+
+		add_filter(
+			'http_request_host_is_external',
+			function ( $external, $host, $url ) use ( $allowed_urls ) {
+				// Check if the URL matches any URLs on the allowed list.
+				foreach ( $allowed_urls as $allowed_url ) {
+					if ( str_starts_with( $url, $allowed_url ) ) {
```
acicovic: `str_starts_with` is a PHP 8 function but we need to support PHP 7.2 and up.

Maybe it would be worth creating a utility function for it in `utils.php` and adding a comment that it can be replaced with `str_starts_with` when the minimum PHP support version becomes 8.

Feel free to implement as you like, I found this:
https://github.com/Automattic/vip-go-mu-plugins-built/blob/43ad825eb2dc3c6a8cd9f12144f162aca2ab6090/vip-helpers/vip-utils.php#L357-L359

vaurdan: Ahh good catch, I completely forgot that caveat with the `str_starts_with`. Will add a fallback in `utils.php` :)

src/Utils/utils.php
```diff
@@ -340,3 +340,23 @@ function convert_endpoint_to_filter_key( string $endpoint ): string {
...
+ * Checks if a string starts with a specific substring.
+ *
+ * This function uses the built-in PHP function `str_starts_with` if it's available (PHP 8.0 and later).
+ * If the function is not available (PHP versions prior to 8.0), it uses the `strpos` function as a fallback.
+ *
+ * @since 3.13.0
+ *
+ * @param string $haystack The string to search in.
+ * @param string $needle The substring to search for at the start of $haystack.
+ *
```
```suggestion
```

acicovic (commented):
Just a line removal comment. This is ready to roll!

src/Utils/utils.php
acicovic: Feel free to merge as-is, or after removing the empty line.

PR #2114, opened 2024-01-11
Make props passed to functions read-only

acicovic:
## Description
This PR makes read-only the props passed to functions.

## Motivation and context
This was initiated by the discussion [here](https://github.com/Parsely/wp-parsely/pull/2106#discussion_r1448681895).

## How has this been tested?
There are no changes in the transpiled code.


vaurdan (approved):
Thanks for doing this! Looks great!

PR #2119, opened 2024-01-15
Fix singular translation issues & accessibility improvements 

### LINKED ISSUE #2033: Incorrect use of singular/plural format for period descriptions
## Describe the bug
Currently, the [getPeriodDescription() utility function](https://github.com/Parsely/wp-parsely/blame/develop/src/content-helper/common/utils/constants.ts#L93-L109) is using `_n()` and the singular and plural forms to return a different string based on the amount of time, i.e. 1 or more than 1.

However, in some languages the singular form can be used for numbers other than 1. For example, in Russian numbers like 21, 31, etc. would also use the singular form, resulting in, for example, `Last 21 Hours` returning the translation for `Last Hour`.

If you would like to use a special case for `[1] Hour/Minute/Day`, check for it specifically:

```js
if ( timeValue === 1 ) {
	__( 'Last Minute', 'wp-parsely' );
} else {
	_n( 'Last %1$d Minute', 'Last %1$d Minutes', timeValue, 'wp-parsely' )
}
```

## Additional context
Please refer to https://codex.wordpress.org/I18n_for_WordPress_Developers#Plurals for more detailed examplanation.

vaurdan:
## Description
This PR introduces some improvements to the Parse.ly plugin. It adds aria labels to previous and next navigations on the Dashboard Widget, enhancing user accessibility and experience. Additionally, it removes an unnecessary translatable string and fixes a translation issue in the `getPeriodDescription()` function. The latter change ensures accurate display of time-related strings in various languages, addressing issues with singular and plural forms.

These issues have been reported in #2033 and closes it. 

## Motivation and context
The update is addresses a specific bug in the `getPeriodDescription()` function, as discussed in the [original issue](https://github.com/Parsely/wp-parsely/issues/2033). This bug affected the handling of singular and plural forms in languages like Russian. Improving accessibility in the Dashboard Widget and refining the overall user experience are also motivations for these changes.

## How has this been tested?
Tested on my local environment and validated if the strings are still correctly displayed, and using Voice Over, if the newly added aria labels are used instead of the button text.

N/A


src/content-helper/common/utils/constants.ts
```diff
@@ -90,20 +90,32 @@ export function getPeriodDescription(
 	/* eslint-disable @wordpress/valid-sprintf */
 	switch ( timeUnit ) {
 		case 'm':
+			if ( timeValue === 1 ) {
```
acicovic: Since those conditions are simple, I'd convert them into Yoda conditions. Also, TIL that [WP Coding Standards ](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#yoda-conditions) actually recommend to avoid Yoda conditions for for `<`, `>`, `<=` and `>=`.

acicovic (approved):
Thank you! approving this with a minor comment.

PR #2121, opened 2024-01-15
Fix PHPStan error in utils.php

acicovic:
## Description
This PR introduces a fix for our pre-commit hooks that would fail with the following PHPStan error:

```
 ------ ---------------------------------------------------- 
  Line   src/Utils/utils.php                                 
 ------ ---------------------------------------------------- 
  360    Casting to string something that's already string.  
  360    Casting to string something that's already string.  
 ------ ---------------------------------------------------- 
```

## Motivation and context
Avoid our workflows from breaking.

## How has this been tested?
No testing, code does not change essentially.

### DIFF
--- a/src/Utils/utils.php
+++ b/src/Utils/utils.php
@@ -357,5 +357,5 @@ function str_starts_with( string $haystack, string $needle ): bool {
 	if ( function_exists( '\str_starts_with' ) ) {
 		return \str_starts_with( $haystack, $needle );
 	}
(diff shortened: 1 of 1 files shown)


vaurdan (approved):
Thanks for fixing it! Not sure how this got through.

acicovic:
>Not sure how this got through.

Yeah, me neither. However, I got the error consistently before and after a `composer update`.

PR #2122, opened 2024-01-15
PCH Title Suggestions: Remove tone and persona from store

acicovic:
## Description
With the introduction of settings in the Editor Sidebar in #2106, we should not need to save tone and persona in the store anymore. This PR removes the bits of code that were responsible for this.

## Motivation and context
Don't keep unneeded code.

## How has this been tested?
Manually. Things seem to work as usual. @vaurdan, I'd appreciate a second look on that though, as you've been the contributor of this feature.


vaurdan (approved):
Looks right to me!

Something that I have been wondering is if we should have the settings save/load login inside the store itself. So it would initially retrieve the value from the settings json (as the default, I would say) and add the logic to `setTone`/`setPersona` to send the HTTP request to save the settings on the database, alongside with updating the the state itself.  That way we would manage all the states, including the settings, from the store itself.

Either way, the PR looks good and I'm approving it, so feel free to merge at your discretion 🙂 But let me know your thoughts, and if my idea makes sense.

acicovic:
Hey @vaurdan, interesting idea. At this point, I'm still very unfamiliar with stores. Two questions that come up are:
1. Besides managing everything from the store, do you think that this approach could have other advantages?
2. Have you seen this been done somewhere? Is this for example a known convention?

acicovic:
I'll be merging this, but let's continue the conversation here or anywhere else.

PR #2131, opened 2024-01-17
PCH Smart Linking: Implement settings system

vaurdan:
## Description
This PR implements the new settings system, introduced in #2106, in the Cross Linker functionality of the Parse.ly Content Helper. 

Worth noting that, in opposition to what we did in the TItle Suggestions where we removed the settings from its store, here we need to keep the individual settings in the store, since this panel is available in two different contexts: in the PCH sidebar, and also on the plugin inspector. Without having these in the store, both instances would get desynchronized. 

## Motivation and context
Include the new settings system in the newest PCH component.

## How has this been tested?
Tested this locally and made sure that the settings are saved and persist across reloads, and that both instances of the Cross Linker panel are sharing the same settings.


acicovic:
Thanks for explaining why the store is needed. I've made a few minor adjustments and will be merging this into its parent PR soon.

acicovic (approved):

PR #2198, opened 2024-02-07
Refactor endpoints availability code

acicovic:
## Description
This PR started as work to harden our User Meta endpoints a bit. Since additional things made sense or needed to be changed, this resulted in a more general refactoring. This PR improves our endpoints situation with the following adjustments:

- User ID defaults to `0` and cannot be passed into User Meta endpoint functions as parameters.
- Replacing `Base_Endpoint`'s `is_public_endpoint`, `user_capability` and `permission_callback()` members with a single `is_available_to_current_user()` function, which must be implemented in child classes. This has the following effects:
  - Simpler `Base_Endpoint` internals and less error-prone structure that could affect a large number of endpoints simultaneously.
  - Every individual endpoint is capable of enforcing its own availability rules in a way that wasn't possible before.
  - More declarative code style is needed, which should be clearer to understand, though it does introduce a small amount of repetition.
- Integration tests have also been accordingly updated. Coverage annotations have also been updated on a best effort basis, and are incomplete. No significant effort was put there as endpoint classes are bound to change in the future, thus probably making that effort a waste of time.

No significant amount of time should be given for reviewing the updated code coverage annotations. This will be done when we start updating them to be correct for the whole project. Currently, we're missing a lot of them.

## Motivation and context
- This work is part of a starting bigger effort to refactor our endpoints code to something that's easier to understand and maintain.
- @mjangda had asked us to see if we can improve our User Meta endpoints so Users IDs are a bit more protected.

## How has this been tested?
- Manual walkthrough of the PCH to see that endpoints still work as expected.
- Some tests have been modified, and all tests pass.


src/RemoteAPI/class-analytics-post-detail-api.php
```diff
@@ -23,10 +24,19 @@ class Analytics_Post_Detail_API extends Base_Endpoint_Remote {
...
+	 * user.
 	 *
-	 * @since 3.7.0
-	 * @var bool
+	 * @since 3.14.0
+	 *
+	 * @return bool
 	 */
-	protected $is_public_endpoint = false;
+	public function is_available_to_current_user(): bool {
```
vaurdan, line 34: Since most of the times the wanted behavior is to check for the `Base_Endpoint::DEFAULT_ACCESS_CAPABILITY` capability, would it make sense to implement this function in the [Base_Endpoint class](https://github.com/Parsely/wp-parsely/blob/2298ad122f1d6ea34dbbfdffd732bda95dfccd66/src/Endpoints/class-base-endpoint.php#L63) and have it be the default behavior?

vaurdan (approved):
Looks great to me, thanks for working on this! 🙂 

I have added a single suggestion, you are free to ignore if you think it doesn't apply or doesn't make sense.

src/RemoteAPI/class-analytics-post-detail-api.php
acicovic: Thanks for chiming in about this. I didn't set a default behavior on purpose, in order to force the implementation in child classes. The idea was to be more explicit about it since we're setting access permissions. I agree it does duplicate some lines of code though.

vaurdan: Sounds reasonable to me. 🙂

PR #2200, opened 2024-02-09
Telemetry: Fix bad reporting when adding or removing Parse.ly blocks

vaurdan:
## Description
There was a problem with the Block Change Monitor component, that monitors the blocks in the editor and sends Tracks events when a Parse.ly block is either added or removed from the content. For more information on this issue, see #2163

This PR refactors the logic of the BCM while also tries to improve the performance by introducing a debounce to the subscribed event. This saves many CPU cycles since this functions runs **every time** there is a change to the store. 

There is also a new initialization delay. This is to prevent reporting blocks that are not actually being inserted, but that are loaded when the block editor loads. This would erroneously send a Track event reporting that a block was added, when it was only being loaded. By introducing this delay, we allow the block editor to load first all the blocks, and then report only the newly added blocks.

## Motivation and context
Have a good correct reporting of the Parse.ly custom blocks usage, and improve the overall performance of the Parse.ly plugin.

## How has this been tested?
Tested locally on my WordPress environment, while looking into the HTTP requests being fired by the block editor. 

N/A


src/js/telemetry/block-change.tsx
```diff
@@ -1,83 +1,95 @@
...
-import { BlockInstance } from '@wordpress/blocks';
-import { select, subscribe } from '@wordpress/data';
 import { useEffect } from '@wordpress/element';
+import { select, subscribe } from '@wordpress/data';
+
+/**
+ * External dependencies
+ */
+// eslint-disable-next-line import/no-extraneous-dependencies
+import debounce from 'lodash.debounce';
```
acicovic: Should this be `lodash/debounce`, or does it make no difference? Either way, let's stick to one convention (we're using `lodash/isequal` in another PR).

src/js/telemetry/block-change.tsx
```diff
@@ -1,83 +1,95 @@
...
- * when the component is mounted.
- * When the block editor changes, it checks if blocks have been added or removed by comparing the current
- * list of blocks with the previous one.
- * If a block has been added or removed, it sends a telemetry event to the server.
- * When the component is unmounted, it unsubscribes from the block editor changes.
+ * BlockChangeMonitor is a React component that monitors changes in the WordPress block editor.
+ * It tracks the addition and removal of blocks that start with the 'wp-parsely/' prefix.
+ * This component does not render anything.
  *
- * @since 3.12.0
```
acicovic, line 24: I see we've removed a  couple of `@since` docblocks. Was this intentional?

src/js/telemetry/block-change.tsx
```diff
@@ -1,83 +1,95 @@
...
+	 * When the component is unmounted, it clears the initialization timeout.
 	 *
 	 * @since 3.12.0
+	 * @since 3.14.0 Improved detection by comparing current and previous block states directly.
 	 */
 	useEffect( () => {
-		const getBlockList = () => select( 'core/block-editor' ).getBlocks() as BlockInstance[];
-		let lastBlockIds = new Set( getBlockList().map( ( block ) => block.clientId ) );
+		// Delay the initialization to avoid reacting to the initial block load.
+		const initializationDelay = 1000; // Delay in milliseconds.
```
acicovic: Just a question for my own understanding: Could there be a small margin of error here for very large pages that could take more than 1 second to load in the editor?

src/js/telemetry/block-change.tsx
vaurdan: Ah yes, since I rewrote a bit of the docblocks I accidently removed some `@since`.

src/js/telemetry/block-change.tsx
vaurdan: That's a fair point. I raised the initialization delay to 2.5 seconds, which hopefully should be more than enough for the all the blocks to load. 

I'm also trying a dynamic solution either using [`__unstableIsEditorReady` ](https://github.com/WordPress/gutenberg/blob/26e2bf6d009431868ee01fa27fb630e9e9514c00/packages/editor/src/store/selectors.js#L1383-L1385), or a different approach.

vaurdan: I just committed a change that uses a couple selectors to determine if the editor is ready. I believe this will be more reliable than having a fixed initialization delay.

I tried `__unstableIsEditorReady` but it was unreliable for our use-case.

acicovic (approved):
Thank you for working on this! 🙂

PR #2220, opened 2024-02-16
Include more metadata in content tracked as a Non-Post

### LINKED ISSUE #1735: Include more metadata for non-posts
Several Parse.ly customers found that their metadata doesn't appear in Dash for non-post pages. 

I would like for the plugin to extract the same metadata fields for non-post pages that posts receive.

I'm simultaneously pursing a request to the Parse.ly Backend team to enable a new version of the Parse.ly crawler to extract metadata from a wider variety of metadata sources and their respective standards.

## Additional context
During initial development of this plugin, Parse.ly Dash did not support metadata for non-post pages but now it does.

acicovic:
## Description
This PR adds more metadata into content being tracked as a `Non-Post`<sup>(1)</sup>, almost bringing it into parity<sup>(2)</sup> with content that is being tracked as a `Post`.

1. This behavior only affects content that is caught in [this block of code](https://github.com/Parsely/wp-parsely/blob/0c2572fe76c0223ac3f44a709bdb9a5144b036a0/src/class-metadata.php#L120). Everything else remains untouched. For example, homepage metadata remains the same, even if the homepage happens to be a WordPress `page`.
2. `Posts` run the `build_type()`, `build_main_entity()` and `build_publisher()` functions that are not executed for `Non-Posts`. If we conclude that these are needed as well, we can include them.

Concerning the internal behavior of the additions, for example how `section` or `keywords` are picked, it is identical to what is being done for `Posts`.

WordPress installations that extend the built-in `page` Post Type to support categories and tags should be supported. This is being tested [here](https://github.com/Parsely/wp-parsely/blob/0c2572fe76c0223ac3f44a709bdb9a5144b036a0/tests/Integration/Metadata/SinglePageTest.php#L415-L443). We've also tested the implementation with [this plugin](https://wordpress.org/plugins/pages-with-category-and-tag/) and it has worked. However, any plugin is subject to change and this is not an officially supported integration.

## Motivation and context
- This has been a long time request, and we know that some users want it.
- Closes #1735.

## How has this been tested?
- Manual testing.
- Updated existing tests and added some more.

### DIFF
--- a/src/Metadata/class-metadata-builder.php
+++ b/src/Metadata/class-metadata-builder.php
@@ -11,6 +11,10 @@ declare(strict_types=1);
 namespace Parsely\Metadata;
 
 use Parsely\Parsely;
+use WP_Post;
+use WP_User;
+
+use function Parsely\Utils\get_default_category;
 
 /**
  * Abstract class that implements modular builders for Metadata.
@@ -20,6 +24,8 @@ use Parsely\Parsely;
  *
  * @since 1.0.0
  * @since 3.3.0 Logic extracted from Parsely\Parsely class to separate file/class.
+ *
+ * @phpstan-import-type Parsely_Options from Parsely
  */
 abstract class Metadata_Builder {
 	/**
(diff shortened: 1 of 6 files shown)


vaurdan (approved):
Amazing work, thank you for moving this forward!

PR #2221, opened 2024-02-16
PCH Smart Linking: Fix disappearing overlay issue

vaurdan:
## Description
After the introduction of the new settings API, when a block is processing the smart links, if you would click on any other block, the block overlay would disappear.

This PR fixes this issue with a small refactor of the overlay, to use the editor.BlockEdit filter instead of relying on a BlockOverlayContainer component to manage the overlays.

## Motivation and context
Make the UX consistent and prevent the block overlay from disappear.

## How has this been tested?
Tested locally on my dev environment, for both single block and full content smart linking.


src/content-helper/editor-sidebar/smart-linking/component-block-overlay.tsx
```diff
@@ -131,16 +134,62 @@ export const BlockOverlayContainer = ( ): JSX.Element => {
...
+			/>
+		);
+	}
+
+	return <></>;
+};
+
+/**
+ * A higher-order component that adds a block overlay over a specific block, flagged by the Smart Linking store.
+ */
```
acicovic, line 153:
```suggestion
 *
 * @since 3.14.0
 */
```

src/content-helper/editor-sidebar/smart-linking/component-block-overlay.tsx
```diff
@@ -131,16 +134,62 @@ export const BlockOverlayContainer = ( ): JSX.Element => {
...
+	return ( props ) => {
+		const { overlayBlocks } = useSelect( ( select ) => {
+			const { getOverlayBlocks } = select( SmartLinkingStore );
+
+			return {
+				overlayBlocks: getOverlayBlocks(),
+			};
+		}, [] );
+
+		// If the block id is currently on the overlayBlocks array, we should render the overlay.
```
acicovic:
```suggestion
		// If the block ID is currently on the overlayBlocks array, we should render the overlay.
```

acicovic (approved):
Thank you!
Left a couple of minor comments so this is good to go. Feel free to merge without further review.

PR #2232, opened 2024-02-22
PCH Redesign: Add tabs interface to the editor sidebar

vaurdan:
## Description
Adds a new interface with tabs to organize the panels in the editor sidebar.

## Motivation and context
- Improve the UX of the Parse.ly Content Helper and make it easier to use.
- To be merged into #2238.

## How has this been tested?
Tested locally and validated that all the existing functionality is still working as expected.


src/content-helper/editor-sidebar/editor-sidebar.scss
```diff
@@ -33,6 +33,54 @@
...
+	.wp-parsely-sidebar-main-panel {
+
+		.wp-parsely-sidebar-tabs {
+
+			.components-tab-panel__tabs {
+
+				// Individual tab.
+				button {
+					display: flex;
+					height: 48px;
```
acicovic: Any reason we're not using `to_rem()` here?

src/content-helper/editor-sidebar/editor-sidebar.tsx
```diff
@@ -336,15 +265,45 @@ const ContentHelperEditorSidebar = (): JSX.Element => {
...
+									onClick={ () => {
+										Telemetry.trackEvent( 'editor_sidebar_view_post_pressed' );
+									} }
+									href={ window.wpParselyPostUrl }
+									rel="noopener"
+									target="_blank"
+								>
+									{
+										/* translators: %s: Post type */
+										sprintf( __( 'View this %s in Parse.ly', 'wp-parsely' ), 'post' )
```
acicovic, line 297: Seems we'll need to get the post type here. If out of scope or not possible, we can just revise the message to "View this in Parse.ly".

src/content-helper/editor-sidebar/editor-sidebar.tsx
```diff
@@ -354,83 +313,36 @@ const ContentHelperEditorSidebar = (): JSX.Element => {
 						<Settings />
 					</PanelBody>
 				</Panel>
```
acicovic: I'd move the settings panel to the bottom if it isn't a hassle. I think it will make things look better in this temporary state.

src/content-helper/editor-sidebar/tabs/sidebar-performance-tab.tsx
```diff
@@ -0,0 +1,56 @@
...
+ */
+export const SidebarPerformanceTab = ( { trackToggle }: SidebarPerformanceTabProps ) => {
+	const { settings, setSettings } = useSettings<SidebarSettings>();
+
+	return (
+		<Panel>
+			<PanelBody
+				title={ __( 'Performance Details', 'wp-parsely' ) }
+				initialOpen={ settings.PerformanceDetailsOpen }
+				onToggle={ ( next ) => {
```
acicovic: Not necessarily something for this PR, but wanted to mention that since this the only content in the Performance tab, the goal is to make it not togglable.

acicovic (commented):
Thank you for this work! Left some comments, nothing blocking.

src/content-helper/editor-sidebar/editor-sidebar.tsx
acicovic: Just pushed a commit that does that.

src/content-helper/editor-sidebar/tabs/sidebar-performance-tab.tsx
acicovic: I've addressed this with a new commit. There's no margin so it doesn't look great, but since the inner component will be receiving a revamp, I've left it this way for now.

src/content-helper/editor-sidebar/editor-sidebar.scss
vaurdan: Nope, it was copy-pasted from Figma and forgot to add `to_rem` here, good catch :)

acicovic (approved):

PR #2238, opened 2024-02-23
PCH: Redesign the PCH Editor Sidebar

acicovic:
## Description
This PR consolidates all the work that has been done in the context of redesigning the Content Helper Editor Sidebar. This was a big amount of work touching many parts of our codebase, so it was made step by step in smaller PRs which are referenced in this PR's comments.

### Side effects
- As part of this work, parts of our settings system were rewritten. Due to this, some settings in the `Related Posts` and `Performance Details` sections will be reset. Crafting a migration path for this would have taken time without important benefits.

## Motivation and context
Closes #2206.

## How has this been tested?
- Manual testing and QA reviews.
- Automated tests have been updated to pass.

## Screenshots

screenshot-3
screenshot-4
screenshot-5
screenshot-6
screenshot-7

PR #2243, opened 2024-02-26
PCH Related Posts: Fix "undefined" author message

acicovic:
## Description
Under certain circumstances, we were getting a	`Related posts by author "undefined" in the last x days.` message in the PCH Related Posts panel. This PR fixes the issue.

## Motivation and context
- Fix bugs.
- To be merged into #2238.

## How has this been tested?
- Manually verified that the error doesn't occur with the fix.
- Updated some tests to make them pass.


vaurdan (approved):
Thanks for digging into this issue! LGTM

PR #2246, opened 2024-02-27
PCH Performance Stats: Update Performance panel UI

vaurdan:
## Description
Updates the Performance Stats panel UI to improve the user experience. 

This PR adds a few new telemetry events, that I'm listing below:

| Event Name                                           | Description                                                      |
|------------------------------------------------------|------------------------------------------------------------------|
| `editor_sidebar_performance_datapoint_hidden`        | Fires when a performance data point is hidden from the editor sidebar. |
| `editor_sidebar_performance_datapoint_shown`         | Fires when a performance data point is shown in the editor sidebar. |
| `editor_sidebar_performance_datapoints_reset`        | Fires when performance data points are reset to default in the editor sidebar. |
| `editor_sidebar_performance_panel_closed`            | Fires when a performance panel is closed in the editor sidebar.  |
| `editor_sidebar_performance_panel_opened`            | Fires when a performance panel is opened in the editor sidebar.  |
| `editor_sidebar_performance_panel_reset`             | Fires when performance panels are reset to default in the editor sidebar. |
| `editor_sidebar_performance_period_changed`          | Fires when the performance period selection is changed in the editor sidebar. |

## Motivation and context
* Improve the UX of the Parse.ly Content Helper

## How has this been tested?


vaurdan:
The duplicated lines warning from SonarCloud will be addressed in the next phase.

.eslintrc
```diff
@@ -45,6 +45,7 @@
 		"@typescript-eslint/prefer-nullish-coalescing": ["error"],
 		// Enabling TS rule and disabling Base rule as it can report incorrect errors.
 		"no-shadow": "off",
-		"@typescript-eslint/no-shadow": ["error"]
+		"@typescript-eslint/no-shadow": ["error"],
+		"@wordpress/no-unsafe-wp-apis": "off"
```
acicovic, line 50: Are we using this only for being able to use experimental Gutenberg components, or are there additional use cases?

src/content-helper/editor-sidebar/editor-sidebar.scss
```diff
@@ -186,6 +186,365 @@
...
+			display: flex;
+			justify-content: center;
+			align-items: center;
+			width: 100%;
+		}
+	}
+
+	.performance-stat-panel {
+		display: flex;
+		padding: to_rem(6px) 0 to_rem(16px) 0;
```
acicovic: Let's discuss about this `to_rem()` function, whether we think it is useful and should be kept as a practice.

src/content-helper/editor-sidebar/editor-sidebar.scss
```diff
@@ -186,6 +186,365 @@
...
+			display: flex;
+			flex: 1 0 0;
+			height: 16px;
+			align-items: center;
+			gap: 6px;
+			align-self: stretch;
+			color: var(--Gutenberg-Gray-700, #757575);
+
+			/* Label */
+			font-family: "SF Pro Text", sans-serif;
```
acicovic: How was the particular font chosen?

acicovic (commented):
Outstanding work!

I've pushed a commit with mostly minor adjustments, see if you agree with the changes.

Left some questions, and I'll also give this a manual test run.

src/content-helper/editor-sidebar/performance-details/component.tsx
```diff
@@ -73,233 +221,82 @@ export function PerformanceDetails(
...
+						onClick={ () => {
+							Telemetry.trackEvent( 'editor_sidebar_view_post_pressed' );
+						} }
+						href={ window.wpParselyPostUrl }
+						rel="noopener"
+						target="_blank"
+					>
+						{
+							/* translators: %s: Post type */
+							sprintf( __( 'View this %s in Parse.ly', 'wp-parsely' ), 'post' )
```
acicovic: This `%s` is always `post`. Is this flagged to be worked on? Also, if this contains the name of CPTs with very long names, text flows outside of the button.

image

IMO it's totally fine to change the text `View in Parse.ly` or `View this in Parse.ly` if there are too many unknowns currently. We also probably need more spacing between the select control and the button? My first impression was that they're too tight.

Also wondering on whether we should actually inverse these two, but we can do that later depending on the feedback we'll get.

cc @danielabloch

.eslintrc
vaurdan: Yes, only for the experimental components. As far as I'm aware, that's the only thing that this specific ESLint rule checks for.

src/content-helper/editor-sidebar/editor-sidebar.scss
vaurdan: It came straight from FIgma - I used the styles provided there with some tweaks. But I'm glad you asked, because that's something I wanted to discuss, since I'm not sure if we should be using this font, and the fall back `sans-serif` I picked was really just temporarily.

src/content-helper/editor-sidebar/performance-details/component.tsx
```diff
@@ -1,50 +1,198 @@
...
+ * @since 3.14.0
  */
-interface PerformanceDetailsProps {
-	period: Period;
-}
+const availablePanels = [
+	{
+		name: 'overview',
+		label: __( 'Overview', 'wp-parsely' ),
+		forced: true,
```
acicovic: Is this `forced` due to any technical constraint or issue?

src/content-helper/editor-sidebar/editor-sidebar.scss
vaurdan: I don't have a strong opinion on this, other than the hassle of having to use `to_rem` everytime I add a dimension. But using rem units is a good practice, I believe. 

As a note, it is possible that I have missed some `to_rem` in some dimensions, since some of the CSS was taken from Figma, and I didn't add the `to_rem` until I had it tested and working as we wanted. I will take a closer look and see if I didn't missed any :)

acicovic (commented):
I gave this a test run. Very impressive work!

I'm leaving some additional feedback, and was also wondering whether it would be simple and make sense to show one spinner instead of three while things are loading in there. If this is currently too much work, we can leave as is.

.eslintrc
acicovic: I'd maybe put a comment there. Something like `// Allow usage of Gutenberg experimental components.`. Not required, but add it if you think it could be useful.

src/content-helper/editor-sidebar/editor-sidebar.scss
acicovic: @dabowman?

src/content-helper/editor-sidebar/performance-details/component.tsx
acicovic: Also tagging @dabowman here.

vaurdan: Interesting point. I think that for simplicity, we can keep "View this in Parse.ly". Looking forward to hear more input on this :)

src/content-helper/editor-sidebar/performance-details/component.tsx
vaurdan: 'forced' means that the panel cannot be disabled. There isn't any technical reason, other than avoiding an empty performance tab if all the panels are disabled. @dabowman, any thoughts on this one?

vaurdan:
> I'm leaving some additional feedback, and was also wondering whether it would be simple and make sense to show one spinner instead of three while things are loading in there. If this is currently too much work, we can leave as is.

That was my first attempt. However, it doesn't look great when we change the period and the whole component transforms into a spinner. 

Additionally, in the future we might need to have individual panels do requests on their own - for example if we change from Page Views to Avg. Time in a single panel - so we can make only the affected panel move to the loading state.

In a next phase, it might be interesting to have a more "complex" loading state, where the components are there but instead of real content, it would have some kind of animation that would imply that it is loading. Something like this:

loading

However, this is completely out of the scope right now :)

src/content-helper/editor-sidebar/performance-details/component.tsx
acicovic: It's still possible to hide all data points, ending up with an empty performance tab. Although technically possible, it doesn't make a lot of sense from a user perspective to just empty everything. We could probably introduce one or more empty states in the future, or give users to get rid of the tab if they don't want it altogether.

I'd probably remove the `forced` property here, just in case there's a legit use case for hiding the overview.

acicovic:
> However, this is completely out of the scope right now :)

Agreed! 🙂

src/content-helper/editor-sidebar/editor-sidebar.scss
vaurdan: I did some improvements on the CSS, adding `to_rem` to everywhere I missed it, and adding a few extra variables.

acicovic: I agree that using relative units is the way to go, unless when there are specific reasons not to.

I think that what the `to_rem` function achieves for us is an easy way to convert `px` to `rem`, though it is hardcoded to `16px`, which would be wrong if the user has custom font sizes in their browser. Directly working with `rem`s (or `em`s) would probably be the correct thing to do, though these units are much harder to understand than `px`.

We can totally keep things as they are currently until we decide on an approach, I just wanted to get this discussion started, and maybe see how this is done in the Gutenberg world.

acicovic: I think that the rem vars you've introduced are useful. This can be a way towards using "real" `rem` values.

acicovic:
Bonus: Our last failing E2E test is now fixed.

vaurdan:
Amazing, thank you so much for your work on fixing the E2E tests!!

src/content-helper/editor-sidebar/performance-details/component.tsx
vaurdan: Sounds good, I will remove the `forced`. I will keep the logic, in case we want to use it for something else in the future, but will remove it from the overview panel.

acicovic:
This last one was only failing on GitHub, so it needed some trial and error.

acicovic:
@vaurdan, I think we're mostly done, things look good to me. Let's wait for any additional feedback to close the outstanding items.

src/content-helper/editor-sidebar/performance-details/component.tsx
danielabloch: Agree. "View this in Parse.ly" is great.

@acicovic do you mean inverse the "View this in Parse.ly" button and the period filter so that the filter is below the button?

acicovic: @danielabloch, yes. As it is right now, it's like the filter is related to the button, but this isn't the case. The filter is related to the information below, so moving it closer towards there makes more sense IMO. Augmenting the distance between them should also help.

dabowman: Initial thoughts: fine to keep remove the name of the post type for now. We should figure out how to properly truncate that if needed. The dynamic names of database stuff has always been a Wordpress feature I've appreciated, so it would be good to keep that convention here instead of just calling it "this." But fine if it needs more work to design around the technical constraints. 

As for the button, I would definitely put it above the period select. I'll need to modify the design if we're intent on moving the view in parse.ly button inside performance so the spacing can get just right.

dabowman: Just updated Figma to take the button out of the header and put it at the bottom of performance, From a UX/markup perspective I think that's what makes the most sense.

acicovic: @dabowman, @vaurdan, Figma seems broken for me right now, but this separation makes sense to me.

acicovic: I think this is how it was before as well.

vaurdan: @dabowman thank you for you input! That makes perfect sense, and looks way better IMO. I will apply that to the PR.

src/content-helper/editor-sidebar/editor-sidebar.scss
vaurdan: As we discussed, I removed the `font-famliy` rules that were introduced to use the default wp-admin fonts. 

As a side-note, on that same commit, I also moved all the Performance Stats rules to its own file in the `performance-stats` directory, to help uncluttering the `editor-sidebar.scss` file and hopefully make it more maintainable going forward.

src/content-helper/editor-sidebar/performance-details/component.tsx
vaurdan: I moved this to the bottom of the performance tab.

image

vaurdan:
@acicovic I believe that all the outstanding items have been addressed. Feel free to take a final look.

acicovic (approved):
Looks good! Thank you for all this amazing work!

PR #2250, opened 2024-02-29
Add option for full metadata in non-posts

acicovic:
## Description
This PR adds a temporary `full_metadata_in_non_posts` option. When it is set to `true`, full metadata will be injected in any content that is being tracked as a non-post.

> This is a temporary measure so customers can upgrade their code to use the new metadata system and set the setting to being enabled. The option will be removed in the future, after which this behavior will be forcefully auto-enabled for all installations.

The reason for this intermediary step is that our testing indicated that adding full metadata to non-posts could result in unexpected metadata or even fatal errors if the metadata was being changed by a filter such as `wp_parsely_metadata`. For this reason, when upgrading to the 3.14 release, this option will be set to:

- `true` if we detect that no metadata-altering filters are called by customer code.
- `false` if we detect that any metadata-altering filters are called by customer code.

This operation will take place only once, when the plugin is upgraded to 3.14 from a prior version, or a new installation of 3.14 or newer is being made. Once this initial setup has been made, the setting is being remembered as any other regular setting, and it must be manually adjusted for its value to change.

The filters that are considered metadata-altering are the following:
- `wp_parsely_metadata`
- `wp_parsely_post_tags`
- `wp_parsely_permalink`
- `wp_parsely_post_category`
- `wp_parsely_pre_authors`
- `wp_parsely_post_authors`
- `wp_parsely_custom_taxonomies`
- `wp_parsely_post_type`

For most cases, the whole operation will be totally transparent, requiring no changes or setting adjustments. However, if any of the above filters are being called by customer code, the setting will need to be enabled manually and tested for any issues or incompatibilities.

## Motivation and context
Related to #1735.

## How has this been tested?
Added several integration tests.

## Screenshots
The following setting has been added to the settings page:
image

### DIFF
--- a/src/Metadata/class-page-builder.php
+++ b/src/Metadata/class-page-builder.php
@@ -51,12 +51,14 @@ class Page_Builder extends Metadata_Builder {
 		$this->build_headline();
 		$this->build_url();
 
-		$this->build_thumbnail_url( $this->post );
-		$this->build_image( $this->post );
-		$this->build_article_section( $this->post );
-		$this->build_author( $this->post );
-		$this->build_keywords( $this->post );
-		$this->build_metadata_post_times( $this->post );
+		if ( true === $this->parsely->get_options()['full_metadata_in_non_posts'] ) {
+			$this->build_thumbnail_url( $this->post );
+			$this->build_image( $this->post );
+			$this->build_article_section( $this->post );
+			$this->build_author( $this->post );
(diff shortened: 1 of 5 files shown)


acicovic:
The failure is a SonarCloud duplication warning. We can ignore it.

vaurdan (approved):

PR #2251, opened 2024-02-29
PCH: Set some props params as Readonly

acicovic:
## Description
This sets a couple of props params as `Readonly`.

## Motivation and context
- Consistency with our other function declarations.
- To be merged into #2238.

## How has this been tested?
No functionality changes, existing tests pass.


vaurdan (approved):
Thanks for this! I wonder if there is a way to check for missing Readonly in props params with ESLint

acicovic:
I hope we can lint more things in the future. For now, much of this is manual unfortunately.

PR #2255, opened 2024-03-01
Complete metadata parity between Posts and Non-Posts

acicovic:
## Description
This PR completes the metadata parity project between Posts and Non-Posts. This feature request was reported in #1735 and implementation started in #2220.

## Motivation and context
Non-posts needed to support metadata that was only available for Posts. Full context can be found in #1735.

## How has this been tested?
Existing tests pass and we updated a test that started failing as a result of the changes.


vaurdan (approved):
Looks good! 👍 
Thanks for working on this!

PR #2257, opened 2024-03-01
PCH Performance Stats: Add Multi-percentage bar hover state

vaurdan:
## Description
Adds a nice tooltip and some CSS animation when hovering a section of the multi percentage bar on the Category panel.

## Motivation and context
* Improve the UX of the Parse.ly Content Helper

## How has this been tested?


vaurdan:
SonarCloud duplication warning can be safely ignored.

src/content-helper/editor-sidebar/performance-details/component-panel-categories.tsx
```diff
@@ -89,42 +89,43 @@ export const PerformanceCategoriesPanel = ( {
...
-									<div className={ 'single-label ' + key } key={ key }>
-										<div className={ 'label-color ' + key }></div>
-										<div className="label-text">{ getKeyTitle( key ) }</div>
-										<div className="label-value">{ formatToImpreciseNumber( value.views ) }</div>
-									</div>
-								) )
-							}
-						</div>
+							return (
+								<Tooltip text={ `${ getKeyTitle( key ) } - ${ value.viewsPercentage }%` } delay={ 150 } key={ key }>
```
acicovic: Some languages put a space before the percent sign, or put it on the left of the value. So let's use `__()` for the `text` property.

acicovic (commented):
Thanks for working on this! Left a comment.

acicovic:
Pushed a commit to bring the Multi-percentage bar animation closer to what we have in the design:

image

Notes:
- Using `!important` isn't my favorite thing, but in this context I think it's OK. Happy to go another way if there are better suggestions.
- I noticed that when the `to_rem()` was being used for `--radius`, the border-radius was not being applied. So I removed the function from there.
- I used `opacity` instead of `brightness`, because the second changes the color (for example, the orange would become yellow).

acicovic (approved):

vaurdan:
Amazing, thanks for working on those improvements!

PR #2263, opened 2024-03-06
PCH Related Posts: Update UI 

vaurdan:
## Description
Update the UI in the Related Post panel of the Editor Sidebar. 

This PR adds a couple new telemetry events:

| Event Name                                           | Description                                                      |
|------------------------------------------------------|------------------------------------------------------------------|
| `related_posts_metric_changed`        | Fires when the metric is changed |
| `related_posts_period_changed`        | Fires when the period is changed |

While the PR is ready for review, there are a few questions we need to answer before merging:
 - [x] How do we handle the ToggleGroup background issue
 - [x] Final decision on the buttons available on the individual related posts
 - [x] Fix tests that have been broken by the significant change in the panel structure
 
## Motivation and context
* Improve the UX of the Parse.ly Content Helper and update Related Posts panel accordingly to the proposed UI.

## How has this been tested?


acicovic:
Thank you for working on this! I've taken a first look to the code here, and I don't see anything major that needs to be changed.

I'll quickly look into the failing integration tests. About the failing structure test, let's wait to have the final version of this.

acicovic:
Working on E2E tests as well.

acicovic:
E2E tests now pass. Should not be difficult to update them if we make any small adjustments like strings etc.

vaurdan:
@acicovic I just submitted a workaround for the animation issue which looks as good as the native animation, so I think we're good on that front :)

acicovic:
All tests should be passing now!

src/content-helper/editor-sidebar/related-posts/component.tsx
```diff
@@ -0,0 +1,413 @@
...
+import { RelatedPostsProvider } from './provider';
+import './related-posts.scss';
+
+const FETCH_RETRIES = 1;
+
+/**
+ * The Related Posts panel in the Editor Sidebar.
+ *
+ * @since 3.14.0
+ */
```
acicovic, line 42: This file contains a few functions from 3.11. If it's a rename of another component, let's indicate it in the `@since 3.14.0` comment.

src/content-helper/editor-sidebar/related-posts/related-posts.scss
```diff
@@ -0,0 +1,245 @@
+@import "../../common/css/variables";
```
acicovic, line 1: There are many units without `to_rem()` conversions in this file. Is this intentional?

acicovic (commented):
This looks and works great! I've left a couple of comments.

src/content-helper/editor-sidebar/related-posts/related-posts.scss
vaurdan: It's not! I forgot to give a final look at the related-posts.scss

acicovic:
Comments for our patch iteration:

Design (@dabowman):
1. Do we need to show the filter type when only one is available? I think we can do without it since it's referenced in the message, but no issue leaving it if you think it makes the design prettier. 🙂

    image

2. We're currently pushing action buttons towards the right, but if we can do it easily, should we pushing them towards the left? I think it could result in quicker access and less mouse movement, but this might be just me.

    image

Implementation (@vaurdan):
1. We're showing our "linked" icon when the URL is in the post, without it being an actual link. Worth thinking if we might want to show the linked icon only when the URL is actually in an `href`.
2. For sites that have `HTTPS`, we're currently copying the URLs with `http://` links. This is obviously related to how the API returns results. We could try tackling this in different ways, but if this proves too complicated we could just default to `https://` links.

src/content-helper/editor-sidebar/related-posts/related-posts.scss
acicovic: I'll take care of that soon.

acicovic: Seems I'll have to be absent for a while starting soon, so feel free to do this if you get to it.

src/content-helper/editor-sidebar/related-posts/component.tsx
vaurdan: It's not really a rename, but a new component that is consolidating a bunch of functions from different places.

acicovic: OK, feel free to ignore then. 🙂

src/content-helper/editor-sidebar/related-posts/related-posts.scss
```diff
@@ -0,0 +1,246 @@
...
+						outline-offset: -3px;
+
+						&[data-active-item] {
+							background: var(--sidebar-black, #1e1e1e);
+							transition:
+								background 0s 0.5s,
+								border-radius 0s 0.5s,
+								box-shadow 0s 0.5s;
+							box-shadow: 0 -3px 0 0 #fff, 0 3px 0 0 #fff;
+							border-radius: 3px;
```
acicovic, line 110: Missing `to_rem`s in 5 instances here, unless we don't want to do that for 2px or 3px.

src/content-helper/editor-sidebar/related-posts/related-posts.scss
```diff
@@ -0,0 +1,246 @@
...
+			}
+		}
+
+
+		.related-posts-list {
+			display: flex;
+			padding: 6px 0 var(--grid-unit-20, 16px) 0;
+			flex-direction: column;
+			align-items: flex-start;
+			gap: var(--grid-unit-20, 16px);
```
acicovic: Missing `to_rem`s in 3 instances here.

src/content-helper/editor-sidebar/related-posts/related-posts.scss
```diff
@@ -0,0 +1,246 @@
...
+				display: flex;
+				flex-direction: column;
+				align-items: flex-start;
+				align-self: stretch;
+				border-radius: 2px;
+				border: 1px solid var(--Gutenberg-Gray-400, #ccc);
+
+				.related-post-title {
+					display: flex;
+					padding: var(--grid-unit-20, 16px);
```
acicovic: Missing `to_rem`.

src/content-helper/editor-sidebar/related-posts/related-posts.scss
```diff
@@ -0,0 +1,246 @@
...
+					flex-direction: column;
+					justify-content: center;
+					align-items: center;
+					gap: to_rem(8px);
+					align-self: stretch;
+					overflow: hidden;
+					text-overflow: ellipsis;
+					font-size: 14px;
```
acicovic: Do we want to use `to_rem` here?

acicovic (approved):

src/content-helper/editor-sidebar/related-posts/related-posts.scss
vaurdan: I believe we cannot use rem units for box-shadow, outlines and border-radius

dabowman:
Yes, I think we're ok to hide the filter when only one option is available. Also yes, let's align buttons to the left. I think that's how it is in Figma.

acicovic:
The patch iteration comments in this thread are addressed.

PR #2286, opened 2024-03-12
Content Suggestions API: Truncate very large strings on the body params

vaurdan:
## Description
The Content Suggestions API only supports content up to 25.000 characters. This PR aims to truncate any string that is longer than 25.000 chars, so it fits the size allowed by the API.

While this means that not all the content would be sent to the Content Suggestions API, still having 25.000 characters worth of data should be sufficient to provide useful AI suggestions.

## Motivation and context
Make the internal API more resilient and aware of the limitations of the external Content Suggestions API. 

## How has this been tested?
Tested locally to validate that the content is being truncated and that very large posts can now be used with the PCH tooling, and also created a integration test for the `truncate_array_content` method.


acicovic:
Thanks for working on this!

Do you think there might be cases where we don't want to truncate the content?

src/RemoteAPI/content-suggestions/class-content-suggestions-base-api.php
```diff
@@ -142,4 +162,31 @@ protected function post_request( array $query = array(), array $body = array() )
 
 		return $decoded;
 	}
+
+	/**
+	 * Truncate the content of an array to a maximum length.
```
acicovic:
```suggestion
	 * Truncates the content of an array to a maximum length.
```

src/RemoteAPI/content-suggestions/class-content-suggestions-base-api.php
```diff
@@ -142,4 +162,31 @@ protected function post_request( array $query = array(), array $body = array() )
...
+	 *
+	 * @param string|array|mixed $content The array to truncate.
```
acicovic:
```suggestion
	 * @param string|array|mixed $content The content to truncate.
```
Should we change `array` another word, since it can be a string? Same for the line below.

src/RemoteAPI/content-suggestions/class-content-suggestions-base-api.php
```diff
@@ -142,4 +162,31 @@ protected function post_request( array $query = array(), array $body = array() )
...
+	 * @return string|array|mixed The truncated array.
+	 */
+	public function truncate_array_content( &$content ) {
```
acicovic: What's the thought behind passing this by reference here? Asking because I also see return statements in the function.

vaurdan: Yeah that's a fair question, I thought I would need to access it by reference initially, and forgot to remove the `&` afterwards. Removed :)

vaurdan:
> Do you think there might be cases where we don't want to truncate the content?

In the Content Suggestions API, I don't think so. However, if the need to bypass the truncation arises, we can set the `TRUNCATE_CONTENT` constant to `false` in the class, and disable it for that endpoint.

src/RemoteAPI/content-suggestions/class-content-suggestions-base-api.php
```diff
@@ -142,4 +162,31 @@ protected function post_request( array $query = array(), array $body = array() )
...
+	/**
+	 * Truncates the content of an array to a maximum length.
+	 *
+	 * @param string|array|mixed $content The content to truncate.
+	 * @return string|array|mixed The truncated content.
+	 */
+	public function truncate_array_content( $content ) {
+		if ( is_array( $content ) ) {
+			// If the content is an array, iterate over its elements.
+			foreach ( $content as &$value ) {
```
acicovic: I don't think we need to reference here either?

src/RemoteAPI/content-suggestions/class-content-suggestions-base-api.php
```diff
@@ -142,4 +162,31 @@ protected function post_request( array $query = array(), array $body = array() )
 
 		return $decoded;
 	}
+
+	/**
+	 * Truncates the content of an array to a maximum length.
+	 *
```
acicovic, line 168: Seems we're missing a `@since` here.

tests/Integration/RemoteAPI/content-suggestions/BaseContentSuggestionsAPITest.php
```diff
@@ -58,4 +58,60 @@ public function test_api_headers(): void {
 		self::assertArrayHasKey( 'X-APIKEY-SECRET', $headers );
 		self::assertEquals( 'my-secret', $headers['X-APIKEY-SECRET'] );
 	}
+
+	/**
+	 * Verifies that the truncate function is properly truncated long content on the body array.
+	 *
```
acicovic, line 64: Add a `@since` here, and check for coverage in case more functions need to be added as `@covers` or `@uses`.

vaurdan: Since only `truncate_array_content` is being called, that's the only requirement in `@covers`. I have validated this with the coverage command.

acicovic: Thank you!

tests/Integration/RemoteAPI/content-suggestions/BaseContentSuggestionsAPITest.php
```diff
@@ -58,4 +58,72 @@ public function test_api_headers(): void {
...
+	}
+
+	/**
+	 * Generate content with a specific length.
+	 *
+	 * @since 3.14.1
+	 *
+	 * @param int $length Length of the generated content.
+	 *
+	 * @return string The generated content
```
acicovic:
```suggestion
	 * @return string The generated content.
```

acicovic (approved):

PR #2287, opened 2024-03-12
PCH Related Posts: Use HTTPS URLs when the site is using HTTPS

vaurdan:
## Description
Currently, all the URLs on the Related Posts component were defaulting to an unsecure link. This PR changes this to force using HTTPS URLs when the site is setup to use HTTPS, i.e. `wp_is_using_https` returns true.

## Motivation and context
Make the Related Posts component more consistent with the site settings, and avoids copying unsecure links with the "Copy to Clipboard" feature.

## How has this been tested?
Tested locally and validated that all the features work as expected.


src/Endpoints/class-base-api-proxy.php
```diff
@@ -237,13 +237,18 @@ protected function extract_post_data( stdClass $item ): array {
...
 			$site_id = $this->parsely->get_site_id();
 			// phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.url_to_postid_url_to_postid
-			$post_id        = url_to_postid( $item->url ); // 0 if the post cannot be found.
-			$data['rawUrl'] = Parsely::get_url_with_itm_source( $item->url, null );
+			$post_id = url_to_postid( $item->url ); // 0 if the post cannot be found.
 
-			$data['dashUrl'] = Parsely::get_dash_url( $site_id, $item->url );
-			$data['id']      = Parsely::get_url_with_itm_source( $item->url, null ); // Unique.
+			$post_url = Parsely::get_url_with_itm_source( $item->url, null );
+			if ( wp_is_using_https() ) {
```
acicovic: It seems that these [HTTPS functions](https://developer.wordpress.org/reference/functions/wp_is_using_https/) were introduced in WP 5.7. Since the plugin supports 5.2 and greater (which might be wrong, but we need to look into that), there are risks of this crashing.

Maybe there's an easy way to check for this (e.g. check home URL protocol), or we could just copy/paste these functions into our own code for now.

vaurdan: Ah, good point! I added a custom fuction that uses `wp_is_using_https` if available, and fallbacks to checking the home and site url.

acicovic (approved):

PR #2290, opened 2024-03-14
PCH Related Posts: Check for links in the content by looking for hyperlinks instead

vaurdan:
## Description
Currently the Related Posts link detection searches for the link string in the content. While this should work in most of the cases, it would detect as linked if simply the URL was somewhere in the content, even without a hyperlink.

This PR changes that logic to search for a hyperlink - `<a>` - instead.

## Motivation and context
Make the Related Posts link detection more robust as less prone to false positives.

## How has this been tested?
Tested locally with both HTTP and HTTPS links, with and without hyperlinks.


src/content-helper/common/utils/functions.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Escapes special characters in a string for use in a regular expression.
+ *
+ * @since 3.14.0
+ * @since 3.14.1 Moved from `editor-sidebar/smart-linking/utils.ts` to `common/utils/functions.ts`.
+ *
+ * @param {string} string - The string to be escaped.
+ *
+ * @return {string} The escaped string.
```
acicovic:
```suggestion
 * @param { string } string The string to be escaped.
 *
 * @return { string } The escaped string.
```

acicovic (approved):

PR #2296, opened 2024-03-15
PCH Editor Sidebar: Fix incorrect AI opt-in message

acicovic:
## Description
An incorrect message was being displayed in the PCH Editor Sidebar for opting into AI features. This PR fixes this by displaying the correct message and also improves the notice's styling which has some display issues.

## Motivation and context
The messages we show should be relevant.

## How has this been tested?
- Manual local testing.
- Automated tests pass.

## Screenshots
| Before | After |
|--------|--------|
| image | image |


src/RemoteAPI/content-suggestions/class-content-suggestions-base-api.php
```diff
@@ -143,14 +143,24 @@ protected function post_request( array $query = array(), array $body = array() )
...
+			$error = json_decode( wp_remote_retrieve_body( $response ), true );
+
+			if ( ! is_array( $error ) ) {
+				return new WP_Error(
+					400,
+					__( 'Unable to decode upstream API error', 'wp-parsely' )
+				);
+			}
+
+			return new WP_Error( $error['error'], $error['detail'] );
```
acicovic, line 163: We're moving from relying on HTTP codes to relying on string API error codes, which is closer to the WordPress guidelines anyway.

src/content-helper/editor-sidebar/smart-linking/component.tsx
```diff
@@ -512,7 +512,7 @@ export const SmartLinkingPanel = ( {
 				</div>
 				{ error && (
 					<Notice status="info" isDismissible={ false } className="wp-parsely-content-helper-error">
-						{ error.message }
+						{ error.Message() }
```
acicovic, line 515: We always need the processed message in our components. Else, we will show the raw API error message which is what was happening before this fix.

src/content-helper/common/content-helper-error.tsx
```diff
@@ -87,23 +103,22 @@ export class ContentHelperError extends Error {
 				'The Parse.ly API did not respond in a timely manner. Please try again later.',
 				'wp-parsely'
 			);
-		} else if ( this.code === ContentHelperErrorCode.ParselySchemaValidationFailed ) {
+		} else if ( this.code === ContentHelperErrorCode.ParselySuggestionsApiSchemaError ) {
 			this.message = __(
 				'The Parse.ly API returned a validation error. Please try again later.',
```
vaurdan: Arthur has suggested that for this error code, it doesn't make sense to tell the users to retry later, since they will hit the same error. Perhaps we can change to something like "Please try again with different parameters"?

vaurdan (commented):
Thank you so much for working of this! I added a comment that's not directly about your changes, but since we're touching this component, I think it makes sense to update it now.

vaurdan:
@acicovic before merging this, can you check #2302 and test if the error notices still need the included CSS fixes? If so, could you move those to the `common.scss` file?

acicovic:
@vaurdan, will do, thanks 🙂

acicovic:
@vaurdan, most of the CSS is still relevant (only removed one line). I agree with you that we might need to improve and move it probably somewhere else. Since moving it though would need some additional testing, I prefer to leave as-is for now and tackle it in a different PR, since we want to release this improvement tomorrow.

That PR would be about errors and the Contact Us message which is considered as an error but isn't. I'd want to make sure there are no clashes happening between the message, the errors, and different parts of the PCH (Sidebar, Dashboard Widget).

Are you OK with that? If yes, let's merge this as-is for now.

vaurdan (approved):
Sounds great! Approved 🙂

PR #2301, opened 2024-03-18
PCH Title Suggestions: Update telemetry event names

vaurdan:
## Description
Updates Title Suggestions' telemetry event names, to make them consistent with the Tracks guidelines.

## Motivation and context
Improve telemetry of the WP Parse.ly plugin.

## How has this been tested?
Tested locally and validated that the events are being correctly sent.


acicovic (approved):
Thank you for working on keeping our Tracks events updated!

PR #2289, opened 2024-03-13
PCH Smart Linking: Remove Beta Badge

acicovic:
## Description
This PR started as a quick attempt to remove the Beta Badge from Smart Linking. However, it was resulting in strange alterations in the UI. After a troubleshooting session, the conclusion was that some styles that were being imported into `common.css` didn't work or didn't register. To get this out of the door, the current solution has been to move these imports to the files that actually use them.

## Motivation and context
Follow our new design guidelines.

## How has this been tested?
- Manually tested that the change is as wanted, and that all side effects have been eliminated.
- Automated tests pass.

## Screenshots
| Before | After |
|--------|--------|
| image | image |

PR #2302, opened 2024-03-18
PCH: Fix issue with common styles not being correctly bundled

vaurdan:
## Description
After the merge of #2289, an issue with the way the `common.scss` file is bundled was introduced. This prevented some critical common styles from being loaded, creating some unexpected UI bugs, such as the ones identified and addressed in #2296 (error notices with wrong paddings and margins).

## Motivation and context
Fix the current issues with PCH UI styles.

## How has this been tested?
Tested locally and gone through all the PCH components to make sure that they are correctly stylized. I also validated that the error messages have the correct margins. 

| Before | After |
|--------|--------|
|image|image|


acicovic (approved):
Thank you for working on this 👍

PR #2324, opened 2024-03-26
PCH Smart Linking: Show "No Block Selected" hint when needed

### LINKED ISSUE #2275: Redesign QA
Just a place to dump all the little Q/A issues I'm seeing.

vaurdan:
## Description
Adds a visual aid, with a hint, when the user tries to select "Selected block" in the Smart Linking UI, without any selected block in the editor. At the same time, the selected option moves back to "All blocks".

This PR also addresses some logic and visual issues with the selected option in the ToggleGroupControl component.

## Motivation and context
- Improve the user experience of the Smart Linking feature.
- Closes https://github.com/Parsely/wp-parsely/issues/2275.

## How has this been tested?
Tested on my local WordPress instance, with different scenarios.

https://github.com/Parsely/wp-parsely/assets/5804923/e6772155-b979-452f-ba1e-bf413867be28


acicovic:
@dabowman: In the video, I see that the hint position is per design guidelines. However, I think that having it there might hinder or prevent visibility (hint being outside of the viewport or at the bottom of the screen comes to mind).

The way the hint implementation was done in this PR (which is great BTW!), I can see the hint being below the `ToggleGroupControl` component. This is not only more visible, but also visually relevant to the operation taking place. So @dabowman, What do you think?

Also, @vaurdan, if we were to do this, would you see moving the hint up causing any glitches or issues?

src/content-helper/editor-sidebar/smart-linking/component-settings.tsx
```diff
@@ -88,8 +77,19 @@ export const SmartLinkingSettings = ( {
 		setMaxLinkWords,
 		setFullContent,
 		setAlreadyClicked,
+		setApplyTo,
 	} = useDispatch( SmartLinkingStore );
 
+	/**
+	 * The value to apply the smart links to.
```
acicovic:
```suggestion
	 * The area to apply the smart links to.
```

src/content-helper/editor-sidebar/smart-linking/store.ts
```diff
@@ -20,13 +20,23 @@ export type SmartLinkingSettingsProps = {
...
 };
 
+/**
+ * Enum for the applyTo setting.
+ *
+ * @since 3.14.3
+ */
+export enum ApplyToOptions {
+	ALL = 'all',
+	SELECTED = 'selected',
```
acicovic: Unless there's a specific reason here, I'd suggest using the casing we use for other TypeScript enums (`All`, `Selected`).

src/content-helper/editor-sidebar/smart-linking/component-settings.tsx
```diff
@@ -98,36 +98,87 @@ export const SmartLinkingSettings = ( {
...
+				moveButtonAndShowHint();
+			}
+		}
+
+		setFullContent( ApplyToOptions.ALL === applyToValue );
+	}, [ animationIsRunning, applyTo, applyToValue, disabled, selectedBlock, setApplyTo, setFullContent, setHint ] );
+
+	/**
+	 * Applies workaround to set the value of the ToggleGroupControl programmatically.
+	 * This is needed because the ToggleGroupControl doesn't update the value when the
```
acicovic, line 161:
```suggestion
	 *
	 * This is needed because the ToggleGroupControl doesn't update the value when the
```

src/content-helper/editor-sidebar/smart-linking/smart-linking.scss
```diff
@@ -161,6 +166,6 @@
 	}
 
 	.wp-parsely-smart-linking-hint {
-		margin: 0;
+		margin-top: var(--grid-unit-20, to_rem(16px));
```
acicovic:
```suggestion
		margin-top: var(--grid-unit-20);
```
Going forward, no need to add a fallback value for variables we're declaring ourselves.

src/content-helper/editor-sidebar/smart-linking/store.ts
```diff
@@ -81,12 +91,18 @@ interface SetWasAlreadyClickedAction {
 	wasAlreadyClicked: boolean;
 }
 
+interface SetApplyToAction {
```
acicovic, line 145: A comment explaining the interface would be nice 🙂

vaurdan:
> Also, @vaurdan, if we were to do this, would you see moving the hint up causing any glitches or issues?

Absolutely not. Should be a pretty trivial change.

src/content-helper/editor-sidebar/smart-linking/store.ts
vaurdan: The other interfaces in the store do not have any comments. But happy to add comments to this interface and the others if you think it's reasonable.

vaurdan:
@acicovic @dabowman I tested moving the hint to below the ToggleGroup, as suggested, and I think it looks better. What are your thoughts? 


https://github.com/Parsely/wp-parsely/assets/5804923/bd993fa7-0123-416a-82bd-4f4a76adef00

src/content-helper/editor-sidebar/smart-linking/store.ts
acicovic: Ah, this had escaped me. I'm always in favor of comments if they can give some clarity. You don't need to do anything at the moment, unless you want to.

acicovic:
> @acicovic @dabowman I tested moving the hint to below the ToggleGroup, as suggested, and I think it looks better. What are your thoughts?

Makes more sense to me personally.

Edit: I think it looks better as well.

acicovic (approved):
This looks good to me and approving this. @vaurdan, as always thank you for your work!

If we ever conclude that we need the hint below the button, we can revert that in another PR.

PR #2325, opened 2024-03-26
PCH Smart Linking: Prevent self-linking

vaurdan:
## Description
In some scenarios, the Smart Linking generation calls could return links for the same post for which it was generating content. Since this is not a desired behavior, this PR aims to address this issue by utilizing the newly added `url_exclusion_list` API parameter to exclude the current post permalink from the Smart Linking results.

## Motivation and context
Improve the reliability of the Smart Linking output.

## How has this been tested?
I have tested locally by faking that my post had a permalink that was usually returned as a Smart Link. After this change, that same link was never returned again for that post.


acicovic (approved):
I've updated `$url_exclusion_list` to be `mixed` per [get_param](https://developer.wordpress.org/reference/classes/wp_rest_request/get_param/)'s return type. This will also handle things gracefully even if a non-array gets passed.

Please take a look at that commit and verify you agree with the changes, I have not performed any testing.

src/content-helper/editor-sidebar/smart-linking/component.tsx
```diff
@@ -346,6 +357,19 @@ export const SmartLinkingPanel = ( {
 		occurrenceCounts: LinkOccurrenceCounts,
 		updatedBlocks: BlockUpdate[],
 	): void => {
+		// Check if any of the links being applied is a self-reference, and remove it if it is.
+		const strippedPermalink = postPermalink
+			.replace( /^https?:\/\//, '' ) // Remove HTTP(s)
+			.replace( /\/+$/, '' ); // Remove trailing slash
```
acicovic:
```suggestion
			.replace( /^https?:\/\//, '' ) // Remove HTTP(S).
			.replace( /\/+$/, '' ); // Remove trailing slash.
```

acicovic (approved):

PR #2326, opened 2024-03-26
Fix isSmall deprecation

acicovic:
## Description
`isSmall` (which we were using in certain component calls) has become deprecated and was causing build problems. In the beginning I tried to support `isSmall` and `size` (the suggested alternative) concurrently for WordPress backwards compatibility. However, this proved tricky and `size: "small"` didn't seem to have any effect at all, and trying `iconSize` didn't produce any results either.

In the end, I decided to implement a CSS solution which is simpler and should work in all WordPress versions.

## Motivation and context
Keep our codebase updated.

## How has this been tested?
Manual testing.

## Screenshots
| Before deprecation | Trying `size`, `iconSize` | Final CSS fix |
|--------|--------|--------|
| image | image | image |


vaurdan (approved):
Thanks for working on this! Approved.

PR #2397, opened 2024-04-18
EmptyCredentialsMessage: Fix empty error

### LINKED ISSUE #2136: Edge Case: Blank error in some PCH panels when no credentials set
## Describe the bug
In the PCH's Title Suggestions tool, when the `Generate Titles` button is pressed and an error occurs, the panel goes blank and no error/info message is displayed.

Update: This also affects the Excerpt Generator when tested similarly. In this case the panel's content will not vanish, but it will display an empty notice which is visible by the blue border on the left.

## To Reproduce
1. With credentials set in plugin settings, go to the `Title Suggestions` panel and click it to make it appear
2. In a new browser tab, remove the credentials
3. In the previous browser tab, press the `Generate Titles` button
4. The panel will become blank and we'll get a `forbidden` error in the browser console

## Expected behavior
We should get an information or error message, as we would for the other panels of the PCH.

acicovic:
## Description
In the very rare edge case that is described in https://github.com/Parsely/wp-parsely/issues/2136, the `EmptyCredentialsMessage` component would show no message. This PR adds a fallback message in the component, to make sure that the user is never greeted with an empty error.

## Motivation and context
- Improve UI/UX clarity, avoid user confusion.
- Fixes https://github.com/Parsely/wp-parsely/issues/2136.

## How has this been tested?
Manually tested, existing tests pass.

## Screenshots
| Before | After |
|--------|--------|
| image | image | 

## Summary by CodeRabbit

- **Enhancements**
	- Improved localization for error messages in the application.


vaurdan (approved):
Thank for working on this, looks great!

PR #2398, opened 2024-04-18
Fix removeEditorPanel() deprecation message in WordPress 6.5

acicovic:
## Description
In the PCH Excerpt Generator, we were using `dispatch( 'core/edit-post' ).removeEditorPanel( 'post-excerpt' )` which is now deprecated in favor of `dispatch( 'core/editor' ).removeEditorPanel( 'post-excerpt' )` on WordPress 6.5 and newer.

This PR allows WordPress 6.5 to use the new code, and keeps the legacy code for older WordPress versions. Unfortunately, a couple of `// ts-ignore` comments needed to be added to stop TypeScript from complaining and failing builds.

## Motivation and context
Resolve deprecation messages while keeping backward compatibility.

## How has this been tested?
- Existing tests pass.
- Manually tested that the correct code is being triggered depending on the WordPress version (tested with WordPress 6.5.2 and 6.4.4).

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved the removal process of the excerpt panel in the WordPress editor to handle deprecations in WordPress 6.5 effectively.


src/content-helper/excerpt-generator/excerpt-generator.tsx
```diff
@@ -44,8 +44,13 @@ const ExcerptGenerator = ( settings: never, name: string ) => {
 		render: ExcerptPanel,
 	} );
 
-	// Remove the excerpt panel by dispatching an action.
-	dispatch( 'core/edit-post' )?.removeEditorPanel( 'post-excerpt' );
+	/* Remove the excerpt panel by dispatching an action. */ // @ts-ignore
```
vaurdan, line 47: Just wondering why we need the extra `@ts-ignore` on this line, since it's a comment.

vaurdan (approved):
Thank you! Added a question, but still approved!

src/content-helper/excerpt-generator/excerpt-generator.tsx
acicovic: We had two TypeScript errors, the first one being in the `if` statement. `ts-ignore` influences the next line, so it has been added to cater for that. I could have added it in a new line but preferred this approach which was the only one that was working. Adding it in the end of a `//` comment didn't work, it needs to be its own comment.

Unfortunately, it's not possible to ignore code blocks using `ts-ignore`, so this was the best way I found to ignore the errors and make the code look OK.

vaurdan: Ahhhhh gotcha! I have dealt with that issue myself a few times as well. Thanks for explaining :)

PR #2428, opened 2024-04-24
PCH Editor Sidebar: Update isDismissible in Notices

acicovic:
## Description
In this PR, we are removing unneeded `isDismissible: true` code from our `Notice`s (this property is set to `true` by default). We're also making a couple of notices dismissable, to be aligned with what we've already done in Smart Linking.

## Motivation and context
UX coherence, remove unneeded code.

## How has this been tested?
Manually tested that code behaves as expected.

## Summary by CodeRabbit

- **Refactor**
	- Removed the `isDismissible` property from notification components across various modules, streamlining user notifications and alerts.


src/content-helper/editor-sidebar/smart-linking/component.tsx
```diff
@@ -676,7 +674,6 @@ export const SmartLinkingPanel = ( {
 				{ error && (
 					<Notice
 						status="info"
-						isDismissible={ true }
 						onRemove={ () => setError( null ) }
```
vaurdan, line 677: Since we no longer can dismiss this notice, we can also remove the onRemove attribute, as it will not be used. 
```suggestion
```

src/content-helper/editor-sidebar/smart-linking/component.tsx
```diff
@@ -686,7 +683,6 @@ export const SmartLinkingPanel = ( {
 				{ suggestedLinks !== null && (
 					<Notice
 						status="success"
-						isDismissible={ true }
 						onRemove={ () => setSuggestedLinks( null ) }
```
vaurdan, line 686: As in my other comment, this line can be removed.
```suggestion
```

vaurdan (commented):
Thank you! I added a couple comments, let me know if they make sense!

PR #2401, opened 2024-04-18
PCH: Enhance Request Handling in Provider Classes

vaurdan:
## Description
This PR refactors all provider classes to inherit from a newly introduced `BaseProvider`. This new base class equips each provider with an enhanced method for fetching data from the API that includes support for cancelling ongoing requests using the `AbortController`.

This update transitions provider methods from static to instance methods, which allows each provider to maintain its own map of `AbortController` instances, one per request. This architecture ensures that cancelling one request does not inadvertently cancel others. While not currently implemented, this setup facilitates future enhancements, such as assigning unique abort IDs for selectively cancelling requests in complex operations like smart linking across multiple paragraphs.

This architecture not only simplifies the management of request cancellations but also positions the codebase for more sophisticated request handling scenarios in the future.

## Motivation and Context
- Implements cancellable requests, thereby resolving issue #1928.
- Enhances the reliability of the codebase and reduces redundancy by utilizing inheritance for common provider functionalities.

## How has this been tested?
Each provider was tested individually through manual checks, end-to-end tests, and unit tests with Jest.

## Summary by CodeRabbit

- **New Features**
	- Introduced a base class for improved data fetching with request cancellation capabilities.
	- Enhanced smart linking and title suggestion functionalities by utilizing a singleton pattern for better resource management.

- **Bug Fixes**
	- Removed the `isDismissible` property from certain UI components to streamline user notifications.

- **Documentation**
	- Standardized JSDoc comments across various components for better consistency and readability.

- **Refactor**
	- Updated several components to use singleton access methods, aligning with modern best practices.

- **Chores**
	- Updated dependencies to improve stability and performance of testing frameworks.

PR #2430, opened 2024-04-24
PCH Smart Linking: Fix removal of existing Smart Links on subsequent runs 

### LINKED ISSUE #2422: Smart Linking: Don't auto-remove existing links
When we use Smart Linking, the existing links of the post or selected block get removed.

I'd assume that in most cases, we do want to keep existing links added by the user, and only regenerate the links injected by Smart Linking. However, the user might wish to keep some of the links injected by Smart Linking, so giving the user granular control seems to make the most sense here.

vaurdan:
## Description
It was reported that when running the Smart Linking a second time in the same block would remove the existing Smart Links from that block and insert the new ones. This seems to be happening since at least WP 6.5.3.

This PR changes the how we were obtaining the block content - using `originalContent` - and now it uses the more appropriate `getBlockContent` function to get the most up-to-date content for that given block.

## Motivation and context
Improve the reliability of the Parse.ly plugin and the Smart Linking feature in the latest WordPress release. 

## How has this been tested?
Tested locally by generating Smart Links on the same block multiple times, none of these tests resulted in the removal of any existing smart links.

## Summary by CodeRabbit

- **Enhancements**
	- Improved the smart linking feature within the editor sidebar, enhancing link generation and application based on content blocks and attributes.


vaurdan:
This is related and closes #2422

src/content-helper/editor-sidebar/smart-linking/component.tsx
```diff
@@ -445,78 +445,80 @@ export const SmartLinkingPanel = ( {
...
 									}
+
+									// Increment the linked count only when a link is applied.
+									occurrenceCount.linked++;
+
+									// Flag the block as updated.
+									blockUpdated = true;
 								}
 							}
-						} );
```
vaurdan: All of these changes are basically removing one level of indentation, since it's no longer inside a `if` condition.

src/content-helper/editor-sidebar/smart-linking/component.tsx
```diff
@@ -445,78 +445,80 @@ export const SmartLinkingPanel = ( {
 				return;
 			}
 
-			if ( block.originalContent ) {
-				const blockContent: string = block.originalContent;
-				const doc = new DOMParser().parseFromString( blockContent, 'text/html' );
+			if ( ! block.originalContent ) {
+				return;
+			}
```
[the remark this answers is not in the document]
vaurdan: There might be some cases where the block doesn't have `originalContent` prop (for example a Group block). The getBlockContent might still return a value, since it always returns a string, hence this validation being required. 

@coderabbitai resolve

src/content-helper/editor-sidebar/smart-linking/component.tsx
acicovic: Thanks for flagging this.

acicovic:
Some questions/thoughts:

1. Would it be easy to implement this solution for a 3.14.4 patch release (base 3.14.3), or does it make things complicated because of other changes?
2. It is mentioned here that this solution keeps previous smart links. I assume this means that it also keeps user-defined links intact.
3. Whether this gets to be a part of a 3.14.4 patch release or 3.15, we should make sure it works in previous WordPress versions. Testing this with WordPress Playground should be easy, and this should work from at least WP 6.3 and newer. But let's decide first if it will be part of a patch or minor release to avoid more duplicate work.

vaurdan:
Thank you for your questions!

> Would it be easy to implement this solution for a 3.14.4 patch release (base 3.14.3), or does it make things complicated because of other changes?

That's a good question. If we are going to push this to 3.14.3 base, it might have conflict with a few recent PRs, namely the #2401 (provider refactoring) and API retries (#2386). Although it should be relatively easy to fix manually. 

> It is mentioned here that this solution keeps previous smart links. I assume this means that it also keeps user-defined links intact.

Yes, that is correct! The user-defined links were never a problem, since they were included in the block's `originalContent` prop already. So these will (still) be preserved. 

> Whether this gets to be a part of a 3.14.4 patch release or 3.15, we should make sure it works in previous WordPress versions. Testing this with WordPress Playground should be easy, and this should work from at least WP 6.3 and newer. But let's decide first if it will be part of a patch or minor release to avoid more duplicate work.

Yeah, agreed. I think it should, since this function - `getBlockContent` - is three years old. But I will spin up a test env and give it a try. I will update here with the results.

vaurdan:
> Yeah, agreed. I think it should, since this function - getBlockContent - is three years old. But I will spin up a test env and give it a try. I will update here with the results.

Update: it works as expected in 6.3, validated with a wp-now environment 🙂

acicovic:
> If we are going to push this to 3.14.3 base, it might have conflict with a few recent PRs, namely the https://github.com/Parsely/wp-parsely/pull/2401 (provider refactoring) and API retries (https://github.com/Parsely/wp-parsely/pull/2386). Although it should be relatively easy to fix manually.

I think the way I'd approach this is to just branch from 3.14.3 and add the updated function/code there, with any needed adaptations. Attempting to merge the current work could bring a lot of code in, and more points of failure which we don't want in a patch release. I hope I'm making sense. What do you think?

vaurdan:
> I think the way I'd approach this is to just branch from 3.14.3 and add the updated function/code there, with any needed adaptations. Attempting to merge the current work could bring a lot of code in, and more points of failure which we don't want in a patch release. I hope I'm making sense. What do you think?

Sounds good to me, yes!

acicovic:
Hello, I see you re-requested a review. I think it might make more sense to make create and merge the 3.14.4 PR first, just in case this PR becomes redundant. No issue approving though 🙂

acicovic:
Looking into this because I see it having conflicts but I wouldn't expect it to.

acicovic:
@vaurdan, after merging `develop` into this, there seems to be only one whitespace change. I propose closing this PR and integrating the whitespace change in any Smart Linking work you're doing, or just ignore the change.

vaurdan:
Sounds good to me! Closing this one.

PR #2506, opened 2024-05-21
PCH Excerpt Suggestions: Allow editing when reviewing

acicovic:
## Description
This PR introduces code to allow editing the excerpt in the PCH Excerpt Suggestions tool while in review mode.

As part of this PR, we have grouped some excerpt state variables in an `excerptData` object. We're also handling updates to the textbox that come from keyboard shortcuts and aren't caught by the `onChange` event.

## Motivation and context
Make Excerpt Suggestions usage more intuitive.

## How has this been tested?
Manually tested.

## Summary by CodeRabbit

- **New Features**
  - Enhanced UI interaction for word count and excerpt changes.

- **Refactor**
  - Improved state management for excerpt data.
  - Refactored word count calculation and display logic.

- **Style**
  - Removed `textarea:read-only` styling from the excerpt textarea.


src/content-helper/excerpt-generator/components/excerpt-panel.tsx
```diff
@@ -144,10 +189,24 @@ const PostExcerptGenerator = () => {
...
+						if ( onChangeFired ) {
+							setOnChangeFired( false );
+							return;
+						}
+
+						const textarea = document.querySelector( '.editor-post-excerpt textarea' );
+						const value = textarea?.textContent ?? '';
+
+						setExcerptData( { ...excerptData, currentExcerpt: value } );
+					} }
```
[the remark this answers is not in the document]
acicovic: Isn't this handled by setOnChangeFired?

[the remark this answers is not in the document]
acicovic: @coderabbitai resolve

src/content-helper/excerpt-generator/components/excerpt-panel.tsx
```diff
@@ -61,24 +79,42 @@ const PostExcerptGenerator = () => {
...
+					_n( '%1$s word', '%1$s words', wordCount, 'wp-parsely' ),
+					wordCount
+				);
+			}
+
+			return '';
+		};
+
+		setWordCountString( getWordCountString() );
+	}, [ excerptData.currentExcerpt, excerpt ] );
```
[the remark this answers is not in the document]
vaurdan: Although this is a pretty valid point, I don't think it would be an issue in this case, since it's restricted to the excerpt textarea, and adding a debounce would prevent the word count to update in real-time - unless the debounce interval would be set to very low, but that kinda defeats the purpose. 

I think we can safely ignore this suggestion. What do you think @acicovic?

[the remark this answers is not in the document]
acicovic: Can you explain what this code will do?

src/content-helper/excerpt-generator/components/excerpt-panel.tsx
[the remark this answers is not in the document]
vaurdan: Unless I'm missing something don't see how this is any different from the current approach.

src/content-helper/excerpt-generator/components/excerpt-panel.tsx
```diff
@@ -144,10 +189,24 @@ const PostExcerptGenerator = () => {
 					__nextHasNoMarginBottom
 					label={ __( 'Write an excerpt (optional)', 'wp-parsely' ) }
 					className="editor-post-excerpt__textarea"
-					onChange={ ( value ) => editPost( { excerpt: value } ) }
-					readOnly={ isLoading || hasGeneratedExcerpt }
+					onChange={ ( value ) => {
+						setOnChangeFired( true );
+						editPost( { excerpt: value } );
+						setExcerptData( { ...excerptData, currentExcerpt: value } );
```
vaurdan, line 196: If we change the generated excerpt and not approve it, nor discard it, will it not still be saved to the post? I think the `editPost` function should only be called when `excerptData.newExcerptGenerated` is false/empty.

src/content-helper/excerpt-generator/components/excerpt-panel.tsx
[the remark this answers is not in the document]
acicovic: Oh, we posted at the same time 🙂

That was what I was pondering about. This might make a slight difference in some cases, but I think it's marginal, unless we set that to 1 second or something. Let's ignore this suggestion.

@coderabbitai resolve

vaurdan (commented):
Thanks for working on this one! Looks great, I just have added a single comment, and once it is addressed I can go ahead and approve it.

src/content-helper/excerpt-generator/components/excerpt-panel.tsx
acicovic: Do you think that this saving would take place due to `editPost()`? In my testing, the only way to persist the excerpt is to press the post's `Update` button. `editPost()` doesn't seem to have a permanent effect.

vaurdan: Ah yeah, that sounds right! This would only happen if the user wouldn't reject nor accept the excerpt, but would still save the post. So maybe not really an issue, I will leave that to your discretion 🙁 

Edit: Although, not sure what would happen when the post autosaves.

vaurdan (approved):
Approved per my last reply.

src/content-helper/excerpt-generator/components/excerpt-panel.tsx
acicovic: > Edit: Although, not sure what would happen when the post autosaves.

Interesting observation. I'll look into it.

acicovic: Testing with a draft post, I can confirm that the textarea's temporary value (shown but not dismissed or accepted) isn't persisted when auto-saving takes place.

Given this, I think we're safe.

acicovic: There's an exception. If the user has typed something manually in that state, changes persist.

acicovic: This seems to be standard WP behavior though. When we type something manually, it gets persisted.

acicovic: So there's a potential for unexpected behavior. I'll leave this open for the time being.

acicovic: This should be fixed now.

acicovic:
@coderabbitai resolve

PR #2508, opened 2024-05-22
PCH Smart Linking: Remove Link Length option from the UI

vaurdan:
## Description
The Link Length option is actually not working in the upstream API, and therefore, it can be removed from the configurable settings in the UI.

This PR removes the option from the UI, and all the associated functions and parameters from the store. 

## Motivation and context
Declutter the UI from having options that do not affect the output.

## How has this been tested?
Tested locally, and validated that this feature is still working as expected, without this option present. 

image


acicovic (approved):
Thanks for working on this, looks good to me 🙂

PR #2510, opened 2024-05-30
Inject post ID into LD-JSON metadata

sanmai:
## Description
Injecting post IDs will let Parsely link back to the edit form of the post.

## Summary by CodeRabbit

- **New Features**
  - Added post ID field to metadata for better identification and tracking of posts.

- **Bug Fixes**
  - Ensured post IDs are correctly populated in metadata.

- **Tests**
  - Updated integration tests to verify the inclusion of post ID in the metadata.

### DIFF
--- a/src/Metadata/class-post-builder.php
+++ b/src/Metadata/class-post-builder.php
@@ -60,6 +60,7 @@ class Post_Builder extends Metadata_Builder {
 		$this->build_publisher();
 		$this->build_keywords( $this->post );
 		$this->build_metadata_post_times( $this->post );
+		$this->build_post_ids( $this->post );
 
 		return $this->metadata;
 	}
--- a/src/Metadata/class-metadata-builder.php
+++ b/src/Metadata/class-metadata-builder.php
@@ -688,4 +688,15 @@ abstract class Metadata_Builder {
 
 		return $all_values;
 	}
+
+	/**
+	 * Populates all the fields related to post ID.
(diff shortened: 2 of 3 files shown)


src/Metadata/class-post-builder.php
```diff
@@ -60,6 +60,7 @@ public function get_metadata(): array {
 		$this->build_publisher();
 		$this->build_keywords( $this->post );
 		$this->build_metadata_post_times( $this->post );
+		$this->build_post_ids( $this->post );
```
rebeccahum, line 63: Is it possible to pass in the post ID directly?

sanmai:
Closing in favor of https://github.com/Parsely/wp-parsely/pull/2511

src/Metadata/class-post-builder.php
sanmai: Thanks for the suggestion! I get where you're coming from, but I think it'll be better to stick with the current pattern of passing the `WP_Post` object around. Here's why:

- All our existing methods use WP_Post, so changing to post ID would break that pattern.
- Using WP_Post keeps all the post-related data and logic neatly contained. 
- If we ever need to pass related post IDs or additional post data, we'd add more methods, increasing the risk of breaking changes.

So it sounds like maintaining the current approach of working with `WP_Post` objects makes the codebase cleaner and easier to maintain. What do you think?

PR #2511, opened 2024-05-30
Include post ID in custom metadata

sanmai:
## Description

Let's add a post ID to the custom metadata. 

https://docs.parse.ly/api-custom-metadata/

CC @medeiros-amanda @rebeccahum @robersongomes 

## Motivation and context

Injecting post IDs will let Parsely link back to the edit form of the post.

As compared to LD-JSON (#2510), we can expect that there are no SEO implications.

## How has this been tested?

LD-JSON works as usual.

## Summary by CodeRabbit

- **New Features**
  - Added a custom metadata field for the current post's ID, enhancing metadata handling.

- **Bug Fixes**
  - Improved metadata accuracy by including the new custom metadata field in the output.

- **Tests**
  - Updated tests to validate the inclusion of the new custom metadata in JSON-LD and repeated meta formats.


src/Metadata/class-metadata-builder.php
```diff
@@ -120,6 +120,21 @@ protected function build_type( WP_Post $post, string $parsely_type ): void {
 		$this->metadata['@type'] = $type;
 	}
 
+	/**
+	 * Populates all the fields related to post ID.
+	 *
+	 * @param WP_Post $post The post/page for which to populate the field.
+	 *
+	 * @since 3.0.x
+	 */
```
acicovic:
```suggestion
	/**
	 * Populates the custom metadata field with the current post's/page's ID.
	 *
	 * @since 3.16.0
	 *
	 * @param WP_Post $post The post/page for which to populate the field.
	 */
```

tests/Integration/Endpoints/RestMetadataTest.php
```diff
@@ -404,7 +404,8 @@ public function test_get_rendered_meta_json_ld(): void {
...
+		$expected    = '<script type="application/ld+json">{"@context":"https:\/\/schema.org","@type":"NewsArticle","headline":"My test_get_rendered_meta_json_ld title","url":"http:\/\/example.org\/?p=' . $post_id . '","mainEntityOfPage":{"@type":"WebPage","@id":"http:\/\/example.org\/?p=' . $post_id . '"},"thumbnailUrl":"","image":{"@type":"ImageObject","url":""},"articleSection":"Uncategorized","author":[],"creator":[],"publisher":{"@type":"Organization","name":"Test Blog","logo":""},"keywords":[],"dateCreated":"' . $date . '","datePublished":"' . $date . '","dateModified":"' . $date . '","custom_metadata":"{\\"postID\\":' . $post_id . '}"}</script>' .
```
acicovic: Do we need the `\\` in `{\\"postID\\"`?

acicovic (commented):
Thanks for working on this! Left a suggestion and a question 🙂

tests/Integration/Endpoints/RestMetadataTest.php
sanmai: Most definitely, even the docs call for it:

> Escape double quotes in JSON item values. Double quotes should be escaped with a backslash symbol like this: `\"`.

And then they provide clearly broken JSON in the example 🤷 

https://docs.parse.ly/api-custom-metadata/#h-json-ld-configuration

acicovic:
I'll be closing this as stale, as well as something we can probably achieve using another approach. We can reopen if we choose to finally use this in the future.

PR #2537, opened 2024-06-04
Revert to using babel-preset-default v7

acicovic:
We're reverting to the previously used version of `babel-preset-default` because of certain issues. See https://github.com/WordPress/gutenberg/pull/62265/files for related information.

Effectively, we cannot use `@wordpress/babel-preset-default` v8 or greater and `@wordpress/scrips` v28 or greater until we require WordPress 6.6 as the minimum version in the plugin.

This PR reverts Parsely/wp-parsely#2532 and adds a couple of related comments in `package.json` for future reference.

## Summary by CodeRabbit

- **Chores**
  - Updated the version of `@wordpress/babel-preset-default` to `^7.42.0`.
  - Added comments to clarify version constraints for `@wordpress/babel-preset-default` and `@wordpress/scripts` until WordPress 6.6 is required.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,7 +27,7 @@
 				"@types/wordpress__wordcount": "^2.4.5",
 				"@typescript-eslint/eslint-plugin": "^6.21.0",
 				"@wordpress/api-fetch": "^7.0.0",
-				"@wordpress/babel-preset-default": "^8.0.0",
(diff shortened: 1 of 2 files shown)


vaurdan (approved):
Thanks! And thank you for adding the dependencies comments, they will be very helpful down the line.

PR #2570, opened 2024-06-17
PCH Smart Linking: Prevent Smart Linking from running on Freeform (Classic) blocks

vaurdan:
## Description
Due to the way the Freeform blocks (aka Classic Editor blocks) works, the parsing and injection of Smart Links do not work properly on this type of block.

This PR checks if the current block (if "Selected Block" is selected) or if the content contains any freeform block (if "All Blocks" is selected), and throws a Snack Bar error if so.

## Motivation and context
Prevent issues when running the Smart Linking on posts with Freeform blocks.

## How has this been tested?
Tested locally

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved reliability of Smart Links by handling cases where they don't work with Freeform blocks.
	- Added error messages for better user feedback when Smart Links cannot be applied.


acicovic:
So, the existence of a single Freeform block can break things for other blocks as well, even if the selected block is not Freeform. Am I understanding this correctly?

vaurdan:
@acicovic exactly. If you are generating smart links for all blocks, and you get a smart link that is to be inserted into a freeform block, there's a high likelihood that either the block editor crashes, or the block content becomes empty. If there is at least one freeform block, only the "Selected Block" option will actually work, and only if you don't select the freeform block

acicovic (approved):

PR #2578, opened 2024-06-26
PCH Sidebar: Move initialization to admin_init

acicovic:
## Description
With this PR, we're moving the initialization of the PCH Sidebar from the `init` to the `admin_init` hook.

## Motivation and context
I think this makes the code more readable (less scatter across different places). In addition, firing this in `admin_init` should make the code execute fewer times. Since we're only using the PCH Sidebar in the admin, this seems a safe change to me.

However @vaurdan, I wanted your confirmation that this looks good to you, just in case I've missed anything and there was a reason behind this initial implementation.

## How has this been tested?
Manual testing and existing tests pass.

## Summary by CodeRabbit

- **Refactor**
  - Improved the initialization process for the editor sidebar, ensuring better performance and reliability.


vaurdan (approved):
This sounds very reasonable! Thanks for working on this.

PR #2579, opened 2024-06-26
PCH Sidebar: Show detailed error messages in Snackbar Notices

### LINKED ISSUE #2424: PCH: Consider forwarding main UI errors to Snackbar Notices
When users attempt to use AI features but they aren't enrolled into the beta program, we display this message within our Sidebar:

image

At the same time, there's a Snackbar notice appearing in the bottom left of the Post Editor:

image

It came to our attention that some users may focus on the Snackbar and wonder about the error, when the reason is stated in the Sidebar. This has been reported once, and we want to avoid this situation as interest in AI features may keep on increasing in the future.

See if we can forward the specific message (or even all error messages) to our Snackbars. This is up to discussion if there are any opinions.

Removing the Snackbar altogether, but we probably don't need to be that radical 🙂 (cc @dabowman, @danielabloch).

acicovic:
## Description
With this PR, we start displaying detailed error messages in our Snackbar Notices in the Title Suggestions and Smart Linking features.

## Motivation and context
- When errors would occur in Title Suggestions and Smart Linking, we were showing a generic `"There was an error"` Snackbar Notice. While we were displaying the detailed error message in the PCH Sidebar, the Snackbar Notice proved to be unhelpful or even confusing to users, with the error within the Sidebar remaining unnoticed in some cases. This change should neutralize this problem.
- Fixes #2424. This was a known issue, and I stumbled upon it while working on other things. While fixing it, I decided to extract it to its own PR for easier review and separation of concerns.

## How has this been tested?
Manual testing.

## Summary by CodeRabbit

- **New Features**
  - Added a new error type `UnknownError` to improve error handling.
  - Dynamic error messages are now displayed based on the actual error encountered.

- **Improvements**
  - Reordered import statements for better code organization.


src/content-helper/editor-sidebar/smart-linking/component.tsx
```diff
@@ -470,7 +470,11 @@ export const SmartLinkingPanel = ( {
 			await processSmartLinks( generatedLinks );
 			setIsReviewModalOpen( true );
 		} catch ( e: any ) { // eslint-disable-line @typescript-eslint/no-explicit-any
-			let snackBarMessage = __( 'There was a problem generating smart links.', 'wp-parsely' );
+			const contentHelperError = new ContentHelperError(
```
acicovic, line 473: Since the error can be any type of object theoretically, we add some special handing to make sure it translates to a `ContentHelperError`.

src/content-helper/editor-sidebar/smart-linking/component.tsx
```diff
@@ -483,7 +487,6 @@ export const SmartLinkingPanel = ( {
 				e.message = snackBarMessage;
 			}
 
-			console.error( e ); // eslint-disable-line no-console
```
acicovic: I found this `console.error( e )` to be quite noisy because it often outputs the same message 3 times due to retries. In addition, we're now showing the actual error message in the snackbar, so I think we're OK with removing this.

Having said that, I'm OK with bringing this back if we think it's better to have.

src/content-helper/editor-sidebar/title-suggestions/component.tsx
```diff
@@ -198,7 +198,7 @@ export const TitleSuggestionsPanel = (): React.JSX.Element => {
 
 		createNotice(
 			'error',
-			__( 'There was an error generating title suggestions.', 'wp-parsely' ),
+			error.message,
```
acicovic, line 201: Here we know that the error is a `ContentHelperError` object, so no special handling needed.

src/content-helper/editor-sidebar/smart-linking/component.tsx
vaurdan: Perhaps we should also remove the one here:

https://github.com/Parsely/wp-parsely/blob/f40cd5fbfacf83aa7549b840a9fbc28d69485dfa/src/content-helper/editor-sidebar/smart-linking/component.tsx#L541-L545

I actually think that the `console.error` above is the one throwing the same errors on each retry. In theory, this one that you removed should only be called if there's an error after all the retries attempts. 

I will leave to your discretion if we want to remove both, or keep this one and remove the other instance.

vaurdan (commented):
Thank you for working on this! I left a comment regarding the `console.error` removal, but other than that, I consider this PR approved.

src/content-helper/editor-sidebar/smart-linking/component.tsx
acicovic: I'm actually restoring the `console.error()`, it will make things simpler for now. If we get noise again, we can look into it.

PR #2604, opened 2024-07-03
Content Helper: Implement AI feature permissions

acicovic:
## Description
With this PR, we're implementing Content Helper AI feature permissions in the back-end and front-end, as the continuation of the work introduced in #2554.

The default settings will be different based on whether the plugin installation is new or existing. For new installations, PCH AI features will be auto-enabled solely for users with the `administrator` User Role. For existing installations, the features will be auto-enabled for all User Roles who have the `edit_posts` capability, as done for all plugin versions before 3.16.0. This is done to prevent unexpected behavior changes while upgrading to 3.16.0. In both cases, administrators will be then able to adjust the related settings from the plugin's Settings page, under the Content Helper tab.

After the upgrade to 3.16.0, any newly added User Roles will need to be manually allowed access to the PCH AI features. This means that we're switching from an "on by default" to an "off by default" approach.

## Motivation and context
Allow website administrator to control which Content Helper AI features are available to users, based on their User Role.

## How has this been tested?
- Manual testing
- Updated existing tests that were failing after the changes.

## Summary by CodeRabbit

- **New Features**
  - Introduced user permission checks for accessing specific features in the Content Helper.
  - Added error handling for access-related issues in the Content Helper.

- **Enhancements**
  - Improved permission management in various components, including the editor sidebar and smart linking functionalities.
  - Updated default user roles to streamline feature access control.

- **Bug Fixes**
  - Fixed incorrect permission checks to ensure users only access features they are authorized to use.

- **Tests**
  - Added comprehensive tests to validate new permission checks and transformations.


acicovic:
The SonarCloud failure is a code duplication warning, which can be ignored.

src/content-helper/common/utils/permissions.ts
```diff
@@ -0,0 +1,27 @@
...
+export function getContentHelperPermissions(): ContentHelperPermissions {
+	if ( ! window.wpParselyContentHelperPermissions ) {
+		return {
+			SmartLinking: false,
+			TitleSuggestions: false,
+		};
+	}
+
+	return JSON.parse( window.wpParselyContentHelperPermissions );
+}
```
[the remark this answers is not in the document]
acicovic: This should be improved now. Can you check again?

src/class-permissions.php
```diff
@@ -39,4 +42,127 @@ public static function get_user_roles_with_edit_posts_cap(): array {
...
+			$permissions[ $key ] = self::current_user_can_use_pch_feature(
+				$value,
+				$pch_options,
+				get_the_ID()
+			);
+		}
+
+		$result = wp_json_encode( $permissions );
+
+		return is_string( $result ) ? $result : '';
```
vaurdan: Since this is a JSON encoded string, I would suggest the fallback value be `'{}'` instead of `''`

src/content-helper/editor-sidebar/tabs/sidebar-tools-tab.tsx
```diff
@@ -31,49 +33,54 @@ type SidebarToolsTabProps = {
...
-					trackToggle( 'smart_linking', next );
-				} }
-			>
-				<VerifyCredentials>
-					<SmartLinkingPanel
-						context={ SmartLinkingPanelContext.ContentHelperSidebar }
-					/>
-				</VerifyCredentials>
-			</PanelBody>
+			{ permissions.SmartLinking &&
```
vaurdan, line 62: This is a suggestion for the future, not really necessary for now, but instead of doing these conditional checks, we can likely create a `<ValidatePermissions>` component, and have it validate the correct permissions for each panel.

vaurdan (approved):
Looks good to me! I have approved the PR, and left a couple notes/suggestions.

src/content-helper/editor-sidebar/tabs/sidebar-tools-tab.tsx
acicovic: Yes, I think this is a good idea, and in line with what we're doing for credential validation. 🙂

src/class-permissions.php
acicovic: Thanks, good catch! Updated.

PR #2606, opened 2024-07-05
PCH: Initialize Content Helper features on 'rest_api_init'

vaurdan:
## Description
This PR is a follow up of #2578. The Content Helper features also need to be initialized on `rest_api_init`, otherwise the custom post type and taxonomies for the Smart Linking feature will not be available in the REST API context.

## Motivation and context
Fixes a bug where the Smart Linking `/get` endpoint does not return any smart links because the `smart_link_source` taxonomy does not exists.

## How has this been tested?
Tested locally, and validated that the Smart Linking endpoint returns the expected results.

## Summary by CodeRabbit

- **New Features**
  - Integrated new functionality to enhance the REST API with the `rest_api_init` action, improving content management capabilities in the editor sidebar.


acicovic:
Thanks for fixing this. Since we're now calling it in `rest_api_init`, do we also need the call in`admin_init`?

vaurdan:
@acicovic yeah, I believe so, so that the initialization also happens on wp-admin and not only just on the REST API.

acicovic (approved):

PR #2609, opened 2024-07-05
PCH Excerpt Suggestions: Improve error handling

vaurdan:
## Description
This PR improves the error handling of the Excerpt Suggestions feature, and updates it to extend the provider from the BaseProvider.
The Generate button will also be disabled when the post content is empty. On the other hand, if the title is empty, since it's still possible to generate an excerpt, it will use the 'Untitled' as a title placeholder.

As a note, the CS API returns the error message as an array if there is an validation error, for example:

```json
{
    "code": "SCHEMA_ERROR",
    "message": [
        {
            "type": "string_too_short",
            "loc": [
                "body",
                "text"
            ],
            "msg": "String should have at least 1 character",
            "input": "",
            "ctx": {
                "min_length": 1
            }
        }
    ],
    "data": null
}
```

The BaseProvider will test if the message is an object, and if so, it will use the first entry of the `message` object as the one to create the ContentHelperError object.

## Motivation and context
Improve the reliability of the Excerpt Suggestions feature, preventing the block editor from crashing on unexpected errors.

## How has this been tested?
Tested locally with an empty post, and a regular post.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling to better process complex error messages in the `BaseProvider` class.
  - Enhanced error handling in the `ExcerptPanel` component, specifically targeting `ContentHelperError`.

- **New Features**
  - Introduced a `getInstance()` method for singleton instance retrieval in the `ExcerptGeneratorProvider` class.
  - Updated the `ExcerptGeneratorProvider` to handle empty titles by setting them to 'Untitled'.


src/content-helper/excerpt-generator/provider.ts
```diff
@@ -35,28 +45,18 @@ export class ExcerptGeneratorProvider {
...
-			return Promise.reject( new ContentHelperError( wpError.message, wpError.code ) );
-		}
-
-		if ( response?.error ) {
-			return Promise.reject( new ContentHelperError(
-				response.error.message,
-				ContentHelperErrorCode.ParselyApiResponseContainsError
-			) );
+		if ( '' === title ) {
+			title = 'Untitled';
```
acicovic, line 49: Should we make that string translatable?

acicovic (approved):
This is excellent! 🙂

src/content-helper/excerpt-generator/provider.ts
vaurdan: I don't think it's necessary, since this string will be sent to the LLM and will not be present to the user.

acicovic: OK, feel free to merge as-is. I wonder if this might be of any significance if we start supporting different languages.

vaurdan: Yeah, might be! However I think the content language would have more impact than the actual WordPress language, eg, if I have my WP install in French, but the content is in English, it wouldn't make sense to send the translated string.

I don't think it's a trivial problem to solve, and in this specific scenario I don't think the gains are actually worth the trouble.

acicovic: Yes, totally agree. Thank you for your work! 🙂

PR #2610, opened 2024-07-05
PCH: Fix potential fatal with 'enqueue_settings_assets' strict parameter

vaurdan:
## Description

Fix a potential fatal from happening, under some very specific scenarios, where the `enqueue_settings_assets` throws a fatal because it's expecting a string, and got a `null` instead.

## Motivation and context
Improve the reliability of the WP Parse.ly codebase.

## How has this been tested?
Tested locally, and validated that the styles and scripts are still being correctly enqueued.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved stability of the settings page by handling potential null values in settings assets.


acicovic (approved):
Looks good to me.

PR #2621, opened 2024-07-09
Use Composer autoloader

### LINKED ISSUE #961: Use an autoloader for the plugin's files

vaurdan:
## Description
This PR implements the Composer autoload into the wp-parsely plugin, to avoid manually requiring all the necessary classes. 

A few notes:
 * The tests have been reorganized to comply with the PSR-4 specification. 
 * The main codebase is not PSR-4 compliant, but it is handled by `classmap` ([see Composer documentation here](https://getcomposer.org/doc/04-schema.md#classmap)). Since this requires an authoritive classmap to be generated each time we add a new class, I have created a `gen-autoload` composer command to take care of it.
 * Utils have been moved to its own class with static methods, so it can be autoloaded instead of manually requiring it.
 
## Motivation and context
Improve the quality, readability and maintenance of the plugin's codebase

## How has this been tested?
Tested locally, and validated that the unit tests are working; and also validated that there are no issues on the plugin inside WordPress.

## Summary by CodeRabbit

- **New Features**
	- Enhanced deployment workflow for the WordPress plugin, optimizing the autoloading process for improved performance.
	- Introduction of a `Utils` class that consolidates various utility functions for better organization and maintainability.

- **Bug Fixes**
	- Improved reliability of end-to-end tests by ensuring Composer autoload files are up to date.

- **Refactor**
	- Refactored utility function calls across multiple classes to use the new `Utils` class, promoting better code organization.
	- Removed unused references to the `get_asset_info` function, streamlining the code structure.
	- Shifted to Composer for dependency management, simplifying the loading process for the Parsely plugin.

- **Chores**
	- Namespace updates across numerous test files to better reflect the organization and structure of tests.


acicovic:
> Do we want to refactor the plugin to also adhere to the PSR-4 spec?

That's an interesting question. Generally, I'd say that the more structured a project is, and the more well-known standards it follows, it's generally easier to work with for external contributors. In our case, this might not translate to any real advantages though, and `classmap` seems pretty standard.

A couple of questions:

- Do you think that the current change could be characterized as a breaking change?
- Do you think that switching to PSR-4 would be a breaking change?

vaurdan:
> Do you think that the current change could be characterized as a breaking change?

Currently, as it is, not really. I have only changed the tests to be PSR-4 compliant, all the remaining code base currently has the same structure and filenames.

> Do you think that switching to PSR-4 would be a breaking change?

I don't think it's a breaking change. Although we would need to change a bunch of file names, all the existing hooks would still function exactly as before, and the same for all the functionalities. The only way I can think of where it could be a breaking change is, if for some reason, someone would be requiring a file directly from the plugin. And I don't think this is something that anybody does, and definitely not something that we support.

wp-parsely.php
```diff
@@ -73,34 +73,10 @@
...
-require_once __DIR__ . '/src/Metadata/class-page-builder.php';
-require_once __DIR__ . '/src/Metadata/class-page-for-posts-builder.php';
-require_once __DIR__ . '/src/Metadata/class-paginated-front-page-builder.php';
-require_once __DIR__ . '/src/Metadata/class-post-builder.php';
-require_once __DIR__ . '/src/Metadata/class-tag-builder.php';
+if ( file_exists( __DIR__ . '/vendor/autoload.php' ) ) {
+	require __DIR__ . '/vendor/autoload.php';
+}
 
-add_action( 'plugins_loaded', __NAMESPACE__ . '\\parsely_initialize_plugin' );
```
GaryJones, line 103: Moving these to below the function makes it harder to read. You need to know the context that the function is being called with, to have the parameters make sense.

vaurdan: Thanks for the feedback! That's a fair point that I didn't consider. 

However, I feel that with the DocBlocks in between the `add_action` and the function definition, it breaks the flow and - personally speaking - makes it a bit more difficult to follow. I feel that having the `add_action`/`add_filter` right below the callback function makes it a bit easier to read, even though we can lose the context of the function parameters, specially on larger functions. 

What do you think @GaryJones? I'm happy to revert back to the original organization, if we think it's an arguably better developer experience.

GaryJones: I've been advocating for filter, documentation, function, since at least 2014 (Tom wrote https://tommcfarlin.com/clean-wordpress-code/ after we discussed it). The docblock can be mentally skipped when reading the code, so having the parameter context, then what is happening to it, makes more sense than having a function that does something to some values, and only then finding out what the values are afterwards.

vaurdan: @GaryJones thanks for sharing that blog post, super interesting! 

> The docblock can be mentally skipped when reading the code

I do agree, although personally I do find slightly more difficult to read through a large file and parse what is happening when I have the docblocks in the middle. But I admit that might be the way my brain works. With that perspective in mind, I think it will make it easier to read as I can logically understand why the filter is above the documentation, and not below the callback, so thanks for sharing it!

Although, playing devil's advocate here, one could argue that with the documentation - function - filter format, all the required information about the function parameters can be found on the docblock, and/or inferred from the parameter’s name.

Anyway, in this case, let's not fix what isn't broken 😄  I will revert that change and go back to the former filter - documentation - function format!

Again, thank you for that context, was a very interesting read and something I will keep on my mind.

acicovic (approved):
Thank you for working on this!

To keep this as simple as possible, let's make a possible PSR-4 migration in another PR.

acicovic:
@vaurdan, just a FYI that I've pushed a [commit](https://github.com/Parsely/wp-parsely/pull/2621/commits/e247799471626f5be0ad5e77d0cb73872056327a) that removes a couple of unneeded `use` declarations below one of your own removals.

PR #2649, opened 2024-07-16
Permissions: Fix PHP Warning when the user role is invalid

vaurdan:
## Description

There are certain scenarios where the user role might be invalid - the role has been deleted, for example - and therefore, the roles array in the `WP_User` object might not have any data on the `0` index.

This PR adds an additional validation that checks if this array key is set, and if not, returns false.

This prevents an issue where a `Warning: Undefined array key 0` warning can be thrown if there is no valid role associated with the user.

## Motivation and context
Improve the reliability of the plugin's codebase and prevent PHP errors and warnings from filling the logs.

## How has this been tested?
Manually tested.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected the user role setting check before verifying post edit capabilities, ensuring proper permission handling.

### DIFF
--- a/src/class-permissions.php
+++ b/src/class-permissions.php
@@ -86,6 +86,11 @@ class Permissions {
 			return false;
 		}
 
+		// Current user's role is not set correctly.
+		if ( ! isset( $current_user->roles[0] ) ) {
+			return false;
+		}
+
 		// Check that the user's role has the capability to edit posts.
 		$current_user_role = $current_user->roles[0];
 		$valid_roles       = array_keys( self::get_user_roles_with_edit_posts_cap() );
--- a/src/class-permissions.php
+++ b/src/class-permissions.php
@@ -80,32 +80,31 @@ class Permissions {
 			return false;
 		}
 
-		// Current user's role is not yet set.
 		$current_user = wp_get_current_user();
-		if ( 0 === count( $current_user->roles ) ) {
+		$user_roles   = $current_user->roles;
+
+		// Current user's role is not yet set.
+		if ( 0 === count( $user_roles ) ) {
 			return false;
 		}
 
-		// Current user's role is not set correctly.
-		if ( ! isset( $current_user->roles[0] ) ) {
+		// Get the roles with the capability to edit posts.
+		$valid_roles = array_keys( self::get_user_roles_with_edit_posts_cap() );
+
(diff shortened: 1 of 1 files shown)


rinatkhaziev (approved):
I think `array_values( $current_user->roles)` might be more appropriate?

WPprodigy:
Also, shouldn't it be checking all of the roles and not just the presumed first one? Or probably better yet, just using `current_user_can()`?

rinatkhaziev:
I've held off on rolling out the hotfix until @vaurdan or @acicovic are online.

acicovic:
Hey @rinatkhaziev and @WPprodigy, thank you for chiming in and for your suggestions.

I don't remember why I ended up using `$current_user->roles[0]`, and we can consider testing the alternatives proposed. Since the current fix should work though, we're happy to merge it as-is and release a 3.16.1 today/tomorrow.

We can then see if any of the proposed solutions are better and act accordingly.

Let me know of your thoughts.

acicovic:
In some cases, this code could be called before the user object would be initialized properly. Haven't tested, but I don't know if `current_user_can()` would work here.

vaurdan:
@rinatkhaziev @WPprodigy thank you both for your feedback!

Since it's actually possible to have multiple roles, I did some changes to the permissions checking to handle that scenario. A good side effect, is that it should also handle when there are users with invalid roles, which seems to be the case of the warning. 

I have tested these changes locally, with a user that had two roles, and it is working as expected. @acicovic what do you think of this solution?

acicovic:
This looks good to me provided we want to have a permissive approach (the role with the most privileges wins).

I'm wondering though if this could raise edge cases (some customers wanting a restrictive approach instead), which would also warrant some UI work. We could provide this in the future though, if we stumble upon it.

Thoughts?

vaurdan:
> I'm wondering though if this could raise edge cases (some customers wanting a restrictive approach instead), which would also warrant some UI work. We could provide this in the future though, if we stumble upon it.

That's a fair point. However, having multiple roles on the same user is not something that is supported out of the box, it requires either a plugin or custom code. I think that, in that scenario, we could offer them a filter, and they can tweak it to their needs.

acicovic:
Agreed that this is an edge case. Hopefully we may never need to deal with it.

acicovic (approved):
Looks good to me. Thanks for working on this!

PR #2650, opened 2024-07-17
PCH Smart Linking: Use the post title as the Smart Link title.

vaurdan:
## Description

This PR improves the Smart Link title returned by the Content Suggestions API. Previously, the title was retrieved from the post title stored in the vector database, which is in lowercase. Now, the Smart Link will use the destination post title if a valid post object is found using the canonical URL/permalink.

## Motivation and context

Improve the quality of the Smart Links attributes.

## How has this been tested?
Tested locally.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved title setting for smart links to ensure accurate display based on the destination post ID.

### DIFF
--- a/src/Models/class-smart-link.php
+++ b/src/Models/class-smart-link.php
@@ -128,7 +128,14 @@ class Smart_Link extends Base_Model {
 		int $post_id = 0
 	) {
 		$this->set_href( $href );
-		$this->title          = $title;
+
+		// Set the title to be the destination post title if the destination post ID is set.
+		if ( 0 !== $this->destination_post_id ) {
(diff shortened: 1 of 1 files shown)


acicovic (approved):
Thanks, makes sense. Does this qualify as a change or a fix?

vaurdan:
Good question @acicovic ! I would say a change, since it wasn't really broken before, and this is more of an improvement than a fix. But I can also see why it could be a fix. Feel free to change the tag if you think it makes more sense as a fix 🙁

acicovic:
Since we're releasing a patch release and this can be flagged as a fix, I'm changing the tag to be a fix.

PR #2652, opened 2024-07-17
PCH Smart Linking: Fix CSS leaking from the Block Preview

### LINKED ISSUE #2627: Smart Linking: Review modal alters styling in PCH Sidebar
## Describe the bug
When the `Review Smart Links` modal is invoked, it alters some styles in the PCH Sidebar.

Initial styling:

image

When the modal is shown:
- Notice how some of the text styling changes in the sidebar.
- The modal's title font changes, and seems to leak in the sidebar.
image

This styling glitch doesn't appear when the close dialog appears:
image

## To Reproduce
- In the PCH Sidebar, click the `Add Smart Links` button.
- Observe the styling changes in the Sidebar.

## Expected behavior
The modal's title font should probably not change, and it should have no impact on the PCH Sidebar's styling.

vaurdan:
## Description
As reported in #2627, some editor styles were leaking outside the block preview context in the Review modal. Although we were replacing all the `body` selectors with the specific `.wp-parsely-preview-editor` class, this was not enough to isolate some selectors, such as headings.

This PR changes that logic, and instead of simply replacing the body with the block preview class, it prefixes all the existing selectors with the `.wp-parsely-preview-editor` class, while also still replacing the `body` selector.

## Motivation and context
* Refine the UI of the Smart Linking modal, and prevent unexpected behaviours.
* Closes #2627 

## How has this been tested?
Tested locally with four different themes. Each theme changed the resulting block preview, and both the Modal title and sidebar headers still retain their original styling. 

image


acicovic:
Congrats on that work, this seems to be more complicated that I expected. Looking into it soon.

acicovic (commented):
Looks good to me, 2 comments:

- I think a couple of introduced `@since` DocBlocks might need to be `3.16.1` instead of `3.16.0`.
- Also wondering if we need a commend above `useEffect`.

vaurdan:
@acicovic thank you, good catch :) 
Fixed the wrong version in the `@since` docblocks, and also added an additional documentation block to the useEffect.

acicovic (approved):
Thanks!

PR #2654, opened 2024-07-18
PCH Smart Linking: Show error when API returns empty list of Smart Link suggestions

### LINKED ISSUE #2626: Smart Linking: Empty Review Smart Links modal
## Describe the bug
It has happened to me a couple of times to get a blank `Review Smart Links` modal. No clear steps to reproduce, but it might have to do with receiving an empty result set, or ending up with empty result set after processing.

image

## Expected behavior
We should show a message stating what has gone wrong.

vaurdan:
## Description
In certain scenarios, it is possible for the API to return an empty set of Smart Links as suggestions. When this happens, the Smart Linking review modal does still open, and if there are no existing Smart Links on the content, it will be empty, as reported in #2626.

This PR adds an additional validation after processing the API smart links, and if there are none, it will prevent opening the modal, and will show an error (Snackbar and Notice) instead.

## Motivation and context
* Improve the UX of the Smart Linking feature
* Closes #2626]

## How has this been tested?
Tested locally, by forcing the API to return an empty response.


acicovic (approved):
Thank you!

PR #2656, opened 2024-07-18
Fix failing E2E tests

acicovic:
## Description
With this PR, we're fixing a couple of failing E2E tests after the release of WordPress 6.6.

## Motivation and context
Make all tests pass, no hindered workflow.

## How has this been tested?
Updated tests that were previously failing now pass.


vaurdan (approved):
Thanks for working on this!

PR #2662, opened 2024-07-23
PCH Smart Linking: Fix bug causing ‘No smart links’ error despite valid API response

vaurdan:
## Description
The #2654 PR introduced a bug where, even though the API would return valid Smart Links, the UI would throw a "No smart links were generated." error.

This was due to how React handles store variables. Even though `processSmartLinks` updates the `smartLinks` variable in the Smart Linking store, the `smartLinks` variable would be empty/outdated when we check its length (`if (smartLinks.length === 0)`).

This PR addresses this issue by returning the processed smart links from the `processSmartLinks` function and using that returned value to determine if an error should be outputted.

## Motivation and context
Fix bug that prevents the proper usage of the Smart Linking feature

## How has this been tested?
Tested manually with different mocked-up API responses.

## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling for smart link processing, improving clarity and user experience.
	- Streamlined flow for displaying processed smart links in the review modal. 

- **Improvements**
	- Improved type safety with explicit return types for smart link processing functions.


acicovic (approved):
Thanks!

PR #2666, opened 2024-07-24
PCH Related Posts: Prevent Related Posts indefinite loading

vaurdan:
## Description

This PR enhances the `usePostData` hook in the Related Posts feature by utilizing `hasFinishedResolution` and `isResolving` to ensure it returns data only when it's fully resolved. This update is particularly useful in scenarios where the `users` REST API endpoint is disabled, preventing the PCH from retrieving a list of users.

Testing indicates that the new approach ensures the hook accurately determines when the data is ready to be rendered, preventing the component from remaining indefinitely in a Loading state. The component will now attempt to render with the available data as soon as all fetching operations are resolved.

## Motivation and context
Improve the reliability of the Related Posts feature, ensuring it functions correctly even in edge cases or when unexpected API data is returned. By checking if the data has been resolved, the feature can gracefully handle scenarios where certain data endpoints, such as the `users` REST API, are disabled or slow to respond. This prevents the application from hanging indefinitely and enhances the user experience by rendering available data promptly.

## How has this been tested?
Tested locally.

## Summary by CodeRabbit

- **New Features**
	- Introduced a timeout mechanism in the post data fetching process, allowing for improved handling of delays. Users can now specify a timeout duration for fetching related post data, enhancing overall performance and responsiveness.


vaurdan:
@coderabbitai review

src/content-helper/editor-sidebar/related-posts/hooks.ts
```diff
@@ -36,13 +37,31 @@ interface PostData {
  *
  * @since 3.14.3
  * @since 3.14.4 Implemented checks to reduce risk of invalid data being processed.
+ * @since 3.16.2 Added a timeout to ensure the hook returns after X seconds.
```
acicovic: Maybe change "seconds" to "milliseconds" as this is what the function expects?

src/content-helper/editor-sidebar/related-posts/hooks.ts
```diff
@@ -81,19 +100,51 @@ export function usePostData(): PostData {
...
+				authors: authorRecords,
+				categories: categoryRecords,
+				tags: tagRecords,
+				isReady: true,
+			} );
+		}
+	}, [ postAttributes, isTimeout ] );
+
+	/**
+	 * Sets a timeout to ensure the hook returns after 5 seconds.
```
acicovic: Description needs improvement here.

acicovic (commented):
Interesting way to tackle this, I like it 🙂

I haven't tested, but if we're hitting WP endpoints by issuing HTTP requests, my gut feeling tells me that 200ms might be too low (e.g. high latency or heavy sites), and that we should raise it if it could be disruptive to user experience.

src/content-helper/editor-sidebar/related-posts/hooks.ts
vaurdan: Good call!

src/content-helper/editor-sidebar/related-posts/hooks.ts
vaurdan: Ah yes, good catch! :)

vaurdan:
@acicovic I agree with you, I am a bit torn here, since this timeout will prevent the filter type UI from rendering for that amount of time, if any of the HTTP requests take to long or fail. Unfortunately, since we're doing the HTTP requests using Gutenberg's `getEntityRecords`, we don't have control/access to the requests, and it's not possible to check if they have either returned an error code, or if they are still running.

Although, and I tested this by throttling the connection in Chrome, if the endpoint is taking too long to respond, when it finally returns a valid response, the component will re-render and work as expected. But I think I agree with you that 200ms might be on the short side. 

For how long do you suggest waiting? 500ms?

acicovic:
If I understand correctly, this timeout will only slow down users when there's a problem (e.g blocked user endpoint). This system will not affect websites that work as expected. So, not all users will be forced to wait for these 200ms.

If the above is correct, in my opinion even waiting 1 second should be fine, since there would be an infinite freeze anyway if this fix wasn't in place. What I'm afraid of is introducing re-renders in configurations that work fine currently, and then having issues regarding that. If the user is using the tool while the re-rendering takes place, it could lead to a weird UX experience.

vaurdan:
I might have found a way to tackle this [using `hasFinishedResolution` and `isResolving`, from the `core-data` store](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-data/#selectors). I will see if I can implement a better solution, and perhaps ditch the need for a timeout :)

vaurdan:
@acicovic I changed the logic for the usePostData hook, and now instead of relying on the timeout, we're using `hasFinishedResolution` and `isResolving` to check if the data we need is actually ready to be used. 

This has fixed the issue of having the UI waiting for the timeout before rendering the filter type Group Toggle, when one of the resolution fails. So it feels nearly instantaneous now :)

src/content-helper/editor-sidebar/related-posts/hooks.ts
```diff
@@ -36,23 +37,42 @@ interface PostData {
  *
  * @since 3.14.3
  * @since 3.14.4 Implemented checks to reduce risk of invalid data being processed.
+ * @since 3.16.2 Added `isLoading` and `hasResolved` to check if the data is still being fetched or if it
+ * 							 has been resolved.
```
acicovic: Is spacing correct here?

src/content-helper/editor-sidebar/related-posts/hooks.ts
```diff
@@ -36,23 +37,42 @@ interface PostData {
...
+		isReady: false,
+	} );
+
+	/**
+	 * Fetches the post attributes from the editor.
+	 * This includes the author, categories, and tags, and if the data is not available, it will be set to `null`.
+	 *
+	 * @since 3.14.4
+	 * @since 3.16.2 Added `isLoading` and `hasResolved` to check if the data is still being fetched or if it
+	 * 							 has been resolved.
```
acicovic: Same question about spacing.

acicovic (approved):
Looks good! Thank you!

src/content-helper/editor-sidebar/related-posts/hooks.ts
vaurdan: Weird, looks fine on my end. I will see if I can adjust it.

PR #2673, opened 2024-07-26
PCH Smart Linking: Prevent suggesting links in unexpected blocks

### LINKED ISSUE #2619: PCH Smart Linking: Improve Classic Block handling
Smart Linking doesn't support Classic Blocks, and we alert users of this.
1. When we try to use Smart Linking on a Classic Block, we get `Smart Linking is not supported for the Freeform block.`
2. When we try to use Smart Linking for the whole post ("All Blocks" setting) and the post contains at least one Classic Block, we get the same error message.

1. In the message, change `Freeform block` to `Classic Block` for clarity, as this is how it's [generally referenced](https://wordpress.org/documentation/article/classic-block/).
2. Come up with a better error message for the second case. For example, `This post contains one or more Classic Blocks, which are unsupported by Smart Linking. Operation aborted`. It would be nice if we could come up with a shorter message.

When the "All Blocks" setting is selected, we could consider removing Classic Blocks and submit the rest of the content. Noting though that this could result in suboptimal results, and that implementation might be tricky.

### LINKED ISSUE #2669: Smart Linking should not suggest links inside headings/captions

There is the possibility of the API returning suggestions in the headings and captions of the post content. This is not wanted, and therefore the plugin should be able to filter these suggestions and ignore them.

When the API returns the suggestions, the suggested links are processed and validated. In this process, it should check if the link is being inserted inside a heading or a caption, and skip/ignore it if so.

Send the post content to the API without any headings or captions. However this could be problematic, since the `offset` returned by the API might not be consistent with the `offset` we're expecting to the actual raw post content.

vaurdan:
## Description
This PR is in response to the #2669, where we want to prevent Smart Links from being suggested in unexpected blocks, such as headings or captions.

It introduces a new constant `BLOCKS_ALLOWED` in the `Smart_Linking` class, which can be customized using the new `wp_parsely_smart_linking_allowed_blocks` filter. This allows developers to specify which blocks are eligible for Smart Linking. Here's an example of how developers can customize the allowed blocks for the Smart Linking feature:

```php
function customize_smart_linking_allowed_blocks( $allowed_blocks ) {
    // Add custom blocks to the allowed list.
    $allowed_blocks[] = 'core/image';
    return $allowed_blocks;
}
add_filter( 'wp_parsely_smart_linking_allowed_blocks', 'customize_smart_linking_allowed_blocks' );
```

The Smart Linking UI was also updated to respect these allowed blocks, and a utility function `isInsideHeadingOrCaption` was added to skip applying links within headings or captions. 

## Motivation and context
* Closes #2669
* Improve the flexibility and accuracy of Smart Linking, by not allowing generating Smart Links on untested blocks, but still giving the option for developers to specify eligible blocks. 

## How has this been tested?
Tested locally by verifying the filter for allowed blocks works as expected, ensuring Smart Links are not applied to inappropriate blocks or within headings or captions, and checking the UI for correct behavior with the new settings and adjustments.

https://github.com/user-attachments/assets/4cf05edb-cae5-4c89-bb53-dbb04054ab0b


src/content-helper/editor-sidebar/smart-linking/component-settings.tsx
```diff
@@ -141,16 +175,17 @@ export const SmartLinkingSettings = ( {
...
-			} else {
-				moveButtonAndShowHint();
-			}
+			moveButtonAndShowHint( __( 'Select a block to apply Smart Links.', 'wp-parsely' ) );
+		}
+
+		// If the selected block is not allowed, move the focus to the
+		// "All Blocks" button and set the hint to the user.
+		if ( selectedBlock && applyTo !== ApplyToOptions.All && ! ALLOWED_BLOCKS.includes( selectedBlock.name ) ) {
+			moveButtonAndShowHint( __( 'This individual block is not supported for Smart Links.', 'wp-parsely' ) );
```
acicovic: The problem isn't the individual block here, but the block type, correct?
In this case, we could update the message to something like "This block type isn't supported by Smart Linking."

acicovic: We could even include the block type name if we wanted, though this isn't a requirement.

vaurdan: Thanks, that a pretty good idea! I did that, and I think it makes it easier to understand the message.
image

acicovic (approved):

PR #2723, opened 2024-08-20
Fix current_user_can_use_pch_feature() issues

acicovic:
## Description
This PR fixes a couple of issue in the `Permissions::current_user_can_use_pch_feature()` function. Notably:
- The function returned a wrong value in certain scenarios when using the `wp_parsely_current_user_can_use_pch_feature` filter.
- The `wp_parsely_current_user_can_use_pch_feature` filter had no effect when the feature in question was disabled.

## Motivation and context
Improve function behavior and make it return the results we expect.

## How has this been tested?
The issues above were discovered while writing tests for the `Permissions` class. These need a final look and will be introduced in a PR very soon. They are to be released as part of 3.17.0.

### DIFF
--- a/src/class-permissions.php
+++ b/src/class-permissions.php
@@ -70,16 +70,6 @@ class Permissions {
 			return false;
 		}
 
-		// All AI features are disabled.
-		if ( true !== $pch_options['ai_features_enabled'] ) {
-			return false;
-		}
-
-		// The specific AI feature is disabled.
-		if ( true !== $feature_options['enabled'] ) {
-			return false;
-		}
-
 		$current_user = wp_get_current_user();
 		$user_roles   = $current_user->roles;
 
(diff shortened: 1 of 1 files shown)


src/class-permissions.php
```diff
@@ -91,21 +81,38 @@ public static function current_user_can_use_pch_feature(
...
+
+		// All AI features are disabled.
+		if ( true !== $pch_options['ai_features_enabled'] ) {
+			return false;
+		}
+
+		// The specific AI feature is disabled.
+		if ( true !== $feature_options['enabled'] ) {
+			return false;
 		}
```
acicovic, line 116: This is marked as added, but it was only moved here from lines 73-81.

vaurdan (approved):
Thanks for working on this!

Issue #1927, opened 2023-10-16
Migrate E2E tests to Playwright


acicovic:
## Is your feature request related to a problem?
WP/Gutenberg are [migrating towards Playwright](https://github.com/WordPress/gutenberg/issues/38851) for E2E tests.

## Describe the solution you'd like
Migrate/convert all our existing E2E tests to Playwright.

## Describe alternatives you've considered
Do nothing, but this doesn't seem like a viable approach. In the [wordpress/e2e-test-utils package page](https://www.npmjs.com/package/@wordpress/e2e-test-utils#e2e-test-utils), it is stated:
> Note that there's currently an ongoing [project](https://github.com/WordPress/gutenberg/issues/38851) to migrate E2E tests to Playwright instead. This package is deprecated and will only accept bug fixes until fully migrated.

PR #2733, opened 2024-08-26
Add Playwright infrastructure for E2E tests

acicovic:
## Description
Per #1927, we're migrating our E2E tests to Playwright. This PR introduces the basis for this work, by adding the essentials to operate with Playwright. It adds:
- The required packages.
- A base configuration file, which doesn't currently differ from WordPress defaults. We can probably delete it if we don't need any overrides.
- An untested GitHub workflow, so Playwright E2E tests can run as checks in PRs. It will need to be tested and finalized, though we might need to merge it as-is before refining it.
- A migrated E2E test (passing locally and located in a default directory, outside our current tests), to validate that things work.

To run our tests with Playwright, we're introducing these 2 new commands:
- `npm run test:e2e:playwright`: Runs Playwright E2E tests.
- `npm run test:e2e:playwright:debug`: Runs Playwright E2E tests in debug mode.

The original E2E test commands have been left untouched, so we can run the existing tests along the new Playwright ones. The commands will be updated and finalized once all the ests have been migrated.

## Motivation and context
Begin the work on #1927.

## How has this been tested?
- The migrated test passes locally, and is to be tested on GitHub.
- The GitHub workflow is pending testing, to verify that it works or follow-up with fixes.

## Summary by CodeRabbit

- **New Features**
  - Introduced an end-to-end testing workflow using Playwright for the WordPress application.
  - Added a new configuration file for Playwright testing, ensuring compatibility with WordPress.
  - Created a test suite for verifying the action link functionality of a WordPress plugin.

- **Improvements**
  - Enhanced test execution control by refining the scope of tests run with Jest and Playwright.
  - Added new scripts for running and debugging Playwright tests, improving testing capabilities.
  - Updated the testing workflow to utilize the latest versions of Node.js and Ubuntu for better performance.


acicovic:
The GitHub workflow worked immediately, but the migrated test was failing due to an oversight. Also, Jest was trying to run the Playwright test, which didn't work well 😄

All these are now fixed, so opening the PR for review.

acicovic:
In the latest update, I've merged the two E2E workflow files into one, which should save some time. I've also updated the tooling (e.g. we now use React 18 instead of 16, and Ubuntu 24 instead of 20).

After the changes, the tests [continue to pass](https://github.com/Parsely/wp-parsely/actions/runs/10566106654/job/29272101829?pr=2733) fine.

PR #2731, opened 2024-08-22
REST API Refactor: Base classes and Content Helper namespace implementation

vaurdan:
## Description
This PR is part of the effort described in #2426, and kicks-off the refactoring of the internal REST API. 

This PR in particular introduces the new base classes for the new internal API, and also the implementation of the `content-helper` namespaces, which includes all the Content Suggestions functionalities, such as the Excerpt Generator, TItle Suggestions and Smart Linking.

I have been also refactoring the tests, so all these classes have proper integration tests.

In the UI, the endpoints in the providers for the Content Suggestions features have been replaced to use the new endpoints.

## Motivation and context
 * Implement the work described in #2426 
 * Improve the reliability of the codebase, and make it easier to maintain and improve upon. 

## How has this been tested?
I have validated locally that all the Content Helper features as still working as expected, and added new integration tests to test all the new classes functionalities.

## Summary by CodeRabbit

- **New Features**
  - Introduced new REST API controllers and endpoints for improved content management, including smart linking, excerpt generation, and title suggestions.
  - Updated API endpoints to utilize the latest version, enhancing functionality and performance.
  - Added validation checks to ensure proper handling of hyperlink references and user permissions.

- **Bug Fixes**
  - Resolved issues related to API endpoint registration and response handling.

- **Tests**
  - Implemented comprehensive integration tests for newly introduced and updated API endpoints to verify functionality and reliability.

PR #2734, opened 2024-08-29
REST API Refactor: Stats namespace implementation

### LINKED ISSUE #2426: Refactor all endpoints code

1. Plan
2. Implement

vaurdan:
## Description
This PR is part of the effort described in #2426, and implements the `stats` endpoint, which is related with specific Parse.ly stats related to a post, such as getting the performance details, post referrers, and related posts. 

There is also a `/stats/related` endpoint that is not necessarily linked to a post ID, and that is available publicly, so it can be used within the Related Posts block.

Here's the overview of the implemented API endpoints:

```
/wp-parsely/v1
  ├── /stats
  │   ├── /related
  │   ├── /posts
  │   ├── /post
  │   │   ├── /{post_id}
  │   │   │   ├── /details
  │   │   │   ├── /referrers
  │   │   │   └── /related
```

The old tests have been migrated to new Endpoints testing, and new tests have been implemented when needed.

The UI was also updated to use these new endpoints, with all the code related with the older endpoints being removed as well.

** Note: since this work as based on the `refactor/internal-api-content-helper` branch, this PR is using that branch as the base. However, once that branch is merged (in #2731), the PR should auto-rebase to `refactor/internal-api`.

## Motivation and context
 * Implement the work described in #2426 
 * Improve the reliability of the codebase, and make it easier to maintain and improve upon. 

## How has this been tested?
I have validated locally that all the features using the `/stats` endpoint are still working as expected, and added new integration tests to test all the new classes functionalities.


acicovic (commented):
Can you elaborate a bit on why we're implementing related posts both as a part of `class-endpoint-post.php` and as a separate `class-endpoint-related.php`? What are their differences, and maybe there are opportunities to simplify or scrap some code?

acicovic (approved):
Thank you for working on this! I'll be merging it momentarily.

vaurdan:
> Can you elaborate a bit on why we're implementing related posts both as a part of class-endpoint-post.php and as a separate class-endpoint-related.php? What are their differences, and maybe there are opportunities to simplify or scrap some code?

Sure! I'm not too fond of this approach either, but I think it's necessary given how the two endpoints behave differently:, although both using the same upstream API:
 
* The `/stats/related` endpoint returns to all the related posts to a given URL. This endpoint is public, and it's used for the Related Content block, for example.
* The `/stats/[post-id]/related` is slightly different. It is a private endpoint, and it is used to retrieve all posts that are related with a given post ID. This endpoint is used in the Related Posts panel in the PCH sidebar. 

The simplest way to uniformize it would be to remove the `/stats/[post-id]/related` endpoint, but that would force us to use the public endpoint in the PCH sidebar, which right now has no side effect since the data return by both is pretty much the same, if I'm not mistaken. However, in the future, if we decide to have different implementations for the public and private `related` endpoints - such as using different upstream API endpoints -, we would need to implement this again. 

I tried as much to avoid code repetition, that's why I implemented a Related Posts trait that has all the business logic for the related posts.

I hope this makes sense. Let me know your thoughts, and if you have any questions @acicovic!

acicovic:
I think it's fine to leave as is then.

PR #2735, opened 2024-08-30
REST API Refactor: Settings namespace implementation

### LINKED ISSUE #2426: Refactor all endpoints code

vaurdan:
## Description
This PR is part of the effort described in #2426, and implements the `settings` endpoint, which is related with specific plugin settings, such as the dashboard widget settings, and the editor sidebar settings.

The endpoint implement both a `/settings/{setting}/get` and a `/settings/{setting}/set` method, but the main endpoint - `/settings/{setting}/`can be used as well, by changing the HTTP verb - `GET` to retrieve the settings and `PUT`/`POST` to update the settings.

Here's the overview of the implemented API endpoints:

```
/wp-parsely/v2
  ├── /settings
  │   ├── /dashboard-widget
  │   │   ├── /get
  │   │   └── /set
  │   └── /editor-sidebar
  │       ├── /get
  │       └── /set
```

The old tests have been migrated to new Endpoints testing, and new tests have been implemented when needed.

The UI was also updated to use these new endpoints, with all the code related with the older endpoints being removed as well.

** Note: since this work as based on the `refactor/internal-api-stats` branch, this PR is using that branch as the base. However, once that branch is merged (in #2734), the PR should auto-rebase to `refactor/internal-api`.

## Motivation and context
 * Implement the work described in #2426 
 * Improve the reliability of the codebase, and make it easier to maintain and improve upon. 

## How has this been tested?
I have validated locally that all the features using the `/settings` endpoint are still working as expected, and added new integration tests to test all the new classes functionalities.


src/rest-api/settings/class-endpoint-dashboard-widget-settings.php
```diff
@@ -1,30 +1,39 @@
...
+	/**
+	 * Returns the endpoint name.
+	 *
+	 * @since 3.17.0
+	 *
+	 * @return string
+	 */
+	public function get_endpoint_name(): string {
+		return 'dashboard-widget';
+	}
```
acicovic, line 30: Do you think there would be an advantage in making these functions static? Maybe they could help eliminate some magic strings?

acicovic (approved):
Thank you for your work! Approving and merging this!

src/rest-api/settings/class-endpoint-dashboard-widget-settings.php
vaurdan: That's a pretty good idea. I will implement that and fix it on a different PR!

Issue #2552, opened 2024-06-12
PCH: include excerpt generator in Content Helper sidebar panel


danielabloch:
## Is your feature request related to a problem?
When using the Parse.ly Content Helper, customers are not aware that we offer smart excerpt generation (powered by their performance data + our ML system/AI). This means one of the capabilities we built is unknown and thereby unused by customers.

## Describe the solution you'd like
Include the existing excerpt generation feature in the Parse.ly Content Helper panel in Block Editor interface

Excerpt generator in main Post tab for WordPress settings
<img width="275" alt="excerpt feature" src="https://github.com/Parsely/wp-parsely/assets/88679146/81cd3d78-d781-4ea8-bd17-09a4f3a62c1f">

Where it should be also added > Content Helper tools tab
<img width="283" alt="add excerpt feature" src="https://github.com/Parsely/wp-parsely/assets/88679146/4e6176f2-c2f8-45e9-881d-ee8b08bcf8f3">

Its placement should be below Title Suggestions and above Smart Linking because that is its logical placement within the publishing workflow

## Describe alternatives you've considered
N/A

## Additional context
After discussing it, we see no harm in offering the feature in both the WP settings tab and PCH tab so please keep it in the former and add it to the latter

acicovic:
@danielabloch, @dabowman, we might choose to integrate this into 3.17.0, relatedly to the work on #2776.

To know how to proceed though, we need to decide if we want to host the Excerpt Suggestions feature under the PCH Sidebar as a UI (like in Post Settings) or as a link to the existing tool.

danielabloch:
i still prefer having it be UI like in Post Settings 
mentioned in slack that it would be helpful to note in the UI if a brief has already been generated when someone is looking at it in PCH or the main sidebar

acicovic:
OK, we'll look into adding it as a UI. 🙂

acicovic:
I'll move this to the 3.17.0 milestone, as it makes sense to pair it with #2776. @vaurdan, let me know about any possible thoughts or objections to that.

PR #2776, opened 2024-09-12
PCH Excerpt Suggestions: Add Persona and Tone settings

### LINKED ISSUE #2692: Excerpt Suggestions: Allow different personas/styles
In Excerpt Suggestions, we currently don't offer a way to customize the persona and style. We recently received internal feedback that this would be desirable, since the current persona/style isn't ideal for writings of certain types (e.g. documentation).

Provide controls to specify persona and style, as we currently do in our Title Suggestions feature.

acicovic:
## Description
This PR adds `Persona` and `Tone` settings to our `Excerpt Suggestions` feature, in a similar manner to what already exists in our `Title Suggestions` feature.

## Motivation and context
- Users want to be able to customize the persona and tone of generated excerpts.
- Closes #2692.


acicovic:
Notes:
- This work is currently based on our [refactor/internal-api](https://github.com/Parsely/wp-parsely/tree/refactor/internal-api) branch in order to avoid duplicate work since it adds a settings endpoint.
- @vaurdan, trying different things, I still couldn't make `SettingsProvider` to work. I suspect it might have to do with this being in Gutenberg Settings, and maybe a higher order component is needed as you've done previously for Smart Linking. Nothing has worked for me so far, but I'm probably missing something. Could you take a look?
- This PR is currently incomplete due to the previous point.

acicovic:
@vaurdan: regarding my previous comment, let's also keep in mind that there's a feature request to have [Excerpt Suggestions in the PCH Sidebar](https://github.com/Parsely/wp-parsely/issues/2552).

As this will probably affect how Excerpt Suggestions settings get saved, it would be good to factor this in during this work. If we're unsure of how this will play out, let's start implementing that request to avoid duplicate/refactoring work.

vaurdan:
@acicovic I went ahead and added the SettingsProvider to the component. It was a bit trickier, because of how we're using a SlotFill to insert the Excerpt panel to the Document sidebar (Using [PluginDocumentSettingPanel](https://developer.wordpress.org/block-editor/reference-guides/slotfills/plugin-document-setting-panel/)). 

This required a bit of reworking on how the code is organized - instead of being a standalone module, I made it part of the `editor-sidebar` module. This will also help when we add the panel to the PCH sidebar, since that move would have to be done nonetheless. 

From your original code, instead of having a new settings namespace for the excerpt generator, I changed it so it's also part of the main SidebarSettings namespace. 

Let me know your thoughts, and if you have any questions!

vaurdan:
Oh - I also renamed the component from `excerpt-generator` to `excerpt-suggestions` on a different commit 🙂

69023030 (approved):

src/content-helper/editor-sidebar/excerpt-suggestions/class-excerpt-suggestions.php
```diff
@@ -42,29 +42,28 @@ public function __construct( Parsely $parsely ) {
 	 * @return string The filter name.
 	 */
 	public static function get_feature_filter_name(): string {
-		return self::get_global_filter_name() . '_excerpt_generator';
+		return self::get_global_filter_name() . '_excerpt_suggestions';
```
acicovic: I think this will break existing usages of the `wp_parsely_enable_content_helper_excerpt_generator` filter. A strategy here would be to deprecate the old filter name, but still support if for some time. We should do the same if we've changed other filter names.

src/content-helper/editor-sidebar/excerpt-suggestions/class-excerpt-suggestions.php
```diff
@@ -42,29 +42,28 @@ public function __construct( Parsely $parsely ) {
...
 	}
-
 	/**
 	 * Returns the feature's script ID.
 	 *
-	 * @since 3.13.0
+	 * @since 3.16.0
```
acicovic: Has this `@since` been updated on purpose? Same for the one below.

src/content-helper/common/settings/types/sidebar-settings.d.ts
```diff
@@ -58,3 +59,14 @@ export interface SmartLinkingSettings {
 	MaxLinks: number;
 	Open: boolean;
 }
+
+/**
+ * Defines the settings structure for the PostExcerptGenerator component.
```
acicovic:
```suggestion
 * Defines the settings structure for the PostExcerptSuggestions component.
```

acicovic (commented):
Thank you for completing this work! I agree with renaming ExcerptGenerator to ExcerptSuggestions anywhere possible.

I've left a couple of comments, the only really important one being the one about the filter.

69023030 (commented):
T

src/content-helper/editor-sidebar/excerpt-suggestions/class-excerpt-suggestions.php
vaurdan: Good call!

vaurdan: I did addressed this on [b8e3978](https://github.com/Parsely/wp-parsely/pull/2776/commits/b8e3978b948c87578da32a098db3e4a6aacfe9bb). I ended up reverting back to the `excerpt_generator`, and it should send a deprecation notice (if WP_DEBUG is enabled), using `apply_filters_deprecated` and suggesting `wp_parsely_current_user_can_use_pch_feature`.

This required some tweaking since the `wp_parsely_enable_content_helper_excerpt_generator` was not even working properly, as now the excerpt generator does not have it's own JS and CSS files, so we can't simply not load it. So I hooked up that filter with the new permission feature of the plugin, so if `wp_parsely_enable_content_helper_excerpt_generator` is set and returning false, is as if the feature was disabled.

vaurdan:
@acicovic I believe I addressed all your feedback! Let me know if you have any additional thoughts 🙂 

I believe since you're the owner of this PR, you can't approve it, but if we have 🟢 , let me know and I can approve it myself.

src/content-helper/editor-sidebar/class-editor-sidebar-feature.php
```diff
@@ -0,0 +1,106 @@
...
+			return false;
+		}
+
+		return Permissions::current_user_can_use_pch_feature(
+			static::get_feature_name(),
+			$this->parsely->get_options()['content_helper'],
+			get_the_ID()
+		);
+	}
+
```
acicovic, line 64: Extra blank line.

```suggestion
```

src/content-helper/editor-sidebar/class-editor-sidebar-feature.php
```diff
@@ -0,0 +1,106 @@
...
+	 * Not in use for Editor Sidebar features.
+	 *
+	 * @since 3.16.0
+	 *
+	 * @return string The filter name.
+	 */
+	public static function get_feature_filter_name(): string {
+		return ''; // Not in use for this feature.
+	}
+
```
acicovic, line 78: Extra blank line.

```suggestion
```

src/content-helper/editor-sidebar/class-editor-sidebar-feature.php
```diff
@@ -0,0 +1,106 @@
...
+
+	/**
+	 * Returns the feature's filter name. The feature filter controls the
+	 * enabled/disabled state of a particular Content Helper feature.
+	 *
+	 * Not in use for Editor Sidebar features.
+	 *
+	 * @since 3.16.0
```
acicovic: Is this `@since` correct?

src/content-helper/editor-sidebar/class-editor-sidebar.php
```diff
@@ -136,7 +133,7 @@ public function init_features(): void {
 
 		// Initialize the features.
 		foreach ( $this->features as $feature ) {
-			$feature->run();
+				$feature->run();
```
acicovic: Extra tab that seems unneed?

```suggestion
			$feature->run();
```

src/content-helper/editor-sidebar/excerpt-suggestions/class-excerpt-suggestions.php
```diff
@@ -0,0 +1,107 @@
...
+				'3.17.0',
+				'wp_parsely_current_user_can_use_pch_feature'
+			);
+
+			// If the feature is disabled via the global or feature filters, return false.
+			if ( false === $feature_enabled_with_filter ) {
+				return false;
+			}
+		}
+		return $current_user_can_use_pch_feature;
```
acicovic, line 97: I think it would be nice to have a blank line above the return.

```suggestion

		return $current_user_can_use_pch_feature;
```

src/content-helper/editor-sidebar/excerpt-suggestions/class-excerpt-suggestions.php
```diff
@@ -0,0 +1,107 @@
...
+	 *
+	 * This throws a deprecated notice if the feature filter name is in use.
+	 *
+	 * @since 3.17.0
+	 *
+	 * @param ?bool  $current_user_can_use_pch_feature Whether the current user can use the feature.
+	 * @param string $feature_name The feature's name.
+	 * @return ?bool  Returns false if the feature is disabled via filters.
+	 */
+	public function callback_disable_feature_via_filters(
```
acicovic: This function returns a bool flag, but it's name indicates that it disables the feature. Maybe we can find a name that's a better fit for it?

acicovic (commented):
Thanks for addressing the filters situation!

I've left some minor comments.

src/content-helper/editor-sidebar/excerpt-suggestions/class-excerpt-suggestions.php
vaurdan: Yeah, I struggled a little bit trying to find the most appropriate name for this callback function. I renamed it to `callback_is_feature_enabled`, since it hooks to `wp_parsely_current_user_can_use_pch_feature`. What do you think?

acicovic: Yes, `callback_is_feature_enabled` makes sense to me!

acicovic: Function summary says `Checks and disables the feature if it is disabled via the global or feature filters.`, which isn't true since this function doesn't disable anything.

acicovic:
I've left a comment in our last discussion, but otherwise I think we're good to go. Let's not merge this though, until our API refactoring branch is merged first.

acicovic:
Closing in favor of #2890.

PR #2847, opened 2024-10-14
Fix AMP validation issues

### LINKED ISSUE #2846: Incompatibilities with AMP plugin
## Describe the bug
We received a report of wp-parsely having [AMP plugin](https://wordpress.org/plugins/amp/) incompatibilities. Since we officially [support this plugin per the docs](https://docs.parse.ly/plugin-common-questions/?itm_source=parsely-api#h-is-wp-parsely-compatible-with-amp-or-google-web-stories), we should fix any compatibilities.

## To Reproduce
1. Install the AMP plugin and follow its setup guide.
2. Go the AMP plugin's menu item and click on it.

There are various indicators of the incompatibilities, such as the badges in the "Error Index" menu item. "Site Scan" in the AMP plugin's settings also lists the Parse.ly plugin as a source of errors.

## Expected behavior
No incompatibilities should be shown for wp-parsely.

## Screenshots
image

## Environment
Online site, but reproducible locally by following the above instructions.

acicovic:
## Description
When using the [AMP plugin](https://wordpress.org/plugins/amp/), wp-parsely would be the source of certain AMP validation issues.

The culprits were found to be the Parse.ly tracker script, as well as our loader script. Both scripts were removed from AMP requests, since:
- Tracking is already being taken care of by `<amp-analytics>` (see [docs](https://docs.parse.ly/google-amp/)).
- The loader script probably doesn't have any practical use under the AMP context, and would probably require some work to be available under it. We can consider adding the loader script in the future if we see that there are practical use cases for it.

Since these scripts were being auto-removed by the AMP plugin anyway, these changes are more to appease warnings, rather than to fix something that wasn't working. This is to say that AMP integration was working, regardless of the warnings.

## Motivation and context
- Offer a user experience without quirky warnings, that might make users believe that their integration isn't working because they're getting warnings.
- Fixes #2846.

## How has this been tested?
Manually tested that we are getting AMP site scanning errors before the fix, and no errors after the fix has been applied. Also checked that the correct markup is still being added to non-AMP pages.

## Summary by CodeRabbit

- **New Features**
	- Enhanced the AMP integration by adding functionality to deregister scripts that could cause validation issues.
	- Updated comments to clarify the necessity of deregistering specific scripts for improved AMP compatibility.

### DIFF
--- a/src/Integrations/class-amp.php
+++ b/src/Integrations/class-amp.php
@@ -85,7 +85,8 @@ class Amp extends Integration {
 	}
 
 	/**
-	 * Adds AMP actions.
+	 * Adds AMP actions and deregisters any scripts that create AMP validation
+	 * issues.
 	 *
 	 * @since 2.6.0
 	 */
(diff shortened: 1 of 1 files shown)


vaurdan (approved):
Looks good!

PR #2873, opened 2024-10-17
PCH Refactor: Refactor the external services

### LINKED ISSUE #2426: Refactor all endpoints code

vaurdan:
## Description
This PR refactors the external services implementation, that are currently two: Parse.ly Content API and Parse.ly Suggestions API.

This PR includes all the new classes, and removes the old classes from the `RemoteAPI` directory. However, the tests are still not implemented, nor the old tests have been removed. Therefore, there are still failing tests and PHPStan warnings flagging issues within the old tests.

## Motivation and context
* Addresses #2426 
* Improve the overall codebase to make it easier to understand, maintain, and extend in the future.

## How has this been tested?
Since integration tests weren't yet implemented, I only relied on manual testing on my environment. From those tests, everything seems to be working as expected.


wp-parsely.php
```diff
@@ -57,6 +57,10 @@
 	require_once __DIR__ . '/vendor/autoload.php';
 }
 
+// Load Telemetry classes.
+require_once __DIR__ . '/src/Telemetry/telemetry-init.php';
```
acicovic, line 61: Can you give a bit of context about this?

src/class-validator.php
```diff
@@ -44,7 +44,7 @@ public static function validate_metadata_secret( string $metadata_secret ): bool
 	 * @param Parsely $parsely The Parsely instance.
 	 * @param string  $site_id The Site ID to be validated.
 	 * @param string  $api_secret The API Secret to be validated.
-	 * @return true|WP_Error True if the API Credentials are valid, WP_Error otherwise.
+	 * @return bool|WP_Error True if the API Credentials are valid, WP_Error otherwise.
```
acicovic, line 47: Should this function be able to return `false`? Do we handle this case in our calling code?

src/content-helper/post-list-stats/class-post-list-stats.php
```diff
@@ -296,12 +299,12 @@ public function get_parsely_stats_response( $analytics_api ) {
 			return null;
 		}
 
-		$response = $analytics_api->get_posts_analytics(
+		$response = $this->content_api->get_posts(
 			array(
-				'period_start'   => Analytics_Posts_API::ANALYTICS_API_DAYS_LIMIT . 'd',
+				'period_start'   => 'max_days',
```
acicovic: I wonder if we're introducing a "magic string" danger here for `period_start` and `limit` params.

src/rest-api/class-base-api-controller.php
```diff
@@ -26,7 +26,7 @@ abstract class Base_API_Controller {
 	 *
 	 * @since 3.17.0
 	 *
-	 * @var Base_Endpoint[]
+	 * @var array<string, Base_Endpoint>
```
acicovic, line 29: I agree that this notation is better, because it allows us to add the key type.

src/rest-api/stats/class-endpoint-posts.php
```diff
@@ -182,6 +180,42 @@ public function register_routes(): void {
...
+	 * @param string|array<string> $string_or_array The string or array to validate.
+	 * @return true|WP_Error
+	 */
+	public function validate_max_length_is_5( $string_or_array ) {
+		if ( is_string( $string_or_array ) ) {
+			$string_or_array = $this->sanitize_string_to_array( $string_or_array );
+		}
+
+		if ( count( $string_or_array ) > 5 ) {
+			return new WP_Error( 'invalid_param', 'The parameter must have at most 5 items.' );
```
acicovic: Message string is untranslatable.

src/rest-api/stats/class-endpoint-posts.php
```diff
@@ -237,6 +260,11 @@ public function get_posts( WP_REST_Request $request ) {
 
 		// Process the data.
 		$posts = array();
+		/**
```
acicovic, line 264: Let's add a blank line above this.

```suggestion

		/**
```

src/rest-api/content-helper/class-endpoint-excerpt-generator.php
```diff
@@ -150,12 +162,38 @@ public function generate_excerpt( WP_REST_Request $request ) {
...
+				'max_characters' => $max_characters,
+			)
+		);
 
 		if ( is_wp_error( $response ) ) {
 			return $response;
 		}
 
+		// TODO: For now, only return the first suggestion. When the UI is ready to handle multiple suggestions, we can
+		// TODO: return the entire array.
```
acicovic: This can be in a single TODO comment I think?

acicovic (commented):
As a first step, I've reviewed all edited files (not new ones) and left some comments.

wp-parsely.php
vaurdan: This was accidentally removed when we introduced the composer autoloading, so effectively the telemetry library was not loading.

src/class-validator.php
vaurdan: It shouldn't return `false`, but the `validate_credentials` allows to return false if there isn't a success in the response, and if there isn't a WP_Error, so I only anticipate this happening on a weird edge case.

However, you're right about the handling of false. I double-checked and the `validate_basic_section` doesn't seem to do anything if it returns false (it doesn't save the credentials, but it doesn't also output any error). I will update that function accordingly!

src/content-helper/post-list-stats/class-post-list-stats.php
vaurdan: Yeah it's not ideal. I did tried a different approach here, to avoid using these magic strings directly. Let me know your thoughts.

acicovic: Maybe there could be ways to use the original constants, but I think this is fine as it is now. Might be simpler that way.

src/rest-api/stats/class-endpoint-posts.php
```diff
@@ -10,12 +10,13 @@
...
 namespace Parsely\REST_API\Stats;
 
-use Parsely\RemoteAPI\Analytics_Posts_API;
 use Parsely\REST_API\Base_Endpoint;
+use Parsely\Services\Content_API\Content_API_Service;
 use WP_Error;
 use WP_REST_Request;
 use WP_REST_Response;
 use stdClass;
+use const _PHPStan_4f7beffdf\__;
```
acicovic: This should be removed. 🙂

src/services/content-api/class-content-api-service.php
```diff
@@ -0,0 +1,235 @@
...
+
+	/**
+	 * Returns the base URL for the Parse.ly Content API, aka Public API.
+	 *
+	 * @since 3.17.0
+	 *
+	 * @return string
+	 */
+	public function get_base_url(): string {
+		return 'https://api.parsely.com/v2';
```
acicovic, line 33: Any specific reason we're not drawing this from `Parsely::PUBLIC_API_BASE_URL`? I wouldn't mint encapsulating this to this class, just asking. To make this transition though, this method should become static.

src/services/suggestions-api/class-suggestions-api-service.php
```diff
@@ -0,0 +1,115 @@
...
+class Suggestions_API_Service extends Base_API_Service {
+	/**
+	 * Returns the base URL for the API service.
+	 *
+	 * @since 3.17.0
+	 *
+	 * @return string
+	 */
+	public function get_base_url(): string {
+		return 'https://content-suggestions-api.parsely.net/prod';
```
acicovic, line 37: Identical to the above.

src/rest-api/stats/class-endpoint-posts.php
acicovic: I'll take care of it.

acicovic (commented):
This is great work! Thank you so much! 🙂

I had lots of DocBlock/whitespace comments, so I addressed them [here](https://github.com/Parsely/wp-parsely/pull/2873/commits/a39172073193b208d1d63e4f723ea348c0dd37e2) instead of leaving them as comments. Please check and let me know if you have any observations.

I've left a couple of comments.

FYI, I'm also getting some PHPStan warnings, not related to tests.

```shell
 ------ ------------------------------------------------------ 
  Line   src/rest-api/class-base-endpoint.php                  
 ------ ------------------------------------------------------ 
  199    Dynamic call to static method                         
         Parsely\REST_API\Base_Endpoint::get_endpoint_name().  
 ------ ------------------------------------------------------ 

 ------ ---------------------------------------------------------------------------- 
  Line   src/rest-api/stats/trait-related-posts.php (in context of class             
         Parsely\REST_API\Stats\Endpoint_Post)                                       
 ------ ---------------------------------------------------------------------------- 
  142    Parameter #1 $callback of function array_map expects (callable(mixed):      
         mixed)|null, Closure(array): array{image_url: mixed, thumb_url_medium:      
         mixed, title: mixed, url: string} given.                                    
         💡 Type array of parameter #1 $item of passed callable needs to be same or   
            wider than parameter type mixed of accepting callable.                   
 ------ ---------------------------------------------------------------------------- 

 ------ ---------------------------------------------------------------------------- 
  Line   src/rest-api/stats/trait-related-posts.php (in context of class             
         Parsely\REST_API\Stats\Endpoint_Related)                                    
 ------ ---------------------------------------------------------------------------- 
  142    Parameter #1 $callback of function array_map expects (callable(mixed):      
         mixed)|null, Closure(array): array{image_url: mixed, thumb_url_medium:      
         mixed, title: mixed, url: string} given.                                    
         💡 Type array of parameter #1 $item of passed callable needs to be same or   
            wider than parameter type mixed of accepting callable.                   
 ------ ---------------------------------------------------------------------------- 
```

src/services/content-api/class-content-api-service.php
vaurdan: > Any specific reason we're not drawing this from Parsely::PUBLIC_API_BASE_URL? I wouldn't mint encapsulating this to this class, just asking.

Yeah, that's the idea: not having any of the external services logic outside the actual services. I'm going to remove the old unused consts from the Parsely class 🙂 

> To make this transition though, this method should become static.

Ah good idea! I will change it to static.

src/services/content-api/endpoints/class-endpoint-analytics-posts.php
```diff
@@ -93,29 +93,29 @@ public function get_endpoint(): string {
 	 * endpoint URL, if they are set. Since the Parse.ly API needs a key for
 	 * every value (e.g. tag=tag1&tag=tag2), we need to append them manually.
 	 *
-	 * @since 3.17.0
+	 * @param array<mixed> $query_args The arguments to pass to the API request.
 	 *
-	 * @param array<mixed> $args The arguments to pass to the API request.
 	 * @return string The endpoint URL for the API request.
+	 * @since 3.17.0
```
acicovic: Misplaced `@since`.

vaurdan: Aghh damn PHPStorm always moving the `@since`tag  around. Fixing it.

acicovic (approved):
Looks good to me!

PR #2893, opened 2024-10-29
PCH Performance Stats: Show stats for posts with a trackable status

### LINKED ISSUE #2881: PCH Sidebar Stats: Investigate displaying data for already published "future" posts
Here's the problem statement, expressed by a customer:

> We utilize the "future" post status in WP for stories that have already been published, but have a future update scheduled to publish. Unfortunately, the Parse.ly plugin sees that "future" status and won't attempt to pull analytics data – the user sees an error message that the story hasn't yet been published, which isn't true.

To give some context, we use the `select( 'core/editor' ).isCurrentPostPublished()` function, which is a core WP/Gutenberg function. This means that Gutenberg itself considers the post non-published.

Having published future posts could be characterized as an edge case, but we've decided to investigate to see if there's a graceful solution for this. To quote @vaurdan:

> Interesting use case. The standard use case for future post status is for posts that are scheduled to be published in a future date, therefore shouldn't have a public URL yet, nor have any analytics. But it seems that the default behavior is overridden so that future posts are indeed published.

> A possible solution to make this more flexible, would be to check only for the permalink - [like we do here](https://github.com/Automattic/vip-go-mu-plugins-built/blob/9ed0bbf924b1ca937515490675c5f8a16b3bd09a/wp-parsely-3.16/src/content-helper/editor-sidebar/performance-stats/provider.ts#L72-L81) - and skip the `isCurrentPostPublished` check. If there is no permalink, it means the post isn't published and never was. This would require some additional testing to make sure we cover any edge cases, but it might be a solution.

As the posts might already have a temporary permalink assigned to them, this solution might not be an option. So the first step is to investigate this potential solution, to determine if it is a valid one.

vaurdan:
## Description
The WP Parse.ly plugin offers a filter - `wp_parsely_trackable_statuses` - that allows defining custom post status that shall be tracked by Parse.ly. On the other hand, the PCH Performance panel does not respect this filter, and will not show stats unless the post is actually published (with the `publish` status).

This PR changes that behaviour, so if the post has any of the statuses set by `wp_parsely_trackable_statuses`, it will attempt to show the performance stats for that post.

I also changed slightly the text when the post has 0 views, so it shows the post title instead of the post ID, as seen on the screenshot. 

## Motivation and context
* Closes #2881 

## How has this been tested?
Tested locally and validated that posts with a trackable status try to fetch performance stats from Parse.ly

image

## Summary by CodeRabbit

- **New Features**
	- Added a new property for tracking statuses in the global `Window` interface.
	- Introduced a method to retrieve allowed post statuses for tracking in the Parsely class.
	- Enhanced the `Editor_Sidebar` to inject new trackable statuses into the frontend.
  
- **Bug Fixes**
	- Improved error handling in performance data fetching, providing clearer messages to users.

- **Refactor**
	- Centralized logic for determining trackable statuses, improving maintainability and readability.


acicovic (approved):
Thank you! This looks good to me and I think it makes sense conceptually.

PR #2895, opened 2024-10-30
Don't allow Super Admins to use disabled PCH features

vaurdan:
## Description
Super Admins, by default, will have no user roles unless they are added to that specific site. By doing a validation if the user is a Super Admin - after validating that the feature is enabled - we can allow super admins to use any PCH feature, unless they are intentionally disabled. 

## Motivation and context
* Fixes undesired behavior for super users using PCH.

## How has this been tested?
Tested locally on a multisite with a super admin account.

## Summary by CodeRabbit

- **New Features**
	- Enhanced user role permissions to allow super administrators access to specific features regardless of their roles.
  
- **Bug Fixes**
	- Maintained existing logic for user access checks, ensuring consistent functionality for all users.

- **Tests**
	- Improved test coverage for permissions, particularly for super administrators accessing enabled and disabled features.


acicovic:
Looking at the code, I have difficulty understanding how this disables features for Super Admins when a feature is disabled. Can you please elaborate?

Alternatively, a quick meeting to show me the behavior with and without the fix is fine, even maybe better.

Thank you! 🙂

acicovic (approved):

PR #3007, opened 2024-12-10
Add identifier to JSON-LD metadata

acicovic:
## Description
There are many plugins that generate JSON-LD. When our plugin's metadata is output as JSON-LD, it's often hard to distinguish if there are several other JSON-LD blocks in the page's source. This can be a hindrance when troubleshooting.

This PR adds a `class="wp-parsely-metadata"` identifier to our JSON-LD's `<script>` tag, so our own metadata can easily be spotted when trying to troubleshoot.

## Motivation and context
Make troubleshooting easier.

## How has this been tested?
Existing tests modified to pass.

## Summary by CodeRabbit

- **New Features**
	- Enhanced JSON-LD script tags to include a new CSS class for improved styling and JavaScript targeting.

- **Bug Fixes**
	- Updated test cases to reflect the new class attribute in JSON-LD script tags, ensuring accurate validation across various tests.


vaurdan (approved):
This is a great idea, thank you!

I'm approving it as it looks good. My only suggestion is that we could use `id` instead of class, since it should be unique. But just a nitpick, `class` works just fine for this purpose as well, so feel free to do it either way :)

acicovic:
@vaurdan, just for context, I did consider an ID when I thought of this PR, and then I looked on how Yoast and Rank Math do it.

Both of them identify their JSON-LD with `class` (`<script type="application/ld+json" class="yoast-schema-graph">` and `<script type="application/ld+json" class="rank-math-schema">`) so I went with class instead.

PR #3090, opened 2025-01-20
Traffic Boost: Fix permalinks and highlighting

vaurdan:
## Description
This PR fixes a couple issues I have detected when testing the TB interface on a real site:

 * It seems that in some cases, the `guid` param of an article might not be the most reliable way to get the link. Sometimes it does not have the permalink, on other cases it might point to a different multisite - this happened to me with a dev environment sync from a VIP environment. The `link` prop, on the other hand, seems more reliable on getting the correct permalink for a post.
 * The link highlighting works by adding a div wrapper around the content. While this work nice on some basic themes, it might break the CSS on certain themes, because the paragraphs might be styled with rules that do not allow a new element in the hierarchy. This was changed, and now the div is injected after the content, being a sibling of the post paragraphs. The content area is, therefore, the parent container of this injected div (called `wp-parsely-preview-marker`).

## Motivation and context
* Improve the reliability and stability of the Traffic Boost UI with real-life sites and data.

## How has this been tested?
Tested locally on my local environment, and on a VIP dev-env synced from a real prod environment.

## Summary by CodeRabbit

- **New Features**
	- Added a new `link` property to hydrated posts.
	- Introduced a utility function to get content area in previews.

- **Bug Fixes**
	- Updated link references across multiple components to use `post.link` instead of `post.guid.raw`.
	- Modified preview wrapper handling to use a marker approach instead of wrapping content.

- **Refactor**
	- Streamlined URL handling and content area selection in preview components.
	- Updated URL construction and link validation methods.
	- Adjusted HTML structure and CSS styling for the preview display.


src/content-helper/dashboard-page/pages/traffic-boost/preview/utils.ts
```diff
@@ -23,3 +23,38 @@ export const isExternalURL = ( link: TrafficBoostLink ): boolean => {
...
+ * The content area is the element that contains the marker as a child.
+ *
+ * @since 3.18.0
+ *
+ * @param {Document} document The document to get the content area from.
+ *
+ * @return {Element | null} The content area element or null if not found.
+ */
+export const getContentArea = ( document: Document ): Element | null => {
+	// Get the content area by checking for the PARSELY_PREVIEW_MARKER_CLASS first.
```
acicovic: Since we don't check with a second way, I think we can remove the `first` here.

acicovic (approved):
Thank you for working on this! Approving with a nitpick comment.

PR #3092, opened 2025-01-21
Fix update metadata endpoint calling issues

acicovic:
## Description
Our logic for calling Parse.ly's update metadata endpoint had multiple flaws. For example, the endpoint would be called for untrackable post types (e.g. revisions) and post statuses (e.g drafts).

This PR makes a bit of a cleanup and updates the function to work as expected, which is the endpoint getting called only for post types and statuses which are trackable. We're also stopping the calls from happening while integration tests are running.

## Motivation and context
Eliminate needless calls to infrastructure and prevent invalid data from being posted to the endpoint.

## How has this been tested?
A few integration tests were added.

## Release Notes

- **Bug Fixes**
    - Improved metadata endpoint handling with more robust checks and error management
    - Simplified post status tracking logic

- **Tests**
    - Added comprehensive integration tests for metadata endpoint functionality
    - Enhanced test coverage for various credential and post type scenarios

- **Refactor**
    - Renamed `update_metadata_endpoint` method to `call_update_metadata_endpoint`
    - Removed unnecessary caching mechanisms

- **Documentation**
    - Improved method documentation and error handling comments

### DIFF
--- a/src/UI/class-settings-page.php
+++ b/src/UI/class-settings-page.php
@@ -543,7 +543,6 @@ final class Settings_Page {
 				'title'      => __( 'Track Post Types as', 'wp-parsely' ),
 				'option_key' => $field_id,
 				'help_text'  => $field_help,
-				'filter'     => 'wp_parsely_trackable_statuses',
 			)
 		);
 
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -282,7 +282,8 @@ class Parsely {
 			update_option( self::OPTIONS_KEY, $options );
 		}
 
-		add_action( 'save_post', array( $this, 'update_metadata_endpoint' ) );
+		// @phpstan-ignore return.void
+		add_action( 'save_post', array( $this, 'call_update_metadata_endpoint' ) );
 	}
 
 	/**
@@ -397,12 +398,6 @@ class Parsely {
 	 *              By default,only 'publish' is allowed.
 	 */
 	public static function post_has_trackable_status( $post ): bool {
-		static $cache = array();
-		$post_id      = is_int( $post ) ? $post : $post->ID;
-		if ( isset( $cache[ $post_id ] ) ) {
-			return $cache[ $post_id ];
-		}
-
 		/**
 		 * Filters whether the post password check should be skipped when getting
 		 * the post trackable status.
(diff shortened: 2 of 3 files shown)


src/UI/class-settings-page.php
```diff
@@ -543,7 +543,6 @@ function (): void {
 				'title'      => __( 'Track Post Types as', 'wp-parsely' ),
 				'option_key' => $field_id,
 				'help_text'  => $field_help,
-				'filter'     => 'wp_parsely_trackable_statuses',
```
acicovic, line 546: This seems wrong. `wp_parsely_trackable_statuses` is to manage post statuses, not types.

src/class-parsely.php
```diff
@@ -397,12 +398,6 @@ public function insert_page_header_metadata(): void {
 	 *              By default,only 'publish' is allowed.
 	 */
 	public static function post_has_trackable_status( $post ): bool {
-		static $cache = array();
-		$post_id      = is_int( $post ) ? $post : $post->ID;
-		if ( isset( $cache[ $post_id ] ) ) {
-			return $cache[ $post_id ];
-		}
```
acicovic, line 404: @vaurdan, I need your opinion on this one.

This function uses caching, which I removed in the moment. I'd guess caching assumed that post status won't change while the different functions of the request are running. For one, I'm not 100% certain that the assumption is correct, and second this actually made some of the new integration tests fail (which is how I became aware of the caching).

The tests can be tinkered to work with caching, so let me know if you think we should restore it.

src/UI/class-settings-page.php
vaurdan: Indeed!

src/class-parsely.php
vaurdan: It is possible for the post status to change while the this request is running, if a post is published, for example, while the request runs. Seems highly unlikely to happen, but it can. 

As for the cache here, I don't really see a point. I guess it could make sense if either `post_has_trackable_status` or even `call_update_metadata_endpoint` would be called within a loop and would be called multiple times for the same post. But from a quick inspection it looks like it should only be called once per post, so there aren't any real gains on having a caching mechanism here, IMO.

vaurdan (approved):
Thank you so much for working on this. Looks good to me!

I have replied to your comments, so feel free to take a look before merging :)

src/class-parsely.php
acicovic: Thanks. Let's go without the caching here.

PR #3095, opened 2025-01-24
PCH: Fix issue with potential duplicate API calls in settings and smart links

vaurdan:
## Description

While working on the Traffic Boost feature, I noticed a re-rendering issue that was causing a bunch of duplicated requests in a row to the `editor-sidebar` settings endpoint. By checking if the data has actually changed, we can only call the endpoint if there are changes in the settings.

I also noticed a double call to the Smart Links `/get` endpoint. This was due to the initialization useEffect running twice. I added a safeguard ref to track if the component was already initialized, so it only calls the `/get` endpoint once.

## Motivation and context
* Improve the reliability of PCH and Parse.ly plugin.
* Save resources and improve performance by avoiding needless API calls. 

## How has this been tested?
* Tested locally on two environments: one where the issue was noticeable, and other where the issue wasn't present. With this solution, none of the environments suffer from this problem.

## Release Notes

- **Performance Improvements**
  - Enhanced settings saving mechanism to prevent unnecessary API calls
  - Optimized Smart Linking component initialization process

- **Code Maintenance**
  - Refactored settings hook implementation
  - Removed deprecated Excerpt Suggestions settings endpoint

- **Stability**
  - Improved component initialization logic to prevent duplicate operations
  - Streamlined state management for settings and component references


src/content-helper/common/settings/provider.tsx
```diff
@@ -59,13 +59,13 @@ const SettingsContext = createContext<SettingsContextType<Settings>>( {
  * // Using the useSettings hook with a specific type
  * const { settings, setSettings } = useSettings<SidebarSettings>();
  */
-export const useSettings = <T = Settings>(): SettingsContextType<T> => {
+export function useSettings<T = Settings>(): SettingsContextType<T> {
```
vaurdan, line 62: Changed this to `function` instead of a arrow function, since the syntax was somehow borking my IDE's syntax highlighting. 

This has the same effect, and I still get pretty colors on this file :)

acicovic (approved):
Thank you!

PR #3178, opened 2025-03-10
Fix issue checking queried object with a false queried object

srtfisher:
This fixes an issue when attempting to browse articles in admin for an author that does not exist. In this situation, this fatal error occurred:

```
Uncaught Error: get_class(): Argument #1 ($object) must be of type object, bool given
in /var/www/wp-content/mu-plugins/wp-parsely-3.17/src/UI/class-admin-bar.php on line 76

Call stack:

get_class(false)
wp-content/mu-plugins/wp-parsely-3.17/src/UI/class-admin-bar.php:76
Parsely\UI\Admin_Bar::admin_bar_parsely_stats_button(WP_Admin_Bar)
wp-includes/class-wp-hook.php:324
WP_Hook::apply_filters(NULL, array)
wp-includes/class-wp-hook.php:348
WP_Hook::do_action(array)
wp-includes/plugin.php:565
do_action_ref_array('admin_bar_menu', array)
wp-includes/admin-bar.php:95
wp_admin_bar_render('')
wp-includes/class-wp-hook.php:324
WP_Hook::apply_filters('', array)
wp-includes/class-wp-hook.php:348
WP_Hook::do_action(array)
wp-includes/plugin.php:517
do_action('in_admin_header')
wp-admin/admin-header.php:267
require_once('/var/www/wp-admin/admin-header.php')
wp-admin/edit.php:411
```

~This is because the queried object here is a `stdClass`. We can prevent this fatal error by simplifying the check.~

Actually, the queried object here is `false` because this author does not exist which throws this fatal error. I was able to prevent the fatal by removing the hooked method.

## Summary by CodeRabbit

- **Refactor**
	- Streamlined admin interface logic for improved reliability and consistency in feature display.

### DIFF
--- a/src/UI/class-admin-bar.php
+++ b/src/UI/class-admin-bar.php
@@ -73,7 +73,7 @@ final class Admin_Bar {
 		 */
 		$queried_object = $GLOBALS['wp_the_query']->get_queried_object();
 
-		if ( null === $queried_object || WP_Post::class !== get_class( $queried_object ) ) {
+		if ( ! $queried_object || ! $$queried_object instanceof WP_Post::class ) {
 			return;
 		}
 
--- a/src/UI/class-admin-bar.php
+++ b/src/UI/class-admin-bar.php
@@ -73,7 +73,7 @@ final class Admin_Bar {
 		 */
 		$queried_object = $GLOBALS['wp_the_query']->get_queried_object();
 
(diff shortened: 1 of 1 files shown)


vaurdan (approved):
Hey @srtfisher, thank you for submitting this fix! 🙂 We have actually identified this issue last week but haven't yet worked on the fix, so I appreciate your PR.

I have approved it, and I'm going to merge it.

PR #3182, opened 2025-03-12
Traffic Boost: Implement generating suggestions in batches

vaurdan:
## Description

In order to speed up how long it takes to get the first generated suggestions, this PR adds a new batch request mechanism to the Traffic Boost suggestions generation, that allows to keep asking for suggesting until a specific number of suggestions is reached (10). 

This also implements and makes use of the new `url_exclusion_list` and `keyword_exclusion_list` parameters of the Suggestions API.

The pagination in the suggestions tab has been replaced with a scroll bar, which makes the UI feel cleaner as well. For inbound smart links, pagination is still being used.

A few other minor issues and bugs have been fixed.

## Motivation and context
* Improve the perceived speed of the Traffic Boost feature.

## How has this been tested?
* Tested locally on my local environment. 

## Summary by CodeRabbit

- **New Features**
  - Introduced an engaging loading indicator with a typewriter animation and spinner, providing clearer, consistent feedback.
  - Enhanced link suggestion generation with batch processing and refined keyword filtering for more relevant recommendations.
  - Added configurable scrolling and pagination in the links list for smoother navigation.
  - Implemented management of canonical URLs for posts, improving link context and SEO.
  - Added a mechanism to ignore specific keywords during link regeneration, refining the process.
  - Expanded API functionality with new parameters for URL and keyword exclusions, allowing for more customizable suggestions.
  - Introduced validation for blending weight parameters, ensuring they meet defined criteria.

- **Style**
  - Improved visual layouts in dashboard headers and sidebar footers for better spacing and overall interface consistency.
  - Added new styles for loading components, enhancing their visual presentation.


src/services/suggestions-api/class-suggestions-api-service.php
```diff
@@ -138,15 +138,16 @@ public function get_inbound_links( \WP_Post $post, $options = array() ) {
...
 	 *                                                            object if the response is an error.
 	 */
-	public function get_inbound_link_positions( \WP_Post $source_post, \WP_Post $destination_post ) {
+	public function get_inbound_link_positions( \WP_Post $source_post, \WP_Post $destination_post, $options = array() ) {
 		/** @var Endpoints\Endpoint_Suggest_Inbound_Link_Positions $endpoint */
 		$endpoint = $this->get_endpoint( '/suggest-inbound-link-positions' );
 
-		return $endpoint->get_inbound_link_positions( $source_post, $destination_post );
+		return $endpoint->get_inbound_link_positions( $source_post, $destination_post, $options );
 	}
```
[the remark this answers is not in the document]
vaurdan: It isn't possible to use type hint in the function signature as these are PHPStan types.

src/Models/class-inbound-smart-link.php
```diff
@@ -196,10 +219,23 @@ private function get_post_data(): array {
...
+				'paragraph'             => '',
+				'is_first_paragraph'    => false,
+				'is_last_paragraph'     => false,
+				'paragraph_offset'      => 0,
+				'permalink'             => '',
+				'parsely_canonical_url' => '',
+				'edit_link'             => '',
+				'author'                => '',
+				'date'                  => '',
+				'image'                 => '',
```
acicovic: This array seems identical to the one also returned below? If yes, we could have that array in a variable and return that.

src/Models/class-inbound-smart-link.php
```diff
@@ -196,10 +219,23 @@ private function get_post_data(): array {
...
-				'paragraph'          => '',
-				'is_first_paragraph' => false,
-				'is_last_paragraph'  => false,
-				'paragraph_offset'   => 0,
+				'id'                    => 0,
+				'title'                 => '',
+				'type'                  => array(
+					'name'  => '',
+					'label' => '',
+					'rest'  => '',
```
acicovic: Is this supposed to hold an endpoint path?

src/class-parsely.php
```diff
@@ -971,6 +972,77 @@ private function set_managed_options(): void {
...
+	 * 
+	 * @since 3.18.0
+	 *
+	 * @param WP_Post|int $post The post object or post ID.
+	 * @param string      $url The canonical URL.
+	 * @return bool True if the canonical URL was set, false otherwise.
+	 */
+	public static function set_canonical_url( $post, string $url ): bool {
+		$post_id = is_int( $post ) ? $post : $post->ID;
+		return update_post_meta( $post_id, self::PARSELY_CANONICAL_URL_META_KEY, $url ) !== false;
```
acicovic: We could be using a Yoda condition here.

src/content-helper/common/components/loading/component.tsx
```diff
@@ -0,0 +1,132 @@
...
+}: LoadingProps ): JSX.Element => {
+	const [ currentMessage, setCurrentMessage ] = useState<string>( '' );
+	const [ currentIndex, setCurrentIndex ] = useState( 0 );
+
+	useEffect( () => {
+		if ( ! loading ) {
+			return;
+		}
+
+		if ( currentMessage === '' ) {
```
acicovic: We could be using a Yoda condition here.

src/content-helper/dashboard-page/pages/traffic-boost/preview/preview.tsx
```diff
@@ -463,30 +466,47 @@ export const TrafficBoostPreview = ( {
...
+
 		try {
 			const updatedLink = await TrafficBoostProvider.getInstance().generateSuggestionForPost(
 				post,
 				activeLink.targetPost,
 				activeLink,
-				[ oldSmartLink?.text ?? '' ],
+				{
+					ignoreKeywords: [ ...ignoredKeywords, oldSmartLink?.text ?? '' ],
+					performanceBlendingWeight: 0.5,
```
acicovic: I'm seeing a couple of places where we set this to `0.5`. We could maybe have a constant for this if there's a suitable place for it.

acicovic (commented):
I've completed the first half of my review.

src/content-helper/dashboard-page/pages/traffic-boost/sidebar/components/links-list/links-list.tsx
```diff
@@ -105,12 +110,49 @@ export const LinksList = ( {
...
+		}
+
+		const activeIndex = links.findIndex( ( link ) => link.uid === activeLink.uid );
+
+		if ( activeIndex !== -1 && itemRefs.current[ activeIndex ] ) {
+			const container = containerRef.current;
+			const activeItem = itemRefs.current[ activeIndex ];
+
+			if ( activeItem ) {
+				if ( activeIndex === 0 ) {
```
acicovic: A couple of places here where we could be using Yoda conditions. Same in other new code within the file. I think we can discuss internally how much we want to use Yoda conditions, so we're all on the same page.

src/content-helper/dashboard-page/pages/traffic-boost/sidebar/components/links-list/single-link.tsx
```diff
@@ -2,6 +2,8 @@
  * WordPress Dependencies
  */
 import { __, sprintf } from '@wordpress/i18n';
+import { forwardRef } from '@wordpress/element';
+import type { ForwardedRef } from 'react';
```
acicovic: Should this be in an external deps section?

src/content-helper/dashboard-page/pages/traffic-boost/sidebar/components/links-list/single-link.tsx
```diff
@@ -21,17 +23,18 @@ type SingleLinkProps = {
...
 /**
- * Displays a single Traffic Boost link.
+ * The SingleLink component, before being forwarded.
  *
  * @since 3.18.0
  *
- * @param {SingleLinkProps} props The component's props.
+ * @param {SingleLinkProps}              props The component's props.
+ * @param {ForwardedRef<HTMLDivElement>} ref   The forwarded ref.
+ * @return {JSX.Element} The SingleLink component.
```
acicovic: Per convention, we haven't been using `@return {JSX.Element}` in DocBlocks. I'm seeing this in a couple of places within this file and `single-link.tsx`.

src/content-helper/dashboard-page/pages/traffic-boost/single-post-component.tsx
```diff
@@ -180,6 +187,17 @@ export const TrafficBoostPostPage = (): React.JSX.Element => {
 		};
 	}, [ setCurrentPost, setSelectedLink ] );
 
+	/**
+	 * Hides the preview if there are no suggestions.
+	 *
+	 * @since 3.18.0
+	 */
+	useEffect( () => {
+		if ( selectedLink?.isSuggestion && suggestions.length === 0 ) {
```
acicovic: Yoda condition?

src/content-helper/dashboard-page/pages/traffic-boost/store.ts
```diff
@@ -157,16 +157,6 @@ interface SetInboundLinksPageAction {
 	page: number;
 }
 
-/**
- * Interface for the SetSuggestionsItemsPerPageAction.
- *
- * @since 3.18.0
- */
-interface SetSuggestionsItemsPerPageAction {
```
acicovic, line 165: It seems we're removing page actions from here, but we still support pagination in `link-list.tsx` with `showPagination`?

These might be unrelated, just wanted to flag it.

src/rest-api/content-helper/class-endpoint-traffic-boost.php
```diff
@@ -120,16 +134,40 @@ public function register_routes(): void {
...
 					'description' => __( 'The keywords to ignore.', 'wp-parsely' ),
 					'required'    => false,
 				),
+				'keyword_exclusion_list'      => array(
+					'type'        => 'array',
+					'description' => __( 'The keywords to exclude from the suggestions.', 'wp-parsely' ),
+					'required'    => false,
+					'default'     => array(),
+				),
+				'performance_blending_weight' => array(
```
acicovic, line 154: `performance_blending_weight` can be a float between 0 and 1 (including these values). Are we validating/sanitizing this anywhere?

src/rest-api/class-base-validation.php
```diff
@@ -0,0 +1,46 @@
...
+	 * @since 3.18.0
+	 *
+	 * @param mixed           $value   The value to sanitize.
+	 * @param WP_REST_Request $request The request object.
+	 * @param string          $param   The parameter name.
+	 * @return mixed The sanitized value.
+	 */
+	public static function sanitize( $value, $request, $param ) { // phpcs:ignore Generic.CodeAnalysis.UnusedFunctionParameter.FoundAfterLastUsed
+		return $value;
+	}
```
acicovic: This method should be declared abstract or throw an exception, forcing child classes to implement it. Else we risk returning unsanitized values if we forget to implement it in child classes.

acicovic (commented):
I've completed my review, and pushed a couple of small nitpicking commits with spacing/comment changes.

src/Models/class-inbound-smart-link.php
vaurdan: Nope, if it's empty, no post object was found, so the rest endpoint is really unnecessary.

src/content-helper/dashboard-page/pages/traffic-boost/sidebar/components/links-list/links-list.tsx
vaurdan: I believe that the code standards recommend yoda conditions every time that there's a risk of assigning a value to a variable by mistake. So `activeIndex === 0` would need yoda condition, while `getActiveIndex() === 0` would not. ([Docs](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#yoda-conditions))

However, in this situation, it's definitely missing a yoda condition 😄

src/content-helper/dashboard-page/pages/traffic-boost/store.ts
vaurdan: Yes, this is intended, but thank you for double checking!

This specific setting was related to the suggestions tabs pagination, which is no longer needed. However, the inbound link tab - for example - still uses pagination and has its own set of selectors and actions to manage the pagination (`SET_INBOUND_LINKS_PAGE` and `SET_INBOUND_LINKS_ITEMS_PER_PAGE`).

src/content-helper/dashboard-page/pages/traffic-boost/sidebar/components/links-list/links-list.tsx
acicovic: Makes sense, let's go with that.

src/rest-api/class-base-validation.php
vaurdan: This was intentional, since there are certain situations where we just want to validate, while no sanitization is needed. A good example is the blending weight. We can validate that is is a float between 0 and 1, but we don't actually need to sanitize, because it has been already validated that it's a float. 

Since the sanitization callback has to be explicitly included in the REST API endpoint parameters, I think it should be pretty difficult to forget about its implementation, if we need to use it:

```php
 'url_exclusion_list'          => array(
	'type'              => 'array',
	'description'       => __( 'The URLs to exclude from the suggestions.', 'wp-parsely' ),
	'required'          => false,
	'default'           => array(),
	'validate_callback' => array( Validations\Validate_Url_Exclusion_List::class, 'validate' ),
	'sanitize_callback' => array( Validations\Validate_Url_Exclusion_List::class, 'sanitize' ),
),
```

I think maybe a good middle ground is to implement the function to throw an exception when it is used but not implemented, as you also suggested. 

Edit: Added an exception to the base `sanitize()` method 🙂

src/rest-api/content-helper/class-endpoint-traffic-boost.php
vaurdan: Not really! I have created a new validation class for the blending weight.

src/rest-api/class-base-validation.php
acicovic: Regardless of the mechanism we pick, we definitely need some protection there. I think that I've seen this error happen, or I may have even done it myself. I'm totally fine with an exception or with children classes returning the unsanitized value, but I'm definitely scared of returning an unsanitized value from a parent function that's called `sanitize()`.

src/Models/class-inbound-smart-link.php
acicovic: Sorry, my question was unclear. I was asking if `'rest'` is supposed to hold a REST point path, but I can see from the code that this is the case.

src/rest-api/class-base-validation.php
vaurdan: Yup, agreed. Added the exception [here](https://github.com/Parsely/wp-parsely/blob/bea1e811bdd301641e3b62e63cb52282067609d8/src/rest-api/class-base-validation.php#L48).

acicovic (approved):
Thank you for working on this! Looks good, I've made an [additional commit](https://github.com/Parsely/wp-parsely/pull/3182/commits/ba863376cea32877bd99cc6a07b8b99ced672c3c) to use the blending weight constant in preview.tsx. This is ready to merge!

PR #3183, opened 2025-03-12
Move plugin Settings page under Parse.ly menu

acicovic:
## Description
With this PR, we're moving the plugin's Settings page under the Parse.ly menu, removing the Parse.ly page from the Settings menu.

The settings page is still accessible using its URL and slug. In other words, trying to access `/wp-admin/options-general.php?page=parsely` will still work. Although we could try to block this, I don't think we need to. @vaurdan, let me know if you have a different opinion.

The remaining issue to tackle is that success/error messages don't show in the page under its new home.

## Motivation and context
Keep all things Parse.ly in a single place.

## How has this been tested?
Manual testing, existing tests pass.

## Screenshots
| Parse.ly Menu | General Menu |
|--------|--------|
| image | image |

### DIFF
--- a/src/UI/class-dashboard-page.php
+++ b/src/UI/class-dashboard-page.php
@@ -180,7 +180,7 @@ final class Dashboard_Page {
 			'Parse.ly Settings',
 			'Settings',
 			Parsely::CAPABILITY, // phpcs:ignore WordPress.WP.Capabilities.Undetermined
-			'parsely-dashboard-page#/settings',
+			Parsely::MENU_SLUG,
 			'__return_null'
 		);
 	}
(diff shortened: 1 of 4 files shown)


vaurdan:
> The remaining issue to tackle is that success/error messages don't show in the page under its new home.

I have been looking into this, and it seems that on pages under `Settings` the settings notices are always displayed. Since we're moving it outside `Settings` it must be manually rendered with `<?php settings_errors(); ?>`.

While I was investigating this, I came across a different solution than yours: instead of removing the menu from the original position, I am registering it under the new Parse.ly menu item instead. I pushed this to a branch so you can take a look if you prefer that approach. Here's the commit: https://github.com/Parsely/wp-parsely/commit/866be8ac668ef8a629ff3232aa113bff34a8cbd4

As for the notices not showing up, you can see the fix in [the settings-page.php file here](https://github.com/Parsely/wp-parsely/commit/866be8ac668ef8a629ff3232aa113bff34a8cbd4#diff-d8812f00b29c22dcec2f473cf7feb1d99682aa9749dde7605d90254b45ca072cR41).

acicovic:
I think we can close this PR and go with your solution, it's probably cleaner as it doesn't need the removal. What do you think?

vaurdan:
Closing in favor of #3186

PR #3187, opened 2025-03-13
Traffic Boost: Add "Boost Traffic" action to the post list

vaurdan:
## Description
This PR adds a new action to the WordPress post list to link directly to Traffic Boost. It is a link with "Boost Traffic" string, and the Parse.ly logo alongside it. 

The link will not be displayed if the user does not have Traffic Boost access or if the post is not published (`draft`, `scheduled`, or `private`).

## Motivation and context
Give Traffic Boost more visibility in the WordPress interface.

## How has this been tested?
Tested locally.

image

## Summary by CodeRabbit

- **New Features**
  - Introduced a new Traffic Boost action link in the row actions for posts and pages. The link is conditionally displayed based on user permissions and the post's status, providing eligible users with quick access to boost traffic.


acicovic:
@dabowman, what's your opinion on the leaf icon? Should we have it? How about for the "Parse.ly Stats" item?

src/UI/class-row-actions.php
```diff
@@ -117,4 +121,57 @@ private function generate_aria_label_for_post( WP_Post $post ): string {
 			$post->post_title
 		);
 	}
+
+	/**
+	 * Adds a link to Traffic Boost in the row actions.
+	 * 
+	 * If the user doesn't have permission to use Traffic Boost, or if the post is a draft, private, or scheduled,
```
acicovic: Let's wrap the function description to column 80.

src/UI/class-row-actions.php
```diff
@@ -117,4 +121,57 @@ private function generate_aria_label_for_post( WP_Post $post ): string {
...
+	 * @param WP_Post               $post    The individual post object the actions apply to.
+	 * @return array<string, string> The amended list of actions.
+	 */
+	public function row_actions_add_traffic_boost_link( array $actions, WP_Post $post ): array {
+		if ( ! Permissions::current_user_can_use_pch_feature( 'traffic_boost', $this->parsely->get_options()['content_helper'] ) ) {
+			return $actions;
+		}
+
+		// Disable for posts that are draft, private, future, pending, or trash.
+		$non_publishable_statuses = array( 'draft', 'private', 'future', 'pending', 'trash' );
```
acicovic: Should `Parsely::post_has_trackable_status()` be taken into account or used here?

vaurdan: I thought about this, but I think that a specific post type might not be tracked but we still might want to boost it either way. That's why I defaulted to hide the boost link to any non-public post status. What do you think?

acicovic: Interesting thoughts, thanks for the explanation and I think it makes sense from that perspective. I mildly remember a case where `future` posts were published for a customer though? Does it ring a bell?

In this case, maybe we could check what you're checking, but also check `Parsely::post_has_trackable_status()` just in case it returns true? I haven't thought of that thoroughly, so my thinking might have a flaw here.

vaurdan: Yes, that should work too! I added that and committed just now. Feel free to take a look.

acicovic (approved):
Thank you!

PR #3210, opened 2025-03-18
PCH: Add category slug support in search

### LINKED ISSUE #2968: PCH Related Posts: Support slugs in search
A customer reported not getting results by section in PCH Related Posts.

Digging deeper, we found that the customer's metadata is registering sections with the category's `slug`, not the `name` field which we currently use to fetch results.

There are a few possible solutions:
1. If possible, query with both `name` and `slug` in one call, and show the results. There's a slight possibility of getting some unwanted results.
2. Perform the `name` query as usual, and if no results are returned then query for `slug`. There's a slighter chance of showing unwanted results, but we use 2 calls and latency increases for results by `slug` and when there are no results at all. 
3. Provide a setting so customers can set whether the query should be made by `name` or `slug`. We could also introduce an option to query both if we think it could be useful, especially if it's possible with one call. This solution eliminates the possibility of showing potentially unwanted results, but it's considerably more work for something that isn't currently a widespread need.

acicovic:
## Description
With this PR, we're introducing a `wp_parsely_use_category_slugs_in_searches` filter. Setting it to `true` (strict bool, any other value will be considered as `false`) will force PCH's Related Posts search to use the category's slug instead of its name.

The filter name is purposefully generic since its use could be expanded to tools outside of PCH Related Posts, for any site setups requiring this configuration.

Edit: Initially, the filter name was `wp_parsely_use_section_slugs_in_searches`, but it was renamed to `wp_parsely_use_category_slugs_in_searches` after @vaurdan's feedback.

## Motivation and context
- Addresses customer feedback to unblock usage of PCH Related Posts.
- Resolves https://github.com/Parsely/wp-parsely/issues/2968.

## How has this been tested?
- Manually tested locally, using settings that emulate the customer's environment. Verified that searching by section works as expected when the `wp_parsely_use_category_slugs_in_searches` filter is set to `true` or `false`, as well as invalid values such as strings, numbers, `null` and `undefined`.
- Existing tests pass.

## Summary by CodeRabbit

- **New Features**
	- Introduced a configurable option that adapts how post category is interpreted during searches, giving users a more tailored search experience.
	- Enhanced category handling in the editor sidebar and related content filters by supplying richer category details, which improves the precision of post categorization and display.


vaurdan (approved):
This looks good to me! Approved.

Regarding the naming of the filter. `wp_parsely_use_section_slugs_in_searches` sounds good to me, however conceptually I'm not sure if we should call it `use_category_slugs` instead, considering that the actual slug comes from the WordPress category, that is matched with the Parse.ly section name. What do you think?

acicovic:
Good point, I'll give it a thought. We probably are using WP terminology internally, and in this case it would make sense to go with `category`.

acicovic:
@vaurdan, I've updated the filter's name per your feedback, and will be merging soon. You can take a look at the changes in https://github.com/Parsely/wp-parsely/pull/3210/commits/f01217e621f252e1480a76c449dc5921ff0d261d if you want.

PR #3213, opened 2025-03-20
Smart Links: Change data structure to use Status taxonomy

vaurdan:
## Description
This PR changes the data structure of the Smart Link, to use a new taxonomy - Smart Link Status - to store the status of a given smart link - either applied or pending.

This is considerable more performant than relying on post meta, since we need to do queries using the smart link status (get all inbound that are pending, get all outbound that are applied, etc).

I have also introduced a function in `wp-parsely.php` that handles data schema updates. In this case it is necessary to prevent breaking changes for a couple reasons:

 * For sites running on Parse.ly 3.17.0 or lower: the Smart Linking feature does not add the `_smart_link_applied` post meta to applied smart links. Therefore, all the smart links that are added to the database at this version are considered applied;
 * For sites running the beta 3.18.0, they are using `_smart_link_applied` to detect if the link is applied (`true` if so, or pending if `false`).

This update callback checks for Smart Links that do not have the `_smart_link_applied` post meta, and if not, add the `accepted` term. If it does have the post meta, it will add the term depending on the meta value. 

The callback only runs once, and only when `PARSELY_DATA_SCHEMA_VERSION` is higher than the current version stored in the `parsely_data_schema_version` option.

## Motivation and context
* Improve performance of Smart Links queries, and makes querying smart links easier overall.
* Add foundation for supporting potential new status.

## How has this been tested?
Tested locally in a few different scenarios:
 * Tested with data from Parse.ly 3.17.0 - data schema was successfully upgraded and existing smart links keep functioning.
 * Tested with data from Parse.ly 3.18.0 beta - data schema was successfully upgraded, existing smart links and traffic boost links were updated to have the new term and the old post meta was removed.
 * Tested and validated that the callback only runs once.

The existing tests also pass.

## Summary by CodeRabbit

- **New Features**
  - Introduced a flexible status system for smart links, offering clear distinctions between states.
  - Added taxonomy support to enhance content tracking and organization.
  - Implemented a method for checking and updating smart link statuses during admin actions.

- **Refactor**
  - Centralized status handling across smart linking functionalities for improved performance and reliability.

- **Chores**
  - Updated the data schema version to streamline migration and ensure smoother processing.


src/Models/class-smart-link.php
```diff
@@ -457,6 +465,68 @@ public function get_link_href( $skip_utm_params = false ): string {
...
+	 *
+	 * @since 3.18.0
+	 *
+	 * @param string $status The status to set.
+	 * @param bool   $save Whether to save the status to the database.
+	 * @throws \InvalidArgumentException If the status is invalid.
+	 */
+	public function set_status( string $status, bool $save = false ): void {
+		if ( ! Smart_Link_Status::is_valid_status( $status ) ) {
+			throw new \InvalidArgumentException( 'Invalid status' );
```
acicovic, line 520: If this is shown to the user, we probably want it to be translatable?

wp-parsely.php
```diff
@@ -227,3 +228,92 @@ function parsely_integrations( $parsely = null ): Integrations {
 
 	return $parsely_integrations;
 }
+
+add_action( 'admin_init', __NAMESPACE__ . '\\parsely_check_data_scheme_updates', 999 );
```
acicovic: I see both scheme and schema being used here. Schema is probably what we want to use?

acicovic (commented):
Thanks for working on this. I've left a couple of minor comments, but also have 2 questions:

1. Did you think to rely on plugin version for the update procedure, and if yes, what made you go with a new option instead?
2. We should think the scenario where the plugin is downgraded. What are your thoughts regarding that?

src/Models/class-smart-link.php
vaurdan: I didn't translate this, as if this error is thrown, it's probably an implementation error, so the end-user shouldn't really never see it. 

But I can add a translation if you think is worth it for consistency.

wp-parsely.php
vaurdan: Ahh good catch! FIxing it.

src/Models/class-smart-link.php
acicovic: In this case, works for me as it is.

vaurdan:
@acicovic to answer your questions:

> Did you think to rely on plugin version for the update procedure, and if yes, what made you go with a new option instead?

That was my initial plan, indeed. However, relying on the plugin version can be tricky in situations where the plugin is downloaded directly through GitHub, for example, where we only bump the version at release. This could be problematic for 3.18.0 beta since it actual version still is 3.17.0, so the schema update would not trigger. And if we need to change the schema in between development builds, this can also become a problem.

That's why I added that new schema version option, so we can track the data layer version separately from the plugin version. 

> We should think the scenario where the plugin is downgraded. What are your thoughts regarding that?

That's a pretty good question, and TBH I haven't considered that scenario. Right now, it isn't really possible, as AFAIK, it's not possible to trigger specific downgrade steps, since those steps would have need to be defined in the version where it is being downgraded to. 

Right now, from my testing (I had to update and downgrade a bunch of times to make sure this was working), what might happen when downgrading is that we lose the data if links have been applied. This means that the queries often return applied smart links as suggestions, and the UI can glitch out here and there. Not great.

I think an easy way to make it backwards compatible, in case of a downgrade, is to not remove any of the existing `_smart_link_applied` post meta. So that, at the very least, the old plugin version still has that meta field where it can rely to understand if a link has been applied. What do you think?

acicovic:
Thanks for explaining about the schema, this makes sense and it's more flexible if we happen to need it.

Regarding the downgrade path, if someone was to apply Smart Links in v3.18 and then go back to v3.17, would this also work for these Smart Links, or only for those who were applied before the upgrade?

vaurdan:
If someone downgrades from this version to 3.17.0, I think the only side effect is that all the existing smart links - even pending ones (suggestions) would be treated as applied, since the meta doesn't yet exists on that version, and each Smart Link in the database is actually an applied Smart Link.

However, I'm pretty sure this would be auto-fixed by the Smart Linking feature, as each time there's a save, it parses all the existing Smart Links in the current article, saves those and discards the remaining. So those "applied" - but actually pending - smart links would get removed from the database without the user even realizing. 

I will change the PR so it doesn't remove the `_smart_link_applied` meta tag when running the schema update. I think as far as backwards compatibility with downgrades, is the best we can do.

acicovic:
Sounds good, we can probably remove the meta tag in the next major/breaking release and flag it as breaking.

BTW, an alternative approach would be to have a 3.17.1 patch, which executes the inverse procedure (if that is possible). But I think that just keeping the meta is better in this case.

The reason for this discussion is that I think that breaking or non-reversible operations that could bring breakage upon downgrade warrant a major/breaking release. But happy to discuss this further if we want.

acicovic (approved):

Issue #2938, opened 2024-11-12
Improve the GitHub actions workflow for releasing


vaurdan:
## Describe the bug
After the 3.17 release - the first time using our new workflow - we've identified a few minor issues that should be addressed before the next release:

 * Generated PR has an empty space before the first line of the changelog (` ### Added`) 
 * Generated PR hasn't a set milesone
 * GitHub action bot not signing commits / PRs (we might need our own bot user for that)
 * When running with dry-run, if the tag already exists it is deleted locally, but when pushing the tag it fails since it already exists on the remote.
 * GitHub release is not published automatically when the zip file is uploaded
 
## Nice to have / Improvements
 * When running the release workflow without dry-run set, it might optionally attempt to remove any old release and tag created by the dry-run execution.
 * When the `Update version number and changelog for x.x.x release` PR is merged, maybe trigger automatically an action that creates a `Release wp-parsely x.x.x` PR that merges `develop` into `trunk`.
 * After the release, automatically create a `Merge trunk into develop after the wp-parsely x.x.x release` PR.

acicovic:
After releasing v3.18, here are my comments:

- We still have a couple of spaces before `## Added` (can be seen [here](https://github.com/Parsely/wp-parsely/pull/3280) when editing the PR description).
- I can confirm that the GitHub release is being published as a draft even for non-dry runs, and that you need to go to the release and publish it manually.

PR #3250, opened 2025-04-10
Improve release workflow

alecgeatches:
## Description

Addresses the release workflow issues outlined in https://github.com/Parsely/wp-parsely/issues/2938:

- [x] Generated PR has an empty space before the first line of the changelog (` ### Added`) 
- [x] Generated PR hasn't a set milesone
- [ ] GitHub action bot not signing commits / PRs (we might need our own bot user for that)
- [x] When running with dry-run, if the tag already exists it is deleted locally, but when pushing the tag it fails since it already exists on the remote.
- [ ] GitHub release is not published automatically when the zip file is uploaded
 
### Nice to have / Improvements

Per discussion, we're skipping these items on this PR.

- [ ] ~When running the release workflow without dry-run set, it might optionally attempt to remove any old release and tag created by the dry-run execution.~
- [ ] ~When the `Update version number and changelog for x.x.x release` PR is merged, maybe trigger automatically an action that creates a `Release wp-parsely x.x.x` PR that merges `develop` into `trunk`.~
- [ ] ~After the release, automatically create a `Merge trunk into develop after the wp-parsely x.x.x release` PR.~

## Summary by CodeRabbit

- **Chores**
  - Enhanced release log formatting for a clearer presentation of changes.
  - Improved the automated release process by associating version milestones and ensuring robust tag management for smoother release operations.
  - Updated sidebar behavior to better handle cases when content is unavailable.


alecgeatches:
@acicovic Howdy! This is ready for review. For now, I'm focusing on the top 5 items and ignoring the nice-to-haves. There are a couple that I haven't completed that I'd like feedback on:

>  GitHub action bot not signing commits / PRs (we might need our own bot user for that)

I found a couple of ways to accomplish this:

1. Using `GITHUB_TOKEN` and manual API requests, see [Solution 1 here](https://github.com/Nautilus-Cyberneering/pygithub/blob/main/docs/how_to_sign_automatic_commits_in_github_actions.md) and the examples [on this post](https://192dot.medium.com/sign-commit-using-github-actions-app-13488f6e76b7). This uses some wacky stuff with GitHub's API directly to [commit files over HTTP](https://github.com/lichao127/actions-pull-request/blob/main/.github/workflows/bot-update.yml#L79-L91). This seems unusual and different enough from rest of our git-managing process to probably not be a good solution.
2. Using a bot, as originally suggested. This is also covered in [Solution 2 on this page](https://github.com/Nautilus-Cyberneering/pygithub/blob/main/docs/how_to_sign_automatic_commits_in_github_actions.md). The steps here would be:
    - Create a bot account and store credentials
    - Generate a PGP signature for that account and store the secrets in the repository
    - [Import the PGP signature](https://github.com/marketplace/actions/import-gpg#sign-commits) in our actions

I haven't implemented either yet, but if we want to go ahead with this I think a bot account makes the most sense. Does that sound good to you, or do you think we should keep things as-is?

> GitHub release is not published automatically when the zip file is uploaded

We use [`ncipollo/release-action@v1`](https://github.com/ncipollo/release-action) in [our release workflow](https://github.com/Parsely/wp-parsely/blob/develop/.github/workflows/release-plugin.yml#L238-L251). From my understanding of the documentation, `draft: ${{ env.DRY_RUN }}` *should* mean that we make a published release when we're running the action outside of a dry run. Do you have any insight on how this wasn't working during the last release?

src/content-helper/editor-sidebar/class-editor-sidebar.php
```diff
@@ -106,11 +106,11 @@ private function get_parsely_post_url( $post_id = null, bool $add_utm_params = t
 		/**
 		 * The post object.
 		 *
-		 * @var WP_Post $post
+		 * @var WP_Post|null $post
 		 */
 		$post = get_post( $post_id );
 
-		if ( ! Dashboard_Link::can_show_link( $post, $this->parsely ) ) {
```
alecgeatches, line 113: This fixes an unrelated test failure due to a type issue I've seen a couple of times [[1]](https://github.com/Parsely/wp-parsely/actions/runs/14390004182/job/40354625985) [[2]](https://github.com/Parsely/wp-parsely/actions/runs/14452501906/job/40528286468#step:12:595).

acicovic:
@alecgeatches, thank you for your work.

I'll share with you internally all the context I've got about the workflow run. I'm unsure it will shed light to your questions though. To be more specific to the questions at hand:

>  I think a bot account makes the most sense.

I agree.

> Do you have any insight on how this wasn't working during the last release?

Unfortunately I don't, and I'm unsure what would cause this. I can see that `DRY_RUN` is evaluated as a string and as a bool in different places, though this might be expected and not the source of the issue.

To conclude, from my understanding none of these are blocking. Thus, we could make our next release without these and see if any of them cause enough friction for us to consider fixing them for v3.19 if there's bandwidth.

acicovic (approved):

alecgeatches:
Thank you @acicovic! I'll commit as-is for now, and we can evaluate if we want the bot changes and understand the published release problem again.

PR #3275, opened 2025-04-15
GitHub Workflows: Update Ubuntu to v24.04 LTS

acicovic:
## Description
I mistakenly thought GitHub Ubuntu v20.04 LTS runner images were being phased on in the end of April, but it turns out this is done today.

This PR updates all our runner images to Ubuntu 24.04. I also opted to substitute `ubuntu-latest` with `ubuntu-24.04` to have a specific version lock-in.

## Motivation and context
Some of our tests started failing, and I discovered that we need to update our runner images now.

## How has this been tested?
As a first step, all our testing workflows should pass. I expect this might impact our release workflows, which can be fixed in another PR.

## Summary by CodeRabbit

- **Chores**
	- Updated all GitHub Actions workflows to use the Ubuntu 24.04 runner environment.
	- Added a step to install Subversion (svn) in the integration tests workflow.

### DIFF
--- a/.github/workflows/build-plugin.yml
+++ b/.github/workflows/build-plugin.yml
@@ -13,7 +13,7 @@ concurrency:
 jobs:
   build:
     name: Build and Commit
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     permissions:
       contents: write
       pull-requests: write
--- a/.github/workflows/bump-version.yml
+++ b/.github/workflows/bump-version.yml
@@ -15,7 +15,7 @@ env:
 jobs:
   validate_version:
     name: Validate the new version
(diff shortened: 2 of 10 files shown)


acicovic:
OK, it seems we'll need to make the extra work now, since the `svn` command is missing.

acicovic:
OK, thanks to @GaryJones, I found this snippet:

```yaml
      - name: Install Subversion
        run: |
          sudo apt-get update
          sudo apt-get install subversion
```

@alecgeatches, FYI this is all we need to add to make subversion work in any other workflows that might fail due to the `svn` command being absent.

alecgeatches (approved):
Thank you for figuring this out!

PR #3293, opened 2025-04-23
Add traffic boost post list stats

alecgeatches:
## Description

Adds campaign tracking for inserted links in Traffic Boost, and displays those results in the UI:

Screenshot 2025-04-23 at 2 40 15 PM

### Smart Link Insertion

Verify links inserted by Traffic Boost have the correct campaign information:

1. Go to WordPress Admin -> Parse.ly menu -> Traffic Boost.
2. Click the "Boost Traffic" button next to a post.
3. Wait for analysis and see a link suggestion. Note the page's title, then click "Accept" on a link suggestion.
4. After accepting the suggestion, open the "Inbound Links" page and select the accepted insertion page. Next, click the "..." symbol on the top right and open the post with the inserted link:

    Screenshot 2025-05-07 at 1 56 34 PM

5. In the modified post, find the inserted link and view the URL. It should look something like this:

    ```
    https://mysite.vipdev.lndo.site/blog/website-downtime-cost/?itm_campaign=wp-parsely&itm_medium=smart-link&itm_term=3fbbdf42ebaab27f68f089bc9ed0ae46&itm_source=traffic-boost
    ```

6. Ensure `itm_campaign=wp-parsely`, `itm_medium=smart-link`, `itm_source=traffic-boost`, and `itm_term` has a value.

### Traffic Boost Results

"Results" here refer to the new UI displaying impact on a boosted page:

Screenshot 2025-05-07 at 2 04 22 PM

This data is provided by Parse.ly directly, so it's difficult to test locally. However, we have a couple of methods we can use to display real data from Parse.ly:

### Test recently-inserted link

For testing purposes, a smart link was inserted into real post data from "[How to Ensure Good Governance for WordPress Block Themes](https://wpvip.com/blog/block-governance/)" to "[Using a Design System with the WordPress Block Editor: Block Types & Styles](https://wpvip.com/blog/using-a-design-system-with-wordpress-block-types-styles/?itm_campaign=wp-parsely&itm_medium=smart-link&itm_term=b6c0844754a216794c5900c621bd69b3&itm_source=traffic-boost)".

The real data is shown in the screenshot above. Using the same local data, you should be able to find these results by searching "theme.json" in the Traffic Boost search bar as shown above.

You can also increment the number by clicking the "Block Editor" smart link inserted in "[How to Ensure Good Governance for WordPress Block Themes](https://wpvip.com/blog/block-governance/)":

image

Next, wait for about 5 minutes for the [local cache to clear](https://github.com/Parsely/wp-parsely/blob/f92e49f8d4a34c177d840c39d165d7eb7324247e/src/services/content-api/class-content-api-service.php#L63) and re-search "theme.json". You should see the counter and overall impact increase.

### Historical Traffic Boost Results

The WPVIP story "Pros and Cons of Using GraphQL on WordPress", a recent post, has some historical testing data available with a small change:

image

This data is over 30 days old, so in order to view it we need to modify $request_params in [`src/rest-api/stats/class-endpoint-posts.php` [before making the analytics request](https://github.com/Parsely/wp-parsely/blob/f92e49f8d4a34c177d840c39d165d7eb7324247e/src/rest-api/stats/class-endpoint-posts.php#L440):

```diff
    // Fill the URLs with the campaign params.
    /** @var array<string, array<string, mixed>> $request_params_with_campaign */
    $request_params_with_campaign['urls'] = $post_urls;

+   $request_params_with_campaign['period_start'] = '60d';

    /**
     * The raw analytics data, received by the API.
     *
     * @var array<array<string, mixed>>|WP_Error $campaign_request
     */
    $campaign_request = $this->content_api->get_posts( $request_params_with_campaign );
```

After making the above change, you should be able to see some click data for "Pros and Cons of Using GraphQL on WordPress", as displayed above.

## Summary by CodeRabbit

- **New Features**
  - Dashboard posts table now displays 30-day view counts, inbound/outbound smart link counts, and traffic boost suggestion bubbles.
  - Added campaign filtering and WordPress permalink support to stats API endpoints.
  - REST API responses for posts/pages now include canonical URL, smart link metrics, and traffic boost suggestion counts.
  - New stats provider enables fetching detailed Parse.ly metrics per post.

- **Improvements**
  - Enhanced pagination and search handling in dashboard and traffic boost pages.
  - Modularized post row rendering and improved error handling for stats loading.
  - Improved caching for smart link data and post ID lookups for better performance.
  - Added support for Parsely metadata in post data and improved type safety for query parameters.
  - Enhanced error handling and cache invalidation after smart link modifications.
  - Refined campaign metrics extraction and URL processing in stats data.

- **Bug Fixes**
  - Fixed issues with taxonomy queries and cache invalidation for smart links.

- **Style**
  - Refined dashboard table layout and styling for metrics, actions, and compact mode.

- **Chores**
  - Updated and reorganized import paths for improved code structure.
  - Added and updated tests to cover new metadata fields and API parameters.
  - Introduced a global cache group constant for consistent cache management.


src/content-helper/dashboard-page/components/posts-table/components/post-details.tsx
```diff
@@ -0,0 +1,75 @@
...
+ * PostDetails component.
+ *
+ * Represents the post details, the first column in the PostsTable.
+ *
+ * @since 3.18.0
+ *
+ * @param {PostDetailsProps} props The component props.
+ */
+export const PostDetails = ( { post, showSuggestionBubble = true }: PostDetailsProps ): React.JSX.Element => {
+	const prettyDate = format( 'M j, o', post.date ?? '' );
```
acicovic: I think we could be using our `getSmartShortDate()` function here.

src/content-helper/dashboard-page/components/posts-table/components/post-details.tsx
```diff
@@ -0,0 +1,75 @@
...
+	const numberOfSuggestions = post.parsely?.traffic_boost_suggestions_count ?? 0;
+
```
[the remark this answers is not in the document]
acicovic: As stated above, I think we should be using the `getSmartShortDate()` function.

src/content-helper/dashboard-page/components/posts-table/components/suggestion-bubble.tsx
```diff
@@ -0,0 +1,55 @@
...
+	 *
+	 * @since 3.18.0
+	 */
+	const handleClick = () => {
+		navigate( `/traffic-boost/${ postId }` );
+	};
+
+	return (
+		<Tooltip
+			text={ `${ numberOfSuggestions } pending suggestions` }
```
acicovic: Maybe we could use `sprintf()` / `_n()` here to support singular/plural as well as number placement? Happy if there are alternatives, this is what just came to mind first.

acicovic (commented):
Hey @alecgeatches, thank you so much for picking this up and continuing the work. 

I'm halfway through my review and leaving a couple of comments. I'll also be pushing a nitpicking PR (spacing/comments) very shortly.

src/content-helper/dashboard-page/components/posts-table/components/post-details.tsx
alecgeatches: Good suggestion! Fixed along with the `_n()` change in https://github.com/Parsely/wp-parsely/pull/3293/commits/94872bbf72977a54e38e2d869ae1f9f916a724c7.

alecgeatches:
Quick summary of an out-of-Github discussion.

When post crawling was enabled for posts locally, this enabled some new UI:

Screenshot 2025-05-08 at 11 01 23 AM

Most of this isn't fully design baked, so the changes in https://github.com/Parsely/wp-parsely/pull/3293/commits/5001e6439ca457ac8b5283fb9215db710914c4ce either remove (in the case of the "=" next to unchanged stories) or comment-out these components until after the initial Traffic Boost release. We're also reverting to a regular button look. Here's the updated UI:

image

src/Models/class-inbound-smart-link.php
```diff
@@ -1055,6 +1055,52 @@
...
+			return $count;
+		}
+
+		$args = array(
+			'post_type'      => 'parsely_smart_link',
+			'posts_per_page' => 0,
+			'fields'         => 'ids',
+			// phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_tax_query
+			'tax_query'      => array(
+				'relation' => 'AND',
```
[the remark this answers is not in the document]
alecgeatches: The `$query->found_posts` count is the only thing we're using from this query, so a `posts_per_page` of `0` is correct.

src/Models/class-smart-link.php
```diff
@@ -1073,6 +1073,97 @@ function ( Smart_Link $smart_link ) {
...
+			return $link_counts;
+		}
+
+		$base_query_args = array(
+			'post_type'      => 'parsely_smart_link',
+			'posts_per_page' => 0,
+			'fields'         => 'ids',
+			// phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_tax_query
+			'tax_query'      => array(),
+		);
```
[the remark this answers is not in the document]
alecgeatches: As you mention, we rely on `->found_posts`, so returning `0` posts is correct.

src/content-helper/dashboard-page/components/posts-table/components/post-details.tsx
```diff
@@ -0,0 +1,75 @@
...
+	const prettyDate = post.date ? getSmartShortDate( new Date( post.date ) ) : '';
+
+	let postTitle = post.title.rendered;
+
+	// If the title is longer than 80 characters, truncate it and add an ellipsis.
+	if ( postTitle !== '' ) {
+		if ( postTitle.length > 80 ) {
+			postTitle = postTitle.substring( 0, 80 ) + '&hellip;';
+		}
+	}
```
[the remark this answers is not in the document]
alecgeatches: Using `postTitle` (`post.title.rendered`) appears to be okay here. Actual HTML is pre-escaped, and it allows expected entities like `&amp;` to render correctly.

image

src/content-helper/dashboard-page/components/posts-table/component.tsx
```diff
@@ -292,42 +287,30 @@ export const PostsTable = ( {
 				{ ! hideHeader && (
 					<thead>
 						<tr>
-							<th className="post-info-header">{ __( 'POST', 'wp-parsely' ) }</th>
+							<th scope="col" className="post-info-header">{ __( 'POST', 'wp-parsely' ) }</th>
 							{ ! compact && (
-								<th className="boost-perf-header">{ __( 'BOOST PERFORMANCE', 'wp-parsely' ) }</th>
+								<th scope="col" className="views-header">
+									{ __( 'VIEWS', 'wp-parsely' ) }
+									<span className="views-header-period">(30 days)</span>
```
acicovic: I think we need to make the `(30 days)` string translatable.

src/content-helper/dashboard-page/components/posts-table/components/single-post-row.tsx
```diff
@@ -0,0 +1,190 @@
...
+	stats,
+	compact,
+	isLoadingStats: initialIsLoadingStats,
+	showStats = true,
+	showActions = true,
+	onErrorLoadingStats,
+}: SinglePostRowProps ): React.JSX.Element => {
+	const [ isLoadingStats, setIsLoadingStats ] = useState( initialIsLoadingStats );
+
+	const viewsWithoutCommas = stats?.views?.replace( /,/g, '' );
```
acicovic: Although this code should be OK, a more defensive method would be stripping out any non-numeric chars. There's an example of this [here](https://github.com/Parsely/wp-parsely/blob/59790dc0b56d0537a4ebb358cea8fd6b37e5c6a4/src/content-helper/common/utils/number.ts#L12), where we also directly convert to an integer.

```js
const number = parseInt( value.replace( /\D/g, '' ), 10 );
```

acicovic (dismissed):

acicovic:
Also, a few notes/observations about caching:

1. I'm noticing [this CodeRabbit comment](https://github.com/Parsely/wp-parsely/pull/3293#discussion_r2078447544), which hasn't been dismissed (resolved as fixed or unnecessary). Is this one to be reviewed or needs discussion?

2. I see we're using `MONTH_IN_SECONDS` [here](https://github.com/Parsely/wp-parsely/blob/fb84dbf2b48c4a08e3013409cb9a2fdf33d56920/src/Models/class-inbound-smart-link.php#L1099). Was there a particular thought leading to that choice of time? The aim of this question is understanding, not questioning the choice.

3. I see that generally our caching naming scheme/methodology might be a bit inconsistent. Searching for functions such as `wp_cache_delete`, `wp_cache_get` and `wp_cache_delete` can reveal this. Some examples:
    - In some cases we use functions, in others constants (for the cache group) or variables. Not the end of the world though, it might make sense depending on the situation.
    - We use a mix of cache keys with hyphens or underscores.
    - Cache keys and groups can be a bit inconsistent.
    - I'm wondering if we should have a standard prefix to all of them.

4. @alecgeatches, you had noted that indefinite caching could be sub suboptimal. We could maybe work around this by clearing the cache when a post gets updated (e.g. on `save_post`).

I don't think that addressing all the above questions is within the scope of this PR, but addressing what we can in the new code in order not to further increase any inconsistencies would definitely be desirable.

I'd be interested to hear your thoughts on all of these.

src/Endpoints/class-rest-metadata.php
```diff
@@ -66,32 +62,38 @@ public function register_meta(): void {
...
+		$response = array(
+			'version'                         => self::REST_VERSION,
+			'canonical_url'                   => \Parsely\Parsely::get_canonical_url_from_post( $post_id ),
+			'smart_links'                     => array(
+				'inbound'  => 0,
+				'outbound' => 0,
+			),
+			'traffic_boost_suggestions_count' => 0,
+		);
+
```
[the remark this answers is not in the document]
alecgeatches: Both `get_link_counts()` and `get_suggestions_count()` are cached values.

src/content-helper/dashboard-page/components/posts-table/component.tsx
alecgeatches: Good catch, fixed in https://github.com/Parsely/wp-parsely/pull/3293/commits/b0f374ca7cf1152d8bd67a921a5b515abe14412d!

src/content-helper/dashboard-page/components/posts-table/components/single-post-row.tsx
alecgeatches: Good improvement, I've added that in https://github.com/Parsely/wp-parsely/pull/3293/commits/b0f374ca7cf1152d8bd67a921a5b515abe14412d.

alecgeatches:
Good questions and comments, @acicovic!

> 1. I'm noticing https://github.com/Parsely/wp-parsely/pull/3293#discussion_r2078447544, which hasn't been dismissed (resolved as fixed or unnecessary). Is this one to be reviewed or needs discussion?

I have now replied to that comment, the area of concern is actually cached.

> 2. I see we're using MONTH_IN_SECONDS [here](https://github.com/Parsely/wp-parsely/blob/fb84dbf2b48c4a08e3013409cb9a2fdf33d56920/src/Models/class-inbound-smart-link.php#L1099). Was there a particular thought leading to that choice of time? The aim of this question is understanding, not questioning the choice.

`MONTH_IN_SECONDS` was sort of arbitrary. A bot was complaining about an unbounded `wp_cache_set()` without an expiration, and I agree that this is typically not desirable. In this code, it's fairly low-impact to recalculate but isn't likely to change. I picked a month as a medium ground "long enough" period to benefit from caching, but not exist forever.

It's possible to rely on side-effects to clear cache (e.g. `save_post` action), but some actions may not fire when we expect (like on database edits), or if our plugin is disabled they may never get called.

If we update the other cache uses, I think `MONTH_IN_SECONDS` is a decent upper bound alternative to a cache that never expires.

> 3. I see that generally our caching naming scheme/methodology might be a bit inconsistent. Searching for functions such as `wp_cache_delete`, `wp_cache_get` and `wp_cache_delete` can reveal this.

This is very true. I'm going to have a go at standardizing these uses.

> 4. @alecgeatches, you had noted that indefinite caching could be sub suboptimal. We could maybe work around this by clearing the cache when a post gets updated (e.g. on `save_post`).

I think this, along with a reasonable upper bound on cache expiration, is the correct way to do things.

acicovic:
Hey @alecgeatches, thank you for your comments. Everything makes sense!

alecgeatches:
@acicovic I've made some updates to `wp_cache_` functions in https://github.com/Parsely/wp-parsely/pull/3293/commits/6389c74cd3a0801f65b358ca33dec31f378f92f6. Here's how it was organized:

- Where possible, all cached values now use the `PARSELY_CACHE_GROUP` (`wp-parsely`) shared cache group constant.
- The one exception is the Smart Link `get_smart_links_post_cache_group()` function, which uses a group cache [for performance reasons](https://github.com/Parsely/wp-parsely/blob/6389c74cd3a0801f65b358ca33dec31f378f92f6/src/Models/class-smart-link.php#L1264). Otherwise, we need to loop through ~12 separate cache values to delete on flush.
- All cache keys are made with dashes, e.g. `smart-link-to-postid-0e7a8793cf7a598dadf42edcc827f8ab` instead of a mix of dashes and underscores
- Any previously unbounded `wp_cache_set()` functions now use a month at maximum.
- Any functions that construct cache keys or groups dynamically in more than one place have been moved into functions.
- `class-smart-link.php` uses three different caching mechanisms:
    - `get_smart_link_object_by_uid()` uses `get_uid_to_smart_link_cache_key()` is used to [get a Smart Link object by UID](https://github.com/Parsely/wp-parsely/blob/6389c74cd3a0801f65b358ca33dec31f378f92f6/src/Models/class-smart-link.php#L196). These are stored under the main `PARSELY_CACHE_GROUP`. These are flushed in `flush_cache()`.
    - `get_smart_links()` uses `get_smart_links_for_post_cache_key()` to return all of the smart links associated with a post. These are saved under a separate group cache that looks like `wp-parsely-smart-links-<post-id>`.
    - `get_link_counts()` uses `get_smart_link_counts_cache_key()` to return the counts of various types of smart links associated with a post. This is also saved under the separate group cache as above, `wp-parsely-smart-links-<post-id>`.

    The latter two caches above are flushed as a group in [`flush_cache_by_post_id()`](https://github.com/Parsely/wp-parsely/blob/6389c74cd3a0801f65b358ca33dec31f378f92f6/src/Models/class-smart-link.php#L1260).

- There were a handful of places [like this function](https://github.com/Parsely/wp-parsely/blob/6389c74cd3a0801f65b358ca33dec31f378f92f6/src/services/class-cached-service-endpoint.php#L70-L73) where the generated key was super long, for example:

    ```
    parsely-api-4d4d213e8114f16525d726080536279f-075a773dfb85dd6abae83bc94a89a3a0-aab17c6a126b5f7f1101eea66f3388b3
    ```

    Cache keys can have various limits [[1]](https://developer.wordpress.org/reference/functions/set_transient/) [[2]](https://github.com/Automattic/wp-memcached/issues/35) imposed by SQL and middleware services, so I changed these to hash the result of concatenation where possible to shorten keys. ~I chose `sha1` hashing for consistency, and that it produces 40-character outputs, and we likely don't need to defend against purposeful collisions. I think any hashing functions from `md5` to `sha256` would work fine here.~

Note: Changed hashing to `sha256` [to make tests happy](https://github.com/Parsely/wp-parsely/pull/3293/commits/0c4edd243239fb47dc291d284acc2da4045784a3), and because it doesn't really matter.

---

One more thing, I just noticed that `get_smart_links()`'s `wp_cache_set()` call [saves the whole Smart_Link object](https://github.com/Parsely/wp-parsely/blob/6389c74cd3a0801f65b358ca33dec31f378f92f6/src/Models/class-smart-link.php#L998) in the cache which is probably not good. I'll take a look at that and see if we can store IDs instead. I'll also address the test failures.

alecgeatches:
> One more thing, I just noticed that get_smart_links()'s wp_cache_set() call [saves the whole Smart_Link object](https://github.com/Parsely/wp-parsely/blob/6389c74cd3a0801f65b358ca33dec31f378f92f6/src/Models/class-smart-link.php#L998) in the cache which is probably not good. I'll take a look at that and see if we can store IDs instead. I'll also address the test failures.

Addressed in https://github.com/Parsely/wp-parsely/pull/3293/commits/32e9869de6de3156ae172c0224790547f5f97f02. The problem was that we were calling into `wp_cache_set()` with an array of objects like this:

Screenshot 2025-05-09 at 1 27 02 PM

This is a bit unusual, and as you can see from the screenshot also includes inner objects like `WP_Post`. I'm surprised this even worked, and I don't think it's a best practice to stuff that much object data into a cache string. Additionally, inner objects will eventually be stale.

The code has been modified to instead cache the smart link IDs. Although this will incur a performance cost to re-hydrate objects, it'll still cache the heavy taxonomy query and avoid potential object serialization and stale data issues.

src/Models/class-smart-link.php
```diff
@@ -15,6 +15,8 @@
 use Parsely\Parsely;
 use Parsely\Utils\Utils;
 
+use const Parsely\PARSELY_CACHE_GROUP;
+
```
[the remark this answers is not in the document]
alecgeatches: Fixed in https://github.com/Parsely/wp-parsely/pull/3293/commits/322d5707c1e931f60c3eb14351d101f48aed9951.

src/Utils/class-utils.php
```diff
@@ -16,6 +16,7 @@
 use WP_Post;
 use WP_Error;
 
+use const Parsely\PARSELY_CACHE_GROUP;
 use const Parsely\PARSELY_FILE;
```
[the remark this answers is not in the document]
alecgeatches: Fixed in https://github.com/Parsely/wp-parsely/pull/3293/commits/322d5707c1e931f60c3eb14351d101f48aed9951.

src/services/class-cached-service-endpoint.php
```diff
@@ -12,6 +12,8 @@
 
 use WP_Error;
 
+use const Parsely\PARSELY_CACHE_GROUP;
+
```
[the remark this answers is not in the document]
alecgeatches: Fixed in https://github.com/Parsely/wp-parsely/pull/3293/commits/322d5707c1e931f60c3eb14351d101f48aed9951.

src/Models/class-smart-link.php
```diff
@@ -1101,24 +1186,66 @@ protected function flush_cache(): void {
...
-			self::flush_cache_by_post_id( $this->destination_post_id );
+			static::flush_cache_by_post_id( $this->destination_post_id );
 		}
 	}
 
 	/**
-	 * Gets the cache group for all smart links in a post.
+	 * Generates a cache key for the smart link.
+	 *
+	 * @since 3.18.1
```
acicovic: Should this be 3.19.0?

src/Models/class-smart-link.php
```diff
@@ -1101,24 +1186,66 @@ protected function flush_cache(): void {
...
+	 * @return string The cache key.
+	 */
+	protected static function get_uid_to_smart_link_cache_key( string $uid ): string {
+		return sprintf( 'smart-link-uid-map-%s', $uid );
+	}
+
+	/**
+	 * Gets the cache key for all smart links in a post.
 	 *
 	 * @since 3.18.0
```
acicovic: Should this be 3.19.0?

acicovic (commented):
I think we're ready to merge, left a couple of minor comments.

acicovic (approved):
Thank you so much for working on this bringing it to its best version!

PR #3299, opened 2025-05-07
Fix some PHPStan errors

acicovic:
## Description
After running a `composer update`, I noticed these PHPStan errors:

```
 ------ ----------------------------------------------------------------------------- 
  Line   src/Endpoints/class-graphql-metadata.php
 ------ ----------------------------------------------------------------------------- 
  128    Parameter #1 $post_id of static method WP_Post::get_instance() expects int,  
         int|null given.
 ------ ----------------------------------------------------------------------------- 

 ------ ------------------------------------------------------------- 
  Line   src/Metadata/class-metadata-builder.php
 ------ ------------------------------------------------------------- 
  414    Property WP_Term::$parent (int) in isset() is not nullable.
 ------ -------------------------------------------------------------
```

This PR fixes them.

## Motivation and context
- Fix broken committing workflow.
- Improve code quality.

## How has this been tested?
- Verified that errors don't appear after the fix.
- Automated testing passes.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved type validation to prevent errors when retrieving post metadata.
	- Enhanced taxonomy hierarchy traversal for more reliable top-level term detection.


src/Endpoints/class-graphql-metadata.php
```diff
@@ -125,7 +125,12 @@ private function register_object_types(): void {
 	private function register_fields(): void {
 		$resolve = function ( \WPGraphQL\Model\Post $graphql_post ) {
 			$post_id = $graphql_post->ID;
-			$post    = WP_Post::get_instance( $post_id );
+
+			if ( ! is_int( $post_id ) ) {
+				return null;
```
alecgeatches, line 130: `register_fields()` has a `void` return type, so it may also make sense to update these returns to match.

```suggestion
				return;
```

alecgeatches (approved):

src/Endpoints/class-graphql-metadata.php
acicovic: @alecgeatches, I went ahead and "fixed" this, but then got a SonarCloud complaint and also realized that this is the `$resolve` variable, which is holding a function result (null or array). So I reverted.

Let me know if this makes sense, or if you think there's something I might be missing.

alecgeatches: Fine with me, whatever works here!

PR #3322, opened 2025-05-12
Traffic Boost: Fix link text selection adjustment when a link is included later in the same parent

alecgeatches:
## Description

This fixes an issue where sometimes allowed text could not be selected for link replacement:

rejected-link-suggestion

This happened under these circumstances:

1. A link is suggested in a paragraph that already contains a link.
2. The user highlights a portion of text that contains the auto-generated suggestion link, but not the existing link in text.
3. The code uses [`range.commonAncestorContainer`](https://github.com/Parsely/wp-parsely/blob/05f365e0cf775025f94ed14b779a102647063507/src/content-helper/dashboard-page/pages/traffic-boost/preview/components/text-selection-tooltip.tsx#L191) to check for existing links, which is the parent `<p>` tag. `containsAnchor()` then [loops through all nodes in the parent](https://github.com/Parsely/wp-parsely/blob/add/traffic-boost/src/content-helper/dashboard-page/pages/traffic-boost/preview/components/text-selection-tooltip.tsx#L148-L157), including the link in the paragraph after the selection. `containsAnchor()` then mistakenly returns `true`, even though the anchor is not in the selected text.

## Fix

link-suggestion-fix

Rather than using `range.commonAncestorContainer` and evaluating all nodes in the parent selection, the new code walks through the node tree from `range.startContainer` to `range.endContainer`. This allows us to stop checking before we run into unrelated links later in the parent element. The code borrows [from this answer](https://stackoverflow.com/a/668058/770938), which recursively visits nodes between the beginning and end nodes.

## Testing

To test:

1. In the `add/traffic-boost` branch, apply Traffic Boost to a story.
2. Find a suggestion in a paragraph that already contains a link, like the first GIF in this issue.
3. Select additional text that contains the suggested text, but none of the actual link later in the paragraph.
4. See incorrect "Select text without existing links." text:

    image

5. Switch to this branch `fix/link-replacement-with-existing-anchor`.
6. Retry the same steps. Ensure that valid highlight changes including the original suggested text can be added.

## Summary by CodeRabbit

- **Refactor**
  - Improved the method for detecting links within selected text, resulting in more accurate identification when highlighting or interacting with text selections.


acicovic (approved):
Thank you! 🙂

PR #3324, opened 2025-05-13
Fix corrupt Traffic Boost option when upgrading

acicovic:
## Description
Upgrading from v3.18 (or earlier) to our Traffic Boost branch was resulting in a corrupt TB option within the plugin's settings. The effect was that the feature's `Enabled` checkbox would be checked, but all user roles checkboxes would be unchecked. The expected behavior in this case was to have the `Administrator` checkbox checked.

This PR fixes the issue. When a PCH feature doesn't exist in the current options, it now gets added from the default options array.

Also, while working on`tests/Integration/OptionsTest.php`, I added 2 new `@uses` which allow the file to pass the `composer coveragewp` command.

## Motivation and context
Prevent a faulty behavior that could result in many undesirable side-effects.

## How has this been tested?
- Tested the upgrade flow manually.
- An integration test was written for this, and the test was ran with and without the fix. Commenting line 586 in `src/class-parsely.php` will make the test fail, as the feature option will be corrupt.

## Screenshots
| After upgrade before the fix | After upgrade after the fix |
|--------|--------|
| image | image | 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of plugin options to ensure all expected Content Helper feature options are present for existing installations.

- **Tests**
  - Added a new test to verify that missing Content Helper feature options are automatically restored to default values for existing plugin installations.
  - Expanded test coverage annotations for related services.

### DIFF
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -555,16 +555,37 @@ class Parsely {
 		 */
 		$options = get_option( self::OPTIONS_KEY, null );
 
-		// @phpstan-ignore isset.offset, booleanAnd.alwaysFalse
+		// Existing plugin installation without full metadata option.
+		/* @phpstan-ignore isset.offset, booleanAnd.alwaysFalse */
 		if ( is_array( $options ) && ! isset( $options['full_metadata_in_non_posts'] ) ) {
-			// Existing plugin installation without full metadata option.
 			$this->set_default_full_metadata_in_non_posts();
 		}
 
-		// @phpstan-ignore isset.offset, booleanAnd.alwaysFalse
+		// Existing plugin installation without Content Helper options.
+		/* @phpstan-ignore isset.offset, booleanAnd.alwaysFalse */
 		if ( is_array( $options ) && ! isset( $options['content_helper'] ) ) {
-			// Existing plugin installation without Content Helper options.
 			$this->set_default_content_helper_settings_values();
+		}
+
+		// Existing plugin installation that's missing a Content Helper feature option.
+		/* @phpstan-ignore isset.offset */
+		if ( is_array( $options ) && isset( $options['content_helper'] ) ) {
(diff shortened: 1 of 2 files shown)


src/class-parsely.php
```diff
@@ -555,18 +555,39 @@ public function get_options() {
...
+		// Existing plugin installation that's missing a Content Helper feature option.
+		/* @phpstan-ignore isset.offset */
+		if ( is_array( $options ) && isset( $options['content_helper'] ) ) {
+			/** @var array<string,Parsely_Options_Content_Helper_Feature> $pch_options */
+			$pch_options = $options['content_helper'];
+
+			/** @var array<string,Parsely_Options_Content_Helper_Feature> $pch_options_defaults */
+			$pch_options_defaults = $this->option_defaults['content_helper'];
+
+			if ( count( $pch_options ) !== count( $pch_options_defaults ) ) {
```
alecgeatches, line 579: I could see cases where an option is removed from defaults where it might fail, or an option is removed but another is added and the count stays the same. Would it be possible to use something like `array_replace_recursive()` to auto-merge defaults? e.g.

```php
$pch_options_defaults = $this->option_defaults['content_helper'];
$options = array_replace_recursive( $pch_options_defaults, $pch_options );
```

alecgeatches (approved):
Awesome that you added a test for this! I left one untested comment above about possibly making defaults more robust, but if that doesn't work then feel free to ignore and merge.

acicovic:
@alecgeatches, I think you're making a great point in your [comment](https://github.com/Parsely/wp-parsely/pull/3324#discussion_r2087691601) that there are better solutions than the current one. Since this works for now and we don't have much time, I'll merge as is and we'll get back to it. I've noted it so I can create an issue about it after release.

Thank you! 🙂

PR #3325, opened 2025-05-13
Do not return Traffic Boost sample settings

acicovic:
## Description
When we first introduced the Parse.ly Dashboard page, we also created the basic machinery for passing settings from the plugin to that page, similarly to what we do for other parts of the PCH, such as the Dashboard Widget or the Post Editor Sidebar.

When we implemented that, we passed some dummy settings as a proof of concept. However, as we approach release we shouldn't do that, so we're just removing the settings in question, while keeping the system so it can be used when needed.

While working on that, I also noticed some incorrect `@since`s, so I fixed those as well (although admittedly I should have done this in another PR).

## Motivation and context
Don't send junk settings to the Parse.ly Dashboard page.

## How has this been tested?
Manually tested the Parse.ly Dashboard page for any errors after the change.

## Summary by CodeRabbit

- **Chores**
  - Updated version annotations in documentation from 3.18.0 to 3.19.0 across multiple components.
- **Bug Fixes**
  - Adjusted settings to return an empty array by default in one configuration area.

### DIFF
--- a/src/UI/class-dashboard-page.php
+++ b/src/UI/class-dashboard-page.php
@@ -3,7 +3,7 @@
  * UI: Dashboard page class
  *
  * @package Parsely
- * @since   3.18.0
+ * @since   3.19.0
  */
 
 declare(strict_types=1);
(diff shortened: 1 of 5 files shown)


alecgeatches (approved):
👍

PR #3328, opened 2025-05-14
Make the Parse.ly menu dynamic

acicovic:
## Description
With this PR we're making the Parse.ly menu dynamic, so it shows only the items for which the current user has rights to. If the current user doesn't have access to any of them (Traffic Boost or Settings page) then the menu doesn't show at all (picture 2). If the user has only access to a single feature, clicking the menu will automatically focus on that feature's submenu. If both features are available, Traffic Boost will be activated by default when clicking on the menu.

## Motivation and context
Don't show menu items that aren't functional for the current user.

## How has this been tested?
Manual testing.

## Screenshots
| Access to all features | Access to no features | Access to TB only | Access to settings only |
|--------|--------|--------|--------|
| image | image | image | image |


acicovic:
It seems that the menu behaves differently for non-admin roles. I'll be looking into that.

acicovic:
I have found the cause of this and will be figuring out how we can fix for it.

acicovic:
Here's the explainer:

When a menu has a single submenu, WordPress doesn't show it. For users with the `manage_options` capability, we always had 2 menus - TB and Settings - even if one of them was getting removed. WordPress was still considering this 2 submenus and was thus showing them, even if in in the end we had only one visible submenu.

This wasn't happening for users without the `manage_optioons` permission, as the Settings submenu wasn't ever being inserted.

I've forced the same behavior across all user roles by introducing a temporary submenu that isn't rendered in the final menu. This makes sure we get a consistent behavior across all users, as well as being clearer on what they're viewing in the main screen through a submenu with a pertinent title. Else they would just have a "Parse.ly" menu that would either land on Settings or Traffic Boost (suboptimal IMO).

As our collection of submenus increases in the future, we may reach a point where this workaround is unneeded and will be removed.

alecgeatches (approved):
The implementation is a bit odd, but I understand [your reasoning](https://github.com/Parsely/wp-parsely/pull/3328#issuecomment-2881240068) and have tested that the menu works under these circumstances:

- As an admin with TB enabled / disabled
- As an editor user with TB enabled / disabled

Works good for me!

PR #3333, opened 2025-05-17
Fix reference to Pendo class

acicovic:
## Description
With this PR, we're porting a fix (originally posted [here](https://github.com/Automattic/remote-data-blocks/pull/553/files)) for the reference to our Pendo class.

## Motivation and context
Both more context and the original fix are [here](https://github.com/Automattic/remote-data-blocks/pull/553).

## How has this been tested?
We expect @LauraKalnicky and @smithjw1 to test the integration to see if this is all we need.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of telemetry functionality by correcting class references for better namespace handling.


alecgeatches (approved):
Thank you for the fix!

PR #3335, opened 2025-05-18
Hide Traffic Boost submenu when API Secret is not set

acicovic:
## Description
We're updating the Traffic Boost submenu behavior to not be shown when the API Secret isn't set in the plugin's options.

## Motivation and context
This follows the hide/show convention we use in other PCH tools, such as the Dashboard Widget and Posts screen stats.

## How has this been tested?
Manually tested by removing and re-adding the API Secret and refreshing the page with and without this code.

## Summary by CodeRabbit

- **Bug Fixes**
	- The "Traffic Boost" submenu in the dashboard will now only be displayed if the Parsely API secret is configured and the user has the appropriate permissions.
- **Style**
	- Improved readability of comments in the dashboard menu code.

### DIFF
--- a/src/UI/class-dashboard-page.php
+++ b/src/UI/class-dashboard-page.php
@@ -142,7 +142,8 @@ final class Dashboard_Page {
 	 * @since 3.19.0
 	 */
 	public function add_dashboard_page_to_menu(): void {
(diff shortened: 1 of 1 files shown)


alecgeatches (approved):
Tested and confirmed this works

PR #3337, opened 2025-05-19
Traffic Boost: Improve error handling

acicovic:
## Description
Traffic Boost error handling was incomplete, with a few important side-effects such as:
- Hard errors were being displayed in (often) multiple snackbar notifications (which disappear after a few seconds) instead of having a permanent place in the main UI. The user could potentially miss them (screenshots row 1).
- Some errors didn't display at all. For example, if credentials weren't eligible for Traffic Boost, this would silently fail and show a UI to generate suggestions, leading the user to believe that the feature was working (screenshots row 2).

This PR solves for these cases by fixing/adding some error handling code, as well as introducing a way to show persistent messages in the UI. Soft errors (those allowing execution to continue) will still appear in a snackbar, while hard errors (those who result in execution stopping) will be displayed as persistent messages.

## Motivation and context
Give correct feedback to the user, with persisting error messages where needed.

## How has this been tested?
The current testing focused on hard errors, especially for the cases where credentials do not work for Traffic Boost (no API Secret present or given credentials don't have TB access). Tested manually using 2 different credential sets (the first eligible for TB, the second not). I haven't tested every possible error. There could be edge cases that can be addressed in the future with more extensive testing, but the current modifications should solve for most cases.

## Screenshots
| Before | After |
|--------|--------|
| image | image |
| image | image | 

## Summary by CodeRabbit

- **New Features**
  - Added explicit error messages in the UI for certain error scenarios in Posts Table and Traffic Boost sidebar components.
  - TabsContent and TrafficBoostSidebar components now support a new prop to display hard error messages directly.

- **Bug Fixes**
  - Improved error handling to distinguish between retryable and non-retryable errors, providing clearer feedback to users and avoiding unnecessary notifications for non-retryable errors.

- **Refactor**
  - Enhanced and streamlined error handling logic across multiple components for better user feedback and maintainability.


acicovic:
A last todo item for this PR is to modify the message when the credentials are not eligible (screenshot 4). I think we should state the tiers where the feature is available, as well as have a link to the Lobby Post and/or docs.

alecgeatches (approved):
Looks great to me! I also verified that hard API errors show up on the main Traffic Boost page, e.g.

Screenshot 2025-05-19 at 11 09 54 AM

acicovic:
This is the error message I've come up with when credentials are ineligible for TB. This is what we'll be shipping with at the moment:

> [Traffic Boost](https://lobby.vip.wordpress.com/2025/05/19/introducing-traffic-boost-a-smarter-way-to-recirculate-your-content/) is currently not enabled for your Site ID.
>
> Information about requesting access to Traffic Boost can be found [here](https://docs.wpvip.com/parse-ly/wp-parsely-features/traffic-boost/#Access).

cc @LauraKalnicky FYI.

acicovic:
image

Issue #3350, opened 2025-05-21
Breaking change in 3.19.0


nikkoi:
## Describe the bug
After our plugin auto updated to 3.19.0, we started to see very weird issues, we're we couldn't publish any new posts on our wordpress sites, nor access already published posts in the edit screen with block editor.

## To Reproduce
1. Have the plugin version 3.19.0
2. Have redis object caching on kinsta

## Environment
Latest WP, latest Parsely (3.19.0), latest Redis object cache on Kinsta.

## Additional context
This seems to be an issue where Redis object cache (on Kinsta at least) and the latest version of the plugin 3.19.0 doesn't work together. Disabling either fixed the problem, but we chose to downgrade back to 3.18.0 to keep Redis working.

This is the error wordpress sent us:
`E_ERROR uppstod på rad 1074 i följande fil: /www/sitename_925/public/wp-content/plugins/wp-parsely/src/Models/class-inbound-smart-link.php. Felorsak: Uncaught TypeError: Parsely\Models\Inbound_Smart_Link::get_suggestions_count(): Return value must be of type int, string returned in /www/sitename_925/public/wp-content/plugins/wp-parsely/src/Models/class-inbound-smart-link.php:1074
Stack trace:
#0 /www/sitename_925/public/wp-content/plugins/wp-parsely/src/Endpoints/class-rest-metadata.php(127): Parsely\Models\Inbound_Smart_Link::get_suggestions_count(688656)
#1 /www/sitename_925/public/wp-includes/rest-api/endpoints/class-wp-rest-controller.php(439): Parsely\Endpoints\Rest_Metadata->get_callback(Array, 'parsely', Object(WP_REST_Request), 'nyheter')
#2 /www/sitename_925/public/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php(2098): WP_REST_Controller->add_additional_fields_to_object(Array, Object(WP_REST_Request))
#3 /www/sitename_925/public/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php(655): WP_REST_Posts_Controller->prepare_item_for_response(Object(WP_Post), Object(WP_REST_Request))
#4 /www/sitename_925/public/wp-includes/rest-api/class-wp-rest-server.php(1292): WP_REST_Posts_Controller->get_item(Object(WP_REST_Request))
[...]

PR #3351, opened 2025-05-21
Make some caching checks more flexible

acicovic:
## Description
In issue #3350 we got a report of a type error, which seems to be due to the way we perform our checks after using `wp_cache_get()`.

This PR makes our checks more flexible, accommodating for the possibility of numeric strings when a number is expected. We're also casting the returned value to int when the function is expected to return an integer.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cache validation to ensure only appropriate numeric or array values are used, reducing the risk of invalid cached data affecting app behavior.
  - Enhanced type checking and value casting for cached data, resulting in more reliable and consistent handling of cached information.

### DIFF
--- a/src/Models/class-inbound-smart-link.php
+++ b/src/Models/class-inbound-smart-link.php
@@ -1069,7 +1069,7 @@ class Inbound_Smart_Link extends Smart_Link {
 		$cache_key = self::get_suggestions_count_cache_key( $post_id );
 		$count     = wp_cache_get( $cache_key, PARSELY_CACHE_GROUP );
 
-		if ( is_numeric( $count ) ) {
+		if ( false !== $count && is_numeric( $count ) ) {
 			return (int) $count;
 		}
 
--- a/src/Models/class-inbound-smart-link.php
+++ b/src/Models/class-inbound-smart-link.php
@@ -1069,9 +1069,8 @@ class Inbound_Smart_Link extends Smart_Link {
 		$cache_key = self::get_suggestions_count_cache_key( $post_id );
 		$count     = wp_cache_get( $cache_key, PARSELY_CACHE_GROUP );
 
-		if ( false !== $count ) {
-			/** @var int $count */
-			return $count;
+		if ( is_numeric( $count ) ) {
+			return (int) $count;
 		}
 
 		$args = array(
--- a/src/Models/class-smart-link.php
+++ b/src/Models/class-smart-link.php
@@ -196,8 +196,9 @@ class Smart_Link extends Base_Model {
 	private function get_smart_link_object_by_uid( string $uid ): int {
 		$cache_key = self::get_uid_to_smart_link_cache_key( $uid );
 		$cached    = wp_cache_get( $cache_key, PARSELY_CACHE_GROUP );
-		if ( is_int( $cached ) && 0 !== $cached ) {
-			return $cached;
+
+		if ( is_numeric( $cached ) && 0 !== (int) $cached ) {
+			return (int) $cached;
 		}
 
 		$smart_links = new \WP_Query(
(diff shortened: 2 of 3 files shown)


src/Models/class-inbound-smart-link.php
```diff
@@ -1069,9 +1069,8 @@ public static function get_suggestions_count( int $post_id ): int {
 		$cache_key = self::get_suggestions_count_cache_key( $post_id );
 		$count     = wp_cache_get( $cache_key, PARSELY_CACHE_GROUP );
 
-		if ( false !== $count ) {
-			/** @var int $count */
-			return $count;
+		if ( is_numeric( $count ) ) {
```
alecgeatches: Can we make all of these checks explicit for checking `false !==`? `wp_cache_get()` is explicitly going to return `false` when it fails, so it's a bit odd to check for typing in those cases.

```suggestion
		if ( false !== $count && is_numeric( $count ) ) {
```

src/Models/class-smart-link.php
```diff
@@ -196,8 +196,9 @@ public function __construct(
 	private function get_smart_link_object_by_uid( string $uid ): int {
 		$cache_key = self::get_uid_to_smart_link_cache_key( $uid );
 		$cached    = wp_cache_get( $cache_key, PARSELY_CACHE_GROUP );
-		if ( is_int( $cached ) && 0 !== $cached ) {
-			return $cached;
+
+		if ( is_numeric( $cached ) && 0 !== (int) $cached ) {
```
alecgeatches: Same here:

```suggestion
		if ( false !== $cached && is_numeric( $cached ) ) {
```

The `0 !== (int) $cached` check seems unnecessary here with a `false` check.

src/Models/class-smart-link.php
```diff
@@ -1097,8 +1098,7 @@ public static function get_link_counts( int $post_id, string $status = Smart_Lin
 		$cache_group = self::get_smart_links_post_cache_group( $post_id );
 		$link_counts = wp_cache_get( $cache_key, $cache_group );
 
-		if ( false !== $link_counts ) {
-			/** @var array<string,int> */
+		if ( is_array( $link_counts ) ) {
```
alecgeatches:
```suggestion
		if ( false !== $link_counts && is_array( $link_counts ) ) {
```

src/Utils/class-utils.php
```diff
@@ -430,8 +430,8 @@ public static function get_post_id_by_url( string $url ): int {
 		$cache_key = sprintf( 'url-to-postid-%s', hash( 'sha256', $url ) );
 		$cache     = wp_cache_get( $cache_key, PARSELY_CACHE_GROUP );
 
-		if ( is_integer( $cache ) ) {
-			return $cache;
+		if ( is_numeric( $cache ) ) {
```
alecgeatches:
```suggestion
		if ( false !== $cache && is_numeric( $cache ) ) {
```

src/Models/class-inbound-smart-link.php
acicovic: I've removed it as unnecessary (e.g. if it's numeric, it's not false) but I'll bring these back if you feel they're more readable.

src/Models/class-smart-link.php
acicovic: I kept it in case there was a reason for it. I didn't dig further on why we're checking for a potential 0 value. Do you think it's irrelevant?

src/Models/class-inbound-smart-link.php
alecgeatches: I'd prefer them in, as `wp_cache_get()` returning `false` has a specific meaning that isn't exactly the same as `is_numeric()`. I'd also prefer it for readability as well, even if it's verbose.

src/Models/class-smart-link.php
alecgeatches: That's a good point! Looking at the cache setting code, it appears we never will set `0` in this cache anyway:

https://github.com/Parsely/wp-parsely/blob/5cf945848bde41842cb5eaec6e4f367354f2decc/src/Models/class-smart-link.php#L355-L366

So I think the suggested change is safe to make.

src/Models/class-inbound-smart-link.php
acicovic: OK, I'll bring these back! 🙂

alecgeatches (approved):

PR #3379, opened 2025-05-27
Fix update-version.sh

acicovic:
## Description
The addition of the `PARSELY_DATA_SCHEMA_VERSION` constant in `wp-parsely.php` line 54, introduced additional spaces on line 52.

This caused our release workflow to stop updating the version number on line 52, requiring [manual intervention](https://github.com/Parsely/wp-parsely/pull/3346/commits/44499445507013bf626ad5766d76480957d95b52) during the release process.

This PR fixes the issue.

## Motivation and context
Prevent our workflow from breaking.

## How has this been tested?
Locally tested that line 52 correctly updates after the fix by directly running the `update-version.sh` script.

## Summary by CodeRabbit

- **Refactor**
  - Improved the accuracy of version update handling to better match the formatting of the target line. No changes to user-facing features.

### DIFF
--- a/bin/update-version.sh
+++ b/bin/update-version.sh
@@ -11,8 +11,8 @@ set -e
 export LC_ALL=C
 
 if [ -z "$1" ]; then
-  echo "Error: You must specify a version number."
-  exit 1
+ echo "Error: You must specify a version number."
+ exit 1
 fi
 
 VERSION=$1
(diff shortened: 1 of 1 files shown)


bin/update-version.sh
```diff
@@ -38,7 +38,7 @@ sed_inplace "s/Stable tag: .*  $/Stable tag: $VERSION  /" README.md
 sed_inplace "s/\"version\": \".*\"/\"version\": \"$VERSION\"/" package.json
 sed_inplace "s/export const PLUGIN_VERSION = '.*'/export const PLUGIN_VERSION = '$VERSION'/" tests/e2e/utils.ts
 sed_inplace "s/ \* Version:           .*$/ \* Version:           $VERSION/" wp-parsely.php
-sed_inplace "s/const PARSELY_VERSION = '.*'/const PARSELY_VERSION = '$VERSION'/" wp-parsely.php
+sed_inplace "s/const PARSELY_VERSION             = '.*'/const PARSELY_VERSION             = '$VERSION'/" wp-parsely.php
```
alecgeatches, line 41: How about this, so if we add another long constant it doesn't break?

```suggestion
sed_inplace "s/const PARSELY_VERSION\( *\)= '.*'/const PARSELY_VERSION\1= '$VERSION'/" wp-parsely.php
```

This takes any number of spaces `( *)` and reuses it in reference `\1`. Try it locally with different spaces, it should work!

We could also do this for `Version:` in the plugin header, but that alignment is less likely to change.

alecgeatches (approved):
Glad you fixed this! I added one suggestion below to make the fix more generalized, but either way 🚀

bin/update-version.sh
acicovic: CodeRabbit had suggested something similar in its "Nitpick comments" but I actually prefer to be more rigid in this case, in order to avoid any possible additional changes (although I admit the possibility is very very low).

Let me know if this is OK with you.

acicovic:
I'll merge as is, and we can implement the suggestion in another PR if we go that route.

PR #3414, opened 2025-06-04
Fix PHPStan errors on PHP 8.3 and above

acicovic:
## Description
@alecgeatches had reported that PHPStan was returning errors on PHP 8.3 and above, similar to:
```
Out of 39 possible constant types, only 1 - 2.5 % actually have it. Add more  
         constant types to get over 99 %
```
Turns out the root of the issue is the introduction of [typed constants](https://php.watch/versions/8.3/typed-constants) in PHP 8.3. Since we need to stick to older PHP versions, I used the `constant_type` setting to neutralize the errors.

## Motivation and context
Fix our commit workflow from breaking on PHP8.3+.

## How has this been tested?
Tested on PHP 8.3/8.4 with and without the fix.

## Summary by CodeRabbit

- **Chores**
  - Updated configuration comments to clarify PHP version requirements for typed properties and constants. No functional changes made.


alecgeatches (approved):
Thank you for figuring out the configuration change we needed!

PR #3419, opened 2025-06-04
Upgrade minimum PHP version to 7.4

### LINKED ISSUE #3418: Upgrade minimum PHP version to 7.4
## Is your request related to a problem?
The combination of PHP 7.2 and 7.3 usage [has dropped below 5%](https://core.trac.wordpress.org/ticket/62622#comment:22), and it's very likely that WordPress 6.9 will require PHP 7.4 as the minimum. Many of VIP's plugins have been requiring PHP 7.4 for quite some time already, and I think it's a good time for us to make that move as well.

Bumping to 7.4 will give us the possibility to use [typed properties, arrow functions and the null coalescing assignment operator](https://www.php.net/manual/en/migration74.new-features.php) among others. It's also worth looking [here](https://php.watch/versions/7.3) in case there's something we can use from PHP 7.3, with the `array_key_*` functions and `is_countable()` coming to mind.

Upgrade the minimum PHP version to 7.4. We can base ourselves on the work in #1401 and #1431.

acicovic:
## Description
With this PR, we're upgrading the required minimum PHP version to v7.4.

## Motivation and context
- More context can be found in #3418.
- Closes #3418.

## How has this been tested?
Automated testing passes. Tests against PHP 7.2/7.3 are expected to stall/fail.

## Summary by CodeRabbit

- **Chores**
	- Updated the minimum required PHP version to 7.4 in documentation, plugin metadata, and configuration files.
	- Adjusted automated test workflows to remove support for PHP 7.2 and 7.3.
	- Updated code analysis tools to check compatibility with PHP 7.4 and above.


alecgeatches (approved):
🎉

PR #3446, opened 2025-06-06
Add action bar highlight hint

alecgeatches:
## Description

Extension to the existing floating actions bar work in #3445. Adds a hint ("Highlight to customize"), a tooltip on the accept/reject actions bar, and minor spacing changes:

Screenshot 2025-06-09 at 1 53 36 PM

Also makes some styling adjustments that I'll note in code below.

## Motivation and context

Users of the Traffic Boost feature can miss that it's possible to retarget or expand link suggestions. This explicit text makes it easier to find that feature.

## How has this been tested?

Locally. There isn't much behavior difference from the work in #3445 other than the bar being longer, and some minor styling changes.

## Summary by CodeRabbit

- **New Features**
  - Added a tooltip with an info icon and hint text to guide users on how to highlight text for customization in the preview actions area.
- **Style**
  - Improved visual styling for the hint area and its text within the actions bar, including updated divider appearance for better clarity.

## Summary by CodeRabbit

- **New Features**
  - Added a tooltip with helpful hints and an info icon when no text is selected in the preview actions area.

- **Style**
  - Improved button spacing and updated icons for a cleaner appearance.
  - Enhanced font consistency across tooltips and action bars for a more unified look.
  - Adjusted the vertical divider for better visual separation.


acicovic:
Hey @alecgeatches, thank you for your work. If this is still a WIP, I recommend setting the PR as draft until it's ready for review.

Thanks! 🙂

src/content-helper/common/components/vertical-divider/component.tsx
```diff
@@ -27,7 +27,7 @@ interface VerticalDividerProps {
  */
 export const VerticalDivider = ( { size = 40, color = '#cccccc' }: VerticalDividerProps ): JSX.Element => {
 	return (
-		<SVG xmlns="http://www.w3.org/2000/svg" width="1" height={ size } viewBox={ `0 0 1 ${ size }` } fill="none">
+		<SVG xmlns="http://www.w3.org/2000/svg" width="3" height={ size } viewBox={ `0 0 1 ${ size }` } fill="none">
```
alecgeatches, line 31: In Firefox this SVG was sometimes invisible with a width of `1`. Using `3` made the divider bar appear consistently.

src/content-helper/dashboard-page/pages/traffic-boost/preview/components/preview-actions.tsx
```diff
@@ -182,7 +190,7 @@ export const PreviewActions = ( {
 									<>
 										<Button
 											variant={ isRemoving ? 'primary' : 'tertiary' }
-											icon={ isRemoving ? null : close }
+											icon={ isRemoving ? null : closeSmall }
```
alecgeatches: Changed the "Reject" button icon to `closeSmall`, which matches the icon size of the "Accept" button. Previously, they mismatched:

image

(notice the X is larger than the check)

src/content-helper/dashboard-page/pages/traffic-boost/preview/components/text-selection-tooltip.tsx
```diff
@@ -58,6 +58,7 @@ const useIframeStyles = ( iframeDocument: Document ) => {
 			}
 
 			.parsely-traffic-boost-iframe-popover {
+				font-family: -apple-system, BlinkMacSystemFont,"Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell,"Helvetica Neue", sans-serif;
```
alecgeatches, line 62: Previously, we did not have an explicit font set for the "Use as Link Text" or actions toolbar components. In testing, this caused both to take the font style of the surrounding content. If I changed `theme.json` to use Comic Sans for post content font, both tools inherit the font incorrectly:

Screenshot 2025-06-09 at 12 51 29 PM


---

Screenshot 2025-06-09 at 1 05 50 PM

---

So why are these hardcoded instead of using a CSS variable? In Gutenberg, the default font is defined in the [`$default-font` SCSS variable](https://github.com/WordPress/gutenberg/blob/faa2c903c95b2d8c25c9521f3b8f0057e41f15cb/packages/base-styles/_variables.scss#L15), but that's not exposed anywhere as a CSS variable we can use. Instead, this is compiled into the [`body` tag's styling](https://github.com/WordPress/gutenberg/blob/faa2c903c95b2d8c25c9521f3b8f0057e41f15cb/packages/block-editor/src/default-editor-styles.scss#L9) and buttons use [`font-family: inherit`](https://github.com/WordPress/gutenberg/blob/faa2c903c95b2d8c25c9521f3b8f0057e41f15cb/packages/components/src/button/style.scss#L11) to implicitly use the same font-family.

In short, Gutenberg only exposes these base set of fonts internally, and internal components expect that a parent tag supplies the correct fonts. Because we're working in somebody else's post content, we can't easily use the same font hierarchy components expect. We solve this by setting these explicitly inside of our components to avoid using the wrong styling.

src/content-helper/common/components/vertical-divider/component.tsx
acicovic: We could put this as a comment for posterity, in case any of us think this should be `1` in the future.

src/content-helper/dashboard-page/pages/traffic-boost/preview/components/text-selection-tooltip.tsx
acicovic: Also worth having a concise version of this as a comment, or having a link to here should we ever wonder why we did this in the first place.

alecgeatches: Good thinking! I put a very condensed version in https://github.com/Parsely/wp-parsely/pull/3446/commits/5a7392bc3427f7af736c9ab701ea77e0a04e1feb:

https://github.com/Parsely/wp-parsely/blob/5a7392bc3427f7af736c9ab701ea77e0a04e1feb/src/content-helper/dashboard-page/pages/traffic-boost/preview/hooks/use-iframe-highlight.ts#L204

src/content-helper/common/components/vertical-divider/component.tsx
alecgeatches: Addressed in https://github.com/Parsely/wp-parsely/pull/3446/commits/5a7392bc3427f7af736c9ab701ea77e0a04e1feb!

alecgeatches:
@acicovic I've reviewed your comments and added some code documentation for the places you pointed out missing context. Should be ready for another review, thank you!

acicovic (approved):
Thank you so much for working on this!

PR #3469, opened 2025-06-18
Fix outbound link highlights in Traffic Boost

### LINKED ISSUE #3468: PCH Traffic Boost: Fix floating toolbar quirks when clicking link buttons
## Describe the bug
When we click the  "N external"/"N internal"/"N smart links" buttons on the top of the Traffic Boost preview, this makes the floating toolbar to either getting misplaced or disappear. @alecgeatches has captured a video of this:

https://github.com/user-attachments/assets/966a8553-0558-4702-b467-c40d30fa8d24

## To Reproduce
Open any post with the Traffic Boost feature and use the buttons.

## Expected behavior
The toolbar should remain in place and visible at all times.

alecgeatches:
## Description

This fixes existing link highlights, which broke in `3.20.0` due to the new Traffic Boost toolbar changes. "Existing link highlights" refers to the "External"/"Internal"/"Smart Links" button at the top of source posts:

*Before this PR*:

https://github.com/user-attachments/assets/e32d6b92-207c-40d1-b27d-6b098249b04c

This contained some odd behavior, including moving the actions bar above unrelated existing link highlights, hiding the actions bar on un-highlight, and placing the post viewer in a broken state afterward.

---

*After this PR*:

https://github.com/user-attachments/assets/227ec353-5bd7-4097-9cbe-7b538d6f0367

Links are highlighted but do not affect the suggested selection.

## Motivation and context

Fixes #3468.

The root cause of the issue is that the logic in `useIframeHighlight()` became coupled with the actions bar, which now has its lifecycle and positioning managed by the hook. We were using the same `highlightRange()` and `removeHighlights()` for both link suggestions and existing link highlighting. After the toolbar logic was introduced, this caused existing link highlighting to interrupt the lifecycle of the actions bar.

The fix was to move existing link highlighting to a separate function that's only for existing link highlights. Although there is some duplication of highlight cleanup code, this allowed us to simplify `use-iframe-highlight.ts` by removing unrelated code and styles from that hook. I'm okay with the minor duplication between these hooks, as these features have very different uses and will likely continue to diverge in the future.

Additionally, the post logic has been changed to reset the selected existing link type when switching between posts. Previously this was persisted between source posts, which didn't really make sense if something like "Smart Links" were selected in one post but the next selected post didn't contain any. I think it's more natural to clear this post-specific selection when switching between source posts.

## Summary by CodeRabbit

- **New Features**
  - Improved link highlighting within the preview iframe, now supporting more precise highlighting of existing links by type.
- **Bug Fixes**
  - Enhanced synchronization between the selected link type in the UI and the underlying preview state.
- **Refactor**
  - Updated internal logic to use a new highlighting mechanism, ensuring better maintainability and performance.


alecgeatches:
@acicovic SonarCloud is showing the code duplication (95 lines) I mentioned in the description. This is entirely in the `removeExistingLinkHighlights()` function (which contains `removeAndClean()` and `unwrapHighlights()`), which is indeed duplicated from `useIframeHighlight()`. We could move functions out of both files to a new shared hook, but the logic is super fiddly and as mentioned above, may well diverge from actionable highlights in the future anyway.

I'd prefer to keep these as separate duplicated logic rather than forcing shared internal code in this case.

acicovic (approved):
Thank you so much for working on this!

PR #3534, opened 2025-07-15
Improve whitespace trimming in get_canonical_url()

acicovic:
## Description
With this PR, we're improving whitespace trimming in the `get_canonical_url()` function, as the previous implementation could result in double slashes when there was trailing whitespace in the canonical. We're also introducing the tests mentioned in #3525.

## Motivation and context
- Fix trailing whitespace edge case.
- Complete and add the integration tests mentioned in #3525.

## How has this been tested?
Through the integration tests introduced.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling and normalization of canonical URL domains to ensure more consistent and accurate URL processing.

* **Tests**
  * Added new integration tests to verify canonical URL handling and normalization across various domain formats.

### DIFF
--- a/src/class-parsely.php
+++ b/src/class-parsely.php
@@ -1043,8 +1043,8 @@ class Parsely {
 		if ( is_string( $canonical_url_domain ) ) {
 			// Get the canonical URL domain without protocol, trailing slashes
 			// or accidental whitespace.
-			$canonical_url_domain = rtrim( $canonical_url_domain, '/' );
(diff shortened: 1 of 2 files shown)


acicovic:
SonarCloud incorrectly complains of "possible security issues" because the integration tests uses `http` in test strings, which is totally acceptable and in line with the canonicals used by many customers. The URLs aren't used to fetch data or perform any operations.

acicovic:
I've marked the issues as "safe" in SonarCloud and will be merging this now.

PR #3636, opened 2025-09-03
Engagement Boost: Make preview work for non-administrators

### LINKED ISSUE #3635: Engagement Boost preview doesn't work for non-administrators
## Describe the bug
Non-administrators see a "-1" displayed in the Engagement Boost preview screen, instead of seeing the post's content.

## To Reproduce
Use Engagement Boost with a non-administrator account, and open to boost any post. The preview pane will display a "-1".

## Expected behavior
Being able to see the post content in the preview.

acicovic:
## Description
When using Engagement Boost, non-administrators would see a "-1" displayed in the Engagement Boost preview screen, instead of seeing the post's content.

This PR creates a predefined changeset post in order to make the Engagement Boost preview work for non-administrator user roles. By creating a predefined changeset post with a known UUID that we use in `iFrameSrc()`, we guarantee that the preview will work for all authorized user roles.

## Motivation and context
- Fixes VIPCMS-1383
- Root issue: When the Engagement Boost preview iframe is loaded, it calls the WordPress Customizer in the `iFrameSrc()` function. The Customizer requires the current user to have the `customize` capability (available to Administrators), or the passed UUID to have a corresponding changeset post in the database. Otherwise [it fails](https://github.com/WordPress/WordPress/blob/f70e032c5f1e38abdbefe993bf21732a29ac1608/wp-includes/class-wp-customize-manager.php#L555-L562) with -1 ("Non-existent changeset UUID") and prevents preview, displaying "-1".

## How has this been tested?
Using the current `develop` branch, use Engagement Boost with a non-administrator account, and notice the "-1" in the preview pane. Repeat the steps with this patch, and the post content will appear as expected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Engagement Boost preview now reliably loads for non‑administrator users by using a stable preview identifier to prevent missing or mismatched previews.

* **Chores**
  * Automatically ensures the required preview draft is created during admin initialization to support the preview flow.

* **Refactor**
  * Consolidated and cleaned up imports in the preview component.


wp-parsely.php
```diff
@@ -114,6 +114,50 @@ function parsely_admin_init_register(): void {
...
+	if ( ! $parsely->api_secret_is_set() || ! Permissions::current_user_can_use_pch_feature(
+		'traffic_boost',
+		$parsely->get_options()['content_helper']
+	) ) {
+		return;
+	}
+
+	// Needs to match the UUID in iFrameSrc() in preview-frame.tsx.
+	$uuid    = '905b130b-4129-4416-919c-9e31433a6f65';
+	$post_id = post_exists( "wp-parsely-$uuid", '', '', 'customize_changeset', 'private' );
```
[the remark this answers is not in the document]
acicovic: These should be addressed now.

wp-parsely.php
```diff
@@ -114,6 +114,54 @@ function parsely_admin_init_register(): void {
...
+ * `class-wp-customize-manager.php` and prevents preview, displaying "-1".
+ *
+ * By creating a predefined changeset post with a known UUID that we use in
+ * `iFrameSrc()`, we guarantee that the preview will work for all authorized
+ * user roles.
+ *
+ * @since 3.21.0.
+ */
+function create_engagement_boost_changeset_post(): void {
+	if ( ! function_exists( 'post_exists' ) ) {
```
alecgeatches, line 136: Does this ever run in a context where `post_exists` is missing? Just curious, as we use a bunch of other functions in this method without a check, so I'm guessing this is a specific fix.

alecgeatches (approved):
Insane bug and fix. I've tested locally and confirmed this fixes the issue for an `editor` user. Great job!

wp-parsely.php
acicovic: @alecgeatches, thanks for checking! 🙂

No, actually this was a CodeRabbit suggestion. During my testing, I never hit a point where the function didn't exist. However, in theory this could happen if we moved the hook quite earlier, but which would probably be an accident.

I'm fine with either keeping or removing this. As you say, we don't do it in other functions, so it does feel like a specific fix when in fact it isn't.

alecgeatches: @acicovic I found [at least one complaint](https://wordpress.stackexchange.com/questions/218715/fatal-error-call-to-undefined-function-post-exists) about this function, so I think it can't hurt to include. I have seen some weird behaviors for `init` functions called by cron tasks similar to this.

acicovic: Great, will leave as is then. Thanks for the review! 🙂

acicovic:
Adding a "Do not merge" label here for the time being, as this will be ported to a fix release instead of 3.21.0.

PR #4097, opened 2026-03-26
PCI Sidebar: Fix multiple contact messages

### LINKED ISSUE #4096: PCI Sidebar: Fix multiple contact messages
## Summary

When no Parse.ly credentials are configured, the plugin sidebar's Tools tab shows the "Contact us" credentials message multiple times — once per expanded panel.

## Current state

Each panel in the Tools tab (Title Suggestions, Excerpt Suggestions, Smart Linking, Related Posts) independently checks for credentials using `<VerifyCredentials>`. When credentials are missing, every expanded panel renders its own copy of the "Contact us / wp-parsely options" message, resulting in the same message appearing up to 4 times simultaneously. The "Boost Engagement" button also remains visible below the repeated messages. The Performance tab does not have this problem — it wraps all its content in a single credential gate.

Image

## Desired state

The credentials message should appear exactly once in the Tools tab, consistent with how the Performance tab handles it. The "Boost Engagement" button should also be hidden when credentials are not set, since it requires credentials to be useful.

acicovic:
## Summary

When no Parse.ly credentials are configured, each panel in the Tools tab independently rendered the "Contact us" credentials message. This was inconsistent with the Performance tab, which shows the message once for the whole tab. The fix lifts the credential gate to the tab level, matching the established pattern, and also moves the "Boost Engagement" button inside the gate since it has no utility without credentials.

As a side effect, the `VerifyCredentials` component's `children` prop type was broadened from `React.JSX.Element|JSX.Element[]` to `React.ReactNode` to correctly support conditional rendering at the wrapper level.

- Fixes https://github.com/Parsely/wp-parsely/issues/4096
- Fixes VIPCMS-1676

## Before

- The Tools tab wrapped each panel's content individually in `<VerifyCredentials>`, causing the "Contact us / wp-parsely options" message to appear once per expanded panel (up to 4 times).
- The "Boost Engagement" button was rendered outside any credentials check and appeared below the repeated messages.
- Behaviour was inconsistent with the Performance tab.

## After

- A single `<VerifyCredentials>` wraps all Tools tab content, showing the message exactly once — consistent with the Performance tab.
- The "Boost Engagement" button is hidden alongside the panels when credentials are absent.

## Screenshots

| Before | After |
|--------|--------|
| image | image | 

## Summary by CodeRabbit

* **Refactor**
  * Broadened a component prop to accept a wider range of React children.
  * Consolidated credential gating into a single wrapper around the tool panels for simpler structure.

* **Tests**
  * Added a reusable sidebar/tab message helper and updated end-to-end tests to use it, simplifying sidebar visibility handling and message retrieval.


acicovic:
@copilot, can you understand why E2E tests fail here (they pass for me locally), and propose a fix? It shouldn't break any functionality, and the tests should pass without compromising their quality.

Name the PR that you'll be opening "Fix failing E2E tests for PR #4097".

PR #4105, opened 2026-03-31
Fix post list title column being squeezed by Stats column width

m:
## Problem

The Parse.ly Stats (7d) column sets `width: 200px`, which is far too wide for the short numeric content it displays (a view count or "—"). On sites with additional plugin columns (e.g. Jetpack Stats), WordPress core's `table-layout: fixed` squeezes the Title column — the only column without an explicit width — to as little as ~50px, making post titles unreadable.

## Fix

- Change the Parse.ly column width from `200px` to `6%` (~69px at typical viewport), which is more than enough for its short content and scales naturally (removing the need for the `991px` media query override).
- Reduce core's `15%` for Categories and Tags to `10%` — these columns typically show short content and `15%` is disproportionate when plugin columns are present.

## Results

| Column | Before | After |
|--------|--------|-------|
| Title | ~50px | **~292px** |
| Categories | 172px (15%) | 115px (10%) |
| Tags | 172px (15%) | 115px (10%) |
| Parse.ly Stats | 200px | 69px (6%) |

No horizontal scrollbar. Tested on WordPress 7.1-alpha (trunk) with Jetpack Stats column active, verified at multiple viewport widths.

## Summary by CodeRabbit

* **Style**
  * Updated column width styling for post list statistics display, improving responsive layout behavior for category and tag columns.


acicovic:
Hi @m, thank you for your contribution.

As a first step, I've rebuilt the project's assets (this is pushed to your branch). I will be testing this momentarily and let you know of any feedback.

Thanks!

acicovic:
The changes cause text wrapping inside the cell. Screenshots for demonstration:

**1900px Viewport** (no issue)
image

**1500px Viewport** (issue visible)
image

**800px Viewport** (issue pronounced)
image

Things we can do:
1. There's some room to make the column width a bit smaller, but not by a lot. The size could be around 140px or so, but we'd need to test what happens with big numbers.
2. If Instead of displaying text, we could resort to displaying icons if we want to limit the column width to the max possible.

Let me know of your feedback. I'm happy to take this to the finish line, or review any new code.

Thank you.

acicovic:
Closing this in favor of https://github.com/Parsely/wp-parsely/pull/4136, which replaces the metric strings with icons, and brings some additional improvements.

Any feedback welcome, and the plan is to release this on Monday.

PR #4521, opened 2026-08-18
Fix Smart Linking save button selector for WordPress 6.5+

### LINKED ISSUE #4518: Smart Linking save-button lock targets a class removed in WordPress 6.5
## Describe the bug

While Smart Links generate, `disableSave()` in `smart-linking/component.tsx` disables the Editor's save buttons so the post cannot be saved mid-generation:

```js
const saveButtons = document.querySelectorAll( '.edit-post-header__settings>[type="button"]' );
```

WordPress 6.5 unified the Editor header and renamed that wrapper to `editor-header__settings`. The old class survives only in a stylesheet; the class actually rendered by `@wordpress/editor` is the new one. The selector therefore matches nothing, and both `disableSave()` and `enableSave()` are no-ops on WordPress 6.5 and later.

The underlying safety is unaffected, because `lockPostSaving()` / `unlockPostSaving()` still run and still prevent the save. What is lost is the visual affordance: the save buttons stay enabled-looking, so the UI gives no indication that saving is blocked.

## To Reproduce

1. On WordPress 6.5 or later, start Smart Link generation on a post.
2. Observe that the save buttons remain visually enabled for the duration.

## Expected behavior

The save buttons appear disabled while generation is in progress, matching the locked save state.

## Environment

- WordPress 6.5 and later (verified against the class names in `@wordpress/editor`)
- wp-parsely 3.23.5

acicovic:
Fixes #4518

## Description

`disableSave()` and `enableSave()` in `smart-linking/component.tsx` target `.edit-post-header__settings`, which WordPress 6.5 renamed to `editor-header__settings` when it unified the Editor header. The selector matches nothing on 6.5 and later, so both functions are no-ops.

The selector now matches both, so this keeps working on the versions below 6.5 that the plugin still supports.

## Motivation and context

The save lock itself is unaffected, because `lockPostSaving()` and `unlockPostSaving()` still run and still prevent the save. What was lost is the visual affordance: the save buttons stayed enabled-looking while Smart Links generated, giving no indication that saving was blocked.

## How has this been tested?

Confirmed the class rename against the installed `@wordpress/editor` package: the class rendered in JS is `editor-header__settings`, and `edit-post-header__settings` survives only in a stylesheet, which is why the selector silently stopped matching.

Build, ESLint, stylelint, PHPCS, PHPStan and `composer lint` all pass. Jest 86 passing. E2E suite unchanged from `develop`.

This is not covered by an automated test. Asserting a `disabled` attribute on core's header markup would couple a test to the same core class names that broke here, which seems like the wrong trade.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with current and legacy WordPress editor layouts.
  * Save-button disabling and re-enabling now work consistently across supported editor versions.


acicovic:
@coderabbitai review.

PR #4522, opened 2026-08-18
Add ambient type declaration for SCSS imports

### LINKED ISSUE #4519: Missing *.scss ambient type declaration causes editor errors on style imports
## Describe the bug

The codebase imports SCSS for its side effects in around a dozen files:

```ts
import './smart-linking.scss';
```

There is no ambient module declaration for `*.scss` anywhere in the project. `src/@types/assets/` declares `*.svg` but nothing for stylesheets, so editors report:

> Cannot find module or type declarations for side-effect import of './smart-linking.scss'

This is cosmetic. The build is unaffected, because webpack resolves these through `sass-loader`, and the project's plain `tsc` invocation does not surface it either. It appears in editors running TypeScript 5.6+ with `noUncheckedSideEffectImports`, which is the default in several setups.

Confirmed with `tsc --noEmit --noUncheckedSideEffectImports`:

| | `*.scss` errors |
|---|---|
| As-is | **13** (TS2307) |
| With a `declare module '*.scss';` declaration | **0** |

Separately, `src/@types/assets/svg.d.ts` declares the module as `'\*.svg'` with an escaped asterisk. This happens to work, because `\*` collapses to `*` in a string literal, and I verified the wildcard still matches. It reads like a typo though, and invites someone to "fix" it while assuming they changed behaviour.

## Expected behavior

No spurious editor errors on SCSS side-effect imports, and no misleading escape sequence in the SVG declaration.

## Environment

- TypeScript 5.9.3
- wp-parsely 3.23.5

acicovic:
Fixes #4519

## Description

Adds `src/@types/assets/scss.d.ts` declaring `*.scss`, alongside the existing `*.svg` declaration. No `tsconfig.json` change is needed, since the file is picked up the same way `svg.d.ts` already is.

Also drops the escaped asterisk in `svg.d.ts`, which declared the module as `'\*.svg'`.

## Motivation and context

Around a dozen files import SCSS for its side effects. With no ambient declaration, editors running TypeScript 5.6+ with `noUncheckedSideEffectImports` report "Cannot find module or type declarations for side-effect import" on every one of them.

The build was never affected, since webpack resolves these through `sass-loader`, and the project's plain `tsc` invocation does not surface it either. This is purely about removing noise in the editor.

The `svg.d.ts` change is cosmetic. `\*` collapses to `*` in a string literal so the wildcard already matched, but it reads like a typo and invites someone to "fix" it while assuming they changed behaviour.

## How has this been tested?

Measured with `tsc --noEmit --noUncheckedSideEffectImports`:

| | `*.scss` errors |
|---|---|
| Before | **13** (TS2307) |
| After | **0** |

For `svg.d.ts`, I verified the escaped and unescaped forms behave identically by resolving a test SVG import against each, and included a deliberately non-matching pattern as a sensitivity check to confirm the test could actually detect a failure:

| Declaration | Errors |
|---|---|
| `'\*.svg'` | 0 — matches |
| `'*.svg'` | 0 — matches |
| `'*.svgZZZ'` | 1 — sensitivity check |

Build output is byte-identical, as expected for type-only changes. The project's existing `tsc` error count is unchanged at its pre-existing baseline. ESLint, stylelint, PHPCS, PHPStan and `composer lint` pass. Jest 86 passing.

## Summary by CodeRabbit

* **Chores**
  * Improved support for importing SCSS assets.
  * Updated SVG asset declarations for more reliable module recognition.


acicovic:
@coderabbitai review.

acicovic:
@coderabbitai review.

